PR #622, opened 2014-11-19
    Commit of partial work (this commit will be amended) to share with J...

aberke:
...ames

```
    - during validation stage:
        Event 1: resource_valid
            - run on each Module in URL path
            - Does this event need a better name, like validate_resource ?

            - binders:
                - cb_users_*
                    - load user_id into context
                - cb_accounts
                    - load account_db/accout_id into context

        Event 2: validate
            - only run on last Module, like resource_exists

    If only resource_exists() method took Context as parameter, could fold o
```

### DIFF
--- a/applications/crossbar/src/api_util.erl
+++ b/applications/crossbar/src/api_util.erl
@@ -802,26 +802,44 @@ does_resource_exist(_Context, _ReqNouns) ->
 %%--------------------------------------------------------------------
 %% @private
 %% @doc
-%% This function gives each noun a chance to determine if
+%% This function gives each {Mod, Params} pair a chance to determine if
 %% it is valid and returns the status, and any errors
+%%
+%% validate_resource for each {Mod, Params} pair
+%% validate for LAST {Mod, Params} pair
+%%
 %% @end
 %%--------------------------------------------------------------------
 -spec validate(cb_context:context()) -> cb_context:context().
-validate(Context0) ->
-    Context1 =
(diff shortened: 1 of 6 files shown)


aberke:
@jamesaimonetti is this PR eligible to go into 3.18?
As follow up to our call @brooksc @littlelies @daviddfm

jamesaimonetti:
@aberke I'll get a list of commit hashes together to cherry-pick to bring these changes into 3.18. Will ping you when 3.18 is ready and a build has been started. Thanks!

aberke:
Great thanks @jamesaimonetti !  How do you want to go about evaluating this list with our team?  I realize some of our PRs are a lot riskier for you to incorporate than others.

jamesaimonetti:
@aberke i don't anticipate there being issues, as we've been using them in our 3.19 dev work quite successfully, and the diff between 3.18 and 3.19 for those touched modules isn't very big. I'll let you know if any of the commits give me trouble getting into 3.18 and we can strategize then. For now, it looks straightforward to accomplish.

jamesaimonetti:
@aberke I've updated 3.18 with the pull requests (included patch for VM since it was so minimal). Please check out the 3.18 branch locally and give it a spin. I'm also building a [3.18 RPM](http://build001-fmt.2600hz.com:8080/view/Kazoo/job/Kazoo/441/) which you should be able to view the progress of via that link.

Let me know if something was missed, isn't working as it should, etc.

PR #671, opened 2014-12-12
Crossbar allow methods fix

aberke:


aberke:
@jamesaimonetti is this PR eligible to go into 3.18?
As follow up to our call @brooksc @littlelies @daviddfm

PR #681, opened 2014-12-18
Added support for an external storage for voicemail medias.

lbueno-daitan:
I wanted to save the voice mail medias in a different place, like the custom URL that we have in Recordings callflow module.

### DIFF
--- a/applications/callflow/src/module/cf_voicemail.erl
+++ b/applications/callflow/src/module/cf_voicemail.erl
@@ -44,12 +44,11 @@
                                    ,[<<"voicemail">>, <<"max_box_number_length">>]
                                    ,15
                                   )).
--define(MAILBOX_DEFAULT_EXTERNAL_STORAGE
+-define(MAILBOX_DEFAULT_STORAGE
         ,whapps_config:get_binary(?CF_CONFIG_CAT
                                   ,[<<"voicemail">>, <<"external_storage">>]
                                   ,'undefined'
                                  )).
-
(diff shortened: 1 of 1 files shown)


applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -1517,6 +1526,29 @@ store_recording(AttachmentName, DocId, Call) ->
...
+                                ,Call),
+    timer:sleep(5000),
+
+    case update_doc(<<"external_media_url">>, Url, DocId, Call) of
+        'ok' -> 'true';
+        {'error', _} -> 'false'
+    end.
+
+-spec get_media_url(ne_binary(), ne_binary(), whapps_call:call(), ne_binary(), ne_binary()) -> ne_binary().
+get_media_url(AttachmentName, DocId, Call, OwnerId, StorageUrl) ->
```
lbueno-daitan: I've already done this test locally. The `get_message/2` function will be called to resolve the media path. At this moment, if the voicemail metadata contains the `external_media_url` property, it will be returned. The `save_meta/4` was changed to add the `external_media_url` property, if it is present in the media document. 
So, when someone calls the mailbox, Kazoo will send Freeswitch the external URL instead of the resolved CouchDB URL. As Freeswitch will issue an http_get, the external application holding the media will have to handle these requests.

jamesaimonetti:
I've merged this in manually with a few formatting changes here and there, but overall this is great! Thanks for working with me to get it ready.

PR #687, opened 2014-12-22
Kazoo 3064

lazedo:


applications/doodle/src/doodle_inbound_listener.erl
```diff
@@ -0,0 +1,187 @@
...
+    Options = ?DOODLE_INBOUND_EXCHANGE_OPTIONS,
+    ED = amqp_util:declare_exchange(Exchange, Type, Options),
+    gen_listener:start_link({'local', ?MODULE}
+                            ,?MODULE
+                            ,[{'bindings', ?BINDINGS(Exchange)}
+                              ,{'responders', ?RESPONDERS}
+                              ,{'queue_name', ?QUEUE_NAME}       % optional to include
+                              ,{'queue_options', ?QUEUE_OPTIONS} % optional to include
+                              ,{'consume_options', ?CONSUME_OPTIONS} % optional to include
+                              ,{'declare_exchanges', [ED]}
```
jamesaimonetti: I don't think gen_listener client modules should be interacting with `#'exchange.declare'{}` records directly like this. I'd prefer to just pass `{Exchange, Type}` or `{Exchange, Type, Options}` tuples in this list and let gen_listener convert those to `#'exchange.declare{}` records via `amqp_util.erl`

core/whistle_amqp-1.0.0/src/amqp_util.erl
```diff
@@ -466,6 +467,22 @@ new_exchange(Exchange, Type, Options) ->
      },
     wh_amqp_channel:command(ED).
 
+-spec declare_exchange(ne_binary(), ne_binary()) -> 'ok'.
```
jamesaimonetti: Wrong spec, these functions return the records.

applications/doodle/src/doodle_inbound_listener.erl
jamesaimonetti: In fact, I think this should probably be in the wapi_* modules only. So wapi_sms should take the 2- or 3-tuple list and declare the exchanges itself, and leave gen_listener out of it.

core/whistle-1.0.0/src/api/wapi_sms.erl
```diff
@@ -174,43 +294,67 @@ resume_v(JObj) -> resume_v(wh_json:to_proplist(JObj)).
...
-    amqp_util:bind_q_to_exchange(Queue, ?DELIVERY_ROUTING_KEY(CallId), ?SMS_EXCHANGE),
-    bind_q(Queue, CallId, Restrict);
-bind_q(Queue, CallId, ['resume'|Restrict]) ->
-    amqp_util:bind_q_to_exchange(Queue, ?RESUME_ROUTING_KEY(CallId), ?SMS_EXCHANGE),
-    bind_q(Queue, CallId, Restrict);
-bind_q(_, _, []) -> 'ok'.
+    CallId = props:get_value('call_id', Props, props:get_value('message_id', Props, <<"*">>)),
+    SystemId = ?LOWER(props:get_value('system_id', Props, <<"*">>)),
+    RouteId = ?LOWER(props:get_value('route_id', Props, <<"*">>)),
+    Exchange = props:get_value('exchange', Props, ?SMS_EXCHANGE),
```
jamesaimonetti: This is probably where you should declare this custom exchange. Maybe a function to get the exchange:

```erlang
get_exchange(Props) ->
    case props:get_value('exchange', Props) of
        'undefined' -> ?SMS_EXCHANGE;
        ?SMS_EXCHANGE -> ?SMS_EXCHANGE;
        {Exchange, Type} ->
            get_exchange(Exchange, Type, []);
        {Exchange, Type, Options} ->
            get_exchange(Exchange, Type, Options)
        end.

get_exchange(Exchange, Type, Options) ->
    amqp_util:new_exchange(Exchange, Type, Options),
    Exchange.
```

Code is obviously rough-draft.

lazedo: comment added here https://github.com/2600hz/kazoo/commit/5daad50f22389b39410e0c1f5bd7e6b9b3691845

PR #693, opened 2014-12-30
KAZOO-3141: Handle 500 errors in cb_context (first draft)

macpie:


applications/crossbar/src/cb_context.erl
```diff
@@ -800,62 +800,140 @@ find_schema(<<_/binary>> = Schema) ->
...
+    build_system_error(500, 'unspecified_fault', <<"unspecified fault">>, Context);
 add_system_error('account_cant_create_tree', Context) ->
-    crossbar_util:response('fatal', <<"account creation fault">>, Context);
+    build_system_error(500, 'account_cant_create_tree', <<"account creation fault">>, Context);
 add_system_error('account_has_descendants', Context) ->
-    crossbar_util:response('fatal', <<"account has descendants">>, Context);
+    build_system_error(500, 'account_has_descendants', <<"account has descendants">>, Context);
 add_system_error('faulty_request', Context) ->
-    crossbar_util:response_faulty_request(Context);
+    build_system_error(404, 'faulty_request', <<"faulty request">>, Context);
```
jamesaimonetti: I feel like "faulty_request" should be 400. 404 is pretty specific to not finding the "thing".

PR #697, opened 2015-01-05
KAZOO-3071: Create an API to get the parents tree of an account

macpie:


applications/crossbar/src/modules/cb_accounts.erl
```diff
@@ -859,6 +866,75 @@ format_account_tree_results(Context, JObjs) ->
...
 
+
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec load_parents(ne_binary(), cb_context:context()) -> cb_context:context().
+load_parents(AccountId, Context) ->
+    case couch_mgr:get_results(?WH_ACCOUNTS_DB, ?AGG_VIEW_SUMMARY, []) of
```
jamesaimonetti: Why isn't this a `crossbar_doc:load_view/4`? You can do the processing of each JObj in the `normalize_fun()` spot

PR #698, opened 2015-01-06
KAZOO-3084: Handle the new 'weight' parameter on group endpoints to mana...

macpie:
...ge the element's order

### DIFF
--- a/applications/callflow/src/module/cf_ring_group.erl
+++ b/applications/callflow/src/module/cf_ring_group.erl
@@ -142,13 +142,12 @@ maybe_order_group_members(Member, JObj, Data) ->
             order_group_members(Member, JObj);
         _ ->
             Endpoints = wh_json:get_ne_value(<<"endpoints">>, JObj, wh_json:new()),
-            lists:foldl(
-                fun(Key, Acc) ->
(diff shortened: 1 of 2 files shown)


applications/callflow/src/module/cf_ring_group.erl
```diff
@@ -52,82 +53,131 @@ attempt_endpoints(Endpoints, Data, Call) ->
...
     end.
+
+-spec maybe_order_group_members(wh_json:object(), wh_json:object(), wh_json:object()) -> wh_json:objects().
+maybe_order_group_members(Member, JObj, Data) ->
+    case wh_json:get_binary_value(<<"strategy">>, Data, <<"simultaneous">>) of
+        <<"single">> ->
+            order_group_members(Member, JObj);
+        _ ->
+            Endpoints = wh_json:get_ne_value(<<"endpoints">>, JObj, wh_json:new()),
+            lists:foldl(
```
jamesaimonetti: If `Endpoints` is a JSON object, I'd use `wh_json:foldl/3` instead:

```erlang
wh_json:foldl(
    fun(Key, Endpoint, Acc) ->
        [create_group_member(Key, Endpoint, Member) | Acc]
    end, [], Endpoints)
```

PR #699, opened 2015-01-09
for create_account:

aberke:
Must create account before user can be validated.
user validation success depends on the existance of views in an account database - that account database must first exist.


aberke:
Please include this commit (will need to be cherry picked) in your 3.18 branch and a forthcoming 3.18 RPM so that we can use it!  Thanks!

jamesaimonetti:
@aberke Thanks! Wrapping up other work, then will merge this in and backport to 3.18.

PR #703, opened 2015-01-13
KAZOO-3229 - ecallmgr_maintenance registrar_summary show not all registrations

onnet:
To continue conversation https://groups.google.com/forum/#!topic/2600hz-dev/SBu5fGmUSpM

Here are some 'unusual' registration strings found in my registrations table:

Bria iOS 3.2.1:

<sip:5519@2.2.2.2:55061;transport=TLS;ob;fs_path=<sip:1.1.1.1:5061;lr;received='sip:1.1.1.1:55061;transport=tls'>>

<sip:User_2pnrza@3.3.3.3:64967;ob;fs_path=<sip:1.1.1.1:5060;lr;received='sip:3.3.3.3:64967;transport=udp'>>

Zoiper r28827:

<sip:5520@4.4.4.4:26070;transport=TLS;rinstance=e6b1e5bda0fcfd30;fs_path=<sip:1.1.1.1:5061;lr;received='sip:4.4.4.4:26070;transport=tls'>>

We could see here a couple of additional fields - "ob" and  "rinstance=e6b1e5bda0fcfd30"

Besides that re:run should be case insensitive because transport protocol name could be in both cases (TLS/tls)

So finally, regex string, which totally satisfied my humble 400 registrations look like this:

-define(SUMMARY_REGEX, <<"^.*?:.*@([0-9.:]*)(?:;transport=(?:udp|tls|tcp))?(?:;ob)?(?:;rinstance=.{16})?(?:;fs_path=.*?:([0-9.:]*);)*">>). 

Regards,
Kirill

### DIFF
--- a/applications/ecallmgr/src/ecallmgr_registrar.erl
+++ b/applications/ecallmgr/src/ecallmgr_registrar.erl
@@ -66,7 +66,7 @@
 -define(REG_QUEUE_NAME, <<>>).
 -define(REG_QUEUE_OPTIONS, []).
 -define(REG_CONSUME_OPTIONS, []).
--define(SUMMARY_REGEX, <<"^.*?:.*@([0-9.:]*)(?:;transport=udp|tcp|tls)(?:;fs_path=.*?:([0-9.:]*);)*">>).
+-define(SUMMARY_REGEX, <<"^.*?:.*@([0-9.:]*)(?:;transport=(?:udp|tls|tcp))?(?:;ob)?(?:;rinstance=.{16})?(?:;fs_path=.*?:([0-9.:]*);)*">>).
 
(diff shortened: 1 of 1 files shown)


k-anderson:
Howdy Kirill :)

So my concern is this regex is just going to keep growing in complexity.   Way back when we used to get registrations from FreeSWITCH the parameters were limited.  As we work to pass the parameters back to the device, each device is going to have unique properties.  Rather than accounting for this in the regex I have just changed the mechanism to pull the contact apart...
https://github.com/2600hz/kazoo/commit/9e9895a38ba501c542f829ada2392336561aac55


Sorry for doubling up on your effort.

PR #706, opened 2015-01-15
Kazoo 3075 - ref external schemas

jamesaimonetti:
To reduce duplication of schema validation objects


core/whistle-1.0.0/src/wh_json.erl
```diff
@@ -588,6 +591,13 @@ get_values(Key, JObj) ->
 set_values(KVs, JObj) when is_list(KVs) ->
     lists:foldr(fun({K,V}, JObj0) -> set_value(K, V, JObj0) end, JObj, KVs).
 
+-spec insert_value(key(), json_term(), object()) -> object().
+insert_value(Key, Value, JObj) ->
```
macpie: love that!

PR #710, opened 2015-01-20
cb_sms was not added to the system_config crossbar document as a default...

aberke:
... module.  Consequently, it did not start.


applications/crossbar/src/crossbar.hrl
```diff
@@ -72,6 +72,7 @@
                           ,'cb_user_auth', 'cb_users'
                           ,'cb_vmboxes'
                           ,'cb_webhooks', 'cb_whitelabel'
+                          ,'cb_sms'
```
jamesaimonetti: After I alphabetized them, @aberke goes and pulls a stunt like this!!! :)

aberke: Sorry man my pattern recognitions skills were off yesterday - didn't even notice!

PR #717, opened 2015-01-21
J5 trunks use separate white black lists for inbound/outbound.

cando-1p:


lazedo:
please make pull request against master, then it can be cherrypicked to 3.18.
your PR would be lost in 3.19, 3.20,....

PR #722, opened 2015-01-22
doodle_exec should also utilize it's gen_listener behavior implementatio...

aberke:
...n, no?

Analogous to cf_exe.erl..

I see that the 'initialize' is caught in `doodle_exec:handle_info` vs `handle_cast` and not quite the same then ensures, but this seems like a good place to ask why.  Why doesn't the flow match callflow?

### DIFF
--- a/applications/doodle/src/doodle_exe.erl
+++ b/applications/doodle/src/doodle_exe.erl
@@ -253,7 +253,7 @@ init([Call]) ->
     process_flag('trap_exit', 'true'),
     CallId = whapps_call:call_id(Call),
     put('callid', CallId),
-    self() ! 'initialize',
+    gen_listener:cast(self(), 'initialize'),
     {'ok', #state{call=Call}}.
 
 %%--------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


applications/doodle/src/doodle_exe.erl
```diff
@@ -253,7 +253,7 @@ init([Call]) ->
     process_flag('trap_exit', 'true'),
     CallId = whapps_call:call_id(Call),
     put('callid', CallId),
-    self() ! 'initialize',
+    gen_listener:cast(self(), 'initialize'),
```
lazedo: self() ! 'initialize' directs to handle_info not handle_cast, you know that right ?

aberke:
Yes: Sorry I was updating commit message as you commented.  In case you didn't see: 

```I see that the 'initialize' is caught in doodle_exec:handle_info vs handle_cast and not quite the same then ensures, but this seems like a good place to ask why. Why doesn't the flow match callflow?```

applications/doodle/src/doodle_exe.erl
aberke: Yes: Sorry I was updating commit message as you commented. In case you didn't see:

I see that the 'initialize' is caught in doodle_exec:handle_info vs handle_cast and not quite the same then ensures, but this seems like a good place to ask why. Why doesn't the flow match callflow?

aberke: Ie, would it be a valid PR if I changed the `handle_info('initialize', #state{call=Call})` to be `handle_cast('initialize', #state{call=Call})` ?

aberke: Yes the main difference is `spawn('cf_singular_call_hooks', 'maybe_hook_call', [CallWithHelpers]),` but I hoped this PR could help me understand the pattern/flow here

aberke:
I amended the commit for the full change @lazedo

PR #727, opened 2015-01-23
doodle_util was not properly taking the Message Body.  This resulted in ...

aberke:
...sms documents with no body property - a bug

### DIFF
--- a/applications/doodle/src/doodle_util.erl
+++ b/applications/doodle/src/doodle_util.erl
@@ -20,7 +20,7 @@
 -export([endpoint_from_sipdb/2, get_endpoint_from_sipdb/2]).
 -export([save_sms/1, save_sms/2]).
 -export([replay_sms/2]).
--export([set_sms_body/2]).
+-export([get_sms_body/1, set_sms_body/2]).
 -export([set_flow_status/2, set_flow_status/3]).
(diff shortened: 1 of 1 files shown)


lazedo:
Hi Alex,
the body is only needed for the initial creation, after that it is taken from pvt_call.
check the views.
would you be more comfortable if we include this ?

aberke:
I don't understand what you mean 
https://github.com/2600hz/kazoo/blob/master/core/kazoo_modb-1.0.0/priv/couchdb/views/sms.json#L5
right?
sms documents have body in the schema but the body is not saved in the SMS document past this point (it is when it's first PUT through crossbar)

lazedo:
the views used to reference the body inside pvt_call, must have changed this :) 
anyway, this will not solve the issue because if you change the body along the flow, the final version will not be saved in the Body field.

add these to doodle_util and remove your change

-export([get_sms_body/1]).  

-spec get_sms_body(whapps_call:call()) -> ne_binary().
get_sms_body(Call) ->
    whapps_call:kvs_fetch(<<"Body">>, Call).


in save_sms/4

--    Body = wh_json:get_value(<<"Body">>, JObj),
++  Body = get_sms_body(Call),

aberke:
I amended the commit

lazedo:
thank you

PR #728, opened 2015-01-24
KAZOO-3168: Fix descendant_count when too many requests

macpie:


applications/crossbar/src/crossbar_util.erl
```diff
@@ -956,33 +956,53 @@ load_descendants_count(ViewOptions) ->
...
 -spec maybe_update_descendants_count(ne_binary(), integer()) -> 'ok'.
+-spec maybe_update_descendants_count(ne_binary(), integer(), integer()) -> 'ok'.
+-spec maybe_update_descendants_count(ne_binary(), wh_json:object(), integer(), integer()) -> 'ok'.
+-spec maybe_update_descendants_count(ne_binary(), wh_json:object(), integer(), integer(), integer()) -> 'ok'.
+
 maybe_update_descendants_count(AccountId, NewCount) ->
+    maybe_update_descendants_count(AccountId, NewCount, 3).
+
+maybe_update_descendants_count(AccountId, _, Try) when Try =< 0 ->
+    io:fromat("too many attempts to update descendants count for ~s~p", [AccountId]);
```
jamesaimonetti: This function does not exist. Also, why no lager? Or, if you do need to use io, you need the `~n` to add an endline.

applications/crossbar/src/crossbar_util.erl
```diff
@@ -956,18 +956,34 @@ load_descendants_count(ViewOptions) ->
...
-                    update_descendants_count(AccountId, JObj, NewCount)
-            end
+            maybe_update_descendants_count(AccountId, JObj, NewCount, Try)
+    end.
+
+maybe_update_descendants_count(AccountId, JObj, NewCount, Try) ->
+    OldCount = wh_json:get_integer_value(<<"descendants_count">>, JObj),
+    maybe_update_descendants_count(AccountId, JObj, NewCount, OldCount, Try).
+
+maybe_update_descendants_count(_, _, Count, Count, _) -> 'ok';
```
lazedo: unused var breaks compilation

PR #729, opened 2015-01-24
Allow Accounts to have preflows that are executed via doodle (sms) as we...

aberke:
...ll as in callflow application.


lazedo:
Hi Alex,
can you please keep the same names as in callflow ?
this will help to merge doodle/callflow in the future.
thanks

aberke:
I didn't like that callflow was using functions names like `maybe_exec_preflow` and `exec_preflow` because from what I can tell no callflow execution happens in the route_req.  It seems the execution happens in the route_win.  It takes a lot of reading code to realize this when it could be immediately obvious from the function names.

lazedo:
i agree and i think your function name is better but for the sake of consistency and possible future merge of _exe into a single executor, we do want to keep the same name. you can choose to change the names in callflow.

aberke:
Ok I did :+1:

aberke:
Also just rebased against master to make a nicer branch for you.

lazedo:
thank you

PR #731, opened 2015-01-27
KAZOO-3252: Fix Delete notification & fetch attachment

macpie:


applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -681,7 +683,7 @@ read_template(Context, Id, Accept) ->
             lager:debug("found attachment ~s in ~s", [AttachmentName, Id]),
 
             cb_context:add_resp_headers(
-              crossbar_doc:load_attachment(Id, AttachmentName, Context)
+              read_account_attachement(Context, Id, AttachmentName)
```
jamesaimonetti: Spelling of 'attachment' is wrong :)

PR #732, opened 2015-01-28
Kazoo 3247

macpie:


applications/crossbar/src/modules/provisioner_v5.erl
```diff
@@ -31,14 +32,20 @@
...
+    AccountId = wh_json:get_value(<<"pvt_account_id">>, JObj),
     Data = provision_data(JObj),
-    MACAddress = wh_json:get_value(<<"mac_address">>, JObj),
-    _ = send_req('devices_put'
-                 ,Data
-                 ,AuthToken
-                 ,wh_json:get_value(<<"pvt_account_id">>, JObj)
-                 ,MACAddress),
-    send_req('files_post', AuthToken, MACAddress).
+    case check_data(wh_json:set_value(<<"merge">>, 'false', Data)) of
```
k-anderson: merge should be set to 'true'

macpie: It is done by the schema default.

PR #735, opened 2015-01-30
J5 multiple flexible trunk definitions.

cando-1p:
Partial implementation as "discussed" in https://2600hz.atlassian.net/browse/KAZOO-3038
and https://docs.google.com/document/d/19JlWKIe53q2e2MQIhsF-ohTkGcaYx6AHwcMHU_AquVo/edit?usp=sharing
See detailed description in new applications/jonny5/doc/trunks.md


applications/jonny5/src/j5_limits.erl
```diff
@@ -64,10 +67,12 @@
 -spec get(ne_binary()) -> limits().
 get(Account) ->
     AccountId = wh_util:format_account_id(Account, 'raw'),
-    case wh_cache:peek_local(?JONNY5_CACHE, ?LIMITS_KEY(AccountId)) of
-        {'ok', Limits} -> Limits;
+    Limits = case wh_cache:peek_local(?JONNY5_CACHE, ?LIMITS_KEY(AccountId)) of
+        {'ok', _Limits} -> _Limits;
```
jamesaimonetti: Using the underscore-prefixed variable name is bad practice (other than when logging). Maybe just use `{'ok', Lims} -> Lims;` or something similarly ghetto :). Using `_Limits` reads as "This Variable contains the limits but I don't care about it here and it should be ignored", which isn't the case.

cando-1p: I made a change just to the var name as you suggested (plus copyright and
contributors) but now for some reason github thinks the whole limits file
has changed.
Trying to work that out.

cando-1p: Looks like commit cb1115fd6343f299e015f4ff3bae2b28ffd25a41 shows the diff the way it should.
I filed a ticket with github.

cando-1p: After wasting a lot of time I found that my editor infact did save the file
sometimes with CRLF and other times with just LF for line endings.
I found a plugin that will show me what it is using and switch it to what I
want.
So the most recent commit now has both j5_limits.erl and trunks.md with
just LF for line endings. Gotta love netbeans. GRRR. Tried switching to
eclipse a little while back but got hung up on many differences. Maybe I'll
try again.

jamesaimonetti:
I've cleaned up your pull request a bit, mostly markdown syntax, formatting, etc, and re-submitted it in https://github.com/2600hz/kazoo/pull/749. Things look good, so far, to me but I want to make sure @k-anderson takes a look too.

PR #736, opened 2015-01-31
Kazoo 3212 - konami

jamesaimonetti:
Konami improvements


applications/ecallmgr/src/ecallmgr_fs_bridge.erl
```diff
@@ -68,42 +70,41 @@ unbridge(UUID, JObj) ->
...
-            case length(wh_json:get_ne_value(<<"Endpoints">>, JObj, [])) > 1 of
-                'false' -> 'ok';
-                'true' ->
-                    lager:debug("bridge is simultaneous to multiple endpoints, starting local ringing"),
-                    %% we don't really care if this succeeds, the call will fail later on
-                    ecallmgr_util:send_cmd(Node, UUID, <<"ring_ready">>, "")
-            end;
-        _Else -> 'ok'
+    Endpoints = wh_json:get_ne_value(<<"Endpoints">>, JObj, []),
+    case ecallmgr_util:get_dial_separator(JObj, Endpoints) of
```
k-anderson: Previously it sent ring_ready when trying more than 1 simultaneous endpoint.  The refactor now appears to send it if its single?

applications/ecallmgr/src/ecallmgr_call_events.erl
```diff
@@ -462,10 +466,32 @@ handle_info('sanity_check', #state{callid=CallId}=State) ->
...
+    NewUUID = handle_bowout(Node, Props, ResigningUUID),
+    {'noreply', State#state{call_id=NewUUID}};
 handle_info(_Info, State) ->
     lager:debug("unhandled message: ~p", [_Info]),
     {'noreply', State}.
 
+-spec handle_bowout(atom(), wh_proplist(), ne_binary()) -> ne_binary().
+handle_bowout(Node, Props, ResigningUUID) ->
+    AcquiringUUID = props:get_value(?ACQUIRED_UUID, Props),
+    case props:get_value(?RESIGNING_UUID, Props) of
```
k-anderson: The bowout will happen on call forwarding and some other use-cases where the call-id may not actually change.  We shouldn't unregister then re-register for events for the same call-id ;)

applications/ecallmgr/src/ecallmgr_call_command.erl
```diff
@@ -767,13 +768,18 @@ get_call_pickup_app(Node, UUID, JObj, Target, Command) ->
                     ,{<<"Fetch-ID">>, wh_util:rand_hex_binary(4)}
                     | wh_api:default_headers(?APP_NAME, ?APP_VERSION)
                    ],
-    wh_amqp_worker:cast(ControlUsurp
-                        ,fun(C) -> wapi_call:publish_usurp_control(Target, C) end
-                       ),
-    lager:debug("published control usurp for ~s", [Target]),
+
+    wh_json:is_true(<<"Publish-Usurp">>, JObj, 'true')
```
k-anderson: For readability I would prefer to use a case-clause here.  It would also allow you to log when the usurp is supressed.

applications/ecallmgr/src/ecallmgr_call_events.erl
```diff
@@ -265,21 +260,20 @@ handle_cast({'transferer', _Props}, State) ->
...
     {'noreply', State#state{other_leg_events=Events}};
-handle_cast({'other_leg', _OtherLeg}, #state{other_leg_events=[]}=State) ->
-    lager:debug("ignoring other leg events for ~s", [_OtherLeg]),
-    {'noreply', State};
 handle_cast({'other_leg', OtherLeg}, #state{other_leg_events=_Events
                                             ,node=Node
                                            }=State) ->
     lager:debug("tracking other leg events for ~s: ~p", [OtherLeg, _Events]),
-    'true' = register_for_events(Node, OtherLeg),
+    _ = ecallmgr_call_sup:start_event_process(Node, OtherLeg),
```
k-anderson: Why start this as a separate process?  I also don't follow why other_leg_events is being ignored here, lets go over this :)

applications/ecallmgr/src/ecallmgr_call_events.erl
```diff
@@ -900,6 +929,7 @@ get_application_name(Props) ->
         <<"sofia::replaced">> -> <<"transfer">>;
         <<"spandsp::rxfax", Event/binary >> -> <<"rxfax",Event/binary>>;
         <<"spandsp::txfax", Event/binary >> -> <<"txfax", Event/binary>>;
+        <<"loopback::bowout">> -> <<"originate">>;
```
k-anderson: Bowout does not necessarily mean originate, for example call forward.

PR #739, opened 2015-02-01
Don't overwrite voicemail greeting media if it was shared media file upl...

vladimirralev:
...oaded from the Web UI

When setting up a mailbox from the Web UI you can use any media file on the account, including shared media files. If somebody with such shared greeting file re-records their voicemail greeting from the phone it will overwrite the shared file (everybody else's greeting as well). This patch checks if the greeting was uploaded from the UI and creates a new media file as opposed to overwriting it when that's the case.

### DIFF
--- a/applications/callflow/src/module/cf_voicemail.erl
+++ b/applications/callflow/src/module/cf_voicemail.erl
@@ -819,7 +819,22 @@ record_unavailable_greeting(AttachmentName, #mailbox{unavailable_media_id='undef
     MediaId = recording_media_doc(<<"unavailable greeting">>, Box, Call),
     record_unavailable_greeting(AttachmentName, Box#mailbox{unavailable_media_id=MediaId}, Call);
 record_unavailable_greeting(AttachmentName, #mailbox{unavailable_media_id=MediaId}=Box, Call) ->
-    lager:info("recording unavailable greeting  as ~s", [AttachmentName]),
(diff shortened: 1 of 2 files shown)


vladimirralev:
Added a second commit for another completely separate issue where an established ringgroup call can be interrupted if a parallel attempt times out is some race conditions.

applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -819,7 +819,22 @@ record_unavailable_greeting(AttachmentName, #mailbox{unavailable_media_id='undef
     MediaId = recording_media_doc(<<"unavailable greeting">>, Box, Call),
     record_unavailable_greeting(AttachmentName, Box#mailbox{unavailable_media_id=MediaId}, Call);
 record_unavailable_greeting(AttachmentName, #mailbox{unavailable_media_id=MediaId}=Box, Call) ->
-    lager:info("recording unavailable greeting  as ~s", [AttachmentName]),
+    case couch_mgr:open_doc(whapps_call:account_db(Call), MediaId) of
```
jamesaimonetti: Please use `couch_mgr:open_cache_doc/2` since the resulting JObj should already have been loaded; we can save a trip to the DB in that case.

vladimirralev:
@jamesaimonetti new revision with cached document is in this pull request now

jamesaimonetti:
Manually merged with some other things. I agree on your last comment and hope you can get some logs so we can fix the ring group properly.

PR #744, opened 2015-02-03
KAZOO-3183: Flush stepswitch_cnam and trunkstore when number doc is modi...

macpie:
...fied

### DIFF
--- a/applications/trunkstore/src/ts_util.erl
+++ b/applications/trunkstore/src/ts_util.erl
@@ -108,6 +108,7 @@ lookup_did(DID, AccountId) ->
             Resp;
         {'error', 'not_found'} ->
             Options = [{'key', DID}],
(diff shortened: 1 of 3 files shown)


applications/trunkstore/src/ts_util.erl
```diff
@@ -108,6 +108,7 @@ lookup_did(DID, AccountId) ->
             Resp;
         {'error', 'not_found'} ->
             Options = [{'key', DID}],
+            CacheProps = [{'origin', [{'db', wnm_util:number_to_db_name(DID), DID}, {'type', <<"number">>}]}],
```
jamesaimonetti: You need to update trunkstore_sup's origin bindings to include `{'type', <<"number">>}`, similar to stepswitch_sup.

PR #748, opened 2015-02-04
KAZOO-3294: Add sub_account_name to transactions

macpie:


core/whistle_transactions-1.0.0/src/wh_transaction.erl
```diff
@@ -49,6 +51,7 @@
                          ,description :: api_binary()
                          ,call_id :: api_binary()
                          ,sub_account_id :: ne_binary()
+                         ,sub_account_name :: 'undefined' | ne_binary()
```
jamesaimonetti: should be api_binary()

PR #749, opened 2015-02-04
Jonny5 multiple flexable trunk def

jamesaimonetti:
From @cando-1p but cleaned up a bit!


fenollp:
Is this dead @jamesaimonetti ?

cando-1p:
No. It is not dead. I have made a bunch of fixes over what is in the pull
request working in production. Unfortunately, in my todo list, I have had a
bunch of other things push their way above making another pull request.
They even made it ahead of upgrading from 3.08 to 3.18. Which I'm finally
back to. Hopefully, I will be able to make the update shortly after that.
Further we have some customers wanting the third stage (as discussed in the
design doc) of supporting multiple trunk definitions per customer. So a
customer can buy 5 US trunks, 8 Western Euro landline trunks.... kind of
idea. But priorities have a strange way of changing. So we will see what
happens.

Dave (cando)

fenollp:
@cando-1p please rebase when you have the time?

fenollp:
Heyo @cando-1p, could you finish this when you get the time?

k-anderson:
@cando-1p as per our IRC discussion we will await a new PR against master! :)

PR #755, opened 2015-02-05
Receiving notification of new voicemail causes document conflicts when w...

aberke:
...e try to fetch that voicemail from the messages list of the vmbox document.  It would be helpful to have a notificaiton published once that metadata is saved in the messages list so that we avoid this race condition.

We are going to implement this either way but would prefer to have this in the 2600hz/kazoo code and to get your feedback via a pull request.  I'm sure you have some holes to poke.

Thanks

### DIFF
--- a/applications/callflow/src/module/cf_voicemail.erl
+++ b/applications/callflow/src/module/cf_voicemail.erl
@@ -1128,8 +1128,9 @@ save_meta(Length, #mailbox{mailbox_id=Id}, Call, MediaId) ->
                            {'ok', JObj} -> wh_json:get_value(<<"external_media_url">>, JObj);
                            {'error', _} -> 'undefined'
                        end,
+    Timestamp = new_timestamp(),
     Metadata = wh_json:from_list(props:filter_undefined(
-                 [{<<"timestamp">>, new_timestamp()}
+                 [{<<"timestamp">>, Timestamp}
                   ,{<<"from">>, whapps_call:from(Call)}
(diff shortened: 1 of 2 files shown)


jamesaimonetti:
Where are you getting the conflict?

Are you bound for the new voicemail notification as well, and when you try to update the message in the voicemail box document, you are racing cf_voicemail to update it? Just trying to get context here.

aberke:
Context:

We have NOT yet implemented my proposed notification above.
We are binding to the `"voicemail_new"` notification.  We are trying to retrieve the new metadata to send out push notifications to devices and then update the metadata with our own added info.  

Yes the issue is that sometimes we are racing cf_voicemail to the update.

jamesaimonetti:
Thanks @aberke, sorry for the delay. I refactored it a bit (and you had a syntax error in wapi_notifications) just to break the larger blocks into functions, but this is now in master. Do you need it backported at all?

aberke:
Thanks for the merge.  If you could put it in 3.18 that would be great!

@jbbichon and @Littlelies

jamesaimonetti:
@aberke I thought you all were targeting the 3.19 builds now? Is this not the case?

aberke:
We are targeting both - for different time frames.  Yes please backport to 3.19.  We'll continue to code around it until we're up to 3.19

jamesaimonetti:
@aberke sorry for the delay, just back ported this to 3.19. Will issue a build in a bit (once I've checked with the others to see if there are other commits for 3.19).

PR #756, opened 2015-02-05
KAZOO-3156: Delete orphaned MODbs & forbid their creation in the first place

fenollp:
There was an issue where accounts would get processed even though
they were soft-destroyed (ie. had the tag pvt_deleted to true).
This tag is now checked (in the account crawler pass).

A tool to remove old orphaned MODbs has been added to
whapps_maintenance: cleanup_orphan_modbs/0.
It uses standalone code that deletes MODbs only if orphaned.
No deletion had previously been implemented in kazoo_modb.

### DIFF
--- a/applications/notify/src/notify_account_crawler.erl
+++ b/applications/notify/src/notify_account_crawler.erl
@@ -126,13 +126,8 @@ handle_info('next_account', [Account|Accounts]) ->
             AccountId ->
                 %% do not open the account def in the account db or we will
                 %% be wasting bigcouch's file descriptors
-                case couch_mgr:open_doc(?WH_ACCOUNTS_DB, AccountId) of
-                    {'ok', JObj} ->
-                        AccountDb = wh_json:get_value(<<"pvt_account_db">>, JObj),
-                        process_account(AccountId, AccountDb, JObj);
-                    {'error', _R} ->
-                        lager:warning("unable to open account definition for ~s: ~p", [AccountId, _R])
-                end
+                Doc = couch_mgr:open_doc(?WH_ACCOUNTS_DB, AccountId),
+                check_then_process_account(AccountId, Doc)
         end,
     Cycle = whapps_config:get_integer(?MOD_CONFIG_CAT, <<"interaccount_delay">>, 10000),
     erlang:send_after(Cycle, self(), 'next_account'),
@@ -150,6 +145,20 @@ handle_info('crawl_accounts', _) ->
 handle_info(_Info, State) ->
     lager:debug("unhandled msg: ~p", [_Info]),
     {'noreply', State}.
+
+-spec check_then_process_account (ne_binary(), {'ok', wh_json:object()} | {'error',_}) -> 'ok'.
+check_then_process_account (AccountId, {'ok', JObj}) ->
+    case wh_json:is_true(<<"pvt_deleted">>, JObj) of
+        'true' ->
+            %% Account has actually been soft-destroyed
+            lager:debug("not processing account ~p (soft-destroyed)", [AccountId]),
+            'ok';
+        'false' ->
+            AccountDb = wh_json:get_value(<<"pvt_account_db">>, JObj),
+            process_account(AccountId, AccountDb, JObj)
+    end;
+check_then_process_account (AccountId, {'error', _R}) ->
+    lager:warning("unable to open account definition for ~s: ~p", [AccountId, _R]).
 
 %%--------------------------------------------------------------------
 %% @private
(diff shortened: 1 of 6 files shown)


applications/notify/src/notify_account_crawler.erl
```diff
@@ -126,13 +126,8 @@ handle_info('next_account', [Account|Accounts]) ->
...
                 %% do not open the account def in the account db or we will
                 %% be wasting bigcouch's file descriptors
-                case couch_mgr:open_doc(?WH_ACCOUNTS_DB, AccountId) of
-                    {'ok', JObj} ->
-                        AccountDb = wh_json:get_value(<<"pvt_account_db">>, JObj),
-                        process_account(AccountId, AccountDb, JObj);
-                    {'error', _R} ->
-                        lager:warning("unable to open account definition for ~s: ~p", [AccountId, _R])
-                end
+                Doc = couch_mgr:open_doc(?WH_ACCOUNTS_DB, AccountId),
```
jamesaimonetti: I would name this `OpenResult` or similar, since it isn't a Doc JSON object.

applications/notify/src/notify_account_crawler.erl
```diff
@@ -151,6 +146,20 @@ handle_info(_Info, State) ->
     lager:debug("unhandled msg: ~p", [_Info]),
     {'noreply', State}.
 
+-spec check_then_process_account (ne_binary(), {'ok', wh_json:object()} | {'error',_}) -> 'ok'.
+check_then_process_account (AccountId, {'ok', JObj}) ->
+    case wh_json:is_true(<<"pvt_deleted">>, JObj) of
+        'true' ->
+            %% Account has actually been soft-destroyed
+            lager:debug("not processing account ~p (soft-destroyed)", [AccountId]),
+            'ok';
```
jamesaimonetti: `lager:debug/2` returns 'ok', so no need to do it explicitly.

applications/notify/src/notify_account_crawler.erl
```diff
@@ -151,6 +146,20 @@ handle_info(_Info, State) ->
     lager:debug("unhandled msg: ~p", [_Info]),
     {'noreply', State}.
 
+-spec check_then_process_account (ne_binary(), {'ok', wh_json:object()} | {'error',_}) -> 'ok'.
+check_then_process_account (AccountId, {'ok', JObj}) ->
```
jamesaimonetti: Don't leave a space between the function name and args

core/kazoo_modb-1.0.0/src/kazoo_modb.erl
```diff
@@ -286,3 +287,26 @@ should_archive(AccountMODb, Year, Month) ->
             ModbMonths = (ModbYear * 12) + ModbMonth,
             (Months - ModbMonths) > whapps_config:get_integer(?CONFIG_CAT, <<"active_modbs">>, 6)
     end.
+
+
```
jamesaimonetti: You and @macpie love all this whitespace! :) Must be a French thing...

fenollp: ;)

core/kazoo_modb-1.0.0/src/kazoo_modb.erl
```diff
@@ -286,3 +287,26 @@ should_archive(AccountMODb, Year, Month) ->
...
+%% Delete an modb if it is no longer associated with its account.
+%% (That is: orphaned).
+%% AccountMODb must be 'encoded' otherwise couch_mgr:db_delete/1 will
+%% fail. Returns whether AccountMODb has been deleted.
+%% @end
+%%--------------------------------------------------------------------
+-spec maybe_delete (ne_binary()) -> boolean().
+maybe_delete (AccountMODb) ->
+    AccountId = wh_util:format_account_id(AccountMODb, 'raw'),
+    AccountIds = whapps_util:get_all_accounts('raw'),
```
jamesaimonetti: I wonder if you could make a `maybe_delete/2` that takes the list of AccountIds, so that the calling function, when it knows it will be doing a lot of operations like this, could pass the list in instead. Consider when we have 10,000+ databases, calculating AccountIds for a large number of MODbs would put a bit of strain on the DB cluster.

fenollp:
https://github.com/2600hz/kazoo/commit/b382a8ddfa2bc473cfac66ab391ccd6aa0153847 : so that's why my changes wouldn't show up. This isn't in the right branch!

jamesaimonetti:
@fenollp When committing, please use the ticket number as a prefix.

"KAZOO-WXYZ: I did this thing" instead of "I did this thing".

Also, title the Pull Request with the ticket number as well.

fenollp:
Er, no problem. I will do this from now on.
I rebased but already merged the branch, so not sure what's going to happen…

PR #759, opened 2015-02-05
Kazoo 3299

macpie:


applications/crossbar/src/cb_context.erl
```diff
@@ -530,21 +529,22 @@ failed_error({'data_invalid'
...
-              ,_FailedSchemaJObj
+              ,FailedSchemaJObj
               ,'not_in_enum'
               ,_FailedValue
               ,FailedKeyPath
              }, Context) ->
     add_validation_error(FailedKeyPath
                          ,<<"enum">>
-                         ,<<"Value not found in enumerated list of values">>
+                         ,[{<<"message">>, <<"Value not found in enumerated list of values">>}
```
jamesaimonetti: As we discussed, I think all calls to `add_validation_error/4` should pass a json object in the third argument.

PR #760, opened 2015-02-06
KAZOO-3170: add a master capture group to phone number regexes

fenollp:
It appears the defaults stored in whapps_config are automatically updated, so no need to flush the defaults in ETS.

### DIFF
--- a/applications/crossbar/doc/internationalization/numbers.md
+++ b/applications/crossbar/doc/internationalization/numbers.md
@@ -83,33 +83,33 @@ This is a set of regexes to group numbers by type and are not used for routing. 
 
     "classifiers":{
         "tollfree_us":{
-            "regex":"^\\+1(800|888|877|866|855)\\d{7}$",
+            "regex":"^\\+1((?:800|888|877|866|855)\\d{7})$",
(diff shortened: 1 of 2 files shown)


applications/crossbar/doc/internationalization/numbers.md
```diff
@@ -83,33 +83,33 @@ This is a set of regexes to group numbers by type and are not used for routing.
 
     "classifiers":{
         "tollfree_us":{
-            "regex":"^\\+1(800|888|877|866|855)\\d{7}$",
+            "regex":"^\\+1((800|888|877|866|855)\\d{7})$",
```
jamesaimonetti: I think the inner capture groups should be non-capturing: `(?:800|888|877|866|855)` in this case., no?

fenollp: You're right. I thought R15B03 didn't support regexp comments, but indeed it does.
And sorry for the response time, I didn't see your comment earlier.

fenollp:
@jamesaimonetti : fixed your comment & rebased.

PR #763, opened 2015-02-09
KAZOO-2137 wakeup sip softphones with cloud message

lazedo:


jamesaimonetti:
@lazedo updated the makefiles and built cleanly here. Give it a test on your system.

Also, how are the beam files for gcm and apn getting loaded into the running VM?

lazedo:
@jamesaimonetti  included modules dir in makefile. everything compiling fine and running. gcm and apns supervisors are loaded in pusher_sup.

jamesaimonetti:
@lazedo I've walked the code, made a few changes but I don't think I've impacted functionality. I'll test actually pushing to my phone and see what happens!

PR #765, opened 2015-02-10
KAZOO-3235: POST hangups JSON data to a URI.

fenollp:
2 new fields are added to hangups' system_config:
* `enable_email_alerts` :: `boolean()`: `true` means receiving emails
on hangups alerts (this is the legacy behaviour).
* `subscriber_url` :: `string()`: when non-empty will be used to send
hangups alerts as JSON in a HTTP POST request.

When both alert methods are deactivated, email is used.

### DIFF
--- a/applications/hangups/src/hangups_channel_destroy.erl
+++ b/applications/hangups/src/hangups_channel_destroy.erl
@@ -32,27 +32,57 @@
 %%
 %% @end
 %%--------------------------------------------------------------------
--spec handle_req(wh_json:object(), wh_proplist()) -> any().
-handle_req(JObj, _Props) ->    
+-spec handle_req(wh_json:object(), wh_proplist()) -> 'ok'.
+handle_req(JObj, _Props) ->
     'true' = wapi_call:event_v(JObj),
     HangupCause = wh_json:get_value(<<"Hangup-Cause">>, JObj, <<"unknown">>),
     case lists:member(HangupCause, ?IGNORE) of
         'true' -> 'ok';
         'false' ->
-            AccountId = wh_json:get_value([<<"Custom-Channel-Vars">>, <<"Account-ID">>], JObj),
-            lager:debug("abnormal call termination: ~s", [HangupCause]),
-            wh_notify:system_alert("~s ~s to ~s (~s) on ~s(~s)"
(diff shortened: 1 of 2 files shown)


applications/hangups/src/hangups_channel_destroy.erl
```diff
@@ -32,28 +32,58 @@
...
-                                     ,AccountId
-                                    ]
-                                   ,maybe_add_hangup_specific(HangupCause, JObj)
-                                  ),
-            add_to_meters(AccountId, HangupCause)
+            alert_about_hangup(HangupCause, JObj)
     end.
 
+
+-spec alert_about_hangup(wh_json:object(), ne_binary()) -> 'ok'.
```
jamesaimonetti: I think the args are backwards here

applications/hangups/src/hangups_channel_destroy.erl
```diff
@@ -32,28 +32,58 @@
...
+alert_about_hangup(HangupCause, JObj) ->
+    lager:debug("abnormal call termination: ~s", [HangupCause]),
+    EmailTrue = whapps_config:get_is_true(?APP_NAME, <<"enable_email_alerts">>, 'true'),
+    SUBUrl    = whapps_config:get_string(?APP_NAME, <<"subscriber_url">>, <<>>),
```
jamesaimonetti: Use `get_string/2` here, since the type cast only applies when a value is found. So your call here will return either the list value of "subscriber_url" or an empty binary. In the case clause below, you test for `<<>>` but in `alert_about_hangup__POST/3` you test for `""` which won't match `<<>>`.

With `get_string/2`, you'd get a list or `'undefined'` which is more explicit to test for.

fenollp:
* Fixed spec args.
* Use get_string/2
* Added a way better way to handle POST failures.

fenollp:
Wait I need to patch something

fenollp:
Ok now it's good

jamesaimonetti:
Can you add a file to `applications/hangups/doc` with directions on how to start receiving the JSON payloads please? What commands does someone need to run on a new system to get their URI being used for these POSTs? Code looks fine otherwise.

fenollp:
On it!

fenollp:
Done

PR #767, opened 2015-02-10
KAZOO-3314

lazedo:
1) fix pagination for 3 part key. 
2) handle authorization


applications/crossbar/src/modules/cb_search.erl
```diff
@@ -105,6 +123,8 @@ validate_search(Context, Q, T) ->
     validate_search(Context, cb_context:req_value(Context, <<"v">>), Q, T).
 
 -spec validate_search(cb_context:context(), api_binary(), ne_binary(), ne_binary()) -> cb_context:context().
+validate_search(Context, 'true', Q, T) ->
```
jamesaimonetti: spec is wrong, doesn't account for 'true'.

PR #771, opened 2015-02-12
Kazoo 3251

fenollp:
**This is not over yet!!**

The part about `maybe_add_information/1` was a hidden bug (there are in fact 2 bugs in the ticket)
This GET request
```
curl -s -H "Accept: application/json" -H "Content-Type: application/json" -H "X-Auth-Token: 963bcd18d187fbee42b0eae4b68d04d3" -X GET http://10.26.0.144:8000/v1/accounts/26528f16f78de7b6b63822b94e266434/phone_numbers/14157125234 | pj
```
would return
```
{
    "auth_token": "963bcd18d187fbee42b0eae4b68d04d3",
    "data": {
        "id": "+14157125234",
        "used_by": "undefined"
    },
    "request_id": "1fdc2da748fa991401f5c00258c43e2f",
    "revision": "undefined",
    "status": "success"
}
```
instead of the complete
```
{
    "auth_token": "428f16de97c5ef91808b5c6cc5f18fae",
    "data": {
        "assigned_to": "26528f16f78de7b6b63822b94e266434",
        "created": 63590344153,
        "features": [
            "local"
        ],
        "id": "+14157125234",
        "state": "in_service",
        "updated": 63590356189,
        "used_by": "undefined"
    },
    "request_id": "25a927c5ad9983af47d5855b2a081a83",
    "revision": "undefined",
    "status": "success"
}
```

---

The other part is about fields not updated on DELETE (in particular the inners of the `features` field)
I think it has to do with [the process dictionary here](https://github.com/2600hz/kazoo/blob/master/core/whistle_number_manager-1.0.0/src/wnm_number.erl#L1169-L1197)
It caches requests but never invalidates them!
I'm working on this now

Update1:
* `whapps_config:flush().` changed nothing.
* However hitting save to bump the account's `phone_numbers` revision did update the result of the following GET request

```
curl -s -H "Accept: application/json" -H "Content-Type: application/json" -H "X-Auth-Token: 428f16de97c5ef91808b5c6cc5f18fae" -X GET http://10.26.0.141:8000/v1/accounts/26528f16f78de7b6b63822b94e266434/phone_numbers/14157125234 | pj
```

thus updating the content of the first GET request (the one with a phone number specified).

Update 2:
Ah! The invalidation issue is not with the process dictionary mentioned earlier as this caching is actually *not useful*. It's a new process executing this code every time!
I'm not sure it can be removed though, just 1 other function calls this code and I haven't looked into it much.

Oh and the bumping trick I pulled earlier does not work every time.
Still looking…

### DIFF
--- a/core/whistle_number_manager-1.0.0/src/wnm_number.erl
+++ b/core/whistle_number_manager-1.0.0/src/wnm_number.erl
@@ -1186,6 +1186,7 @@ load_phone_number_doc(Account, 'false') ->
         {'ok', J} ->
             lager:debug("loaded phone_numbers from ~s", [AccountId]),
             JObj = wh_json:set_values(PVTs, J),
+            lager:debug(">>> AccountId = ~p  JObj = ~p", [AccountId,JObj]),
             erlang:put({'phone_number_doc', AccountId}, JObj),
             {'ok', JObj};
         {'error', 'not_found'} ->
--- a/core/whistle_number_manager-1.0.0/src/wnm_number.erl
+++ b/core/whistle_number_manager-1.0.0/src/wnm_number.erl
@@ -150,7 +150,7 @@ get(Number) -> get(Number, 'undefined').
 get(Number, PublicFields) ->
     Num = wnm_util:normalize_number(Number),
     Routines = [fun check_reconcilable/1
-                ,fun(#number{}=N) -> R=fetch_number(N, PublicFields), lager:debug(">>> fetch_number -> ~p",[R]), R end
+                ,fun(#number{}=N) -> fetch_number(N, PublicFields) end
                 ,fun maybe_add_information/1
                ],
     lists:foldl(fun(F, N) -> F(N) end
@@ -1182,7 +1182,6 @@ load_phone_number_doc(Account, 'false') ->
         {'ok', J} ->
             lager:debug("loaded phone_numbers from ~s", [AccountId]),
             JObj = wh_json:set_values(PVTs, J),
-            lager:debug(">>> AccountId = ~p  JObj = ~p", [AccountId,JObj]),
             erlang:put({'phone_number_doc', AccountId}, JObj),
             {'ok', JObj};
         {'error', 'not_found'} ->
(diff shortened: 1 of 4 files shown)


core/whistle_number_manager-1.0.0/src/wnm_number.erl
```diff
@@ -181,36 +181,32 @@ fetch_number(#number{number_db=Db
...
+    case wh_json:get_value(Number, JObj) of
+        'undefined' -> N;
+        NumberInfo ->
+            lager:info("merging doc data into number_doc for number ~s", [Number]),
+            %% This also makes sure used_by field is correct
+            NewNumberDoc = wh_json:merge_jobjs(NumberInfo, NumberDoc),
+            save_number_doc(N#number{number_doc=NewNumberDoc})
     end.
 
+
```
jamesaimonetti: Why??? :(

core/whistle_number_manager-1.0.0/src/wnm_number.erl
```diff
@@ -181,34 +181,29 @@ fetch_number(#number{number_db=Db
...
-            NewNumberDoc = wh_json:set_value(<<"used_by">>, NewUsedBy, NumberDoc),
-            save_number_doc(N#number{used_by=NewUsedBy
-                                     ,number_doc=NewNumberDoc
-                                    })
+                      ,JObj) ->
+    case wh_json:get_value(Number, JObj) of
+        'undefined' -> N;
+        NumberInfo ->
+            lager:info("merging doc data into number_doc for number ~s", [Number]),
+            %% This also makes sure used_by field is correct
```
jamesaimonetti: How does the ensure `used_by` is up to date in the record?

fenollp: The older code was just fetching this field from `JObj` and setting it in `N`'s `number_doc` if it was different, if I'm correct.
So what this is doing now is the same behaviour, plus merging other fields. Private ones get filtered out later on.

PR #782, opened 2015-02-15
KAZOO-3334 - [ERR] switch_core_file.c:149 Invalid file format [prompt]

onnet:
Looks like default prompts are acceptable only if they reside in cache.
This is what solved this problem to me.
Please take a look if it could be helpfull.
Sorry if it is just a problem of my local misconfuguration.

Regards,
Kirill

### DIFF
--- a/core/whistle_media-1.0.0/src/wh_media_map.erl
+++ b/core/whistle_media-1.0.0/src/wh_media_map.erl
@@ -362,7 +362,10 @@ get_map(?WH_MEDIA_DB = Db, PromptId) ->
             #media_map{id=MapId
                        ,account_id=Db
                        ,prompt_id=PromptId
-                       ,languages=wh_json:new()
+                       ,languages=wh_json:set_value(wh_media_util:default_prompt_language()
+                                                    ,<<"/", ?WH_MEDIA_DB/binary, "/", (wh_media_util:default_prompt_language())/binary, "%2F", PromptId/binary>>
+                                                    , wh_json:new()
+                                                   )
                       }
     end;
 get_map(AccountId, PromptId) ->


core/whistle_media-1.0.0/src/wh_media_map.erl
```diff
@@ -362,7 +362,10 @@ get_map(?WH_MEDIA_DB = Db, PromptId) ->
             #media_map{id=MapId
                        ,account_id=Db
                        ,prompt_id=PromptId
-                       ,languages=wh_json:new()
+                       ,languages=wh_json:set_value(wh_media_util:default_prompt_language()
+                                                    ,<<"/", ?WH_MEDIA_DB/binary, "/", (wh_media_util:default_prompt_language())/binary, "%2F", PromptId/binary>>
```
jamesaimonetti: This should probably be `wh_media_util:prompt_path(?WH_MEDIA_DB, PromptId)` instead of building the binary. See if that works and update the pull request if so. Otherwise, this looks good!

jamesaimonetti:
Also, can you desrcibe why you think this code is needed? I can issue, on a fresh VM, the commands to get the prompt and it returns successfully (though you need media_mgr running):

```erlang
1> wh_media_map:prompt_path(<<"10c403f4b6a70e404bf6def333df3ae4">>, <<"agent-resume">>, <<"en-us">>).
<<"/system_media/en-us%2Fagent-resume">>
```
This VM only had media_mgr, crossbar, and sysconf running on startup.

onnet:
Hi James,

My prompt_path gives me 'undefined' and I am falling through to default_prompt_path/2:

Let's take a look at  default_prompt_path/2

default_prompt_path(PromptId, Language) ->
    #media_map{languages=Langs} = get_map(PromptId),
    lager:debug("checking default langs ~p", [default_language_keys(Language)]),
    wh_json:get_first_defined(default_language_keys(Language), Langs).


Langs variable will be empty in any case because of the next definition in get_map/2: #media_map{languages=wh_json:new()}
Because of that, the result of default_prompt_path/2 execution will be also empty forever if nothing found in ets table, which seems odd....

get_map(PromptId) ->
    get_map(?WH_MEDIA_DB, PromptId).

get_map(?WH_MEDIA_DB = Db, PromptId) ->
    MapId = mapping_id(Db, PromptId),
    case ets:lookup(table_id(), MapId) of
        [Map] -> Map;
        [] ->
            #media_map{id=MapId
                       ,account_id=Db
                       ,prompt_id=PromptId
                       ,languages=wh_json:new()
                      }
    end;


In my both installations which are upgraded from ancient 3.0x I dont have the results you have on your new VM, but anyway, if default_prompt_path/2 exists, it should function properly :)

core/whistle_media-1.0.0/src/wh_media_map.erl
onnet: No language in path if I use wh_media_util:prompt_path(?WH_MEDIA_DB, PromptId)

(whistle_apps@kz527.onnet.su)4> wh_media_util:prompt_path(<<"system_media">>, <<"vm-saved">>).                                    
<<"/system_media/vm-saved">>

jamesaimonetti: This is expected from the prompt_path function. Can you try `wh_media_util:get_prompt(<<"vm-saved">>).`? On mine I get `<<"prompt://system_media/vm-saved/en-us">>` which is the Kazoo-encoded prompt path.

For testing what media_mgr should return to ecallmgr, you can use:

```erlang
media_mgr_maintenance:prompt_url(<<"vm-saved">>).
media_mgr_maintenance:prompt_url(<<"vm-saved">>, <<"{ACCOUNT_ID}">>).
media_mgr_maintenance:prompt_url(<<"vm-saved">>, <<"{ACCOUNT_ID}">>, <<"la-ng">>).
```
I'll do a diff between 3.18 and master for the media_mgr app and the whistle_media core lib and see if there's an obvious difference that could account for the different results you and I are seeing.

onnet: Actually, I've got the same problem on 3.18, 3.19 and master (this weekend), so it looks more like a Kazoo misconfiguration instead of code difference. And since this problem appears not ony with installations of mine, probably a kind of failover should be done. May be it should be inserted  into ets also

onnet: It's OK with 
(whistle_apps@kz527.onnet.su)1> wh_media_util:get_prompt(<<"vm-saved">>).
<<"prompt://system_media/vm-saved/en-us">>
(whistle_apps@kz527.onnet.su)2> 

I think we need to start from the point, that prompts could be found only when they are in cache.
It it is absent Kazoo fails to build path

onnet: Or just polish provided crutch and forget about problem :)))))

jamesaimonetti: I think any fix should target `wh_media_map:init_map/0` and its execution, as I would expect that to be the cause of this issue.

The whole point of building the mapping as a cache is to avoid querying the database over and over for a prompt that may not exist (perhaps it was improperly imported, or the code referencing the prompt was typed in wrong). So I don't know that code should 'assume' a prompt exists if it is missing from the ETS table.

onnet: Got you, 
let us  see if there will be more victims of such "misconfiguraton" and decide if we need to spend time for this. If there will be no more complaints, I will debug my setup deeper next weekend and let you know if something will be found.

jamesaimonetti: Thanks @onnet!

onnet:
Hi!

Decided to dig it a bit deeper this weekend as promised. Pulled fresh 3.19. Found that this problem disappeared. No configuration was changed since previous week, so it looks like it also wasn't a problem of my local configuration. Believe this pull request could be closed now.

Thank you for help!

Regards,
Kirill

PR #783, opened 2015-02-17
KAZOO-3283: make so that channels that are destroyed on one media server

fenollp:
do not impact a call on another server with the same Call-ID

Drop the CHANNEL_DESTROY message before it kills the working server
(when the message comes from the non-working server).

### DIFF
--- a/applications/ecallmgr/src/ecallmgr_fs_event_stream.erl
+++ b/applications/ecallmgr/src/ecallmgr_fs_event_stream.erl
@@ -332,6 +332,17 @@ maybe_send_event(<<"loopback::bowout">> = EventName, 'undefined', Props, Node) -
     ResigningUUID = props:get_value(?RESIGNING_UUID, Props),
     put('callid', ResigningUUID),
     maybe_send_event(EventName, ResigningUUID, Props, Node);
+maybe_send_event(<<"CHANNEL_DESTROY">> = EventName, UUID, Props, Node) ->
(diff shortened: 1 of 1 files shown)


applications/ecallmgr/src/ecallmgr_fs_event_stream.erl
```diff
@@ -332,6 +332,17 @@ maybe_send_event(<<"loopback::bowout">> = EventName, 'undefined', Props, Node) -
     ResigningUUID = props:get_value(?RESIGNING_UUID, Props),
     put('callid', ResigningUUID),
     maybe_send_event(EventName, ResigningUUID, Props, Node);
+maybe_send_event(<<"CHANNEL_DESTROY">> = EventName, UUID, Props, Node) ->
+    wh_util:put_callid(UUID),
+    case ecallmgr_fs_channel:node(UUID) of
+        {'ok', Node} ->
+            gproc:send({'p', 'l', ?FS_EVENT_REG_MSG(Node, EventName)}, {'event', [UUID | Props]}),
+            maybe_send_call_event(UUID, Props, Node);
+        {'ok', OtherNode} ->
```
jamesaimonetti: When we only use a Var for logging purposes, we prepend it with '_'. This should be `{'ok', _OtherNode}`.

fenollp:
@jamesaimonetti amended on your comment

PR #786, opened 2015-02-18
KAZOO-3270: stop saving notification template on preview req

macpie:


applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -916,15 +916,18 @@ on_successful_validation(Id, Context) ->
...
         {'success', _} -> Context1;
         {_Status, _Code} ->
             lager:debug("load/merge of ~s failed with ~p / ~p", [_Status, _Code]),
             Context1
     end.
 
--spec handle_missing_account_notification(cb_context:context(), ne_binary()) -> cb_context:context().
-handle_missing_account_notification(Context, Id) ->
+-spec handle_missing_account_notification(cb_context:context(), ne_binary(), wh_proplist()) -> cb_context:context().
+handle_missing_account_notification(Context, Id, [{<<"notifications">>, [Id ,<<"preview">>]}|_]) ->
```
fenollp: Is it really safe to match against this proplist?

macpie: for req_nouns yes

jamesaimonetti: @fenollp Just so you know, ReqNouns will be a list of modules and their args built from the URI. See https://github.com/2600hz/kazoo/blob/master/applications/crossbar/doc/basics.md#resources for how that gets built.

So the head of the ReqNouns list will be the last module/args combo from the URI, hence why we can match the head like this.

PR #788, opened 2015-02-19
KAZOO-2581: fix hangups thresholds & use POST in addition to email for alerts in the hangups, johnny5 and ecallmgr applications.

fenollp:
Fixes KAZOO-3235 & KAZOO-3310 & KAZOO-3311

* Thresholds handling/applicatio had some issues related to keys getting
missed due to not being of binary() type or not upper case.
* Email or POST alerts are configurable (kazoo-wide) in system_config's
notify.system_alert. wh_notify:detailed_alert/3 makes use of these changes.
* JSON POSTing logic is moved to a concurrent worker (notify_system_alert)
and code to notify_util (as email sending code is there too).
* Documentation is in notify's doc/


applications/ecallmgr/src/ecallmgr_call_command.erl
```diff
@@ -981,6 +975,12 @@ stream_over_http(Node, UUID, File, Method, Type, JObj) ->
         'fax' -> send_store_fax_call_event(UUID, Result)
     end.
 
+send_detailed_alert(Type, File, UUID, Node, Reason) ->
```
jamesaimonetti: Missing spec

core/whistle-1.0.0/src/wh_notify.erl
```diff
@@ -132,12 +134,8 @@ transaction(Account, Transaction, ServicePlan) ->
     wh_amqp_worker:cast(Notify, fun wapi_notifications:publish_transaction/1).
 
 -spec system_alert(atom() | string() | binary(), [term()]) -> 'ok'.
--spec system_alert(atom() | string() | binary(), [term()], wh_proplist()) -> 'ok'.
-
 system_alert(Format, Args) ->
-    system_alert(Format, Args, []).
-
-system_alert(Format, Args, Props) ->
+    Props = [],
```
jamesaimonetti: If you aren't going to have an arity/3 version, better to remove Props and `<<"Details">>` entirely (why send an empty JSON object?). That said, I do think having the arity/3 available to send Details is worthwhile to keep.

jamesaimonetti: Ah, I see now, you're using detailed_alert/3 instead. Remove `<<"Details">>` from system_alert/2 then

jamesaimonetti:
Typically one pull request for one JIRA ticket.

fenollp:
Fixed. And sorry about that!

PR #789, opened 2015-02-20
KAZOO-3355: Frontier app

siplabs:
This app has a history of it's own, it was re-written at least twice and renamed :)

This is a alpha-version, request for feedback. Some minor fixes are to be commited soon while we re-test after your feedback and code style changes.

Let's make KAZOO world a safer world :)


siplabs:
relates to 2600hz/kazoo-configs#30

applications/crossbar/doc/access_lists.md
```diff
@@ -0,0 +1,57 @@
...
+
+This URI is used to manipulate the access_lists for the entire account.
+
+#### _GET_ - Fetch account-level access lists:
+
+    curl -v -X GET -H "X-Auth-Token: {AUTH_TOKEN}" http://server:8000/v1/accounts/{ACCOUNT_ID}/access_lists
+
+#### _POST_ - Update account-level access lists:
+
+    curl -v -X POST -H "X-Auth-Token: {AUTH_TOKEN}" -H "Content-Type: application/json" http://server:8000/v1/accounts/{ACCOUNT_ID}/access_lists -d '{"data": {"order": "AD","cidrs": ["127.0.0.3/32"]}}'
```
jamesaimonetti: Why is `"order":"AD"` here when the docs above say "allow,deny" or "deny,allow"?

applications/crossbar/doc/access_lists.md
```diff
@@ -0,0 +1,57 @@
...
+
+Here, `{THINGS}` would be "accounts" or "devices" and `{THING_ID}` would be a device or account id. Let's look at adding access lists to a device.
+
+#### _GET_ - Fetch device-level access lists:
+
+    curl -v -X GET -H "X-Auth-Token: {AUTH_TOKEN}" http://server:8000/v1/accounts/{ACCOUNT_ID}/devices/{DEVICE_ID}/access_lists
+
+#### _POST_ - Update device-level access lists:
+
+    curl -v -X POST -H "X-Auth-Token: {AUTH_TOKEN}" -H "Content-Type: application/json" http://server:8000/v1/accounts/{ACCOUNT_ID}/devices/{DEVICE_ID}/access_lists -d '{"data": {"order": "AD","cidrs": ["127.0.0.3/32"]}}'
```
jamesaimonetti: Same question on `order` here

applications/crossbar/doc/rate_limits.md
```diff
@@ -0,0 +1,124 @@
...
+                "registrations": 2
+                "invites": 4
+                "total_packets": 10
+            }
+        }
+    }
+}
+````
+
+ * device-pevel document contains is one level "higher" and contains only the "device" part
```
jamesaimonetti: Should be 'device-level'

applications/crossbar/priv/couchdb/account/access_lists.json
```diff
@@ -0,0 +1,9 @@
+{
+    "_id": "_design/access_lists"
+    ,"language": "javascript"
+    ,"views": {
+        "crossbar_listing": {
+            "map": "function(doc) { if ((doc.pvt_type != 'account' && doc.pvt_type != 'device') || doc.pvt_deleted || !doc.access_lists) return; var name = doc.realm || doc.sip.username; var type = ''; if (doc.pvt_type == 'account') type = 'realm'; else type = 'device'; emit(name, { 'type': type, 'acls': doc.access_lists }); };"
```
jamesaimonetti: I wouldn't bother limiting `doc.pvt_type` to account and device types. Just check for pvt_deleted or no access_lists. It could be, in the future, that other doc types will want access_lists added.

applications/crossbar/priv/couchdb/account/rate_limits.json
```diff
@@ -0,0 +1,12 @@
+{
+    "_id": "_design/rate_limits"
+    ,"language": "javascript"
+    ,"views": {
+        "crossbar_listing": {
+            "map": "function(doc) {if (doc.pvt_type != 'rate_limits' || doc.pvt_deleted) return;var emit_records = function(Name, Type, Limits) {for (SipMethod in Limits.per_minute) {emit([Name, SipMethod], {'type': Type,'per_minute': Limits.per_minute[SipMethod]})};for (SipMethod in Limits.per_second) {emit([Name, SipMethod], {'type': Type,'per_second': Limits.per_second[SipMethod]})};};if (doc.pvt_owner_type == 'account') {emit_records('device-default-rate-limits', 'device', doc.device);emit_records(doc.pvt_queryname, 'realm', doc.account);}if (doc.pvt_owner_type == 'device') emit_records(doc.pvt_queryname, 'device', doc);};"
+        },
+        "list_by_owner": {
+            "map":"function(doc) {if (doc.pvt_type != 'rate_limits' || doc.pvt_deleted) return; emit(doc.pvt_owner_id, doc._id || doc.id);}"
```
jamesaimonetti: `doc._id` will always be present, so not sure why you have `doc.id`. Also, you will get the ID in the view results, so I don't know that you need to emit it here; you could emit `null` instead.

applications/crossbar/priv/couchdb/schemas/access_lists.json
```diff
@@ -0,0 +1,33 @@
+{
+  "_id": "access_lists",
+  "$schema": "http://json-schema.org/draft-04/schema#",
```
jamesaimonetti: The jesse JSON validator only handles draft-03 schemas. Please change this otherwise jesse won't validate the data using this schema doc.

applications/crossbar/priv/couchdb/schemas/access_lists.json
```diff
@@ -0,0 +1,33 @@
...
+  "type": "object",
+  "required": false,
+  "name": "ACL Entry",
+  "description": "Access Control List entries for device or account",
+  "additionalProperties": false,
```
jamesaimonetti: Crossbar generally allows client requests to append their own data to requests. If I want to store meta-data relevant to my app along side these ACL entries, the `additionalProperties:false` will bar me from doing that. I don't see a reason to restrict this, as you validate the actual properties you're interested in and unknown properties will be unused by the underlying code.

applications/crossbar/priv/couchdb/schemas/access_lists.json
```diff
@@ -0,0 +1,33 @@
...
+    "cidrs": {
+      "type": "array",
+      "required": true,
+      "name": "CIDR",
+      "description": "Classless Inter-Domain Routing IP notation for use on the access lists"
+    },
+    "order":{
+      "type":"string",
+      "enum":[
+        "AD",
```
jamesaimonetti: I don't think it is worth using `AD` and `DA`; it obfuscates the meaning. I support "allow,deny" and "deny,allow" as being acceptable and expected (esp since other software like Apache uses it).

applications/crossbar/priv/couchdb/schemas/account_rate_limits.json
```diff
@@ -0,0 +1,137 @@
+{
+  "_id": "account_rate_limits",
+  "$schema": "http://json-schema.org/draft-04/schema#",
```
jamesaimonetti: Change to draft-03

applications/crossbar/priv/couchdb/schemas/account_rate_limits.json
```diff
@@ -0,0 +1,137 @@
...
+  "type": "object",
+  "required": true,
+  "name": "Rate limits",
+  "description": "describe limits for SIP requests",
+  "additionalProperties": false,
```
jamesaimonetti: Remove

applications/crossbar/priv/couchdb/schemas/account_rate_limits.json
```diff
@@ -0,0 +1,137 @@
...
+      "additionalProperties": false,
+      "properties": {
+        "per_minute": {
+          "type": "object",
+          "required": true,
+          "name": "Per minute",
+          "description": "The limits to apply by a minute",
+          "additionalProperties": false,
+          "properties": {
+            "invites": {
```
jamesaimonetti: Take a look at the work to use nestable schemas. You could create a schema for the properties of "per_minute" and "per_second" and include them here and use the "$ref":"sub-schema-name". See the "metaflows" key in the `accounts` schema.

applications/crossbar/src/modules/cb_access_lists.erl
```diff
@@ -0,0 +1,186 @@
...
+                                    cb_context:context().
+-spec validate_set_acls(cb_context:context(), wh_json:object(), api_object()) ->
+                                    cb_context:context().
+validate_set_acls(Context) ->
+    lager:debug("access lists data is valid, setting on thing"),
+    validate_set_acls(Context, cb_context:doc(Context), thing_doc(Context)).
+
+validate_set_acls(Context, _, 'undefined') ->
+    lager:debug("no doc found"),
+    crossbar_util:response_faulty_request(Context);
```
jamesaimonetti: Should probably return a 404 since Thing is not found. See `cb_context:add_system_error('bad_identifier', JObj, Context)` in places like `cb_devices_v2`.

applications/crossbar/src/modules/cb_rate_limits.erl
```diff
@@ -0,0 +1,231 @@
...
+%%    /rate_limits/foo/bar => [<<"foo">>, <<"bar">>]
+%% @end
+%%--------------------------------------------------------------------
+-spec resource_exists() -> 'true'.
+resource_exists() -> 'true'.
+
+-spec only_if_reseller_or_master(cb_context:context()
+                                 ,fun((cb_context:context()) -> cb_context:context())
+                                ) -> cb_context:context().
+only_if_reseller_or_master(Context, Continue) ->
```
jamesaimonetti: I think this should be part of an `authorize` callback. Check against req_nouns to see if it should authorize, then do this check.

applications/crossbar/src/modules/cb_rate_limits.erl
```diff
@@ -0,0 +1,231 @@
...
+    case cb_context:req_nouns(Context) of
+        [{<<"rate_limits">>, []}, {Type, [_ThingId]} | _] ->
+            Type;
+        _Nouns -> 'undefined'
+    end.
+
+-spec thing_id(cb_context:context()) -> api_binary().
+thing_id(Context) ->
+    case cb_context:req_nouns(Context) of
+        [{<<"rate_limits">>, []}, {<<"accounts">>, [AccountId]} | _] ->
```
jamesaimonetti: Why in `thing_type/1` do you use Type and _ThingId to match any preceding noun while here you are explicit about accounts/devices? I'd do something like:

```erlang
thing_type(Context) ->
    case thing_type_id(Context) of
        {Type, _Id} -> Type;
        'undefined' -> 'undefined'
end.

thing_id(Context) ->
    case thing_type_id(Context) of
        {_Type, Id} -> Id;
        'undefined' -> 'undefined'
    end.

thing_type_id(Context) ->
    case cb_context:req_nouns(Context) of
        [{<<"rate_limits">>, []}, {<<"accounts">> = Type, [Id]} | _] -> {Type, Id};
        [{<<"rate_limits">>, []}, {<<"devices">> = Type, [Id]} | _] -> {Type, Id};
        _ReqNouns -> 'undefined'
    end.
```
Now, if rate_limits gets extended to users or carriers or whatever else, its one place to change to permit this.

applications/frontier/Makefile
```diff
@@ -0,0 +1,57 @@
+PROJECT = frontier
+ROOT = ../..
+REBAR = $(ROOT)/utils/rebar/rebar
+DIALYZER = dialyzer
+
+EBINS = $(shell find $(ROOT)/core -maxdepth 2 -name ebin -print) $(shell find $(ROOT)/deps -maxdepth 2 -name ebin -print)
```
jamesaimonetti: See app Makefiles like Crossbar; they've been re-worked to be a bit more efficient in including ebin/ paths that are actually necessary for compilation (rather than everything like this does).

applications/frontier/src/frontier_ensure_rates_in_syscfg.erl
```diff
@@ -0,0 +1,59 @@
...
+                                      ,{?SECOND, DeviceSec}
+                                     ]),
+
+    wh_json:from_list([{<<"account">>, AccountRates}
+                       ,{<<"device">>, DeviceRates}
+                      ]).
+
+-spec start_link() -> startlink_ret().
+start_link() ->
+    gen_server:start_link({'local', ?MODULE}, ?MODULE, [], []).
```
jamesaimonetti: I think start_link() can return 'ignore' instead of calling `gen_server:start_link/4`.

Typically we have a `app_init.erl` (like acdc_init.erl) that will spawn a process to do app initialization stuff (and returning 'ignore' to the supervisor). You may want a frontier_init.erl instead, and have it call to here to init system_config. Then you can also add frontier_maintenance functions to set/update these settings.

applications/frontier/src/frontier_handle_acl.erl
```diff
@@ -0,0 +1,129 @@
...
+    end.
+
+-spec lookup_acl_records(ne_binary()) -> wh_json:objects().
+lookup_acl_records(Entity) ->
+    lookup_acl_records(Entity, 'true').
+
+-spec run_acl_query(ne_binary(), boolean()) -> wh_json:objects().
+run_acl_query(Entity, IncludeRealm) ->
+    Keys = binary:split(Entity, <<"@">>),
+    ViewOpts = case Keys of
```
jamesaimonetti: Put this in a function `build_view_options(Entity, IncludeRealm)` and do the splitting of Entity in there.

applications/frontier/src/frontier_handle_rate.erl
```diff
@@ -0,0 +1,329 @@
...
+        'false' -> construct_records(Method, Realm, PerMinute, PerSecond)
+    end.
+
+-type status() :: 'ok' | 'need_account'.
+-type rates_ret() :: {status(), wh_json:objects()}.
+-spec handle_db_response(wh_json:objects(), boolean()) -> rates_ret().
+handle_db_response(JObjs, IncludeRealm) ->
+    {DefaultRates, OtherRates} = lists:partition(fun is_device_defaults/1, JObjs),
+    {AccountRates, DeviceRates} = lists:partition(fun frontier_utils:is_realm/1, OtherRates),
+    DeviceResult = case length(DeviceRates) > 0 of
```
jamesaimonetti: Getting the length of a list just to see if it is empty or not isn't a great thing to do, in general. I would change this to be more like:

```erlang
case DeviceRates of
    [] -> DefaultRates;
    _ -> DeviceRates
end
```
Still testing emptiness, but without having to count the length of the list (afaik Erlang doesn't track the length and has to compute it when `length/1` is called).

applications/frontier/src/frontier_handle_rate.erl
```diff
@@ -0,0 +1,329 @@
...
+                       'true' ->
+                           lager:info("Found rates in the device doc"),
+                           DeviceRates;
+                       'false' ->
+                           lager:info("Found default rates for the device"),
+                           DefaultRates
+                   end,
+    Status = case IncludeRealm andalso length(AccountRates) =:= 0 of
```
jamesaimonetti: I would suggest:
```erlang
Status = case IncludeRealm andalso AccountRates =:= [] of
    'true' -> {'need_account', DeviceResult};
    'false' -> {'ok', AccountRates ++ DeviceResult}
end.
```

applications/frontier/src/frontier_maintenance.erl
```diff
@@ -0,0 +1,46 @@
...
+%%% @contributors
+%%%   SIPLABS, LLC (Maksim Krzhemenevskiy)
+%%%-------------------------------------------------------------------
+-module(frontier_maintenance).
+
+-include("frontier.hrl").
+
+%% API
+-export([lookup_acls/1
+         ,lookup_ratelimits/1
```
jamesaimonetti: Suggest `lookup_rate_limits/1`, reads better.

Would also suggest adding setters for updating system_config.

applications/frontier/src/frontier_sup.erl
```diff
@@ -0,0 +1,58 @@
...
+
+-behaviour(supervisor).
+
+-export([start_link/0]).
+-export([init/1]).
+
+-include("frontier.hrl").
+
+%% Helper macro for declaring children of supervisor
+-define(CHILDREN, [?WORKER('frontier_ensure_rates_in_syscfg')
```
jamesaimonetti: As I said in the module, I'd have a generic `frontier_init` module that spawns a process and returns 'ignore' instead.

applications/frontier/src/wapi_frontier.erl
```diff
@@ -0,0 +1,118 @@
...
+-define(OPTIONAL_RESP_HEADERS, [<<"Realm">>, <<"Device">>]).
+
+
+-define(RATELIMITS_REQ_VALUES, [{<<"Event-Category">>, <<"rate_limit">>}
+                                ,{<<"Event-Name">>, <<"query">>}
+                               ]).
+-define(RATELIMITS_RESP_VALUES, [{<<"Event-Category">>, <<"rate_limit">>}
+                                ,{<<"Event-Name">>, <<"query_resp">>}
+                               ]).
+-define(RATELIMITS_RESP_TYPES, [{<<"Realm">>, fun(V) -> wh_json:is_json_object(V) end}
```
jamesaimonetti: No need to create an anon function:

```erlang
{<<"Realm">>, fun wh_json:is_json_object/1}
```

jamesaimonetti:
@siplabs Overall, first pass looked good! Mostly smaller things to improve/change. Hopefully we'll get you some feedback on the Kamailio side, get changes made here (please be sure to keep rebasing master into this until it is merged), and get Frontier into master for everyone to play with!

siplabs:
@jamesaimonetti, it's ready for second pass.

applications/crossbar/src/modules/cb_modules_util.erl
```diff
@@ -75,6 +77,14 @@ range_view_options(Context, MaxRange) ->
         _N -> {CreatedFrom, CreatedTo}
     end.
 
+-spec maybe_clear_outbound_flags(cb_context:context()) -> cb_context:context().
```
jamesaimonetti: What is the purpose of this function? Why, when a device fails schema validation, would you need to clear these flags? This is unexpected behaviour and I don't think we can accept this change (and the cb_devices_* changes) unless there's a significant issue we don't realize.

Please enlighten! :)

applications/frontier/src/frontier_handle_rate.erl
```diff
@@ -0,0 +1,330 @@
...
+                    'false' -> Realm
+                end,
+    case IncludeRealm andalso Realm =/= Entity of
+        'true' -> [Realm, Username];
+        'false' -> [QueryName]
+    end.
+
+
+-spec run_rate_limits_query(ne_binary(), ne_binary(), boolean(), ne_binary()) -> wh_json:objects().
+run_rate_limits_query(Entity, AccountDB, IncludeRealm, MethodList) ->
```
jamesaimonetti: No queries are being run, that I can see. Perhaps this should be named something more like 'build_rates' or something more description?

applications/frontier/src/frontier_handle_rate.erl
```diff
@@ -0,0 +1,330 @@
...
+                            fetch_rates_from_sys_config(Realm, <<"realm">>, MethodList)
+                            ++ fetch_rates_from_sys_config(Entity, <<"device">>, MethodList);
+                        'false' ->
+                            Type = frontier_utils:get_entity_type(Entity),
+                            fetch_rates_from_sys_config(Entity, Type, MethodList)
+                    end,
+    Rates ++ FromSysCOnfig.
+
+-spec to_first_upper(ne_binary()) -> ne_binary().
+to_first_upper(<<L, Ls/binary>>) ->
```
jamesaimonetti: see `wh_util:ucfirst_binary/1`

applications/frontier/src/frontier_handle_rate.erl
```diff
@@ -0,0 +1,330 @@
...
+
+-spec key_join(ne_binary(), ne_binary()) -> ne_binary().
+key_join(X, Acc) ->
+    list_to_binary([Acc, "-", X]).
+
+-spec to_json_key(ne_binary()) -> ne_binary().
+to_json_key(Token) ->
+    Tokens = binary:split(Token, <<"_">>),
+    [H | Tail] = lists:map(fun to_first_upper/1, Tokens),
+    lists:foldl(fun key_join/2, H, Tail).
```
jamesaimonetti: Could this function be written:

```erlang
to_json_key(Token) ->
    Tokens = binary:split(Token, <<"_">>),
    wh_util:join_binary([wh_util:ucfirst_binary(Token) || Token <- Tokens], <<"-">>).
```

applications/frontier/src/frontier_handle_rate.erl
```diff
@@ -0,0 +1,330 @@
...
+                          'true' -> Acc;
+                          'false' -> construct_records(Method, Entity, RPM, RPS) ++ Acc
+                      end
+                 end, [], MethodList).
+
+-spec fetch_rates(ne_binary(), boolean(), ne_binaries(), ne_binary()) -> wh_json:objects().
+fetch_rates(_, _, [], _) ->
+    lager:info("document: Empty request - empty response"),
+    [];
+fetch_rates([], _, _, _) ->
```
jamesaimonetti: Dialyzer reports this clause as being uncallable and I agree. I only see this arg being filled as a result of a call to `build_list_of_querynames/2` which always returns a non-empty list.

jamesaimonetti:
@siplabs I opened pull request #812 with most of the changes I want (including removing the device outbound flag removal). Would be great if you could look it over, try it out, and see if anything changed unexpectedly. Please issue pull requests against this branch if there are changes you'd like to propose.

I think there's minimal work left to get this accepted, so hopefully you can give the OK and we'll merge this in!

PR #790, opened 2015-02-20
KAZOO-3250: callflow app callflow submodule does not to allow a call to ...

macpie:
...be transferred to another callflow if the call is hungup

### DIFF
--- a/applications/callflow/src/module/cf_callflow.erl
+++ b/applications/callflow/src/module/cf_callflow.erl
@@ -32,11 +32,17 @@ handle(Data, Call) ->
 maybe_branch_callflow(Data, Call) ->
     Id = wh_json:get_value(<<"id">>, Data),
     case couch_mgr:open_doc(whapps_call:account_db(Call), Id) of
-        {'ok', JObj} ->
-            lager:info("branching to callflow ~s", [Id]),
-            Flow = wh_json:get_value(<<"flow">>, JObj, wh_json:new()),
-            cf_exe:branch(Flow, Call);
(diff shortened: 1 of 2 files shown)


lazedo:
this seems to introduce overhead between callflow steps and it will impact overall perfomance

macpie:
This is why it is done in the callflow module so that it will only impact callflows with callflow. Otherwise in some cases it could enter into a tight loop.

applications/callflow/src/module/cf_callflow.erl
```diff
@@ -32,11 +32,17 @@ handle(Data, Call) ->
...
-        {'ok', JObj} ->
-            lager:info("branching to callflow ~s", [Id]),
-            Flow = wh_json:get_value(<<"flow">>, JObj, wh_json:new()),
-            cf_exe:branch(Flow, Call);
         {'error', R} ->
             lager:info("could not branch to callflow ~s, ~p", [Id, R]),
-            cf_exe:continue(Call)
+            cf_exe:continue(Call);
+        {'ok', JObj} ->
+            case whapps_call_command:channel_status(Call) of
```
jamesaimonetti: I think this needs to be `b_channel_status`. You'll only get the error tuple if the Call has no call_id set in it. b_channel_status will actually query and wait for the response.

lazedo:
@macpie please ammend this. cf_voicemail uses branch and that may introduce delays in voicemail handling.

maybe create a branch/3 to allow actions to override this behavior. then change in (at least) cf_voicemail.

PR #797, opened 2015-02-24
Gateway dst/src port

Dinkor:
Added fields "port" and "force_port" in the documents of resources.
At outgoing call field "port" is added to the SIP URI.
At incoming call (if "force_port" = true), the source port is compared with the field of "port".
This allows at one IP-address have multiple gateways separated on different ports.

To work correctly, also requires changes in the configuration file kamailio.
File "default.cfg":
```
route[EXTERNAL_TO_INTERNAL_RELAY]
...
    remove_hf_re("^X-.*");
    append_hf("X-AUTH-IP: $si\r\n");
    append_hf("X-AUTH-PORT: $sp\r\n"); <--- add this!!!

```

### DIFF
--- a/applications/ecallmgr/src/ecallmgr_util.erl
+++ b/applications/ecallmgr/src/ecallmgr_util.erl
@@ -17,7 +17,7 @@
 -export([export/3, bridge_export/3]).
 -export([get_expires/1]).
 -export([get_interface_properties/1, get_interface_properties/2]).
--export([get_sip_to/1, get_sip_from/1, get_sip_request/1, get_orig_ip/1]).
+-export([get_sip_to/1, get_sip_from/1, get_sip_request/1, get_orig_ip/1, get_orig_port/1]).
 -export([custom_channel_vars/1, custom_channel_vars/2]).
 -export([eventstr_to_proplist/1, varstr_to_proplist/1, get_setting/1, get_setting/2]).
 -export([is_node_up/1, is_node_up/2]).
@@ -210,6 +210,10 @@ get_sip_request(Props) ->
 -spec get_orig_ip(wh_proplist()) -> api_binary().
 get_orig_ip(Prop) ->
     props:get_first_defined([<<"X-AUTH-IP">>, <<"ip">>], Prop).
+
+-spec get_orig_port(wh_proplist()) -> api_binary().
+get_orig_port(Prop) ->
+    props:get_first_defined([<<"X-AUTH-PORT">>, <<"port">>], Prop).
 
 %% Extract custom channel variables to include in the event
 -spec custom_channel_vars(wh_proplist()) -> wh_proplist().
--- a/applications/ecallmgr/src/ecallmgr_util.erl
+++ b/applications/ecallmgr/src/ecallmgr_util.erl
@@ -211,9 +211,12 @@ get_sip_request(Props) ->
 get_orig_ip(Prop) ->
     props:get_first_defined([<<"X-AUTH-IP">>, <<"ip">>], Prop).
 
--spec get_orig_port(wh_proplist()) -> api_binary().
+-spec get_orig_port(wh_proplist()) -> api_integer().
 get_orig_port(Prop) ->
-    props:get_first_defined([<<"X-AUTH-PORT">>, <<"port">>], Prop).
+    case props:get_first_defined([<<"X-AUTH-PORT">>, <<"port">>], Prop) of
+        'undefined' -> 'undefined';
+        Port -> wh_util:to_integer(Port)
+    end.
 
 %% Extract custom channel variables to include in the event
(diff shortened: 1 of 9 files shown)


lazedo:
@Dinkor Hi, maybe i'm missing something obvious but, what is this trying to solve ? for outgoing you just put the port in the uri right ? what does it solve on inbound ?
Best

Dinkor:
Sorry for unclear description.
For example we have two global offnet resources:
```json
{
   "_id": "Offnet-E1-1",
   "gateways": [
       {
           "server": "192.168.1.1",
           "realm": "192.168.1.1",
           "port": "5071",
           "force_port": true
       }
   ],
   "pvt_type": "resource"
}
```
```json
{
   "_id": "Offnet-E1-2",
   "gateways": [
       {
           "server": "192.168.1.1",
           "realm": "192.168.1.1",
           "port": "5072",
           "force_port": true
       }
   ],
   "pvt_type": "resource"
}
```
Both of this resources point to one PSTN-gateway (192.168.1.1), which connected to different operators (E1). This GW route calls, based on src/dst-port and ip to appropriate E1 and vice versa.
 In this setup we can distinguish two calls to same number coming through different E1, and CDR have "Offnet-E1-1" or "Offnet-E1-2" in "resource_id".

applications/stepswitch/src/stepswitch_resources.erl
```diff
@@ -778,6 +785,7 @@ gateway_from_jobj(JObj, #resrc{is_emergency=IsEmergency
              ,fax_option=wh_json:is_true([<<"media">>, <<"fax_option">>], JObj, T38)
              ,codecs=wh_json:get_value(<<"codecs">>, JObj, Codecs)
              ,bypass_media=wh_json:is_true(<<"bypass_media">>, JObj, BypassMedia)
+             ,force_port=wh_json:get_value(<<"force_port">>, JObj, false)
```
jamesaimonetti: This should be `wh_json:is_true/3` to ensure a 'true' or 'false' atom.

applications/stepswitch/src/stepswitch_resources.erl
```diff
@@ -766,6 +772,7 @@ gateway_from_jobj(JObj, #resrc{is_emergency=IsEmergency
     EndpointType = wh_json:get_ne_value(<<"endpoint_type">>, JObj, <<"sip">>),
     #gateway{endpoint_type=EndpointType
              ,server=wh_json:get_value(<<"server">>, JObj)
+             ,port=wh_json:get_value(<<"port">>, JObj)
```
jamesaimonetti: Shouldn't this use `wh_json:get_integer_value/2` ? Will return either the Port as an integer or 'undefined'.

applications/stepswitch/src/stepswitch_resources.erl
```diff
@@ -195,11 +197,14 @@ reverse_lookup(JObj) ->
     IP = wh_json:get_first_defined([<<"From-Network-Addr">>
                                     ,<<"Orig-IP">>
                                    ], JObj),
-    case maybe_find_global(IP, Realm) of
+    Port = wh_json:get_first_defined([<<"From-Network-Port">>
```
jamesaimonetti: Should put this in a function to `find_port(JObj)` that checks for undefined and casts to integer if not undefined.

applications/stepswitch/src/stepswitch_resources.erl
```diff
@@ -21,6 +21,7 @@
 
 -record(gateway, {
            server :: api_binary()
+          ,port :: api_binary()
```
jamesaimonetti: Should be api_integer()

applications/ecallmgr/src/ecallmgr_util.erl
```diff
@@ -235,6 +235,10 @@ get_sip_request(Props) ->
 get_orig_ip(Prop) ->
     props:get_first_defined([<<"X-AUTH-IP">>, <<"ip">>], Prop).
 
+-spec get_orig_port(wh_proplist()) -> api_binary().
+get_orig_port(Prop) ->
+    props:get_first_defined([<<"X-AUTH-PORT">>, <<"port">>], Prop).
```
jamesaimonetti: Should check for 'undefined', and cast to integer if not 'undefined'

lazedo:
``` "In this setup we can distinguish two calls to same number coming through different E1, and CDR have "Offnet-E1-1" or "Offnet-E1-2" in "resource_id".
```

when would the same number be routed from "Offnet-E1-1" or "Offnet-E1-2". shouldn't it be routed from the number provider ? i really miss to see the use case here.

Dinkor:
In our case - we are number provider. We do not buy these numbers from anyone else.

PR #798, opened 2015-02-24
KAZOO-3358: User external caller id instead of internal for user views

macpie:


macpie:
I change "internal" by "external" for the caller id feature on the view

PR #808, opened 2015-03-04
KAZOO-3382: trunkstore: skip route requests for bridged calls

siplabs:
Trunksore sometimes steal call processing from callflow. Example: REFER-ed calls.

### DIFF
--- a/applications/trunkstore/src/ts_callflow.erl
+++ b/applications/trunkstore/src/ts_callflow.erl
@@ -40,11 +40,18 @@
 
 -type ts_state() :: #ts_callflow_state{}.
 
+-spec is_bridged(wh_json:object()) -> boolean().
+is_bridged(JObj) ->
+    case wh_json:get_value([<<"Custom-Channel-Vars">>, <<"Bridge-ID">>], JObj) of
+        'undefined' -> 'false';
+        _BridgeId -> 'true'
+    end.
+
 -spec init(wh_json:object()) -> ts_state() | {'error', 'not_ts_account'}.
 init(RouteReqJObj) ->
     CallID = wh_json:get_value(<<"Call-ID">>, RouteReqJObj),
     put('callid', CallID),
-    case is_trunkstore_acct(RouteReqJObj) of
+    case is_trunkstore_acct(RouteReqJObj) andalso not is_bridged(RouteReqJObj) of
         'false' ->
             lager:info("request is not for a trunkstore account"),
             {'error', 'not_ts_account'};
--- a/applications/trunkstore/src/ts_callflow.erl
+++ b/applications/trunkstore/src/ts_callflow.erl
@@ -5,6 +5,8 @@
 %%% @end
 %%% @contributors
 %%%   James Aimonetti
+%%%
+%%% Fix KAZOO-3382 sponsored by Conversant, Ltd. Implemented by SIPLABS, LLC.
 %%%-------------------------------------------------------------------
 -module(ts_callflow).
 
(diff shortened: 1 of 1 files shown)


applications/trunkstore/src/ts_callflow.erl
```diff
@@ -40,11 +40,18 @@
 
 -type ts_state() :: #ts_callflow_state{}.
 
+-spec is_bridged(wh_json:object()) -> boolean().
+is_bridged(JObj) ->
+    case wh_json:get_value([<<"Custom-Channel-Vars">>, <<"Bridge-ID">>], JObj) of
```
jamesaimonetti: Why not just use `=/= 'undefined'` instead of the case?

jamesaimonetti:
@siplabs Can you provide a more concrete scenario for showing the undesired behaviour? It seems if you have a call for a number configured in trunkstore, that trunkstore should process it, no?

siplabs:
@jamesaimonetti 

One of our clients stepped over this bug:
they have a SIP TAPI adaptor that makes outbound calls by sending INVITE and then REFER-ing to the external mobile number. First time callflow wins and everything's fine. But when making second and further attempts it doesn't work because trunkstore intercepts Route Wins. 
NB: First Route Req contains Auth-ID and Auth-Type==device, second and further Route Reqs (initiated by REFER) have no Auth-Type and that's why trunkstore takes it

lazedo:
@siplabs it would be preferable to fix the SIP/TAPI adaptor

siplabs:
Well, the application is well known - (http://www.ipcom.at/en/telephony/siptapi/). We can't say it's misbehaving and it definitely looks like KAZOO bug. Trunkstore steals Route Wins from callflow

lazedo:
as you said the REFER is not authenticated, if trunkstore is "stealing" as you said, is because it can handle it, and should it handle it ? maybe the REFER is not well formed ?

jamesaimonetti:
@siplabs I agree with @lazedo that if trunkstore tries to route a call, it is because it has been configured to do so. Also, there is no way to "steal" a route_win message; it is sent to a targeted queue (one consumer). It would be great to see the SIP packets and the logs related to the issue put on the JIRA ticket.

siplabs:
If trunkstore could handle it - it would and there was no bug.
The situation looks like the following:

I do the first call and both (from the two) RouteReqs are being served by callflow.
With next and further calls trunkstore intercepts second RouteReq (from the two)

This is (at minimum) inconsistant behaviour. And looks very much like race condition.

Both places [here](https://github.com/2600hz/kazoo/blob/master/applications/callflow/src/cf_route_req.erl#L159) and [here](https://github.com/2600hz/kazoo/blob/master/applications/trunkstore/src/ts_callflow.erl#L263) say both think they have right to handle RouteReq.

I see two ways out of this complicated situation:

1) Forbid trunkstore to handle already-bridged calls (why should it finally!)
2) Fix the code so second RouteReq has Auth-Type together with Auth-ID

Both ways resolve the race condition

Maybe we are missing something here and there are other ways.

P.S. If we reload the whistle apps node first call goes ok despite SIPTAPI. So we don't think we should fix it for any reason - it behaves as expected and violates nothing.

lazedo:
i don't agree. 

the REFER sent from SIP/TAPI must be wrong orelse we would have serious problems with call transfer . does it have a From-Tag, To-Tag, Refer-To , Referred-By,  Replaces ?

siplabs:
@lazedo, @jamesaimonetti 
Attached everything to JIRA ticket.

jamesaimonetti:
Please try https://github.com/2600hz/kazoo/pull/828 instead and see if that solves this issue (and maintains trunkstore's functionality).

siplabs:
@jamesaimonetti 
Works well for my case! Very thanks! :)

PR #810, opened 2015-03-04
KAZOO-3380: Simplify pivot debug

macpie:


applications/pivot/src/pivot_call.erl
```diff
@@ -148,9 +150,11 @@ init([Call, JObj]) ->
 
     ?MODULE:new_request(self(), VoiceUri, Method, BaseParams),
 
+    Counter = whapps_call:kvs_fetch('pivot_counter', 1, Call),
+
     {'ok'
      ,#state{cdr_uri=wh_json:get_value(<<"CDR-URI">>, JObj)
-             ,call=Call
+             ,call=whapps_call:kvs_update_counter('pivot_counter', Counter, Call)
```
jamesaimonetti: `Counter` should be `1`, as that parameter is the Increment to add, rather than setting the value to `Counter`.

Meaning, as you have it now, `pivot_counter` would be 1, 2, 4, 8, 16,...

applications/crossbar/priv/couchdb/account/pivot.json
```diff
@@ -6,7 +6,10 @@
             "map": "function(doc) { if(doc.pvt_type != 'pivot' || doc.pvt_deleted) return; emit(doc._id, null);}"
         },
         "debug_listing": {
-            "map": "function(doc) {if (doc.pvt_type != 'pivot_debug' || doc.pvt_deleted) return; emit([doc.call_id, doc.pvt_created], {'call_id': doc.call_id, 'created': doc.pvt_created,'uri': doc.uri,'status_code': doc.resp_status_code,'node': doc.pvt_node});}"
+            "map": "function(doc) {if (doc.pvt_type != 'pivot_debug' || doc.pvt_deleted) return; emit([doc.call_id, doc.pvt_created], null);}"
+        },
+        "first_iteration": {
```
jamesaimonetti: Should this be in the account's pivot view? I think its MODB-only, no?

macpie: the views where both the same so I was thinking of keeping it that way.

PR #814, opened 2015-03-09
It seems like an inconsistency that the only storage of owner_id is in k...

aberke:
...vs for doodle.  This means we can't use the whapps_call:owner_id function in the same way as callflow.

SMS documents only have owner_id/Owner-Id values in the pvt_call['Key-Value-Store']


aberke:
This is a PR for the `3.19` branch

lazedo:
"When I look at my saved SMS docs on staging the owner_id is not set in the call object."
are these received or sent ? if sent, were they sent by app/api/device ?
thanks

aberke:
they were sent via device

aberke:
repoened!

k-anderson:
As per our conversation we determined the issue was related to ecallmgr_util using the wrong keys for the view results.  We addressed this in #888.  Thanks!

PR #815, opened 2015-03-09
KAZOO-3054: Add a flag in the "new user" api to send an email on user cr...

macpie:
...eation

### DIFF
--- a/applications/crossbar/src/modules_v2/cb_users_v2.erl
+++ b/applications/crossbar/src/modules_v2/cb_users_v2.erl
@@ -236,7 +236,7 @@ post(Context, _) ->
 
 -spec put(cb_context:context()) -> cb_context:context().
 put(Context) ->
-    DryRun = not(cb_context:accepting_charges(Context)),
+    DryRun = (not wh_json:is_true(<<"accept_charges">>, cb_context:req_json(Context), 'false')),
     put_resp(DryRun, Context).
 
 put_resp('true', Context) ->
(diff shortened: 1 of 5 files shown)


applications/crossbar/src/modules_v2/cb_users_v2.erl
```diff
@@ -248,7 +248,42 @@ put_resp('true', Context) ->
...
-    crossbar_doc:save(Context).
+    Context1 = crossbar_doc:save(Context),
+    case cb_context:resp_status(Context1) of
+        'success' ->
+            _ = maybe_send_email(Context1),
+            Context1;
+        _ -> Context1
+    end.
+
+maybe_send_email(Context) ->
```
jamesaimonetti: spec

applications/crossbar/src/modules_v2/cb_users_v2.erl
```diff
@@ -248,7 +248,42 @@ put_resp('true', Context) ->
...
+    ReqJObj = cb_context:req_data(Context),
+    case wh_json:is_true(<<"send_email_on_creation">>, ReqJObj, 'true') of
+        'false' -> 'ok';
+        'true' -> send_email(Context)
+    end.
+
+send_email(Context) ->
```
jamesaimonetti: spec

applications/teletype/src/templates/teletype_new_user.erl
```diff
@@ -0,0 +1,155 @@
...
+-define(TEMPLATE_MACROS
+        ,wh_json:from_list(
+           [?MACRO_VALUE(<<"user.first_name">>, <<"first_name">>, <<"First Name">>, <<"First Name">>)
+            ,?MACRO_VALUE(<<"user.last_name">>, <<"last_name">>, <<"Last Name">>, <<"Last Name">>)
+            ,?MACRO_VALUE(<<"user.password">>, <<"password">>, <<"Password">>, <<"Password">>)
+            | ?SERVICE_MACROS
+           ])
+       ).
+
+-define(TEMPLATE_TEXT, <<"Welcome {{user.first_name}} {{user.last_name}}.\n\nYour password is : {{user.password}}.\n Please keep it safe and secret.">>).
```
jamesaimonetti: We should probably instruct them to change the password as soon as they can.

jamesaimonetti:
@macpie have you tested to be sure teletype will send this and notify won't, when running at the same time?

PR #816, opened 2015-03-10
KAZOO-3406: ability to record call at X kHz

siplabs:


applications/ecallmgr/src/ecallmgr_call_command.erl
```diff
@@ -180,8 +182,12 @@ get_fs_app(Node, UUID, JObj, <<"record_call">>) ->
             case wh_json:get_value(<<"Record-Action">>, JObj) of
                 <<"start">> ->
                     FollowTransfer = wh_json:get_binary_boolean(<<"Follow-Transfer">>, JObj, <<"true">>),
+                    ConfigSampleRate = ecallmgr_config:get(<<"record_sample_rate">>, ?DEFAULT_SAMPLE_RATE),
+                    SampleRate = wh_json:get_value(<<"Record-Sample-Rate">>, JObj, ConfigSampleRate),
```
jamesaimonetti: You can use `wh_json:get_binary_value/3` which should do the cast for you

core/whistle_media-1.0.0/src/wh_media_recording.erl
```diff
@@ -42,6 +44,7 @@
 
 -record(state, {url                        :: ne_binary()
                 ,format                    :: ne_binary()
+                ,sample_rate = undefined   :: integer() | 'undefined'
```
jamesaimonetti: No need to initialize to 'undefined', that is the default behaviour.

applications/ecallmgr/src/ecallmgr_call_command.erl
```diff
@@ -180,8 +182,12 @@ get_fs_app(Node, UUID, JObj, <<"record_call">>) ->
             case wh_json:get_value(<<"Record-Action">>, JObj) of
                 <<"start">> ->
                     FollowTransfer = wh_json:get_binary_boolean(<<"Follow-Transfer">>, JObj, <<"true">>),
+                    ConfigSampleRate = ecallmgr_config:get_integer(<<"record_sample_rate">>, ?DEFAULT_SAMPLE_RATE),
+                    SampleRate = wh_json:get_integer_value(<<"Record-Sample-Rate">>, JObj, ConfigSampleRate),
```
siplabs: @jamesaimonetti 
If we talk about data types I'd go this way. It helps to be sure that we use appropriate values on every step

jamesaimonetti:
Manually merged, thanks!

PR #817, opened 2015-03-10
Siplabs kazoo 3355

siplabs:
If this method is not good enough, can you suggest something?


jamesaimonetti:
@siplabs As I said in the previous pull request, we can't accept this clearing of outbound flags without some justification from your side. Please explain why you have included it so we can determine if it is indeed useful functionality.

siplabs:
Explained [here](https://github.com/2600hz/kazoo/commit/c5716199caee5d855401dfbff32b6cdf13aab159#diff-8de6c8f403f1c03d5abaf3d4fdecb518L421)

PR #818, opened 2015-03-10
KAZOO-3377

fenollp:
After conversing with @jamesaimonetti I resolved to cleanup `system_media` lines that could be found on older systems [as per this code](https://github.com/2600hz/kazoo/commit/3ca29dbc7a2ac0f806de92949d33c87cf56dd3aa)

After conversing with @k-anderson I made it so `wh_call_response:send*` accepts a `whapps_call` while providing the old interface to `trunkstore` functions.

### DIFF
--- a/core/whistle-1.0.0/src/wh_call_response.erl
+++ b/core/whistle-1.0.0/src/wh_call_response.erl
@@ -12,11 +12,16 @@
 -export([send_default/2]).
 -export([get_response/2]).
 -export([default_response/2]).
+-export([config_doc_id/0]).
 
 -include_lib("whistle/include/wh_types.hrl").
 -include_lib("whistle/include/wh_log.hrl").
 
 -define(CALL_RESPONSE_CONF, <<"call_response">>).
+
+-spec config_doc_id() -> ne_binary().
+config_doc_id() ->
+    ?CALL_RESPONSE_CONF.
 
 %%--------------------------------------------------------------------
 %% @public
--- a/core/whistle-1.0.0/src/wh_call_response.erl
+++ b/core/whistle-1.0.0/src/wh_call_response.erl
@@ -1,5 +1,5 @@
 %%%-------------------------------------------------------------------
-%%% @copyright (C) 2013-2014, 2600Hz
+%%% @copyright (C) 2013-2015, 2600Hz
 %%% @doc
 %%%
 %%% @end
@@ -26,25 +26,23 @@
 %% @end
 %%--------------------------------------------------------------------
 -spec send(ne_binary(), ne_binary(), api_binary()) ->
-                        {'ok', ne_binary()} |
-                        {'error', 'no_response'}.
+                  {'ok', ne_binary()} |
+                  {'error', 'no_response'}.
 -spec send(ne_binary() | whapps_call:call(), ne_binary(), api_binary(), 'undefined' | binary()) ->
-                        {'ok', ne_binary()} |
-                        {'error', 'no_response'}.
+                  {'ok', ne_binary()} |
+                  {'error', 'no_response'}.
 -spec send(ne_binary(), ne_binary(), api_binary(), 'undefined' | binary(), 'undefined' | binary()) ->
-                        {'ok', ne_binary()} |
-                        {'error', 'no_response'}.
+                  {'ok', ne_binary()} |
+                  {'error', 'no_response'}.
 
 send(CallId, CtrlQ, Code) ->
     send(CallId, CtrlQ, Code, 'undefined').
 
-send(CallId, CtrlQ, Code, Cause)
-  when is_binary(CallId) ->
+send(<<_/binary>> = CallId, CtrlQ, Code, Cause) ->
     send(CallId, CtrlQ, Code, Cause, 'undefined');
-
 send(Call, Code, Cause, Media) ->
     CallId = whapps_call:call_id(Call),
-    CtrlQ  = whapps_call:control_queue(Call),
+    CtrlQ = whapps_call:control_queue(Call),
     send(CallId, CtrlQ, Code, Cause, Media).
 
 send(_, _, 'undefined', 'undefined', 'undefined') ->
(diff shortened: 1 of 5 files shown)


lazedo:
@fenollp i believe the intent to use whapps_call was to fetch the language from it and get the prompt for the language (not_available in french) but i don't see it ?

fenollp:
You're right @lazedo but from what I understood from @jamesaimonetti the language would not be picked up because on old configs there are still some `system_media` nonsense.
I guess I got something wrong somewhere though…

fenollp:
My main problem is this:
once I get the prompt for the language, how do I modify `do_send` to use it?
And what do I do with trunkstore calls? they don't have language information…

core/whistle-1.0.0/src/wh_call_response.erl
```diff
@@ -116,18 +122,19 @@ do_send(CallId, CtrlQ, Commands) ->
...
+send_default(Call, Cause) ->
     lager:debug("attempting to send default response for ~s", [Cause]),
-    Response = get_response(Cause),
-    send(CallId, CtrlQ
-         ,wh_json:get_value(<<"Code">>, Response)
-         ,wh_json:get_value(<<"Message">>, Response)
-         ,wh_json:get_value(<<"Media">>, Response)).
+    Response = get_response(Cause, Call),
+    send(cf_exe:callid(Call)
+        ,cf_exe:control_queue(Call)
```
lazedo: cf_exe belongs to callflow. this way we are making core dependent of callflow.
can you use whapps_call  ?

core/whistle_apps-1.0.0/src/whapps_call_command.erl
```diff
@@ -439,9 +439,7 @@ response(Code, Call) ->
 response(Code, Cause, Call) ->
     response(Code, Cause, 'undefined', Call).
 response(Code, Cause, Media, Call) ->
-    CallId = whapps_call:call_id(Call),
-    CtrlQ = whapps_call:control_queue(Call),
-    wh_call_response:send(CallId, CtrlQ, Code, Cause, Media).
+    wh_call_response:send(Call, Code, Cause, Media).
```
jamesaimonetti: Why the reorder of args?

core/whistle_apps-1.0.0/src/whapps_maintenance.erl
```diff
@@ -949,3 +951,27 @@ show_status(CallId, 'true', Resp) ->
     lager:info("Media Server: ~s", [wh_json:get_value(<<"Switch-Hostname">>, Resp)]),
     lager:info("Responding App: ~s", [wh_json:get_value(<<"App-Name">>, Resp)]),
     lager:info("Responding Node: ~s", [wh_json:get_value(<<"Node">>, Resp)]).
+
+
+-spec delete_system_media_references() -> 'no_return'.
+delete_system_media_references() ->
+    Db = <<"media">>,
```
jamesaimonetti: Not a database but a doc id.

core/whistle_apps-1.0.0/src/whapps_maintenance.erl
```diff
@@ -949,3 +951,27 @@ show_status(CallId, 'true', Resp) ->
...
+    {'ok', MediaDoc} = couch_mgr:open_cache_doc(?WH_CONFIG_DB, Db),
+    TheKey = <<"default">>,
+    MediaDefault = wh_json:get_value(TheKey, MediaDoc),
+    Deleter =
+        fun (Key, Value, Acc) ->
+                case starts_with_system_media(Key) orelse
```
jamesaimonetti: We dropline `orelse`, `andalso`, etc, AND we dropline the `of` in multi-line case:

```erlang
case starts_with_system_media(Key)
        orelse starts_with_system_media(Value)
of
    'true' -> Acc;
    'false' -> [{Key, Value} | Acc]
end
```
Also, I don't think you need to check Key in this case.

core/whistle_apps-1.0.0/src/whapps_call_command.erl
fenollp: `Call` replaces `CallId, CtrlQ`. I uh… that's life

fenollp:
I'll have to rebase this before merging

core/whistle-1.0.0/src/wh_call_response.erl
```diff
@@ -38,6 +38,12 @@
 send(CallId, CtrlQ, Code) ->
     send(CallId, CtrlQ, Code, 'undefined').
 
+send(Call, Code, Cause, Media)
+  when element(1, Call) =:= 'whapps_call' ->
```
jamesaimonetti: I'd prefer to test CallId for is_binary/1 and have the second clause be `send(Call,...`

fenollp:
Mergeable

core/whistle_apps-1.0.0/src/whapps_maintenance.erl
```diff
@@ -949,3 +951,20 @@ show_status(CallId, 'true', Resp) ->
...
+delete_system_media_references() ->
+    DocId = <<"media">>,
+    {'ok', MediaDoc} = couch_mgr:open_cache_doc(?WH_CONFIG_DB, DocId),
+    TheKey = <<"default">>,
+    MediaDefault = wh_json:get_value(TheKey, MediaDoc),
+    Deleter =
+        fun (_Key, <<"/system_media/",_/binary>>, Acc) -> Acc;
+            (Key, Value, Acc) -> [{Key, Value} | Acc]
+        end,
+    NewMediaDefault = wh_json:foldl(Deleter, [], MediaDefault),
```
jamesaimonetti: This should be `wh_json:new()` instead of `[]`. `wh_json:set_value/3` should be setting `TheKey` to a JSON object, not a list of Key/Value pairs. `Deleter` should use `wh_json:set_value/3` to set them in the Acc too.

Finally, the code should detect if any changes were made and only save if so.

fenollp: amended

PR #819, opened 2015-03-10
KAZOO-3407 support multiple connection for doodle listener

lazedo:
created supervisor multiple inbound connections

### DIFF
--- a/applications/doodle/src/doodle_inbound_listener_sup.erl
+++ b/applications/doodle/src/doodle_inbound_listener_sup.erl
@@ -32,7 +32,6 @@
 %% ===================================================================
 %% API functions
 %% ===================================================================
-
 -spec start_inbound_listener(amqp_listener_connection()) -> startlink_ret().
 start_inbound_listener(Connection) ->
     supervisor:start_child(?MODULE, [Connection]).
@@ -56,7 +55,6 @@ start_link() ->
         _Other -> lager:error("error starting inbound_listeneres sup : ~p", [_Other])
     end,
     R.
-        
 
 %% ===================================================================
 %% Supervisor callbacks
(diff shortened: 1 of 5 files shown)


applications/doodle/src/doodle_inbound_listener_sup.erl
```diff
@@ -0,0 +1,109 @@
...
+%% Whenever a supervisor is started using supervisor:start_link/[2,3],
+%% this function is called by the new process to find out about
+%% restart strategy, maximum restart frequency and child
+%% specifications.
+%% @end
+%%--------------------------------------------------------------------
+-spec init([]) -> sup_init_ret().
+init([]) ->
+    wh_util:set_startup(),
+    RestartStrategy = 'one_for_one',
```
jamesaimonetti: Why not use simple_one_for_one? Since you're just giving them "random" names anyway, why bother naming them at all?

applications/doodle/src/doodle_inbound_listener_sup.erl
```diff
@@ -0,0 +1,109 @@
...
+-module(doodle_inbound_listener_sup).
+
+-behaviour(supervisor).
+
+-export([start_link/0]).
+-export([init/1]).
+
+-include("doodle.hrl").
+
+-define(JSON(Proplist), {Proplist}).
```
jamesaimonetti: Please don't do this. Use `wh_json:to_proplist(JObj)` and `wh_json:from_list(Props)` for converting.

jamesaimonetti:
Can you add something in doodle/doc/configuration.md with examples on how to configure the inbound listeners settings as well?

applications/doodle/src/doodle_inbound_listener_sup.erl
lazedo: true but i wanted to supply the CHILDREN list and that's not possible in simple_one_for_one. maybe there's another strategy for initializing the CHILDREN list ?

PR #821, opened 2015-03-11
KAZOO-2941: Add phone number cleanup

macpie:


core/whistle_number_manager-1.0.0/src/wh_number_fix.erl
```diff
@@ -0,0 +1,334 @@
...
+%%%-------------------------------------------------------------------
+-module(wh_number_fix).
+
+-export([fix_account_numbers/1]).
+
+-include("wnm.hrl").
+
+-define(PHONE_NUMBERS, <<"phone_numbers">>).
+
+-define(TIME_BETWEEN_ACCOUNTS, whapps_config:get_integer(?WNM_CONFIG_CAT, <<"time_between_accounts">>, 1000)).
```
jamesaimonetti: Add a suffix to the keys for the unit of time:
*  "_s" -> seconds
* "_ms" -> milliseconds

jamesaimonetti: Also, instead of 1000, use one of the time macros in wh_types like ?MILLISECONDS_IN_SECONDS etc, so we don't have "magic" numbers.

core/whistle_number_manager-1.0.0/src/whistle_number_manager_maintenance.erl
```diff
@@ -154,6 +156,35 @@ reconcile_providers([], Config) ->
...
 %%--------------------------------------------------------------------
+%% @public
+%% @doc
+%% exist
+%% @end
+%%--------------------------------------------------------------------
+-spec cleanup_phone_numbers() -> 'ok'.
+-spec cleanup_phone_numbers(ne_binary() | ne_binaries()) -> 'ok'.
+cleanup_phone_numbers() ->
+    Databases = get_databases(),
```
jamesaimonetti: Why not use `whapps_util:get_all_accounts('encoded')` ?

macpie: did not know it existed.

PR #823, opened 2015-03-12
Spec which serves as documentation seems incompatible with erlang applic...

aberke:
...ation docs.  This is confirmed when running code.

Changed spec for whapps_controller:stop_app and whapps_controller:restart_app

### DIFF
--- a/core/whistle_apps-1.0.0/src/whapps_controller.erl
+++ b/core/whistle_apps-1.0.0/src/whapps_controller.erl
@@ -59,13 +59,12 @@ start_app(App) ->
 stop_app(App) when not is_atom(App) ->
     stop_app(wh_util:to_atom(App));
 stop_app(App) ->
-    _ = application:stop(App),
-    lager:info("stopped kazoo application ~s", [App]).
+    case application:stop(App) of
+      ok -> lager:info("stopped kazoo application ~s", [App]);
+      Err -> lager:error("Error stopping ~p application: ~p~n", [App, Err]), Err
+    end.
 
--spec restart_app(atom() | nonempty_string() | ne_binary()) ->
-                         {'ok', pid() | 'undefined'} |
-                         {'ok', pid() | 'undefined', term()} |
-                         {'error', term()}.
+-spec restart_app(atom() | nonempty_string() | ne_binary()) -> 'ok' | {'error', term()}.
 restart_app(App) when not is_atom(App) ->
     restart_app(wh_util:to_atom(App));
 restart_app(App) when is_atom(App) ->


core/whistle_apps-1.0.0/src/whapps_controller.erl
```diff
@@ -59,13 +59,12 @@ start_app(App) ->
 stop_app(App) when not is_atom(App) ->
     stop_app(wh_util:to_atom(App));
 stop_app(App) ->
-    _ = application:stop(App),
-    lager:info("stopped kazoo application ~s", [App]).
+    case application:stop(App) of
+      ok -> lager:info("stopped kazoo application ~s", [App]), ok;
```
jamesaimonetti: lager's log levels all return 'ok', so no need to have the explicit ok afterwards.

I'd say to log the error tuple as well since these functions are typically called via the SUP and not in the shell.

I'd also consider making these io:format/2 for that reason too...thoughts?

core/whistle_apps-1.0.0/src/whapps_controller.erl
```diff
@@ -59,13 +59,12 @@ start_app(App) ->
...
+      Err -> Err
+    end.
 
--spec restart_app(atom() | nonempty_string() | ne_binary()) ->
-                         {'ok', pid() | 'undefined'} |
-                         {'ok', pid() | 'undefined', term()} |
-                         {'error', term()}.
+-spec restart_app(atom() | nonempty_string() | ne_binary()) -> 'ok' | {'error', term()}.
```
lazedo: this spec seems wrong, it doesn't consider the values returned by start_app which is what is returned.

shouldn't this be
-spec restart_app(atom() | nonempty_string() | ne_binary()) -> 'ok' | 'error' | 'exists'.

aberke: @lazedo I don't follow.
This function calls `application:start(atom())` which returns `'ok' | {'error', term()}.`
As per documentation: http://www.erlang.org/doc/apps/kernel/application.html#start-1

core/whistle_apps-1.0.0/src/whapps_controller.erl
aberke: ok I'll amend

core/whistle_apps-1.0.0/src/whapps_controller.erl
lazedo: @aberke you're right. it seemed to me more appropriate to call stop_app/start_app and i misread it. 
@jamesaimonetti shouldn't we call stop_app/start_app instead ?

jamesaimonetti: Probably? :) I'm fine with that change too.

aberke: Ok I don't have an opinion on that.  Want to merge? :)

jamesaimonetti:
Manually merged, thanks @aberke

PR #824, opened 2015-03-12
KAZOO-3418: Obtain rate for call to particular number through crossbar

onnet:
Please consider adding this feature. 

Remember James' words: "calling other applications' modules is a no-no", but still not sure if I need to duplicate code over here. Please let me know if I need to copy several functions from hon_util.erl

### DIFF
--- a/applications/crossbar/src/modules/cb_rates.erl
+++ b/applications/crossbar/src/modules/cb_rates.erl
@@ -453,7 +453,20 @@ rate_for_number(Context, Phonenumber) ->
                                     ,fun wapi_rate:publish_req/1
                                     ,fun wapi_rate:resp_v/1
                                     ,10000) of
-               {'ok', Rate} ->  Rate;
+               {'ok', Rate} ->  
+                   Routines = [fun(J) -> wh_json:set_value(<<"Base-Cost">>, wh_util:to_binary(wh_util:to_float(wh_json:get_value(<<"Base-Cost">>, J))/10000), J) end
+                               ,fun(J) -> wh_json:set_value(<<"Rate">>, wh_util:to_binary(wh_util:to_float(wh_json:get_value(<<"Rate">>, J))/10000), J) end
+                               ,fun(J) -> wh_json:delete_key(<<"Event-Category">>, J) end
+                               ,fun(J) -> wh_json:delete_key(<<"Event-Name">>, J) end
+                               ,fun(J) -> wh_json:delete_key(<<"App-Name">>, J) end
+                               ,fun(J) -> wh_json:delete_key(<<"App-Version">>, J) end
+                               ,fun(J) -> wh_json:delete_key(<<"Msg-ID">>, J) end
+                               ,fun(J) -> wh_json:delete_key(<<"Node">>, J) end
+                               ,fun(J) -> wh_json:delete_key(<<"Call-ID">>, J) end
+                               ,fun(J) -> wh_json:delete_key(<<"Pvt-Cost">>, J) end
+                               ,fun(J) -> wh_json:delete_key(<<"Update-Callee-ID">>, J) end
+                              ],
+                   lists:foldl(fun(F, J) -> F(J) end, Rate, Routines);
                _ -> wh_json:new() 
            end,
     crossbar_util:response(Resp, Context).
--- a/applications/crossbar/src/modules/cb_rates.erl
+++ b/applications/crossbar/src/modules/cb_rates.erl
@@ -9,10 +9,10 @@
 -module(cb_rates).
 
 -export([init/0
-         ,allowed_methods/0, allowed_methods/1
-         ,resource_exists/0, resource_exists/1
+         ,allowed_methods/0, allowed_methods/1 ,allowed_methods/2
+         ,resource_exists/0, resource_exists/1 ,resource_exists/2
          ,content_types_accepted/1
-         ,validate/1, validate/2
+         ,validate/1, validate/2, validate/3
          ,put/1
          ,post/1, post/2
          ,delete/2
@@ -22,6 +22,7 @@
 
 -define(PVT_FUNS, [fun add_pvt_type/2]).
 -define(PVT_TYPE, <<"rate">>).
+-define(NUMBER_RATE, <<"number_rate">>).
 -define(CB_LIST, <<"rates/crossbar_listing">>).
 
 -define(UPLOAD_MIME_TYPES, [{<<"text">>, <<"csv">>}
@@ -62,6 +63,10 @@ allowed_methods() ->
 allowed_methods(_) ->
     [?HTTP_GET, ?HTTP_POST, ?HTTP_DELETE].
 
+-spec allowed_methods(path_token(),path_token()) -> http_methods().
+allowed_methods(?NUMBER_RATE,_) ->
+    [?HTTP_GET].
+
 %%--------------------------------------------------------------------
 %% @private
 %% @doc
@@ -75,6 +80,9 @@ resource_exists() -> 'true'.
 
 -spec resource_exists(path_token()) -> 'true'.
 resource_exists(_) -> 'true'.
+
+-spec resource_exists(path_token(),path_token()) -> 'true'.
+resource_exists(?NUMBER_RATE,_) -> 'true'.
 
 -spec content_types_accepted(cb_context:context()) -> cb_context:context().
 content_types_accepted(Context) ->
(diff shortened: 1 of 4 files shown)


lazedo:
@onnet why don't you make wapi call to hotornot instead ?

onnet:
Just an easy way was chosen as usually :)
Than you Luis. Will look at it

applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -22,6 +22,7 @@
 
 -define(PVT_FUNS, [fun add_pvt_type/2]).
 -define(PVT_TYPE, <<"rate">>).
+-define(NUMBER_RATE, <<"number_rate">>).
```
jamesaimonetti: How do you feel about `/rates/number/+74956063602`. We typically try not to put verbs in the URI path

jamesaimonetti:
Agree with @lazedo on the AMQP request; @onnet let me know if you need any help with that but I suspect you know your way around well enough!

onnet:
Hopefully added fields wouldn't break rater's functionality

applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -435,3 +446,27 @@ save_processed_rates(Context, Count) ->
...
+                               ,fun(J) -> wh_json:set_value(<<"Rate">>, wh_util:to_binary(wh_util:to_float(wh_json:get_value(<<"Rate">>, J))/10000), J) end
+                               ,fun(J) -> wh_json:delete_key(<<"Event-Category">>, J) end
+                               ,fun(J) -> wh_json:delete_key(<<"Event-Name">>, J) end
+                               ,fun(J) -> wh_json:delete_key(<<"App-Name">>, J) end
+                               ,fun(J) -> wh_json:delete_key(<<"App-Version">>, J) end
+                               ,fun(J) -> wh_json:delete_key(<<"Msg-ID">>, J) end
+                               ,fun(J) -> wh_json:delete_key(<<"Node">>, J) end
+                               ,fun(J) -> wh_json:delete_key(<<"Call-ID">>, J) end
+                               ,fun(J) -> wh_json:delete_key(<<"Pvt-Cost">>, J) end
+                               ,fun(J) -> wh_json:delete_key(<<"Update-Callee-ID">>, J) end
```
lazedo: maybe you can create a macro at the top of module with the keys you want to delete and just call delete_keys with that macro, or create a macro with keys you want to include and call wh_json:map

applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -435,3 +446,27 @@ save_processed_rates(Context, Count) ->
...
+                              ],
+                   lists:foldl(fun(F, J) -> F(J) end, Rate, Routines);
+               _ -> wh_json:new()
```
lazedo: add the error to cb_context

applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -95,10 +109,13 @@ content_types_accepted_by_verb(Context, ?HTTP_POST) ->
 %%--------------------------------------------------------------------
 -spec validate(cb_context:context()) -> cb_context:context().
 -spec validate(cb_context:context(), path_token()) -> cb_context:context().
+-spec validate(cb_context:context(), path_token(), path_token()) -> cb_context:context().
 validate(Context) ->
     validate_rates(Context, cb_context:req_verb(Context)).
 validate(Context, Id) ->
     validate_rate(Context, Id, cb_context:req_verb(Context)).
+validate(Context, ?NUMBER, Phonenumber) ->
+    rate_for_number(Context, Phonenumber).
```
lazedo: check if its a valid number before passing to rate_for_number, return error if not valid

applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -435,3 +452,17 @@ save_processed_rates(Context, Count) ->
...
+rate_for_number(Context, Phonenumber) ->
+    case wh_amqp_worker:call([{<<"To-DID">>, wnm_util:normalize_number(Phonenumber)}| wh_api:default_headers(?APP_NAME, ?APP_VERSION)]
+                             ,fun wapi_rate:publish_req/1
+                             ,fun wapi_rate:resp_v/1
+                             ,10000) of
+        {'ok', Rate} ->  
+            Routines = [fun(J) -> wh_json:set_value(<<"Base-Cost">>, wh_util:to_binary(wh_util:to_float(wh_json:get_value(<<"Base-Cost">>, J))/10000), J) end
+                        ,fun(J) -> wh_json:set_value(<<"Rate">>, wh_util:to_binary(wh_util:to_float(wh_json:get_value(<<"Rate">>, J))/10000), J) end
+                       ],
+            crossbar_util:response(lists:foldl(fun(F, J) -> F(J) end, wh_json:filter(fun({K,_}) -> lists:member(K, ?NUMBER_RESP_FIELDS)  end, Rate), Routines), Context);
```
lazedo: this works but i would break down in small functions,
something like 
-> normalize_view(Rate, Context)

normalize_view(Rate, Context) ->
   crossbar_util:response(filter_view(Rate) , Context).

filter_view(Rate) ->
  normalize_fields(wh_json:filter(fun({K,_}) -> lists:member(K, ?NUMBER_RESP_FIELDS)  end, Rate))

normalize_fields(Rate) ->
   Routines = [...],
   lists:foldl(fun(F, J) -> F(J) end, Rate, Routines).

thoughts ?

applications/crossbar/src/modules/cb_rates.erl
onnet: Hi Luis.
Not sure what is the best way to validate number.
Looks like wnm_util doesn't provides anything special.
So I just clean up string from anything but digits and check if it not very few of it.

applications/crossbar/src/modules/cb_rates.erl
onnet: Looks better now :)

applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -133,6 +150,23 @@ put(Context) ->
 delete(Context, _RateId) ->
     crossbar_doc:delete(Context).
 
+-spec validate_number(ne_binary(), cb_context:context()) -> cb_context:context().
+validate_number(Phonenumber, Context) ->
+    case length(re:replace(Phonenumber, "[^0-9]", "", [global, {return, list}])) >= 7 of
```
jamesaimonetti: I think a better test is `wnm_util:is_reconcilable/1`. A number that is considered "reconcilable" is a "global" number for routing purposes, which is when rates would be applied.

applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -133,6 +150,23 @@ put(Context) ->
 delete(Context, _RateId) ->
     crossbar_doc:delete(Context).
 
+-spec validate_number(ne_binary(), cb_context:context()) -> cb_context:context().
+validate_number(Phonenumber, Context) ->
+    case length(re:replace(Phonenumber, "[^0-9]", "", [global, {return, list}])) >= 7 of 
+        'true' -> 
+            rate_for_number(re:replace(Phonenumber, "[^0-9]", "", [global, {return, list}]), Context);
```
jamesaimonetti: Once you know it is reconcilable, why not use `wnm_util:to_e164/1` to process it into a known format?

jamesaimonetti:
@onnet This looks really close! The last addition is docs added in the rates.md for how to use the new API.

applications/crossbar/src/modules/cb_rates.erl
onnet: James, let me disagree with you. By default reconcilable numbers are only US numbers.
-define(DEFAULT_RECONCILE_REGEX, <<"^\\+?1?\\d{10}$">>).

This means all other directions with fail with this check.
International users usually just change +1 code to its own country. Which means that they also will be able to check rates to national calls but not international 

Please let me know if I'm wrong

onnet: May be it is time to change default to something like this: <<"^\+?(\d{10,})$">>
I use it for long time without complaints :)

onnet:
Oops! Docs... OK, will do, but don't blame me for my English after that :)

applications/crossbar/src/modules/cb_rates.erl
jamesaimonetti: @onnet The whole point of the reconcile_regex is to distinguish between globally-routable and locally-routable numbers. I'm all for having a better, all-inclusive default, but that doesn't relieve the kazoo administrator of the responsibility of making sure they have a proper reconcile_regex configued.

The first thing hotornot does is run `wnm_util:to_e164/1` on the numbers supplied. So this work is being done on the query anyway; I don't see the problem of restricting the API in a similar way as callflows would when routing a number.

Anyway, we're happy to have a pull request with better defaults too! I think the call to `is_reconcilable/1` is more in line with how a kazoo cluster is expected to be configured for numbers than just checking length after a regex. Why waste time rating a number if the system doesn't believe it is reconcilable anyway?

onnet: OK, got you. 
Will go your way.
And will make pull request to remove US country code from default reconcile_regex since Kazoo spreads around the globe with speed of the light. And we'll see what you guys will tell about that :)

onnet: https://2600hz.atlassian.net/browse/KAZOO-3431

onnet:
Hi
Hope it looks like nothing to add any more :)
Regards,

jamesaimonetti:
Hi @onnet I've made adjustments and created a new pull request with them: (https://github.com/2600hz/kazoo/pull/847)

Please try it and if it works for you as before, we can merge this in!

PR #835, opened 2015-03-16
Fix: assign value to alredy bound variable (Filename)

Dinkor:


lazedo:
@Dinkor nice catch! thank you

PR #836, opened 2015-03-16
KAZOO-3428: fix `abandon_reason`s

siplabs:


applications/acdc/src/acdc_agent_fsm.erl
```diff
@@ -750,12 +750,10 @@ ringing({'channel_hungup', MemberCallId, _Cause}, #state{agent_listener=AgentLis
                                                          ,account_id=AccountId
                                                          ,agent_id=AgentId
                                                          ,member_call_id=MemberCallId
-                                                         ,member_call_queue_id=QueueId
                                                         }=State) ->
     lager:debug("caller's channel (~s) has gone down, stop agent's call: ~s", [MemberCallId, _Cause]),
     acdc_agent_listener:channel_hungup(AgentListener, MemberCallId),
 
-    acdc_stats:call_abandoned(AccountId, QueueId, MemberCallId, ?ABANDON_HANGUP),
```
jamesaimonetti: Perhaps instead of removing these (as others may want these data points), you could introduce calls to `whapps_config:get_is_true(?CONFIG_CAT, <<"track_member_timeout">>, 'true')` to preserve backwards compatibility. For each `ABANDON_*` reason, system admins can select which hangup causes they'd prefer to be tracked.

siplabs: James,

in thiss pull I'm trying to solve the problem of stats being written from the different places. Here are some logs:

>|1159721268@10.11.100.40|acdc_queue_fsm:587 (\<0.4840.0\>) **Abandon call 1159721268@10.11.100.40 with reason timeout**
>|mgr_84f64b6c4f2b7ca2bfcb884641db10b1|acdc_queue_manager:397 (\<0.4826.0\>) **Abandon call 1159721268@10.11.100.40 with reason member_timeout**
>|0000000000|Undefined:Undefined (<0.4835.0>) Channel (\<0.4835.0\>): dropping method {{'basic.deliver',<<"amq.ctag-qGGJhNppolWzSeJy0Pbvvg">>,13,false,<<"targeted">>,<<"whistle_apps@ringtwo.siplabs.local-<0.11668.0>-d89eb19b">>},{content,60,none,<<128,0,16,97,112,112,108,105,99,97,116,105,111,110,47,106,115,111,110>>,rabbit_framing_amqp_0_9_1,[<<"{\"Agent-ID\":\"dd864c9650d534a270ecb08a69ffbfe5\",\"Process-ID\":\"whistle_apps@ringtwo.siplabs.local-<0.4840.0>\",\"Call-ID\":\"1159721268@10.11.100.40\",\"Queue-ID\":\"84f64b6c4f2b7ca2bfcb884641db10b1\",\"Account-ID\":\"d551fb161a6b66c8524902188dc1ee1c\",\"Server-ID\":\"\",\"Node\":\"whistle_apps@ringtwo.siplabs.local\",\"Msg-ID\":\"68f90bf7aad6ff04d02823d2f5825864\",\"App-Version\":\"1.0.0\",\"App-Name\":\"acdc\",\"Event-Name\":\"call_fail\",\"Event-Category\":\"member\"}">>]}} from server because channel is closing#012
>|1159721268@10.11.100.40|acdc_agent_fsm:758 (\<0.4867.0\>) **Abandon call 1159721268@10.11.100.40 with reason member_hangup**

Here we can see that three different abandon_reasons are being written from three different places. In this case the call was terminated due to timeout. 
We chose to use the simplest method - leave the places that are most actual. Maybe other people want to see something else, but I suppose we should expand the number of stat fields in this case.

PR #837, opened 2015-03-16
Patches for review

lazedo:


core/whistle_config-1.0.0/src/whapps_account_config.erl
```diff
@@ -32,7 +32,11 @@ get_global(Account, Category, Key, Default) ->
     AccountDb = wh_util:format_account_id(AccountId, 'encoded'),
     case couch_mgr:open_cache_doc(AccountDb, config_doc_id(Category), [{'cache_failures', ['not_found']}]) of
         {'ok', JObj} -> get_global_from_doc(Category, Key, Default, JObj);
-        {'error', _} -> whapps_config:get(Category, Key, Default)
+        {'error', _} ->
```
jamesaimonetti: I don't think this is appropriate for whapps_account_config. It could possibly leak a reseller's reseller to a customer via whitelabeled settings, while the system_config version could be generic. Talked with Karl and he agreed. Maybe we create a `get_from_tree` or `get_chained` or similar to achieve this?

PR #845, opened 2015-03-16
KAZOO-3433: Fix cb_limits dry_run

macpie:


core/whistle_services-1.0.0/src/services/wh_service_limits.erl
```diff
@@ -27,18 +27,44 @@ reconcile(Services) ->
...
-                        ,fun(S) ->
-                                 Quantity = wh_json:get_integer_value(<<"inbound_trunks">>, JObj, 0),
-                                 wh_services:update(<<"limits">>, <<"inbound_trunks">>, Quantity, S)
-                         end
-                         ,fun(S) ->
-                        	 Quantity = wh_json:get_integer_value(<<"outbound_trunks">>, JObj, 0),
-                                 wh_services:update(<<"limits">>, <<"outbound_trunks">>, Quantity, S)
-                         end
-                       ],
+            Routines = [
```
jamesaimonetti: Why aren't you calling `reconcile(Services, JObj)` here instead of repeating yourself?

PR #846, opened 2015-03-17
KAZOO-3388

lazedo:


applications/ecallmgr/src/ecallmgr_fs_event_stream.erl
```diff
@@ -147,14 +149,32 @@ handle_cast(_Msg, #state{idle_alert=Timeout}=State) ->
...
+                                         ,switch_uri='undefined'
+                                         }=State) ->
+    try    
+        SwitchURL = ecallmgr_fs_node:sip_url(Node),    
+        [_, SwitchURIHost] = binary:split(SwitchURL, <<"@">>),
+        SwitchURI = <<"sip:", SwitchURIHost/binary>>,
+        handle_info({'tcp', Socket, Data}, State#state{switch_uri=SwitchURI, switch_url=SwitchURL})
+    catch
+        E1:E2 -> lager:warning("failed to include switch_url/uri for node ~s : ~p : ~p",[Node, E1, E2]),
+                 handle_info({'tcp', Socket, Data}, State)
```
jamesaimonetti: Won't this just loop continually?

lazedo: oops! yes it will

applications/ecallmgr/src/ecallmgr_fs_route.erl
```diff
@@ -347,6 +347,9 @@ start_message_handling(_Node, _FetchId, CallId, JObj) ->
 
 -spec route_req(ne_binary(), ne_binary(), wh_proplist(), atom()) -> wh_proplist().
 route_req(CallId, FetchId, Props, Node) ->
+    SwitchURL = ecallmgr_fs_node:sip_url(Node),
+    [_, SwitchURIHost] = binary:split(SwitchURL, <<"@">>),
```
jamesaimonetti: Why is this extraction not wrapped in a try/catch when the ecallmgr_fs_event_stream's version is?

lazedo: ecallmgr_fs_event_stream's version has a try/catch because of the startup sequence.
ecallmgr_fs_node may not be initialized and running (this happened when tried to initialize on startup, after connection). i had to move it to first occurence of an event. the tryt/catch here is more like a sanity check

in ecallmgr_fs_route, ecallmgr_fs_node:sip_url is already initialized so there isn't a problem here.

jamesaimonetti: So should the handle_info/2 of ecallmgr_fs_event_stream, when it sets `switch_info` to `true` to process the event, get the return of the sub-call to handle_info and set `switch_info` back to `false`? Otherwise, if it is missed the first time, the switch info won't be available on that event stream.

PR #849, opened 2015-03-17
KAZOO-3422: Correct teletype previews

macpie:


lazedo:
@macpie what about attachments ?

jamesaimonetti:
I'd say not to worry about attachments until its requested. We have plenty of other things to do.

PR #851, opened 2015-03-18
Rewriting request validation for better understanding

Dinkor:


jamesaimonetti:
Thanks @Dinkor ! I was inspired and refactored the leaking fields flow to follow suit with a list of funs to fold over.

PR #853, opened 2015-03-19
Kazoo 523

macpie:


macpie:
We need to decide where to start the crawler

core/whistle_number_manager-1.0.0/src/wh_port_request.erl
```diff
@@ -200,3 +203,126 @@ enable_number(N) ->
...
+
+-spec send_request(ne_binary(), wh_json:object()) -> 'error' | 'ok'.
+send_request(Url, JObj) ->
+    Id = wh_json:get_value(<<"_id">>, JObj),
+
+    Headers = [{"Content-Type", "application/json"}
+               ,{"User-Agent", wh_util:to_list(erlang:node())}
+              ],
+
+    Uri = binary:bin_to_list(<<Url/binary, "/", Id/binary>>),
```
jamesaimonetti: what's wrong with `wh_util:to_list/1` ?

core/whistle_number_manager-1.0.0/src/wh_port_request.erl
```diff
@@ -200,3 +203,126 @@ enable_number(N) ->
...
+            lager:debug("submitted_port_request successfully sent");
+        _Other ->
+            lager:error("failed to send submitted_port_request ~p", [_Other]),
+            'error'
+    end.
+
+-spec send_attachements(ne_binary(), wh_json:object()) -> 'error' | 'ok'.
+send_attachements(Url, JObj) ->
+    Id = wh_json:get_value(<<"_id">>, JObj),
+    Attachments = wh_json:get_value(<<"_attachments">>, JObj),
```
jamesaimonetti: Use `wh_doc:attachments/1`

core/whistle_number_manager-1.0.0/src/wh_port_request.erl
```diff
@@ -200,3 +203,126 @@ enable_number(N) ->
...
+    wh_json:foldl(
```
jamesaimonetti: This fold will only return the result of the last attachment upload attempt. If there are 3 attachments, and the first one fails, should the rest be attempted anyway? If so, and they succeed, should 'ok' really be returned?

core/whistle_number_manager-1.0.0/src/wh_port_request.erl
```diff
@@ -200,3 +203,126 @@ enable_number(N) ->
...
+        {'ok', "200", _Headers, _Resp} ->
+            lager:debug("attachment ~s for submitted_port_request successfully sent", [Name]);
+        _Other ->
+            lager:error("failed to send attachment ~s for submitted_port_request ~p", [Name, _Other]),
+            'error'
+    end.
+
+-spec set_flag(wh_json:object()) -> 'ok'.
+set_flag(JObj) ->
+    Doc = wh_json:set_value(<<"sent">>, 'true', JObj),
```
jamesaimonetti: Shouldn't this be a pvt_ field too? Would this be needed in the listing/UI component?

core/whistle_number_manager-1.0.0/src/wh_port_request_crawler.erl
```diff
@@ -0,0 +1,152 @@
...
+%% Starts the server
+%%
+%% @spec start_link() -> {ok, Pid} | ignore | {error, Error}
+%% @end
+%%--------------------------------------------------------------------
+start_link() ->
+    gen_server:start_link(?MODULE, [], []).
+
+stop() ->
+    self() ! 'stop'.
```
jamesaimonetti: Would prefer using `gen_server:cast/2`

core/whistle_number_manager-1.0.0/src/wh_port_request_crawler.erl
```diff
@@ -0,0 +1,152 @@
...
+%% @spec init(Args) -> {ok, State} |
+%%                     {ok, State, Timeout} |
+%%                     ignore |
+%%                     {stop, Reason}
+%% @end
+%%--------------------------------------------------------------------
+init([]) ->
+    put('callid', ?MODULE),
+    lager:debug("started ~s", [?MODULE]),
+    self() ! 'start_crawling',
```
jamesaimonetti: Prefer `gen_server:cast/2`; handle_info is typically for out-of-band messages, things from external processes that are unexpected or don't know they're sending messages to a gen_server process.

core/whistle_number_manager-1.0.0/src/wh_port_request_crawler.erl
```diff
@@ -0,0 +1,152 @@
...
+%%--------------------------------------------------------------------
+handle_info('start_crawling', State) ->
+    lager:debug("crawling"),
+    Cycle = whapps_config:get_integer(?MOD_CONFIG_CAT, <<"crawler_delay_time">>, ?MILLISECONDS_IN_MINUTE),
+    wh_port_request:send_submitted_requests(),
+    erlang:send_after(Cycle, self(), 'start_crawling'),
+    {'noreply', State, 'hibernate'};
+handle_info('stop', State) ->
+    lager:debug("crawler has been stopped"),
+    {'stop', 'ok', State};
```
jamesaimonetti: Use 'normal' for Reason instead of 'ok'.

core/whistle_number_manager-1.0.0/src/wh_port_request_crawler.erl
```diff
@@ -0,0 +1,152 @@
...
+%% @spec handle_info(Info, State) -> {noreply, State} |
+%%                                   {noreply, State, Timeout} |
+%%                                   {stop, Reason, State}
+%% @end
+%%--------------------------------------------------------------------
+handle_info('start_crawling', State) ->
+    lager:debug("crawling"),
+    Cycle = whapps_config:get_integer(?MOD_CONFIG_CAT, <<"crawler_delay_time">>, ?MILLISECONDS_IN_MINUTE),
+    wh_port_request:send_submitted_requests(),
+    erlang:send_after(Cycle, self(), 'start_crawling'),
```
jamesaimonetti: I prefer `erlang:start_timer/3`, capturing the Ref returned, and putting it in the State. Then you can `handle_info({'timeout', Ref, Message}, #state{field=Ref}=State) ->` to make sure you're handling the message you sent yourself.

If you do this, then in `init/1` you can start the timer and not message yourself to do so.

core/whistle_number_manager-1.0.0/src/wh_port_request_crawler.erl
jamesaimonetti: After viewing the handle_info clause for this, please start the initial timer here instead of messaging yourself.

jamesaimonetti: For reference, you can view konami_event_listener's use of `cleanup_timer/0` in `init/1` and `handle_info/2`

core/whistle_number_manager-1.0.0/src/wh_port_request.erl
macpie: I think it could be relevant for the UI.

jamesaimonetti: I think it should still be a pvt_ field that is then leaked. We don't want the UI setting "sent" on the port request document and expecting it to be valid.

macpie: Valid point

PR #856, opened 2015-03-20
KAZOO-3103: topup teletype template

fenollp:


fenollp:
Note: this adds things in util that upcoming KAZOO-3100 will depend on.

applications/teletype/src/teletype_util.erl
```diff
@@ -308,29 +313,24 @@ account_params(DataJObj) ->
 
 -spec find_account_params(wh_json:object(), ne_binary()) -> wh_proplist().
 find_account_params(DataJObj, AccountId) ->
-    case teletype_util:open_doc(<<"account">>, AccountId, DataJObj) of
-        {'ok', AccountJObj} -> build_account_params(AccountJObj);
+    case ?MODULE:open_doc(<<"account">>, AccountId, DataJObj) of
```
lazedo: you changed teletype_util:open_doc to ?MODULE:open_doc , what does this solve ?
why not change to open_doc ? if we move open_doc outside this module we will get compilation error, which i believe is better then getting the errors when running.

applications/teletype/src/teletype_shared_listener.erl
```diff
@@ -55,12 +55,13 @@
                       }
                      ,{{'teletype_system_alert', 'handle_system_alert'}
                        ,[{<<"notification">>, <<"system_alert">>}]
+                     ,{{'teletype_topup', 'handle_topup'}
```
lazedo: missing } before the new entry ?

fenollp: Ah. Rebase failed me

applications/teletype/src/teletype_util.erl
fenollp: You're very right

fenollp:
Ready for merging!

PR #860, opened 2015-03-20
KAZOO-3100: teletype low_balance template

fenollp:


applications/teletype/src/templates/teletype_low_balance.erl
```diff
@@ -0,0 +1,179 @@
...
+handle_low_balance(JObj, _Props) ->
+    'true' = wapi_notifications:low_balance_v(JObj),
+    wh_util:put_callid(JObj),
+
+    %% Gather data for template
+    DataJObj0 = wh_json:normalize(wh_api:remove_defaults(JObj)),
+
+    ServerID = wh_json:get_first_defined([<<"server_id">>, <<"Server-ID">>], JObj),
+    DataJObj = wh_json:set_value(<<"Server-ID">>, ServerID, DataJObj0),
+
```
lazedo: why not use just wh_json:normalize ?

server_id will be handled in send_update if that's the case.

fenollp:
Ready to merge!

applications/teletype/src/templates/teletype_low_balance.erl
fenollp: Ah! Just done this and yes, it fixes the timeouts I kept getting.

PR #862, opened 2015-03-23
KAZOO-3097: Add new template for port_request

macpie:


applications/teletype/src/teletype_util.erl
```diff
@@ -785,6 +785,7 @@ find_account_id(JObj) ->
 find_account_db(<<"account">>, JObj) -> find_account_db_from_id(JObj);
 find_account_db(<<"user">>, JObj) -> find_account_db_from_id(JObj);
 find_account_db(<<"fax">>, JObj) -> find_account_db(JObj);
+find_account_db(<<"port_request">>, _JObj) -> <<"port_requests">>;
```
jamesaimonetti: Use the DB macro instead of hardcoding it

lazedo: @jamesaimonetti what do you think to default to  find_account_db(JObj) ?
@macpie if the port_request sets the Account-DB in the notification you can can call find_account_db

core/whistle-1.0.0/src/api/wapi_notifications.erl
```diff
@@ -399,6 +399,8 @@ headers(<<"skel">>) ->
     ?SKEL_HEADERS ++ ?OPTIONAL_SKEL_HEADERS;
 headers(<<"topup">>) ->
     ?TOPUP_HEADERS ++ ?OPTIONAL_TOPUP_HEADERS;
+headers(<<"port_request">>) ->
```
jamesaimonetti: @macpie FYI, you forgot to add this clause in for the preview stuff to work. Please be sure to test both the Preview in the UI and the actual generating of the notification by the system.

PR #863, opened 2015-03-23
Errors saving SMS documents should be higher log error than debug.

aberke:


applications/doodle/src/doodle_util.erl
```diff
@@ -161,7 +161,7 @@ save_sms(JObj, DocId, Doc, Call) ->
         {'ok', Saved} ->
             whapps_call:kvs_store(<<"_rev">>, wh_json:get_value(<<"_rev">>, Saved), Call);
         {'error', E} ->
-            lager:debug("error saving sms doc , wazzup ? ~p", [E]),
+            lager:info("Error saving sms doc: ~p", [E]),
```
lazedo: @aberke i would agree more with lager:error

aberke:
Please don't change it to lager:error

aberke:
There's a race condition caused by the fact that you save the SMS in doodle_exe:stop in a separate process.  TLDR sometimes we purposely cause this error

lazedo:
"There's a  race condition caused by the fact that you save the SMS in doodle_exe:stop in a separate process". can you elaborate on that ?

what's the purpose on changing from debug to info ?

aberke:
Debug seems too casual for this - this is still important information.

lazedo:
@aberke  "There's a race condition caused by the fact that you save the SMS in doodle_exe:stop in a separate process". can you elaborate on that ?

aberke:
Background: Each time call doodle_exec:stop(Call) --> save in separate process

Say for given callflow Module:

1. Modify (SMS) Call
2. Save modified (SMS) Call as SMS doc
3. Spawn process that fetches and does something with modified SMS doc
4. Stop call ... saves Call that doodle started module with when above modified and saved call differently
5. If process spawned in (3) tries to save SMS doc again --> conflict

lazedo:
so the problem is with your spawned process, no?

aberke:
But we must finish getting through the flow of that callflow module. How else would we trigger an event on this call to handle the newly saved SMS doc and finish the flow of the callflow module without following the above steps?

aberke:
Our steps are certainly a hacky solution.  But for the time being, can't see a better way given the limitations.  I know i haven't given you much detail, but if you have any insights then please share :)

lazedo:
yep, more detail would be useful.

did you take a look at usurp_control in doodle_exe ? it stops the flow without saving the sms. if you're spawning a new process and pass along the Call object, then you can take the responsabilty for saving the sms. 

or you could wait until the spawned process ends and return the updated Call object to the flow.

lazedo:
merged as lager:error . thanks

aberke:
I'm looking at using `doodle_exe:control_usurped` instead of `doodle_exe:stop` in certain places.
At the end of normal shutdown (for stop) the hangup call command is sent out.  This doesn't happen for `doodle_exe:control_usurped`.  what does the hangup call command do?

PR #864, opened 2015-03-23
Kazoo 3098

macpie:


applications/teletype/src/templates/teletype_port_cancel.erl
```diff
@@ -0,0 +1,193 @@
...
+-spec get_attachments(wh_json:object()) -> attachments().
+-spec get_attachments(wh_json:object(), boolean()) -> attachments().
+get_attachments(DataJObj) ->
+    get_attachments(DataJObj, teletype_util:is_preview(DataJObj)).
+
+get_attachments(_DataJObj, 'true') -> [];
+get_attachments(DataJObj, 'false') ->
+    PortReqId = wh_json:get_value(<<"port_request_id">>, DataJObj),
+    Attachments = wh_doc:attachments(wh_json:get_value(<<"port_request">>, DataJObj)),
+    AttachmentNames = wh_doc:attachment_names(Attachments),
```
jamesaimonetti: I don't think this will work. `get_value(<<"port_request">>, DataJObj)` would be the `{"_id":port_request",...,"_attachments":{...}}` doc, making Attachments the JSON object under the `_attachments` key. `wh_doc:attachment_names/1` wants to load attachments then get the names, which will fail here because `_attachments` won't exist in `Attachments`, so you'll only ever get `[]` as a return from `attachment_names/1`.

applications/crossbar/src/modules/cb_port_requests.erl
```diff
@@ -427,7 +427,19 @@ post(Context, Id, ?PORT_SUBMITTED) ->
...
+    try send_ported_notification(Context, Id) of
+        _ ->
+            lager:debug("ported notification sent"),
+            do_post(Context, Id)
+    catch
+        _E:_R ->
+            lager:debug("failed to send the ported notification: ~s:~p", [_E, _R]),
+            cb_context:add_system_error(
+                'bad_gateway'
+                ,<<"failed to send ported email to system admins">>
```
jamesaimonetti: Do we want to fail the request if the email fails to send? I don't know at the moment. What's your thought process here?

macpie: It is already the case for the others (states), So I would not want to put my port request to complete state if the email was not sent (could forget about it)

PR #867, opened 2015-03-24
KAZOO-3460: fix bug with eavesdroping on user

siplabs:


lazedo:
@siplabs please rebase. thank you.

siplabs:
@lazedo was in the process

applications/callflow/src/module/cf_eavesdrop_feature.erl
```diff
@@ -100,28 +100,37 @@ get_target_for_extension(Exten, Call) ->
...
+maybe_correct_target(_Target, 'undefined', _Call) ->
+    'true';
+maybe_correct_target('error', _GroupId, _Call) ->
+    'false';
+maybe_correct_target({'ok', TargetId, _}, GroupId, Call) ->
+    lists:member(TargetId, find_group_members(GroupId, Call)).
+
+-spec find_group_members(ne_binary(), whapps_call:call()) -> ne_binaries().
+find_group_members(GroupId, Call) ->
+    GroupsJObj = cf_attributes:groups(Call),
```
jamesaimonetti: Why not just open the GroupId doc instead? The view's "value" is doc.endpoints, so it seems like you could just call `couch_mgr:open_cache_doc(whapps_call:account_db(Call), GroupId)` and get the "endpoints" out from there, vs iterating over all groups just to find the one you know you want already.

jamesaimonetti:
@siplabs I've made the change to lookup the group JObj directly and return the endpoint keys.

PR #868, opened 2015-03-24
KAZOO-3303 build proxy from dedicated ip list

lazedo:


applications/stepswitch/src/stepswitch_resources.erl
```diff
@@ -344,25 +345,42 @@ maybe_resource_to_endpoints(#resrc{id=Id
...
-                                ,update_ccvs(Endpoint, Updates)
-                               )
-             || Endpoint <- gateways_to_endpoints(Number_Match, Gateways, JObj, [])
-            ] ++ Endpoints;
+            Updates = [{<<"Name">>, Name}
+                       ,{<<"Weight">>, Weight}
+                      ],
+            EndpointList = [update_endpoint(Endpoint, Updates, CCVUpdates)
+                 || Endpoint <- gateways_to_endpoints(Number_Match, Gateways, JObj, [])],
+            maybe_add_proxies(EndpointList, Proxies, []) ++ Endpoints;
```
jamesaimonetti: If you Endpoints as the accumulator, you can re-work `maybe_add_proxies/3` to do the appends there.

applications/stepswitch/src/stepswitch_resources.erl
```diff
@@ -344,25 +345,42 @@ maybe_resource_to_endpoints(#resrc{id=Id
...
         {'error','no_match'} -> Endpoints
     end.
 
+-spec update_endpoint(wh_json:object(), wh_proplist(), wh_proplist()) -> wh_json:object().
+update_endpoint(Endpoint, Updates, CCVUpdates) ->
+    wh_json:set_values(Updates ,update_ccvs(Endpoint, CCVUpdates)).
+
+-spec maybe_add_proxies(wh_json:objects(), wh_proplist(), wh_json:objects()) -> wh_json:objects().
+maybe_add_proxies([], _, Acc) -> Acc;
+maybe_add_proxies(Endpoints, [], _) -> Endpoints;
```
jamesaimonetti: This would become `maybe_add_proxies(Endpoints, [], Acc) -> Endpoints ++ Acc;`

applications/stepswitch/src/stepswitch_resources.erl
```diff
@@ -584,7 +602,8 @@ fetch_local_resources(AccountId) ->
         {'ok', JObjs} ->
             CacheProps = [{'origin', fetch_local_cache_origin(JObjs, AccountDb, [])}],
             Resources = resources_from_jobjs([wh_json:get_value(<<"doc">>, JObj) || JObj <- JObjs]),
-            LocalResources = [Resource#resrc{global='false'} || Resource <- Resources],
+            Proxies = fetch_account_dedicated_proxies(AccountId),
```
jamesaimonetti: Could we pass Proxies into `resources_from_jobjs/2` and set them there, rather than building the list of resources, followed by iterating over the list again to set the proxies property?

lazedo: we are already iterating the returned resources to set global='false' and i opted to set proxies on that iteration. resources_from_jobjs is also used by fetch_global_resources that's why we have the iterator after the return, why not use that iterator to set the proxies ?

jamesaimonetti: Perhaps we should also pass the `global` boolean in? So global resources will call `resources_from_jobjs/1` which will call `resources_from_jobj(Resources, 'true', [])`. Meanwhile local resources will call `resources_from_jobjs(Resources, 'false', Proxies)`.

I'm just looking to get rid of unnecessary iteration; if this isn't too cumbersome to do: great!

PR #869, opened 2015-03-24
KAZOO-3437 & KAZOO-3450: Adding new port request state

macpie:


core/whistle_number_manager-1.0.0/src/wh_port_request.erl
```diff
@@ -120,6 +125,8 @@ maybe_transition(PortReq, ?PORT_SCHEDULED) ->
 maybe_transition(PortReq, ?PORT_COMPLETE) ->
     transition_to_complete(PortReq);
 maybe_transition(PortReq, ?PORT_REJECT) ->
+    transition_to_rejected(PortReq);
+maybe_transition(PortReq, ?PORT_CANCELED) ->
```
jamesaimonetti: Boo copy/paste without testing! This clause calls `transition_to_rejected` when it should be `transition_to_canceled`

core/whistle_number_manager-1.0.0/src/wh_port_request.erl
```diff
@@ -18,6 +18,7 @@
          ,transition_to_scheduled/1
          ,transition_to_complete/1
          ,transition_to_rejected/1
+         ,transition_to_canceled/1
```
jamesaimonetti: Why are any of these exported?

core/whistle_number_manager-1.0.0/src/wh_port_request.erl
macpie: bad peter!

core/whistle_number_manager-1.0.0/src/wh_port_request.erl
macpie: only complete is used (by stepswitch)

PR #880, opened 2015-03-27
Functionality of dynamically updating the callflow, like branching.

aberke:
Does not add extra flow to Flows the way branching does.

### DIFF
--- a/applications/callflow/src/cf_exe.erl
+++ b/applications/callflow/src/cf_exe.erl
@@ -17,6 +17,7 @@
 -export([queue_name/1]).
 -export([control_queue/1, control_queue/2]).
 -export([continue/1, continue/2]).
+-export([continue_with_flow/2]).
 -export([branch/2]).
 -export([stop/1]).
 -export([hard_stop/1]).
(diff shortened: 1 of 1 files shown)


applications/callflow/src/cf_exe.erl
```diff
@@ -106,6 +106,13 @@ continue(Key, Call) ->
     Srv = whapps_call:kvs_fetch('consumer_pid', Call),
     continue(Key, Srv).
 
+-spec continue_with_flow(wh_json:object(), whapps_call:call() | pid()) -> 'ok'.
```
jamesaimonetti: This function is not exported and not used internally. I assume you meant to export it?

applications/callflow/src/cf_exe.erl
```diff
@@ -353,6 +360,9 @@ handle_cast('transfer', State) ->
     {'stop', {'shutdown', 'transfer'}, State};
 handle_cast('control_usurped', State) ->
     {'stop', {'shutdown', 'control_usurped'}, State};
+handle_cast({'continue_with_flow', NewFlow}, #state{flow=Flow}=State) ->
```
jamesaimonetti: `Flow` is unused; either don't bind for it or prefix it with an '_'.

aberke:
This was very sloppy.  I apologize.  I needed to put something down to come back to later.  Will fix now.

aberke:
I amended the commit @jamesaimonetti

jamesaimonetti:
Thanks @aberke, merged.

PR #881, opened 2015-03-30
Want to be able to see from other applications whether or not the defaul...

aberke:
...t caller id is being used. Rather than copying over define statement, safer to export getter function in case that DEFAULT_CALLER_ID_NUMBER changes.


aberke:
ahh i see in the callflow.hrl we don't need this!

PR #882, opened 2015-03-30
KAZOO-2840: Correct email attachment encoding

macpie:


applications/teletype/src/templates/teletype_port_utils.erl
```diff
@@ -34,7 +34,31 @@ get_attachments(DataJObj, 'false') ->
 get_attachment_fold(Name, Acc, PortReqId, Doc) ->
     {'ok', Attachment} = couch_mgr:fetch_attachment(?KZ_PORT_REQUESTS_DB, PortReqId, Name),
     ContentType = wh_doc:attachment_content_type(Doc, Name),
-    [{ContentType, Name, Attachment}|Acc].
+    [{ContentType, Name, decode_base64(Attachment)}|Acc].
+
+-spec decode_base64(ne_binary()) -> {api_binary(), ne_binary()}.
+decode_base64(Base64) ->
```
jamesaimonetti: Since this functionality is used in Crossbar too, should probably put this in core somewhere.

macpie: Yes sounds good, I will do that

jamesaimonetti:
Rebased and manually merged.

PR #884, opened 2015-03-30
PropEr buckets: proper testing of token buckets

jamesaimonetti:


core/kazoo_token_buckets-1.0.0/src/kz_token_bucket_pqc.erl
```diff
@@ -0,0 +1,289 @@
...
+-module(kz_token_bucket_pqc).
+
+-ifdef(TEST).
+
+-include("kz_buckets.hrl").
+
+-include_lib("proper/include/proper.hrl").
+-include_lib("eunit/include/eunit.hrl").
+
+-behaviour(proper_statem).
```
lazedo: missing dep ?

jamesaimonetti: We used to include proper, but since it was only for testing and is licensed as GPL, there were concerns with including it. So we stopped including it. This will only run if you exec 'make test' so I would assume you'd have proper in your Erlang load path. I have it in among the other core Erlang/OTP libs myself.

PR #891, opened 2015-04-01
KAZOO-3426: fix Cowboy sending duplicate headers

fenollp:
The fix actually relies on setting headers to lower case. But for future additions I'd say just lowercasing header names here is the best.


applications/crossbar/src/cb_context.erl
```diff
@@ -311,10 +311,10 @@ set_resp_header(#cb_context{resp_headers=RespHeaders}=Context, K, V) ->
 set_resp_header_fold({K, V}, Hs) -> lists:keystore(K, 1, Hs, {K, V}).
 
 add_resp_headers(#cb_context{resp_headers=RespHeaders}=Context, Headers) ->
-    Context#cb_context{resp_headers=lists:foldl(fun add_resp_headers_fold/2, RespHeaders, Headers)}.
+    Append = fun ({K,V}, Acc) -> [{wh_util:to_lower_binary(K),V} | Acc] end,
```
jamesaimonetti: Why remove the named function? Looks like you only needed to change `add_resp_headers_fold/2` to do the `to_lower_binary` call.

fenollp:
There you go @jamesaimonetti

jamesaimonetti:
manually merged as I added using `props:set_value/3`

PR #893, opened 2015-04-01
Kazoo 3465

macpie:


macpie:
I think we should allow user to change pvt_notification_preference on POST ?

PR #894, opened 2015-04-02
Disconnect call if rate.resp is empty

Dinkor:


jamesaimonetti:
Thanks @Dinkor !

applications/ecallmgr/src/ecallmgr_fs_authz.erl
```diff
@@ -338,5 +348,6 @@ rating_req(CallId, Props) ->
      ,{<<"Call-ID">>, CallId}
      ,{<<"Account-ID">>, AccountId}
      ,{<<"Direction">>, props:get_value(<<"Call-Direction">>, Props)}
+     ,{<<"Send-Empty">>, 'true'}
```
lazedo: @Dinkor where is this used ?

jamesaimonetti: HotOrNot will, by default, not send a response if it doesn't find a rate. Using the Send-Empty flag means HotOrNot will respond (rather than having ecallmgr wait for a timeout). Similar to ecallmgr_fs_channels' `Active-Only` flag

lazedo: @jamesaimonetti thanks, i misinterpreted because of the way github presented the changes. it seemed it was part of set_rating_ccvs.

PR #899, opened 2015-04-03
Kazoo 3396 - move number to deleted state

jamesaimonetti:
If the system is configured to do so, instead of moving a number to 'disconnected' move it to 'deleted' for subsequent hard-removal from the system.


lazedo:
@Dinkor patched in master. thank you

PR #902, opened 2015-04-04
KAZOO-3490: only send template preview when enabled

fenollp:
And some cleansing too

### DIFF
--- a/applications/crossbar/src/modules/cb_notifications.erl
+++ b/applications/crossbar/src/modules/cb_notifications.erl
@@ -275,16 +275,6 @@ do_post(Context) ->
 
 post(Context, Id, ?PREVIEW) ->
     Notification = cb_context:doc(Context),
-    case wh_json:is_true(<<"enabled">>, Notification) of
-        'true' ->
-            handle_preview(Context, Id, Notification);
-        'false' ->
-            lager:debug("ending request: preview is disabled"),
-            cb_context:add_system_error('disabled', Context)
-    end.
-
--spec handle_preview(cb_context:context(), ne_binary(), wh_json:object()) -> cb_context:context().
-handle_preview(Context, Id, Notification) ->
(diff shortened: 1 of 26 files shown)


applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -274,13 +274,22 @@ do_post(Context) ->
...
 
 post(Context, Id, ?PREVIEW) ->
-    Preview = build_preview_payload(Context),
+    Notification = cb_context:doc(Context),
+    case wh_json:is_true(<<"enabled">>, Notification) of
+        'true' ->
+            handle_preview(Context, Id, Notification);
+        'false' ->
+            lager:debug("ending request: preview is disabled"),
+            crossbar_util:response_202(<<"Normally ending notification processing">>, Context)
```
jamesaimonetti: This should be an error instead of a 202

applications/teletype/src/teletype_util.erl
```diff
@@ -903,15 +903,15 @@ should_handle_notification_for_account(AccountId, ResellerId) ->
...
-    case {wh_json:get_value([<<"notifications">>
-                             ,NoticeKey
-                             ,<<"enabled">>
-                            ], AccountJObj)
-          ,wh_json:is_true(<<"Preview">>, ApiJObj, 'false')
-         }
-    of
+-spec is_notice_enabled(ne_binary(), wh_json:object(), ne_binary()) -> boolean().
+is_notice_enabled(AccountId, ApiJObj, NoticeKey) ->
+    AccountDb = wh_util:format_account_id(AccountId, 'encoded'),
```
lazedo: hon_rater uses wh_notify:system_alert that doesn't send account_id in the notify payload.
ne_binary() should be api_binary() and AccountId = 'undefined' can handled instead of waiting for couch error

PR #903, opened 2015-04-04
A User should be able to send an SMS to themself.  Currently, if user se...

aberke:
...nds SMS to self, it stays as status 'pending' forever since should_recreate_endpoint blocks it.

Using owner_id in kvs_store seems like an inconsistency.  Would be clearer throughout codebase to use whapps_call:owner_id

### DIFF
--- a/applications/callflow/src/cf_endpoint.erl
+++ b/applications/callflow/src/cf_endpoint.erl
@@ -517,10 +517,12 @@ should_create_endpoint([Routine|Routines], Endpoint, Properties, Call) when is_f
                                      'ok' |
                                      {'error', 'owner_called_self'}.
 maybe_owner_called_self(Endpoint, Properties, Call) ->
+    Source = wh_json:get_value(<<"source">>, Properties)
     CanCallSelf = wh_json:is_true(<<"can_call_self">>, Properties),
     EndpointOwnerId = wh_json:get_value(<<"owner_id">>, Endpoint),
-    OwnerId = whapps_call:kvs_fetch('owner_id', Call),
+    OwnerId = whapps_call:owner_id(Call),
     case CanCallSelf
+        orelse (Source =:= 'cf_sms_user')
         orelse (not is_binary(OwnerId))
         orelse (not is_binary(EndpointOwnerId))
         orelse EndpointOwnerId =/= OwnerId


applications/callflow/src/cf_endpoint.erl
```diff
@@ -517,10 +517,12 @@ should_create_endpoint([Routine|Routines], Endpoint, Properties, Call) when is_f
                                      'ok' |
                                      {'error', 'owner_called_self'}.
 maybe_owner_called_self(Endpoint, Properties, Call) ->
+    Source = wh_json:get_value(<<"source">>, Properties)
     CanCallSelf = wh_json:is_true(<<"can_call_self">>, Properties),
     EndpointOwnerId = wh_json:get_value(<<"owner_id">>, Endpoint),
-    OwnerId = whapps_call:kvs_fetch('owner_id', Call),
+    OwnerId = whapps_call:owner_id(Call),
     case CanCallSelf
+        orelse (Source =:= 'cf_sms_user')
```
lazedo: please remove this

applications/callflow/src/cf_endpoint.erl
```diff
@@ -517,10 +517,12 @@ should_create_endpoint([Routine|Routines], Endpoint, Properties, Call) when is_f
                                      'ok' |
                                      {'error', 'owner_called_self'}.
 maybe_owner_called_self(Endpoint, Properties, Call) ->
+    Source = wh_json:get_value(<<"source">>, Properties)
     CanCallSelf = wh_json:is_true(<<"can_call_self">>, Properties),
     EndpointOwnerId = wh_json:get_value(<<"owner_id">>, Endpoint),
-    OwnerId = whapps_call:kvs_fetch('owner_id', Call),
+    OwnerId = whapps_call:owner_id(Call),
```
lazedo: this is not the same as whapps_call:owner_id

lazedo:
doesn't this work is user has can_call_self = 'true' ?

lazedo:
@aberke this changes the way things work. it would be better to open a JIRA ticket so if its worth trying to patch.

applications/callflow/src/cf_endpoint.erl
aberke: can you please explain the difference to me so that I can better understand the code base?

aberke:
Do you mean I should open a support ticket?
Yes it works if callflow has can_cal_self = true.  The issue is that a single callflow is shared for SMS and calls.  A user shouldn't necessarily be able to call themself, but a user should always be able to send themselves an SMS - this is what all other carriers allow.

How should I proceed with fixing this issue?

lazedo:
if you want to propose system changes, please file a ticket in JIRA https://2600hz.atlassian.net

aberke:
Ticket: https://2600hz.atlassian.net/browse/KAZOO-3509

PR #908, opened 2015-04-06
A User should be able to send an SMS to themself.  Currently, if user se...

aberke:
...nds SMS to self, it stays as status 'pending' forever since should_recreate_endpoint blocks it.

Using owner_id in kvs_store seems like an inconsistency.  Would be clearer throughout codebase to use whapps_call:owner_id


aberke:
update of https://github.com/2600hz/kazoo/pull/903

lazedo:
duplicate of #903 which was closed as the initiative is not valid

PR #911, opened 2015-04-06
doodle_maintenance tries to replay pending SMS documents via doodle_util...

aberke:
...:replay_sms.

This is a nice feature, but if there is something wrong with the SMS doodle_maintenance will retry again each time, forever.  This is an unnecessary consumption of resources - a buggy message should be stopped after given number of tries.

Fix via this commit:
doodle_util:replay_sms checks if SMS has already been replayed/saved more than 100 times.
If so: saves status as "stopped" and does not replay
Otherwise: replays as usual.

Also fixed typo/oversight: Changed Flow -> JObj when JObj was pvt_call result

### DIFF
--- a/applications/doodle/src/doodle_util.erl
+++ b/applications/doodle/src/doodle_util.erl
@@ -231,9 +231,26 @@ get_endpoint_from_sipdb(Realm, Username) ->
 replay_sms(AccountId, DocId) ->
     lager:debug("trying to replay sms ~s for account ~s",[DocId, AccountId]),
     {'ok', Doc} = kazoo_modb:open_doc(AccountId, DocId),
-    Flow = wh_json:get_value(<<"pvt_call">>, Doc),
+    JObj = wh_json:get_value(<<"pvt_call">>, Doc),
     Rev = wh_json:get_value(<<"_rev">>, Doc),
-    replay_sms_flow(AccountId, DocId, Rev, Flow).
+    replay_sms_flow_or_stop(AccountId, DocId, Rev, JObj).
+
+-spec replay_sms_flow_or_stop(ne_binary(), ne_binary(), ne_binary(), api_object()) -> any().
+replay_sms_flow_or_stop(AccountId, DocId, <<_:1/binary, "-", _/binary>> = Rev, JObj) ->
+    replay_sms_flow(AccountId, DocId, Rev, JObj);
+replay_sms_flow_or_stop(AccountId, DocId, <<_:2/binary, "-", _/binary>> = Rev, JObj) ->
+    replay_sms_flow(AccountId, DocId, Rev, JObj);
+replay_sms_flow_or_stop(_AccountId, _DocId, _Rev, JObj) ->
+    lager:info("Stopping overplayed SMS ~s with revision ~p",[_DocId, _Rev]),
+    stop_sms(JObj).
+
+-spec stop_sms(api_object()) -> whapps_call:call().
+stop_sms(JObj) ->
+    Routines = [
+        fun(C) -> set_flow_status(<<"stopped">>, C) end
+        ,fun save_sms/1
+    ],
+    whapps_call:exec(Routines, whapps_call:from_json(JObj)).  
 
 -spec replay_sms_flow(ne_binary(), ne_binary(), ne_binary(), api_object()) -> any().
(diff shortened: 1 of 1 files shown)


applications/doodle/src/doodle_util.erl
```diff
@@ -231,9 +231,26 @@ get_endpoint_from_sipdb(Realm, Username) ->
...
+replay_sms_flow_or_stop(AccountId, DocId, <<_:2/binary, "-", _/binary>> = Rev, JObj) ->
+    replay_sms_flow(AccountId, DocId, Rev, JObj);
+replay_sms_flow_or_stop(_AccountId, _DocId, _Rev, JObj) ->
+    lager:info("Stopping overplayed SMS ~s with revision ~p",[_DocId, _Rev]),
+    stop_sms(JObj).
+
+-spec stop_sms(api_object()) -> whapps_call:call().
+stop_sms(JObj) ->
+    Routines = [
+        fun(C) -> set_flow_status(<<"stopped">>, C) end
```
aberke: use "failed" instead

applications/doodle/src/doodle_util.erl
```diff
@@ -231,9 +231,26 @@ get_endpoint_from_sipdb(Realm, Username) ->
...
     lager:debug("trying to replay sms ~s for account ~s",[DocId, AccountId]),
     {'ok', Doc} = kazoo_modb:open_doc(AccountId, DocId),
-    Flow = wh_json:get_value(<<"pvt_call">>, Doc),
+    JObj = wh_json:get_value(<<"pvt_call">>, Doc),
     Rev = wh_json:get_value(<<"_rev">>, Doc),
-    replay_sms_flow(AccountId, DocId, Rev, Flow).
+    replay_sms_flow_or_stop(AccountId, DocId, Rev, JObj).
+
+-spec replay_sms_flow_or_stop(ne_binary(), ne_binary(), ne_binary(), api_object()) -> any().
+replay_sms_flow_or_stop(AccountId, DocId, <<_:1/binary, "-", _/binary>> = Rev, JObj) ->
```
aberke: use "attempts" instead

Have system config document -- doodle config document?  with the max_attempts

aberke:
Jira Ticket: https://2600hz.atlassian.net/browse/KAZOO-3510
Alex will finish

k-anderson:
As per our call I would request that rather than use the rev to track attempts a specific parameter is stored on the document for this purpose. I would also request that we compare the attempts value to a system_config parameter as a threshold for the maximum retries.

Additionally, I request that rather than 'stopped' we flag it as something more descriptive to why it was stopped. For example, 'max_retries', 'undelieverable', or 'failed'.

aberke:
@k-anderson yup agreed - see notes above
plan to use "failed"
for this system_config variable -- is it okay to add this to the doodle system_config doc?
What should the default be?  100?

lazedo:
@aberke hey alex, as  @k-anderson  commented on the jira ticket, we are implementing a rescheduler based on rules in https://2600hz.atlassian.net/browse/KAZOO-3506 that will address this and other issues. stay tuned, almost done.

the can_send_text_to_himself is also being addressed so you can focus on kazoo-grid stuff

aberke:
@lazedo so I shouldn't work on this?  You guys are on it instead?

lazedo:
@aberke yes, focus on kazoo-grid for now, you can help us test it and improve when we issue the PR :)

k-anderson:
@aberke Actually, I talked to @lazedo and he is wrapping up work now (KAZOO-3506) that will implement the same failure/retry mechanism that we use for faxes.  His approach is more consistent within the project and I believe address the issue identified here.  I added a link to your Jira ticket, but he is about to issue a pull request (waiting on travis).  Want to take a look and comment on that?  

Seems like there was a misscommunication that we should attempt to resolve by way of the Jira tickets to avoid duplicating work.  I think it was also triggered by the closure of the other pull request but not commented that he would be addressing this.  I think we are getting better at communicating but clearly more to be done here ;)

aberke:
Ok can you send me a link to the PR when he is finished please?

lazedo:
@aberke missing some tweaks https://github.com/2600hz/kazoo/pull/922. feel free to close this pr.

PR #912, opened 2015-04-06
Kazoo 543

jamesaimonetti:


jamesaimonetti:
I need to clean this up but didn't want to lose track of it. Please don't merge yet.

PR #913, opened 2015-04-06
KAZOO-3179: As an API consumer, I would like to get a list of objects re...

k-anderson:
...lated to prompt language files so that I can see metadata about the prompt language files


k-anderson:
@macpie I found this this code while cleaning the old branches, the Jira ticket is marked closed (but not wont-fix).  If we accept this change it should only impact v2 APIs.  You should also make sure the front-end is updated (or not impacted).

macpie:
You mean you want me to add that only to v2 ?

jamesaimonetti:
@k-anderson I don't think that's necessary, as this is prompts-only stuff, which wasn't introduced until v2 anyway. No one using /v1 would have been using the prompts stuff anyway.

@macpie just make sure there's a ticket filed with the UI to handle the new format of the response for prompts.

macpie:
ok

PR #916, opened 2015-04-07
KAZOO-3507: System-based call waiting feature

siplabs:


siplabs:
Please, generate media files for cw-* prompts. Thanks!

siplabs:
relates to 2600hz/kazoo-configs#39

applications/ecallmgr/src/ecallmgr_fs_channels.erl
```diff
@@ -38,6 +38,8 @@
 -export([handle_query_channels/2]).
 -export([handle_channel_status/2]).
 
+-export([has_channels_for_owner/1, find_by_authorizing_id/1]).
```
jamesaimonetti: Why export `find_by_authorizing_id/1`?

k-anderson:
We prefer that new features are represented as object on the documents.  Historically we have also used 'enabled' to toggle functionality so for consistency I re-factored and submitted pull request 930.  Please take a look and lets have a discussion regarding that.  If it is still acceptable for you let me know and we will merge it.  I am aware this might mean you need to migrate the old parameter if it is already in use on your cluster.  If that is the case comment on 930 and we can discuss backward support for 'disable_call_waiting'.  Thanks!

PR #918, opened 2015-04-07
Kazoo 3438

jamesaimonetti:
Macros and formatting updates


k-anderson:
I am worried this will lead to 911 code spattered about.  How would you feel about a wnm_util:is_emergency_configured(ne_binary() | wh_json:object()) -> boolean() to contain the logic for checking for the appropriate feature flags?

jamesaimonetti:
@k-anderson updated

PR #920, opened 2015-04-08
KAZOO-3348: flush FreeSWITCH XML cache after parameter updates

fenollp:


applications/crossbar/src/crossbar_util.erl
```diff
@@ -1010,6 +1013,63 @@ format_emergency_caller_id_number(Context) ->
...
+              wh_json:get_value(<<"owner_id">>, Doc))
+    ) andalso
+        refresh_fs_xml(Realm, Doc),
+    'ok'.
+
+-spec refresh_fs_xml(ne_binary(), wh_json:object()) -> 'ok'.
+refresh_fs_xml(Realm, Doc) ->
+    Username = sip(<<"username">>, Doc),
+    lager:debug("flushing fs xml for user '~s' at '~s'", [Username,Realm]),
+    ecallmgr_fs_nodes:flush(Username, Realm).
```
k-anderson: AMQP please

applications/crossbar/src/crossbar_util.erl
```diff
@@ -1010,6 +1013,71 @@ format_emergency_caller_id_number(Context) ->
...
+            lager:debug("flushing fs xml for user '~s' at '~s'", [Username,Realm]),
+            Req = [{<<"Username">>, Username}
+                   ,{<<"Realm">>, Realm}
+                   | wh_api:default_headers(?APP_NAME, ?APP_VERSION)
+                  ],
+            wh_amqp_worker:cast(Req, fun wapi_switch:publish_fs_xml_flush/1)
+    end.
+
+-spec sip(ne_binary(), wh_json:object()) -> api_binary().
+sip(Key, JObj) ->
```
jamesaimonetti: Not a very descriptive name for a function. Seems like it should be `get_sip_attribute` or something along those lines.

jamesaimonetti: See also `kz_device:sip_username/1` when getting data from a device doc

k-anderson:
Despite the fact that we already worked on this together I would like to propose we find a way to bind for these switch events in a way that will not cause message delivery duplication.  My concern is since the events are consumed in a per-freeswitch node process, one published event will be delivered to multiple consumers (if there are 12 freeswitch nodes then it would be a 1:12 ratio).  We should also probably bind for this in a federated environment.  Thoughts?

applications/ecallmgr/src/ecallmgr_fs_node.erl
```diff
@@ -132,12 +132,15 @@
 -define(RESPONDERS, [{{?MODULE, 'handle_reload_acls'}
                       ,[{<<"switch_event">>, <<"reload_acls">>}]
                      }
-                     ,{{?MODULE, 'handle_reload_gtws'}
-                       ,[{<<"switch_event">>, <<"reload_gateways">>}]
-                      }
+                    ,{{?MODULE, 'handle_reload_gtws'}
```
jamesaimonetti: Just call it `handle_reload_gateways`

jamesaimonetti: We should rename it

applications/ecallmgr/src/ecallmgr_fs_nodes.erl
```diff
@@ -290,6 +312,14 @@ capability_to_json(#capability{node=Node
                        ,{<<"is_loaded">>, IsLoaded}
                       ]).
 
+-spec handle_fs_xml_flush(wh_json:object(), wh_proplist()) -> 'ok'.
+handle_fs_xml_flush(JObj, _Props) ->
+    'true' = wapi_switch:fs_xml_flush_v(JObj),
+    Username  = wh_json:get_value(<<"Username">>, JObj),
+    Realm = wh_json:get_value(<<"Realm">>, JObj, <<>>),
+    _ = wh_cache:erase_local(?ECALLMGR_AUTH_CACHE, ?CREDS_KEY(Realm, Username)),
```
k-anderson: Since we are using a named Q this will round-robbin, if there are multiple ecallmgrs they should flush their cache as well....

applications/ecallmgr/src/ecallmgr_registrar.erl
```diff
@@ -923,8 +922,13 @@ query_authn(#registration{username=Username
             AccountId = wh_json:get_value(<<"Account-ID">>, CCVs),
             AccountDb = wh_util:format_account_id(AccountId, 'encoded'),
             AuthorizingId = wh_json:get_value(<<"Authorizing-ID">>, CCVs),
+            OwnerIdProp = case wh_json:get_value([<<"Custom-Channel-Vars">>, <<"Owner-ID">>], JObj) of
+                              'undefined' -> [];
+                              OwnerId -> [{'db', OwnerId}]
```
k-anderson: Did you test this, surely you would need to bind the cache to user document updates ;)

fenollp: I tested this and it was never called. I guessed it wasn't always run..
I'm looking into binding this

PR #923, opened 2015-04-08
KAZOO-3513: new tts engine

siplabs:


siplabs:
This adds support for Russian TTS engine VoiceFabric

core/whistle_apps-1.0.0/src/whapps_speech.erl
```diff
@@ -138,6 +138,70 @@ create(<<"ispeech">> = Engine, Text, Voice, Format, Options) ->
...
+        VFabricVoice ->
+            BaseUrl = whapps_config:get_string(?MOD_CONFIG_CAT, <<"tts_url">>, <<"https://voicefabric.ru/WSServer/ws/tts">>),
+            ApiKey = whapps_config:get_binary(?MOD_CONFIG_CAT, <<"tts_api_key">>, <<>>),
+            Data = [{<<"apikey">>, ApiKey}
+                    ,{<<"ttsVoice">>, VFabricVoice}
+                    ,{<<"textFormat">>, <<"text/plain">>}
+                    ,{<<"text">>, Text}
+                   ],
+            ArgsEncode = whapps_config:get_binary(?MOD_CONFIG_CAT, <<"tts_args_encode">>, <<"urlencode">>),
+            {Headers, Body} = case ArgsEncode of
```
jamesaimonetti: Can you put this in a separate function to build the headers and body? This function has gotten quite large.

core/whistle_apps-1.0.0/src/whapps_speech.erl
```diff
@@ -290,6 +354,37 @@ create_response(_Engine, {'error', _R}) ->
 create_response(_Engine, {'ibrowse_req_id', ReqID}) ->
     lager:debug("speech file streaming as ~p", [ReqID]),
     {'ok', ReqID};
+create_response(<<"voicefabric">> = _Engine, {'ok', "200", Headers, Content}) ->
+    [lager:debug("hdr: ~p", [H]) || H <- Headers],
```
jamesaimonetti: Probably don't need this if you're ready to commit the code

core/whistle_apps-1.0.0/src/whapps_speech.erl
```diff
@@ -290,6 +354,37 @@ create_response(_Engine, {'error', _R}) ->
...
+    lager:debug("converting media"),
+    RawFile = tmp_file_name(<<"raw">>),
+    WavFile = tmp_file_name(<<"wav">>),
+    _ = file:write_file(RawFile, Content),
+    From = "raw -r 8000 -e signed-integer -b 16",
```
jamesaimonetti: Take a look at `wh_media_util:normalize_media/3`. It does what you're doing here, but via a port_command, avoiding creating temporary files to cleanup.

siplabs: @jamesaimonetti 
 We tried it but got some problems:
1. sox cannot get file length which is important
2. sox generates warning message about [1], this messages prepends encoded data
3. sox cannot create right header for wav file because of [1]
4. port returned after 20 second timeout, I think caused by [1]

core/whistle_apps-1.0.0/src/whapps_speech.erl
siplabs: @jamesaimonetti 
Without headers it will be very hard to resolve problems with this service.
It doesn't have format option, and it may change raw to something else by default in future.
Also it may deliver media in one response or by chunks.
Long story short - we guarantee these logs are needed when you deal with voicefabric :)

jamesaimonetti: Noted :)

core/whistle_apps-1.0.0/src/whapps_speech.erl
```diff
@@ -138,9 +138,70 @@ create(<<"ispeech">> = Engine, Text, Voice, Format, Options) ->
...
+            HTTPOptions = [{'response_format', 'binary'} | Options],
+            Response = ibrowse:send_req(BaseUrl, Headers, 'post', Body, HTTPOptions),
+            create_response(Engine, Response)
+    end;
 create(_, _, _, _, _) ->
     {'error', 'unknown_provider'}.
 
+voicefabric_request_body(<<"urlencode">>, Data) ->
+    Headers_ = [{"Content-Type", "application/x-www-form-urlencoded"}],
+    <<$&, Body_/binary>> = iolist_to_binary([[$&, Key, $=, Val] || {Key, Val} <- Data]),
```
jamesaimonetti: `props:to_querystring/1` should build this for you.

core/whistle_apps-1.0.0/src/whapps_speech.erl
```diff
@@ -138,9 +138,70 @@ create(<<"ispeech">> = Engine, Text, Voice, Format, Options) ->
...
+    {Headers_, Body_};
+voicefabric_request_body(<<"multipart">>, Data) ->
+    Boundary = iolist_to_binary([<<"--boundary--">>
+                                 ,couch_mgr:get_uuid(),
+                                 <<"--boundary--">>
+                                ]),
+    Headers_ = [{"Content-Type"
```
jamesaimonetti: I'm curious why you suffix the Variables with the '_'. When prefixed, the compiler won't complain about unused vars but as far as I know, '_' as a suffix does nothing special.

core/whistle_apps-1.0.0/src/whapps_speech.erl
```diff
@@ -138,9 +138,70 @@ create(<<"ispeech">> = Engine, Text, Voice, Format, Options) ->
...
+            ArgsEncode = whapps_config:get_binary(?MOD_CONFIG_CAT, <<"tts_args_encode">>, <<"urlencode">>),
+            {Headers, Body} = voicefabric_request_body(ArgsEncode, Data),
+            HTTPOptions = [{'response_format', 'binary'} | Options],
+            Response = ibrowse:send_req(BaseUrl, Headers, 'post', Body, HTTPOptions),
+            create_response(Engine, Response)
+    end;
 create(_, _, _, _, _) ->
     {'error', 'unknown_provider'}.
 
+voicefabric_request_body(<<"urlencode">>, Data) ->
```
jamesaimonetti: Don't forget specs! :)

core/whistle_apps-1.0.0/src/whapps_speech.erl
jamesaimonetti: Try applying this diff: https://gist.github.com/jamesaimonetti/62d7267d0265d6eabe12 to wh_media_util.

Regarding your points:
1. I'm confused where sox needs length? I would guess byte-order and encoding would work from stdin too, but maybe not?
2. If 1 isn't addressed with the diff above, we'll roll with this for now.
3. Ok
4. Configurable timeout now in the Options list.

Give it a try and see if it works for you; if not, we'll get this merged and revisit it another time.

siplabs: @jamesaimonetti 
I tried your patch and still got a warning message and timeout.
Debug message in timeout clause looks like: `sending error response: [[[],WarnMsg],WavData]`
I tried to remove `'stderr_to_stdout'` from `PortOptions` and it removes WarnMsg, but function still returns `{'error', ...}` and function returns after timeout. In this case I recieve WAV data without prepending but with incorrect header

The exact warning message is: `sox WARN wav: Length in output .wav header will be wrong since can't seek to fix it.`
In shell you can get this message running sox using pipes on both sides: http://sox.10957.n7.nabble.com/run-sox-in-a-pipe-with-header-based-formats-tt1469.html#a1472

So if we trying to convert raw data using erlang port:
1. `wh_media_util:normalize_media/4` will return {'error', WavData}.
2. `WavData` will have incorrect length in header.
3. `wh_media_util:normalize_media/4` will return after timeout, not after processing all data.
4.`wh_media_util:normalize_media/4` may return before processing all data.

jamesaimonetti: Thanks for trying!

PR #925, opened 2015-04-08
KAZOO-3379: remove media_mgr doc after migrate

jamesaimonetti:


core/whistle_media-1.0.0/src/whistle_media_maintenance.erl
```diff
@@ -207,12 +207,24 @@ refresh() ->
...
+        {'error', 'not_found'} -> 'ok';
+        {'ok', JObj} ->
+            migrate_system_config(wh_doc:public_fields(JObj)),
+            maybe_delete_system_config(ConfigId, DeleteAfter)
     end.
 
+-spec maybe_delete_system_config(ne_binary(), boolean()) -> 'ok'.
+maybe_delete_system_config(_ConfigId, 'false') -> 'ok';
+maybe_delete_system_config(ConfigId, 'true') ->
+    {'ok', _} = couch_mgr:del_doc(?WH_CONFIG_DB, ConfigId),
```
lazedo: if we run a 2nd time will it not crash here ?

jamesaimonetti: ConfigId won't exist, so line 216 will take care of not running this.

PR #926, opened 2015-04-09
KAZOO-3518: PATCH method for crossbar modules

siplabs:


jamesaimonetti:
@siplabs Manually merged this in. Thanks! Really great to have more PATCH support.

PR #928, opened 2015-04-09
log_stacktrack is a very useful function - but it logs at the debug leve...

aberke:
...l.

Would like this to log at a higher level so that debug level messages can be suppressed from logs and stacktrace information can continue to come through.


aberke:
Thanks for merging this!  Grid would like to be able to use this soon in a coming RPM

PR #930, opened 2015-04-09
Siplabs kazoo 3507

k-anderson:
Updated SIPLabs pull request for server-side call forwarding for consistency with features.  We prefer that all new features are objects on the respective documents so future parameters could be added in addition to the logical organization (rather than cluttering the root level of the document).


applications/ecallmgr/src/ecallmgr_util.erl
```diff
@@ -571,9 +571,34 @@ build_simple_channels(Endpoints) ->
...
     build_bridge_channels(EPs, []).
 
+-spec maybe_apply_call_waiting(wh_json:objects()) -> wh_json:objects().
+maybe_apply_call_waiting(Endpoints) ->
+    lists:map(fun call_waiting_map/1, Endpoints).
+
+-spec call_waiting_map(wh_json:object()) -> wh_json:object().
+call_waiting_map(Endpoint) ->
+    CCVs = wh_json:get_value(<<"Custom-Channel-Vars">>, Endpoint, wh_json:new()),
+    case wh_json:is_true(<<"Call-Waiting-Disabled">>, CCVs) of
```
jamesaimonetti: Who sets this CCV? I don't see it being set or overridden elsewhere in this pull request.

siplabs:
@k-anderson 
Thank you for updates! This is very acceptable for us.

PR #937, opened 2015-04-13
Kazoo 3514 and Kazoo 3515

jamesaimonetti:
Sends check_sync AMQP payload when device is updated in provisioner.

Exposes API at `/devices/{DEVICE_ID}/sync` to send the check_sync via Crossbar.


lazedo:
@jamesaimonetti added AOR support. see if it works for you.

PR #938, opened 2015-04-14
FIX: unable refresh without system_config/call_response document

Dinkor:
When start whistle_apps with empty DB, whapps_maintenance:refresh() hangs on delete_system_media_references()


lazedo:
@Dinkor thank you.

PR #939, opened 2015-04-14
KAZOO-3550 do not force 'include_docs' 

lazedo:
KAZOO-3550 do not force 'include_docs' when there is a normalize function


applications/crossbar/src/modules/cb_media.erl
```diff
@@ -902,7 +902,7 @@ load_media_binary(Context, MediaId) ->
                       crossbar_doc:load_attachment(cb_context:doc(Context1), Attachment, Context1)
                       ,[{<<"Content-Disposition">>, <<"attachment; filename=", Attachment/binary>>}
                         ,{<<"Content-Type">>, wh_doc:attachment_content_type(cb_context:doc(Context1), Attachment)}
-                        ,{<<"Content-Type">>, wh_doc:attachment_length(cb_context:doc(Context1), Attachment)}
+                        ,{<<"Content-Length">>, wh_doc:attachment_length(cb_context:doc(Context1), Attachment)}
```
aberke: was this related to this PR or just an added fix while you were in here?

jamesaimonetti: unrelated but easy fix

PR #943, opened 2015-04-14
KAZOO-3533

fenollp:


applications/crossbar/src/modules/cb_apps_store.erl
```diff
@@ -346,6 +377,40 @@ load_app(Context, AppId) ->
              )
     end.
 
+%% @private
+-spec load_app_from_master_account(cb_context:context(), ne_binary()) -> cb_context:context().
+load_app_from_master_account(Context, AppId) ->
+    {'ok', MasterAccountDb} = whapps_util:get_master_account_db(),
+    {'ok', MasterAccountId} = whapps_util:get_master_account_id(),
+    try
```
jamesaimonetti: I don't get the point of using `try` here.

Why not just get the JObjs from couch_mgr and case ListComp of [JObj] -> good; _ -> bad end ?

fenollp: It was to shorten the code. And because I thought that if it can't read the doc from the master account something is really wrong (and would be noticed some time earlier).
I'll amend.

jamesaimonetti: But a DB read error is different from asking for a non-existent AppId. Different errors should be returned in that case.

fenollp:
@jamesaimonetti want to take another look?

PR #945, opened 2015-04-15
Fix recording

Dinkor:
1. Store recording file only when we started recording.
2. When store file to 'local' DB dont append filename to Url (its already here).
3. Store to third_party only when enabled store_recordings and set third_party_bigcouch_host


Dinkor:
Need additional testing.

PR #947, opened 2015-04-15
KAZOO-3496: Send provision/feature keys to provisioner

macpie:


applications/crossbar/src/modules/provisioner_v5.erl
```diff
@@ -310,6 +311,10 @@ settings_datetime(JObj) ->
              ),
     wh_json:from_list(Props).
 
+-spec settings_feature_keys(wh_json:object()) -> wh_json:object().
+settings_feature_keys(JObj) ->
```
k-anderson: @macpie We should talk about this.  You will need to change what is sent based on the type of key selected as well as resolve the presence_id for the given owner in some cases.

PR #948, opened 2015-04-16
Bug: When timeout error comes from gen_server, it comes as {error, {erro...

aberke:
...r, timeout}}.  Code before then tried to convert {error, timeout} to binary --> process dies.

Similar issues could occur when other errors are surfaced by gen_server.
Note: this gen_server is from beneath the gen_listener behavior.


aberke:
Found by Alex and @k-anderson today during Babble failover testing.

When merged, please include in 3.19.

jamesaimonetti:
@aberke picked back to 3.19

PR #949, opened 2015-04-17
Allow to use <<"timeout">> as menu item

Dinkor:
With this patch we can create menus like:
"Press 1 for ...
Press 2 for ...
**Or wait to connect with operator**"


jamesaimonetti:
I added a quick doc in crossbar/doc/menus.md to mention the "timeout" child. Hopefully this can be fleshed out more fully soon. Thanks @Dinkor !

PR #951, opened 2015-04-17
Fix recording

Dinkor:
1. Store recording file on call stop, only if we started recording early.
2. When store file to 'local' DB dont append filename to Url (its already here).
3. Store to third_party only when enabled store_recordings and set third_party_bigcouch_host


fenollp:
No need for the extra `ok`s on the `lager` lines. The `lager` functions already return `ok`.

jamesaimonetti:
Added a doc for the storage options. Thanks @Dinkor !

PR #952, opened 2015-04-17
Previously if there was a module before the user module (like preflow) i...

aberke:
...n the callflow, then view would not emit the doc for the recipient user.


jamesaimonetti:
Hi @aberke, please rebase and maybe check with @lazedo to discuss your changes to this view (if they still apply after rebasing).

aberke:
Ok I see but I'll still need to test to see if the issue I noted on PR https://github.com/2600hz/kazoo/pull/953

PR #956, opened 2015-04-17
When there are multiple Dbs and no docs are found in one of them, rather...

aberke:
... than returning 404, move on to searching in the next DB.

### DIFF
--- a/applications/crossbar/src/crossbar_doc.erl
+++ b/applications/crossbar/src/crossbar_doc.erl
@@ -349,9 +349,12 @@ load_view(#load_view_params{view=View
             'false' -> IncludeOptions;
             _V -> props:delete('include_docs', IncludeOptions)
         end,
-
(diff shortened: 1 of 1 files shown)


aberke:
@jamesaimonetti related to this...
If I create a new view for the modbs, is there an easy way to refresh the modb views to load the view into the previous months' dbs?
The fact that `kazoo_modb:refresh_views` only refreshes the views for the most recent months is why I ran into this issue.
I'm still going to miss out on those previous months...

jamesaimonetti:
@aberke regarding MODb view propagation, `whapps_maintenance:refresh/0` will walk all DBs and refresh their views (only if the view(s) have changed). If there's a specific DB you want to update, `whapps_maintenance:refresh(Db)` will do so. Just use the URL-encoded version for account, number, and month-only databases.

PR #958, opened 2015-04-18
KAZOO-3565: acdc attended transfer

siplabs:
Hello,

We are trying to make it possible to transfer ACDC-managed calls (e.g. transfer to another queue). It works but demonstrates wierd behaviour at rare cases: when first transfer to another queue works ok as does the second one, but the third one fails. The agent FSM acts in wrong way and the state is wrong. Probably we are not unsubscribing from AMQP or something like that.... we have a little trouble here due to lack of deep knowledge of ACDC guts. Can you suggest something?

Logs attached to JIRA ticket.


jamesaimonetti:
@siplabs I merged the PR because everything looked good, but I'd like to keep the ticket open while we discuss the issue you've seen with multiple transfers. You can issue a second PR when the issue is identified and corrected.

PR #959, opened 2015-04-20
KAZOO-128: per account ratedeck

siplabs:


jamesaimonetti:
@siplabs Please update the `rates.md` in `crossbar/doc/` with commands for account-specific requests.

applications/crossbar/src/crossbar_doc.erl
```diff
@@ -284,15 +284,15 @@ load_view(View, Options, Context, StartKey, PageSize) ->
     load_view(View, Options, Context, StartKey, PageSize, 'undefined').
 
 load_view(View, Options, Context, StartKey, PageSize, FilterFun) ->
+    Dbs = props:get_value('databases', Options, [cb_context:account_db(Context)]),
+    ExistingDbs = lists:filter(fun couch_mgr:db_exists/1, Dbs),
```
jamesaimonetti: When `couch_mgr:get_results/3` is actually called below, the code now branches on `{'error', 'not_found'}` and continues to process the view for the next DB (if any). I think this filter is no longer necessary. Thoughts?

applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -53,18 +53,25 @@ init() ->
...
+-spec init_db() -> 'ok'.
 init_db() ->
-    _ = couch_mgr:db_create(?WH_RATES_DB),
-    couch_mgr:revise_doc_from_file(?WH_RATES_DB, 'crossbar', "views/rates.json").
+    init_db(?WH_RATES_DB).
 
+-spec authorize(cb_context:context()) -> boolean().
 authorize(Context) ->
-    authorize(Context, cb_context:req_nouns(Context)).
+    AccId = cb_context:account_id(Context),
```
jamesaimonetti: Please use `AccountId` instead; we're trying to avoid abbreviations, especially when it could conflict with other contexts (in this case `Acc` is usually short for `Accumulator` in a fold).

applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -53,18 +53,25 @@ init() ->
...
+authorize(?HTTP_GET, AccId, AccId) ->
     'true';
-authorize(_Context, _Nouns) ->
-    'false'.
+authorize(_, AccId, AuthId) ->
+    (
+     wh_util:is_in_account_hierarchy(AuthId, AccId, 'true')
+     andalso wh_services:is_reseller(AuthId)
+    )
+    orelse cb_modules_utils:is_superduper_admin(AuthId).
```
jamesaimonetti: This is the wrong module name

applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -252,10 +264,45 @@ on_successful_validation(Id, Context) ->
 %% resource.
 %% @end
 %%--------------------------------------------------------------------
--spec summary(cb_context:context()) -> cb_context:context().
-summary(Context) ->
-    crossbar_doc:load_view(?CB_LIST, [], Context, fun normalize_view_results/2).
-
+-spec summary(cb_context:context(), ne_binary(), ne_binary(), list()) -> cb_context:context().
```
jamesaimonetti: `list()` should be `ne_binaries()`.

applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -252,10 +264,45 @@ on_successful_validation(Id, Context) ->
...
+           | lists:map(
+               fun account_rates_db/1
+               ,Parents ++ [AccId]
+              )],
+    summary(Context, DBs);
+summary(Context, AccId, _AuthId, _Parents) ->
+    lager:debug("loading client rates"),
+    summary(Context, [account_rates_db(AccId)]).
+
+summary(Context, DBs) ->
```
jamesaimonetti: Missing spec

applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -252,10 +264,45 @@ on_successful_validation(Id, Context) ->
...
+    Context1 = crossbar_doc:load_view(?CB_LIST, [{databases, DBs}], Context, fun normalize_view_results/2),
```
jamesaimonetti: Tickies missing on `databases`

applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -252,10 +264,45 @@ on_successful_validation(Id, Context) ->
...
+                   _ ->
+                       Context1
+               end,
+    Data = lists:foldl(fun merge_acc/2
+                      ,[]
+                      ,cb_context:resp_data(Context2)),
+    Data1 = [Val || {_K, Val} <- Data],
+    cb_context:set_resp_data(Context2, Data1).
+
+merge_acc({K, Val}, Acc) ->
```
jamesaimonetti: Missing spec

applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -252,10 +264,45 @@ on_successful_validation(Id, Context) ->
...
+    UnDef = case Val of
+                'undefined' -> 'null';
```
jamesaimonetti: I don't understand your intent here. What is the goal of this function?

applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -300,8 +347,16 @@ check_uploaded_file(Context, _ReqFiles) ->
 %% @end
 %%--------------------------------------------------------------------
 -spec normalize_view_results(wh_json:object(), wh_json:objects()) -> wh_json:objects().
-normalize_view_results(JObj, Acc) ->
-    [wh_json:get_value(<<"value">>, JObj)|Acc].
+normalize_view_results(JObj, Acc) when is_list(Acc) ->
```
jamesaimonetti: Spec is now wrong. A 4-tuple is not a JSON key. I'd type it something more like:

```erlang
-type view_result_key() :: {api_binary(), ne_binaries(), api_binary(), ne_binaries()}.  
-type view_result_value() :: wh_json:json_term().                                          
-type view_result() :: {view_result_key(), view_result_value()}.                          
-type view_results() :: [view_result(),...] | [].           

-spec normalize_view_results(wh_json:object(), view_results()) -> view_results().
```

applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -555,3 +610,32 @@ normalize_field(<<"Surcharge">> = K, Surcharge) ->
     {K, wht_util:units_to_dollars(Surcharge)};
 normalize_field(K, V) ->
     {K, V}.
+
+-spec account_rates_db(ne_binary()) -> ne_binary().
+account_rates_db(AccountId) ->
+    FormatedAccountDb = wh_util:format_account_db(AccountId),
+    <<FormatedAccountDb/binary, "-", ?WH_RATES_DB/binary>>.
+
+-spec account_parents(cb_context:context()) -> list().
```
jamesaimonetti: Should be `ne_binaries()` instead of `list()`

applications/hotornot/src/hon_util.erl
```diff
@@ -8,9 +8,10 @@
 %%%-------------------------------------------------------------------
 -module(hon_util).
 
--export([candidate_rates/1, candidate_rates/2
-         ,matching_rates/2, matching_rates/4
+-export([candidate_rates/3, candidate_rates/4
```
jamesaimonetti: This breaks `hotornot_maintenance`. Please update that module's functions that rely on these exports.

applications/hotornot/src/hon_util.erl
```diff
@@ -20,59 +21,82 @@
...
+build_keys(Direction, <<D:1/binary, Rest/binary>>) ->
+    build_keys(Direction, Rest, D, [[D, Direction]]).
+
+build_keys(Direction, <<D:1/binary, Rest/binary>>, Prefix, Acc) ->
+    build_keys(Direction, Rest, <<Prefix/binary, D/binary>>, [[<<Prefix/binary, D/binary>>, Direction] | Acc]);
+build_keys(_, <<>>, _, Acc) -> Acc.
+
+-spec rate_dbs(ne_binary()) -> list().
+rate_dbs(AccountId) ->
+    {ok, A} = couch_mgr:open_doc(?WH_ACCOUNTS_DB, AccountId),
```
jamesaimonetti: The account doc in the account database is authoritative, not the version in the `accounts` database. Pull from the account DB instead.

applications/hotornot/src/hon_util.erl
```diff
@@ -20,59 +21,82 @@
...
 
-    lager:debug("searching for prefixes for ~s: ~p", [E164, Keys]),
-    case couch_mgr:get_results(?WH_RATES_DB, <<"rates/lookup">>, [{'keys', Keys}
-                                                                  ,'include_docs'
-                                                                 ])
-    of
-        {'ok', []}=OK -> OK;
+-spec acc_candidate_rates(ne_binary(), {'ok', list(), integer(), list()} | {'error', any()}) ->
+    {'ok', list(), integer(), list()} | {'error', any()}.
+acc_candidate_rates(RateDb, {'ok', Keys, Lvl, Acc}) ->
```
jamesaimonetti: I don't think `Lvl` is a descriptive variable name. What does it represent? Can it be named better? At the very least, don't abbreviate it.

applications/hotornot/src/hon_util.erl
```diff
@@ -20,59 +21,82 @@
...
-build_keys(<<D:1/binary, Rest/binary>>) ->
-    build_keys(Rest, D, [wh_util:to_integer(D)]).
-
-build_keys(<<D:1/binary, Rest/binary>>, Prefix, Acc) ->
-    build_keys(Rest, <<Prefix/binary, D/binary>>, [wh_util:to_integer(<<Prefix/binary, D/binary>>) | Acc]);
-build_keys(<<>>, _, Acc) -> Acc.
+normalize_rate(Rate, Lvl) ->
+    Rate1 = wh_json:get_value(<<"value">>, Rate),
+    Rate2 = wh_json:merge_jobjs(Rate1, wh_json:get_value(<<"doc">>, Rate)),
+    wh_json:set_value(<<"lvl">>, Lvl, Rate2).
```
jamesaimonetti: Same here, the `lvl` key isn't a programmer-friendly name.

applications/hotornot/src/hon_util.erl
```diff
@@ -112,8 +127,15 @@ sort_rate(RateA, RateB) ->
 
     case PrefixA =:= PrefixB of
         'true' ->
-            wh_json:get_integer_value(<<"weight">>, RateA, 100) >
-                wh_json:get_integer_value(<<"weight">>, RateB, 100);
+            LvlA = wh_json:get_integer_value(<<"lvl">>, RateA),
```
jamesaimonetti: Let's put this clause in a function like `sort_matching_prefix(RateA, RateB)`

applications/hotornot/src/hon_util.erl
```diff
@@ -20,59 +21,82 @@
...
+rate_dbs(AccountId) ->
+    {ok, A} = couch_mgr:open_doc(?WH_ACCOUNTS_DB, AccountId),
+    lists:foldl(fun acc_rate_dbs/2, [], lists:reverse([AccountId | kz_account:tree(A)]))
+    ++ [?WH_RATES_DB].
+
+-spec acc_rate_dbs(ne_binary(), list()) -> list().
+acc_rate_dbs(AccountId, Acc) ->
+    Db = wh_util:format_account_db(<<AccountId/binary, "-", ?WH_RATES_DB/binary>>),
+    case couch_mgr:db_exists(Db) of
+        true -> [Db | Acc];
```
jamesaimonetti: Don't forget to ticky the atoms! :)

applications/hotornot/src/hon_util.erl
```diff
@@ -20,59 +21,82 @@
...
-                                                                 ])
-    of
-        {'ok', []}=OK -> OK;
+-spec acc_candidate_rates(ne_binary(), {'ok', list(), integer(), list()} | {'error', any()}) ->
+    {'ok', list(), integer(), list()} | {'error', any()}.
+acc_candidate_rates(RateDb, {'ok', Keys, Lvl, Acc}) ->
+    case couch_mgr:get_results(RateDb, <<"rates/lookup">>, [{'keys', Keys}, 'include_docs']) of
         {'ok', ViewRows} ->
-            {'ok', [wh_json:get_value(<<"doc">>, ViewRow) || ViewRow <- ViewRows]};
+            {'ok', Keys, Lvl + 1, Acc ++ [normalize_rate(Rate, Lvl) || Rate <- ViewRows]};
```
jamesaimonetti: Why not build the list up and then reverse it, rather than appended each time?

applications/hotornot/src/hon_util.erl
```diff
@@ -20,59 +21,82 @@
...
+
+build_keys(Direction, <<D:1/binary, Rest/binary>>, Prefix, Acc) ->
+    build_keys(Direction, Rest, <<Prefix/binary, D/binary>>, [[<<Prefix/binary, D/binary>>, Direction] | Acc]);
+build_keys(_, <<>>, _, Acc) -> Acc.
+
+-spec rate_dbs(ne_binary()) -> list().
+rate_dbs(AccountId) ->
+    {ok, A} = couch_mgr:open_doc(?WH_ACCOUNTS_DB, AccountId),
+    lists:foldl(fun acc_rate_dbs/2, [], lists:reverse([AccountId | kz_account:tree(A)]))
+    ++ [?WH_RATES_DB].
```
jamesaimonetti: This is going to be horribly expensive for any account with a non-trivial amount of ancestor accounts (plus the global view read). We need to rethink how the appropriate ratedeck for an account will be selected. @k-anderson, thoughts or strategies?

applications/crossbar/src/crossbar_doc.erl
siplabs: @jamesaimonetti 
I removed filter and got error.
Also this code still exists in master branch. This modification is only for better looking. I can revert it :)

applications/hotornot/src/hon_util.erl
siplabs: @jamesaimonetti 
This code is just simplier for testing.
It's not too hard to rewrite it.
I'll wait for k-anderson opinion about way and strategies. :)

applications/crossbar/src/crossbar_doc.erl
jamesaimonetti: Thanks for trying, was just curious if it would work.

applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -252,10 +268,47 @@ on_successful_validation(Id, Context) ->
...
+                   _ ->
+                       Context1
+               end,
+    Data = lists:foldl(fun merge_acc/2
+                      ,[]
+                      ,cb_context:resp_data(Context2)),
+    Data1 = [Val || {_K, Val} <- Data],
+    cb_context:set_resp_data(Context2, Data1).
+
+-spec merge_acc({ne_binary(), ne_binary()}, list()) -> list().
```
jamesaimonetti: This spec still seems wrong. `Val` can't be only ne_binary() if you're testing it for 'undefined' below. And `Acc` is what? Seems like `view_results()` is the right type?

applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -555,3 +617,32 @@ normalize_field(<<"Surcharge">> = K, Surcharge) ->
...
+            'ok'
+    end.
+
+-spec set_account_db(cb_context:context()) -> cb_context:context().
+set_account_db(Context) ->
+    Db = case account_parents(Context) of
+             [] -> ?WH_RATES_DB;
+             _Parents -> account_rates_db(cb_context:account_id(Context))
+         end,
+    init_db(Db),
```
jamesaimonetti: I'd rather detect, when loading the view, if it was a 'not_found' error and init the DB (or check the DB), rather than checking the DB on every request.

applications/hotornot/src/hotornot_maintenance.erl
```diff
@@ -24,12 +24,12 @@
...
 
--spec rates_for_did(ne_binary()) -> 'ok'.
--spec rates_for_did(ne_binary(), api_binary(), trunking_options()) -> 'ok'.
-rates_for_did(DID) ->
-    rates_for_did(DID, 'undefined', []).
-rates_for_did(DID, Direction, RouteOptions) when is_list(RouteOptions) ->
-    case hon_util:candidate_rates(DID) of
+-spec rates_for_did(ne_binary(), ne_binary(), api_binary()) -> 'ok'.
+-spec rates_for_did(ne_binary(), ne_binary(), api_binary(), trunking_options()) -> 'ok'.
+rates_for_did(AccountId, DID, Direction) ->
```
jamesaimonetti: We should keep arity-2 for querying the global ratedeck since that's what most people have.

applications/hotornot/src/hon_util.erl
```diff
@@ -20,74 +21,97 @@
...
+
+build_keys(Direction, <<D:1/binary, Rest/binary>>, Prefix, Acc) ->
+    build_keys(Direction, Rest, <<Prefix/binary, D/binary>>, [[<<Prefix/binary, D/binary>>, Direction] | Acc]);
+build_keys(_, <<>>, _, Acc) -> Acc.
+
+-spec rate_dbs(ne_binary()) -> list().
+rate_dbs(AccountId) ->
+    AccountDb = wh_util:format_account_db(AccountId),
+    {'ok', A} = couch_mgr:open_doc(AccountDb, AccountId),
+    lists:foldl(fun acc_rate_dbs/2, [], lists:reverse([AccountId | kz_account:tree(A)]))
```
jamesaimonetti: While we wait for @k-anderson to weigh in, I was thinking it should, at most, query the Account, the Reseller, and then the global ratedeck.

Another thought is to add a pvt_ratedeck_db to the account doc which points to Account/Reseller/Global ratedecks.

Thoughts?

lazedo: @jamesaimonetti i would also like to see pvt_ratedeck_id to support multiple ratedecks per ratedeck db. one reseller can have several ratedeckas and apply A to a customer and B to another.

jamesaimonetti: Good point @lazedo.

siplabs: @jamesaimonetti @lazedo 
It's possible to apply different ratedecks for A and B in this patch. Reseller can modify their ratedecks. Also he may add common rates to his own ratedeck.
For example

| RateDeck | Prefix | Cost |
|----------|--------|------|
| Reseller | 123 | 10 |
| A | 1234 | 100 |
| B | 12 | 1 |

Calling to number 12345 for A will cost 100 and 10 for B.
I think it's quite easy for Reseller to maintain this ratedecks.

applications/crossbar/src/modules/cb_rates.erl
siplabs: @jamesaimonetti 
Wrong spec for `couch_mgr:db_exists/1`.
May be I got a error in `crossbar_doc:load_view/6` because in `lists:filter/2` it should be used fun which returns only `boolean()`?

applications/crossbar/priv/couchdb/views/rates.json
```diff
@@ -2,7 +2,7 @@
   ,"language":"javascript"
   ,"views":{
     "lookup": {
-      "map": "function(doc) { if (doc.deleted || doc.pvt_type !== 'rate') return; emit(parseInt(doc.prefix), null); }"
+      "map": "function(doc) { if (doc.deleted || doc.pvt_type !== 'rate') return; do_emit = function(direction) {emit([(''+doc.prefix), direction], {prefix: ''+doc.prefix, routes: (doc.routes || []).length ? doc.routes : ['^\\\\+?' + (''+doc.prefix).replace(/[^0-9]/g, '')], rate_name: doc.rate_name || 'prefix:'+doc.prefix}); }; if ((doc.direction || 'both') == 'both') { do_emit('inbound'); do_emit('outbound'); do_emit('any');} else {do_emit(doc.direction); do_emit('any');}}"
```
jamesaimonetti: I think JSON object keys must be enclosed by quotes. So should be `{\"prefix\": ''+doc.prefix,...`

siplabs: @jamesaimonetti 
I thought it's inside JS function and it's a JS-object ;)
JS-object keys MAY be quoted.

applications/hotornot/src/hon_util.erl
siplabs: @jamesaimonetti 
I fixed this.

siplabs:
@jamesaimonetti 
I rebased this pull request to current master.

aaavoip:
Are there plans to merge this in the near future?

siplabs:
@aaavoip 
this patch still need some work, we are waiting for @k-anderson and @jamesaimonetti design about  an implemented of storing and retriving rates

aaavoip:
@siplabs  Thanks for the update, much appreciated!

onnet:
Hi, any news on that? It is such a showstopper... At least several ratedecks per cluster would be a giant step forward....

siplabs:
@k-anderson Howdy! Any ideas about the way this one should move? We are ready to fix it when we have your feedback...

callforamerica:
@jamesaimonetti We're also interested in this feature.  Are there any updates on recommendations on how to better implement it?  We're currently running multiple instances of kazoo in order to rate accounts differently for our 3 different brands.  We would like to collaborate with @siplabs in further redevelopment  on the appropriate direction you suggest.

onnet:
Hi guys,

Is there any thoughts to add an option to hon_rater to do rate lookup at a third_party API.

For example we can make an API request to a separate service/database by providing caller and callee numbers and recieve needed rate back.
May be we can also add an account_id to such API call or whatever...

This can eliminate need of running multiple instances of kazoo :))  

Whats your thoughts? Will such PR be of any interest?

Regards,
Kirill

lazedo:
@onnet i think that is not a good option, you can however create an alternative whapp that listens to rate requests and does whatever you want.

onnet:
@lazedo Absolutely agree. But one ratedeck per cluster is more of a demo instead of production app. And it feels like something should be done to this app to make it usable.

lazedo:
@onnet if there's something to do to this app (and there is), it should be done within the app and not extend it only to call outside service. you can create for example extend cgrates (or other) to listen/reply to amqp requests. if there's a need for some tweaks in the amqp paylooad, that would go into wapi_rates.
anyway, i think this app will move forward, will have multiple ratedecks tied to service plans / carriers with inheritance, auto calculation of margins, etc...

onnet:
@lazedo I keep fingers crossed, but since this PR is almost year old it's difficult to predict its perspective.

PR #960, opened 2015-04-20
Update service_plans.md

onnet:
Some notes just to not to keep local


jamesaimonetti:
Thanks @onnet, keep it coming!

PR #961, opened 2015-04-20
Kazoo 3496

macpie:


fenollp:
Looks great to me! (also I need this)

applications/crossbar/src/modules/provisioner_v5.erl
```diff
@@ -310,6 +309,82 @@ settings_datetime(JObj) ->
...
+            Type = wh_json:get_binary_value(<<"type">>, Value),
+            V = wh_json:get_binary_value(<<"value">>, Value),
+            FeatureKey = get_feature_key(Type, V, Brand, Family, AccountId),
+            maybe_add_feature_key(Key, FeatureKey, Acc)
+        end
+        ,wh_json:new()
+        ,FeatureKeys
+    ).
+
+-spec get_feature_key(ne_binary(), wh_json:object(), ne_binary(), ne_binary(), ne_binary()) -> wh_json:object().
```
jamesaimonetti: Wrong spec. `Value` isn't a JSON object, is it? And `Brand` isn't ne_binary(). And `Family` isn't ne_binary()

jamesaimonetti: It also can return 'undefined' so should be `api_object()` for the return.

applications/crossbar/src/modules/provisioner_v5.erl
```diff
@@ -310,6 +309,82 @@ settings_datetime(JObj) ->
...
+    end;
+get_feature_key(<<"parking">> = Type, Value, Brand, Family, _AccountId) ->
+    wh_json:from_list([
+        {<<"label">>, <<"Park ", Value/binary>>}
+        ,{<<"value">>, <<"*3", Value/binary>>}
+        ,{<<"type">>, get_feature_key_type(Type, Brand, Family)}
+    ]).
+
+-spec get_feature_key_type(ne_binary(), ne_binary(), ne_binary()) -> wh_json:object().
+get_feature_key_type(Type, Brand, Family) ->
```
jamesaimonetti: I think `Brand` and `Family` are JSON objects, no? And the return could be 'undefined'

applications/crossbar/src/modules/provisioner_v5.erl
```diff
@@ -310,6 +309,82 @@ settings_datetime(JObj) ->
...
+    wh_json:get_first_defined([[Brand, Family, Type], [Brand, <<"_">>, Type]], ?FEATURE_KEYS).
+
+-spec get_user(ne_binary(), ne_binary()) -> wh_json:object().
+get_user(AccountId, UserId) ->
```
jamesaimonetti: `couch_mgr:open_doc/2` returns `{'ok', wh_json:object()} | {'error', _}`

applications/crossbar/src/modules/provisioner_v5.erl
```diff
@@ -310,6 +309,82 @@ settings_datetime(JObj) ->
...
+    AccountDb = wh_util:format_account_id(AccountId, 'encoded'),
+    couch_mgr:open_doc(AccountDb, UserId).
```
jamesaimonetti: Should probably use `open_cache_doc/2`

macpie: I want to keep it open_doc just to make sure data is accurate. Function is spawned anyway

PR #965, opened 2015-04-21
KAZOO-3520: Added eradius library and the Circlemaker app skeleton.

siplabs:
The eradius library was adapted for Kazoo and added into deps to support working with RADIUS protocol. Also was added base code of the Circlemaker application which will provide functionality to interoperate with an AAA-servers using RADIUS and DIAMETER protocols.

### DIFF
--- a/applications/callflow/src/module/cf_aaa.erl
+++ b/applications/callflow/src/module/cf_aaa.erl
@@ -30,6 +30,7 @@
 -include("../callflow.hrl").
 
 -define(CCV, <<"Custom-Channel-Vars">>).
+-define(CMSG, <<"Custom-Msg">>).
 
 %%--------------------------------------------------------------------
 %% @public
@@ -49,7 +50,8 @@ handle(Data, Call) ->
     CallProp2 = wh_json:set_value(<<"AAA-Request-Type">>, <<"cf_aaa">>, CallProp1),
     CallProp3 = wh_json:set_values(RequestFields1, CallProp2),
     CallProp4 = wh_json:set_values(wh_api:default_headers(?APP_NAME, ?APP_VERSION), CallProp3),
-    maybe_do_radius_request(CallProp4, Data, Call).
+    CallProp5 = wh_json:set_value(?CMSG, CallProp4, wh_json:new()),
+    maybe_do_radius_request(CallProp5, Data, Call).
 
 -spec maybe_do_radius_request(wh_json:object(), wh_json:object(), whapps_call:call()) -> any().
 maybe_do_radius_request(Request, Data, Call) ->
@@ -74,8 +76,9 @@ maybe_do_radius_request(Request, Data, Call) ->
 get_last_part_of_key(Key) when is_list(Key) -> lists:last(Key);
 get_last_part_of_key(Key) when is_binary(Key) -> Key.
 
-process_response(Response, Data, Call) ->
-    lager:debug("Response received: ~p", [Response]),
+process_response(ResponseWrapped, Data, Call) ->
+    lager:debug("Response received: ~p", [ResponseWrapped]),
+    Response = wh_json:get_value(?CMSG, ResponseWrapped),
     ResponseFields = wh_json:get_value(<<"response_fields">>, Data),
     lager:debug("Response fields: ~p", [ResponseFields]),
     Channel = wh_json:get_value(<<"channel">>, Data, <<"a">>),
--- a/applications/callflow/src/module/cf_aaa.erl
+++ b/applications/callflow/src/module/cf_aaa.erl
@@ -50,7 +50,7 @@ handle(Data, Call) ->
     CallProp2 = wh_json:set_value(<<"AAA-Request-Type">>, <<"cf_aaa">>, CallProp1),
     CallProp3 = wh_json:set_values(RequestFields1, CallProp2),
     CallProp4 = wh_json:set_values(wh_api:default_headers(?APP_NAME, ?APP_VERSION), CallProp3),
-    CallProp5 = wh_json:set_value(?CMSG, CallProp4, wh_json:new()),
+    CallProp5 = wh_json:set_value(?CMSG, CallProp4, CallProp4),
     maybe_do_radius_request(CallProp5, Data, Call).
 
 -spec maybe_do_radius_request(wh_json:object(), wh_json:object(), whapps_call:call()) -> any().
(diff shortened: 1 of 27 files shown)


jamesaimonetti:
@siplabs can you rebase to the latest master? I'd prefer to review it knowing it would merge in properly!

applications/circlemaker/doc/index.md
```diff
@@ -0,0 +1,148 @@
+/*                                                                                                                                                                                                                                                                                                 
+Section: Circlemaker
+Title: Circlemaker
+Language: en-US
+*/
+
+# Overview
```
jamesaimonetti: This isn't really an overview of Circlemaker ;)

applications/circlemaker/priv/couchdb/views/aaa.json
```diff
@@ -0,0 +1,12 @@
+{
+  "_id": "_design/aaa",
+  "language": "javascript",
+  "views": {
+    "fetch_doc": {
+      "map": "function(doc) { if (doc.pvt_type != 'aaa' || doc.pvt_deleted) return; emit(doc._id, null); }"
```
jamesaimonetti: Why two views that do the exact same thing?

applications/circlemaker/src/cm_config.erl
```diff
@@ -0,0 +1,143 @@
...
+get_aaa_mode(Doc) ->
+    wh_json:get_json_value(<<"aaa_mode">>, Doc, <<"off">>).
+
+%%--------------------------------------------------------------------
+%% @public
+%% @doc
+%% Get servers list.
+%% @end
+%%--------------------------------------------------------------------
+-spec get_servers_list(wh_json:object()) -> ne_binary().
```
jamesaimonetti: This should return `wh_json:object()` no?

applications/circlemaker/src/cm_config.erl
```diff
@@ -0,0 +1,143 @@
...
+get_servers_list(Doc) ->
+    wh_json:get_json_value(<<"servers">>, Doc).
+
+%%--------------------------------------------------------------------
+%% @public
+%% @doc
+%% Get AAA configuration document from a database
+%% @end
+%%--------------------------------------------------------------------
+-spec get_aaa_doc(wh_json:object() | atom()) -> ne_binary().
```
jamesaimonetti: Wrong return in the spec

applications/circlemaker/src/cm_radius.erl
```diff
@@ -0,0 +1,42 @@
...
+-export([add_request_authz/2, add_request_start/2, add_request_stop/2]).
+
+%%--------------------------------------------------------------------
+%% @public
+%% @doc
+%%
+%% @end
+%%--------------------------------------------------------------------
+-spec add_request_authz(wh_json:object(), boolean()) -> any().
+add_request_authz(_JObj, 'true') ->
```
jamesaimonetti: I don't understand why all the handlers are 'not_implemented' for authz requests. What does Circlemaker *do* ?

siplabs: Yes, you are right, currently the Circlemaker app does nothing. We are decided to send this part of already written code to reduce size of code for the future review of the [KAZOO-3521](https://2600hz.atlassian.net/browse/KAZOO-3521).

applications/circlemaker/doc/index.md
siplabs: A list of TODOs was added instead of existing text, and it will be filled later.

applications/circlemaker/priv/couchdb/views/aaa.json
siplabs: Fixed. The second view is used for the "aaa_dict" document type.

siplabs:
@jamesaimonetti Done. The branch was rebased to the latest master.

siplabs:
The branch was updated. Authentication (authn) via RADIUS protocol was implemented.

k-anderson:
@siplabs Sorry this needs a rebase again.

One last request before it is accepted.   I would really like to see the authn reply originate from the circlemaker application.  That is to say, if the aaa mode enables circlemaker then the registrar will forward the authn request and die.  After that circlemaker would be authoritative sending an authn reply or  failure.  This would remove the need for the message to return via registrar and the tracking/correlation in the listener state.  Thoughts?

siplabs:
@k-anderson 
When I started to implement this algorithm, I thought about some kind of single point to make a request and catch an response. I mean if we made a request from registrar, the registrar should process corresponding response. So we get a predictable behavior here.
But yes, if the registrar died, then it couldn't process a response from circlemaker. If the registrar died, then it means that all subsequent attempts to register will fail, because the registrar is responsible for registration process, and it's reasonable, I suppose.
What do you think about it? If you still like to see sending the authn reply from the circlemaker, then I'll begin this refactoring.

k-anderson:
@siplabs 
I see what you are seeing however I propose that it is not registrar that is processing the request.  IMHO registrar is proxing the request and handing authority to circlemaker, the request is from ecallmgr.

ecallmgr: Hey, can anybody auth this?
registrar: Oh oh oh I can I can....
registrar: Actually I can't the account is configured for circlemaker....
registrar: hey circlemaker can you auth this for ecallmgr?
circlemaker: hey ecallmgr, here are the creds
   -or-
circlemaker: hey ecallmgr, sorry that is a no go


Ideally we would be able to bypass registrar so it would look more like this:
ecallmgr: Hey, can anybody auth this?
registrar: hey ecallmgr, I can't
circlemaker: hey ecallmgr, here are the creds
   -or-
circlemaker: hey ecallmgr, nope

This approach is the end goal (and would let us move the GSM auth to its own application as well and future alternate auth types such as 'webhook') but we will need resolve how to  block the amqp worker but wait for all applications that might return.  Therefore, for the timebeing proxing the request via registrar is a good approach.  We can then file a ticket to work out the long term strategy and separate circlemaker, GSM auth, and registrar.

k-anderson:
@siplabs Howdy guys, just wondering what the status of this is.  Did my last commit make sense?

siplabs:
This one is to be re-submitted soon.

jamesaimonetti:
@siplabs any updates here?

k-anderson:
@siplabs This was opened almost a year ago! Last we discussed about this you mentioned you were going to push a much improved version so I am going to close this (and other stale pull requests).  We are still interested in this module if you are still willing to push the new version! ;)

PR #966, opened 2015-04-21
KAZOO-3223: Fix multipart data

macpie:
for @lazedo to review

### DIFF
--- a/applications/crossbar/doc/multipart.md
+++ b/applications/crossbar/doc/multipart.md
@@ -1,0 +1,31 @@
+/*
+Section: Crossbar
+Title: Multi-part Request
+Language: en-US
+*/
+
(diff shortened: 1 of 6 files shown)


jamesaimonetti:
@macpie added some fixes.

Also, can you include, in `faxes.md` or `multipart.md` or somewhere, how you tested this code? How would an API user do requests using multipart bodies?

lazedo:
@macpie you can include the script and files i sent you. just make sure to change the url and account names. i test it and i'm good.
multipart/mixed is not needed in crossbar.hrl because it is handled in cb_faxes, but we can leave there as a reference.

jamesaimonetti:
Already fixed and merged before your commit @macpie

PR #971, opened 2015-04-22
KAZOO-3573: VoiceFabric TTS engine body encoding fix

siplabs:


core/whistle_apps-1.0.0/src/whapps_speech.erl
```diff
@@ -205,13 +205,13 @@ voicefabric_request_body(<<"multipart">>, Data) ->
     Headers = [{"Content-Type"
                  ,"multipart/form-data; charset=UTF-8; boundary=" ++ erlang:binary_to_list(Boundary)
                 }],
-    Body = iolist_to_binary([[<<Boundary/binary,
+    Body = iolist_to_binary([[<<"--", Boundary/binary,
```
jamesaimonetti: Something I missed initially, but I think you can pass the iolist() to ibrowse directly, rather than running `iolist_to_binary/1` on it first. May save a bit of time!

PR #973, opened 2015-04-22
Run tests

jamesaimonetti:
Running `make test` from the root will run tests in `core` and `applications` (skipping `deps`). You can also run `make test` in `core` or `applications`, as well as in the individual core libs or apps.


jamesaimonetti:
You will need PropEr in the path (I put it in $ERLANG/lib/proper-X.Y.Z myself)

v6:
// , Whoa, excellent. 

Is this how the [tests are being run currently][∅]? 

[∅]: https://travis-ci.org/2600hz/kazoo

jamesaimonetti:
@v6 Travis does a host of things for us, including compiling to enure no build errors, running tests, Dialyzer, etc. You can run the tests yourself by issuing `make test` in the root Kazoo directory or in one of the core/applications directories (if that folder has tests to run).

PR #978, opened 2015-04-23
KAZOO-3570: replace number manager`s global cache with a local one

fenollp:


core/whistle_number_manager-1.0.0/src/wh_number_manager.erl
```diff
@@ -136,12 +136,16 @@ lookup_account_by_number('undefined') ->
...
-    case wh_cache:peek({'account_lookup', Number}) of
+    Key = {'account_lookup', Number},
+    case wh_cache:peek_local(?WNM_NUMBER_CACHE, Key) of
         {'ok', Ok} -> Ok;
         {'error', 'not_found'} ->
             case fetch_account_by_number(Number) of
                 {'ok', _, _}=Ok ->
-                    wh_cache:store({'account_lookup', Number}, Ok),
+                    CacheProps = [{'origin', [{'db', wnm_util:number_to_db_name(Number), Number}
+                                              ,{'type', <<"number">>}
```
jamesaimonetti: I don't think we want to flush this entry if *any* number is changed.

jamesaimonetti:
Manually merged

PR #983, opened 2015-04-27
KAZOO-3499 alternative lookup by account realm

lazedo:


applications/fax/src/fax_smtp.erl
```diff
@@ -458,6 +459,82 @@ maybe_faxbox_owner(#state{faxbox=FaxBoxDoc}=State) ->
             end
     end.
 
+-spec maybe_faxbox_domain(state()) -> state().
+maybe_faxbox_domain(#state{faxbox_email=Domain}=State) ->
+    ViewOptions = [{'key', Domain}],
+    case couch_mgr:get_results(?WH_ACCOUNTS_DB, <<"accounts/listing_by_realm">>, ViewOptions) of
+        {'ok', [JObj]} -> AccountId = wh_json:get_value([<<"value">>, <<"account_id">>], JObj),
+                          maybe_faxbox_by_owner_email(AccountId, State);
+        _ ->
```
jamesaimonetti: Is there any value in differentiating `{'ok', []}` vs `{'error', _E}` in the case clauses?

applications/crossbar/priv/couchdb/account/users.json
```diff
@@ -16,7 +16,10 @@
        },
        "creds_by_sha": {
            "map": "function(doc) { if(!doc.pvt_sha1_auth || (doc.pvt_type != 'user' || doc.pvt_deleted)) return; emit( doc.pvt_sha1_auth, {'owner_id': doc._id, 'account_id': doc.pvt_account_id} ); }"
-       }
+       },
+       "list_by_email": {
+            "map": "function(doc) { if( doc.pvt_type != 'user' || doc.pvt_deleted) return; if( !doc.email || doc.email == '' ) return ;emit( doc.email, {'owner_id': doc._id, 'account_id': doc.pvt_account_id} ); }"
```
jamesaimonetti: Why not just `emit(doc.email, null);` ? `doc._id` will be in the `id` of the view result, and this is in the account's DB, right? So we already know the account.

lazedo: followed the logic for other views like creds_by_sha, do you want to change all the views ?

applications/fax/src/fax_smtp.erl
lazedo: i don't think there is a value here.

applications/crossbar/priv/couchdb/account/users.json
jamesaimonetti: No, just this one would be fine by me, for now.

applications/fax/src/fax_smtp.erl
```diff
@@ -458,6 +459,82 @@ maybe_faxbox_owner(#state{faxbox=FaxBoxDoc}=State) ->
...
+-spec maybe_faxbox_by_owner_email(ne_binary(), state()) -> state().
+maybe_faxbox_by_owner_email(AccountId, #state{from=From}=State) ->
+    ViewOptions = [{'key', From}],
+    AccountDb = wh_util:format_account_db(AccountId),
+    case couch_mgr:get_results(AccountDb, <<"users/list_by_email">>, ViewOptions) of
+        {'ok', []} -> 
+            lager:debug("user ~s does not exist in account ~s, trying by rules",[From, AccountId]),
+            maybe_faxbox_by_rules(AccountId, State);
+        {'ok', [JObj]} -> 
+            OwnerId = wh_json:get_value([<<"value">>, <<"owner_id">>], JObj),
```
jamesaimonetti: If the view changes, this becomes `wh_json:get_value(<<"id">>, JObj)`

PR #984, opened 2015-04-27
KAZOO-3504: Not loading skel/system notifications for non system admins

macpie:
I know using process dictionary is not the best way but as I dont have access to the context. The only solution would be after loading the view which would mean that I have to re-do the resp breaking the paginations and stuff ...

### DIFF
--- a/applications/crossbar/src/modules/cb_notifications.erl
+++ b/applications/crossbar/src/modules/cb_notifications.erl
@@ -819,8 +819,6 @@ attachment_name_by_media_type(CT) ->
 %%--------------------------------------------------------------------
 -spec summary(cb_context:context()) -> cb_context:context().
 summary(Context) ->
-    Account = cb_context:auth_account_id(Context),
-    erlang:put(<<"is_system_admin">>, wh_util:is_system_admin(Account)),
     case cb_context:account_db(Context) of
         'undefined' -> summary_available(Context);
         _AccountDb -> summary_account(Context)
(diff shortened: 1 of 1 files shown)


applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -819,6 +819,8 @@ attachment_name_by_media_type(CT) ->
 %%--------------------------------------------------------------------
 -spec summary(cb_context:context()) -> cb_context:context().
 summary(Context) ->
+    Account = cb_context:auth_account_id(Context),
+    erlang:put(<<"is_system_admin">>, wh_util:is_system_admin(Account)),
```
jamesaimonetti: Why not, when calling `crossbar_doc:load_view/3` in `summary_{available,account}`, set the normalize function based on `wh_util:is_system_admin(Account)`?

```erlang
NomalizeFun = select_normalize_function(wh_util:is_system_admin(Account)),
crossbar_doc:load_view(?CB_LIST, [], Context, NormalizeFun)
...

select_normalize_function('true') -> fun normalize_available/2;
select_normalize_function('false') -> fun normalize_non_admin_available/2.
```

Naming is not the greatest but that should avoid the process dictionary and be clearer in intent.

applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -908,8 +910,20 @@ merge_fold(Overridden, Acc) ->
...
 normalize_available(JObj, Acc) ->
-    [wh_json:get_value(<<"value">>, JObj) | Acc].
+    IsSystemAdmin = erlang:get(<<"is_system_admin">>),
+    normalize_available(JObj, Acc, IsSystemAdmin).
+
+normalize_available(JObj, Acc, 'true') ->
+    [wh_json:get_value(<<"value">>, JObj) | Acc];
+normalize_available(JObj, Acc, 'false') ->
+    Value = wh_json:get_value(<<"value">>, JObj),
+    case wh_json:get_value(<<"category">>, Value) of
```
jamesaimonetti: use `kz_notification:category(Value)`

macpie:
Good idea, I'll do that.

PR #985, opened 2015-04-28
KAZOO-3586 fix permissions for non super duper admin accounts

lazedo:


applications/crossbar/src/api_util.erl
```diff
@@ -720,6 +736,30 @@ is_permitted_nouns(Req, Context0, _Nouns) ->
...
+is_permitted_nouns(Req, Context, _ReqVerb, []) ->
+    lager:debug("no one authz'd the request"),
+    ?MODULE:halt(Req, cb_context:add_system_error('forbidden', Context));
+is_permitted_nouns(Req, Context0, _ReqVerb, [{Mod, Params} | ReqNouns]) ->
+    Event = api_util:create_event_name(Context0, <<"authorize.", Mod/binary>>),
+    Payload = [Context0 | Params],
+    case crossbar_bindings:succeeded(crossbar_bindings:map(Event, Payload)) of
+        [] ->
+            lager:debug("failed to authorize : ~p", [Mod]),
+            is_permitted_nouns(Req, Context0, _ReqVerb, ReqNouns);
```
k-anderson: I dont think we should loop the ReqNouns.  Contrived example but this would cause the auth for:

v1/accounts/123/search
as well as
v1/search

which may not be expected.  I think using the final token, which is the end of the URL chain, is expected (as we do for execute).

applications/crossbar/src/modules/cb_search.erl
```diff
@@ -62,6 +66,14 @@ allowed_methods() ->
 -spec resource_exists() -> 'true'.
 resource_exists() -> 'true'.
 
+-spec authenticate(cb_context:context()) -> 'true'.
+authenticate(Context) ->
```
k-anderson: This (and the binding) is not required.  cb_token_auth will take care of this.  You should only need to bind for authorization since there is no account id.

PR #989, opened 2015-04-29
KAZOO-3529: if the inception indicates a forwarded call originating from...

k-anderson:
... on-net then use the external caller id


fenollp:
Manually merged?

fenollp:
Ah, moved into #991

PR #992, opened 2015-04-29
KAZOO-3430 handle CHANNEL_DISCONNECTED in waiting for unpark call

lazedo:


lazedo:
mistyped the branch, should be KAZOO-3430

PR #993, opened 2015-04-29
KAZOO-3593 issue registration sync

lazedo:


applications/ecallmgr/src/ecallmgr_registrar.erl
```diff
@@ -435,6 +437,14 @@ handle_cast({'flush', Realm}, State) ->
 handle_cast({'flush', Username, Realm}, State) ->
     _ = ets:delete(?MODULE, registration_id(Username, Realm)),
     {'noreply', State};
+handle_cast({'gen_listener', {'created_queue', Q}}, State) ->
+    {'noreply', State#state{queue=Q}};
+handle_cast({'gen_listener',{'is_consuming', 'true'}}, #state{queue=Q}=State) ->
+    wapi_registration:publish_sync(
+      [{<<"Queue">>, Q}
```
jamesaimonetti: Typically Server-ID is used to identify the targeted queue name that responders should use when replying.

On that note, shouldn't it be 2 wapi functions, `sync_req` and `sync_resp`? `sync_req` being this publish to a topic exchange, and `sync_resp` being targeted exchange publishes to the `Server-ID` of the `sync_req`.

Finally, I don't see any processing binding for these sync messages; is that to come?

lazedo: Hi,

sync replies are issued by kamailio who doesn't use the bindings.

but we can add it.

jamesaimonetti: I see. I wonder how we can let future readers know this? Perhaps for now, add a comment in the wapi module?

PR #994, opened 2015-04-29
KAZOO-3592: Correct activation charges/ only apply min amount on quick s...

macpie:
...ale

### DIFF
--- a/applications/crossbar/src/crossbar_services.erl
+++ b/applications/crossbar/src/crossbar_services.erl
@@ -120,13 +120,13 @@ extract_item_from_category(_, ItemJObj, Acc) ->
                           ,wh_transaction:transactions()
                           ,integer()) -> wh_transaction:transactions().
 create_transactions(Context, Item, Acc) ->
-    Quantity = wh_json:get_value(<<"quantity">>, Item),
+    Quantity = wh_json:get_integer_value(<<"quantity">>, Item, 0),
     create_transactions(Context, Item, Acc, Quantity).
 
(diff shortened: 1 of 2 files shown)


applications/crossbar/src/crossbar_services.erl
```diff
@@ -120,24 +120,21 @@ extract_item_from_category(_, ItemJObj, Acc) ->
                           ,wh_transaction:transactions()
                           ,integer()) -> wh_transaction:transactions().
 create_transactions(Context, Item, Acc) ->
-    Quantity = wh_json:get_integer_value(<<"quantity">>, Item, 0),
+    Quantity = wh_json:get_value(<<"quantity">>, Item),
```
jamesaimonetti: I think you need to rebase your branch or something. I don't think this is a good change, nor the one below for getting `activiation_charges`

core/braintree-1.0.0/src/braintree_transaction.erl
```diff
@@ -146,11 +141,20 @@ sale(CustomerId, Transaction) ->
     create(Transaction#bt_transaction{type=?BT_TRANS_SALE, customer_id=CustomerId}).
 
 quick_sale(CustomerId, Amount) ->
-    sale(CustomerId, #bt_transaction{amount=wh_util:to_binary(Amount)}).
+    MinAmount = whapps_config:get_float(<<"braintree">>, <<"min_amount">>, 5.00),
```
jamesaimonetti: Should make this a macro or its own function

PR #995, opened 2015-04-30
KAZOO-3594: get users VCards

siplabs:


siplabs:
This one is the first of patches to enable more profile data both for Users and for List elements (contact lists and in future - Roster lists).
VCARD format support is introduced as well.

jamesaimonetti:
@siplabs The v1 API is locked and being deprecated once v2 is finished. Please code any new APIs to use v2

applications/crossbar/priv/couchdb/schemas/accounts.json
```diff
@@ -175,12 +175,18 @@
...
-            "default":"en-us"
-        }
+    "language": {
+      "type": "string",
+      "required": false,
+      "name": "Language",
+      "description": "The language for this account",
+      "default":"en-us"
+    },
+    "org": {
```
jamesaimonetti: What is the difference between "name" and "org"? Why do we need both?

applications/crossbar/priv/couchdb/schemas/profile.json
```diff
@@ -0,0 +1,81 @@
...
+                        "description": "To specify the address"
+                    }
+                }
+            }
+        },
+        "note": {
+            "type": "string",
+            "required": false,
+            "name": "Note",
+            "description": "To specify supplemental information or a comment that is associated with the"
```
jamesaimonetti: Seems the description got truncated :)

applications/crossbar/src/modules_v1/cb_users_v1.erl
```diff
@@ -297,6 +341,180 @@ get_channels(Context) ->
...
+    {Key, <<"3.0">>};
+card_field(Key, _)
+  when Key =:= <<"END">> ->
+    {Key, <<"VCARD">>};
+card_field(Key, JObj)
+  when Key =:= <<"FN">> ->
+    FirstName = wh_json:get_value(<<"first_name">>, JObj),
+    LastName = wh_json:get_value(<<"last_name">>, JObj),
+    MiddleName = wh_json:get_value(<<"middle_name">>, JObj),
+    {Key, iolist_join(<<" ">>, [X || X <- [FirstName, MiddleName, LastName], X =/= undefined, X =/= [], X =/= <<>>])};
```
jamesaimonetti: Maybe use `props:filter_empty([Firstname, MiddleName, Lastname])` instead of all the guards and the list comprehension?

jamesaimonetti: Why not use `wh_util:join_binary(Bins, <<" ">>)` as well?

applications/crossbar/src/modules_v1/cb_users_v1.erl
```diff
@@ -297,6 +341,180 @@ get_channels(Context) ->
...
+vcard_normalize_val(Val) when is_binary(Val) ->
+    vcard_escape_chars(Val).
+
+-spec vcard_normalize_type(list() | {ne_binary(), ne_binary()} | ne_binary()) -> ne_binary().
+vcard_normalize_type(T) when is_list(T) -> iolist_join(<<";">>, [vcard_normalize_type(X) || X <- T]);
+vcard_normalize_type({T, V}) -> iolist_join(<<"=">>, [T, V]);
+vcard_normalize_type(T) -> T.
+
+%-spec card_field(ne_binary(), wh_json:object()) -> {vcard_type_spec(), vcard_val()}.
+card_field(Key, _)
```
jamesaimonetti: One, why is the spec commented out?

Two, why not pattern match in the function header? `card_field(<<"BEGIN">> = Key, _) -> {Key, <<"VCARD">>};` and `card_field(<<"VERSION">> = Key, _) -> {Key, <<"3.0">>};`, etc...

applications/crossbar/src/modules_v1/cb_users_v1.erl
```diff
@@ -297,6 +341,180 @@ get_channels(Context) ->
...
+card_field(Key = <<"NOTE">>, JObj) ->
+    {Key, element(2, card_field(<<"note">>, JObj))};
+card_field(Key = <<"TITLE">>, JObj) ->
+    {Key, element(2, card_field(<<"title">>, JObj))};
+card_field(Key = <<"ROLE">>, JObj) ->
+    {Key, element(2, card_field(<<"role">>, JObj))};
+card_field(Key = <<"TZ">>, JObj) ->
+    {Key, element(2, card_field(<<"timezone">>, JObj))};
+card_field(Key = <<"NICKNAME">>, JObj) ->
+    Val = case element(2, card_field(<<"nicknames">>, JObj)) of
```
jamesaimonetti: I'd write this as:

```erlang
Val = case card_field(<<"nicknames">>, JObj) of
    {_, 'undefined'} -> [];
    {_, V} -> V
end,
```

jamesaimonetti: Or why not just access the property directly? `wh_json:get_value(<<"nicknames">>, JObj, [])`

siplabs:
@jamesaimonetti I'll move code to cb_users_v2. And I want to move vcard-related code somewhere else, but I don't know where. Can you tell appropriate place/module?

jamesaimonetti:
@siplabs In `core/kazoo_documents/src` create a kzd_users.erl (I'm moving the prefix from `kz_` to `kzd_` so ignore the ones currently prefixed by `kz_`). I would then create a function `kzd_user:to_vcard(JObj)` which takes the `UserJObj` and creates the vCard-formatted version.

So `cb_users_v2:convert_to_vcard(Context)` would become:
```erlang
convert_to_vcard(Context) ->
    VCard = kzd_user:to_vcard(cb_context:doc(Context)),
    cb_context:set_resp_data(Context, VCard).
```

Something like that :)

applications/crossbar/priv/couchdb/schemas/accounts.json
siplabs: org is a full legal name of the organization. Names (as practice shows) are usually shortened or even can be not connected to organization name in any possible way

applications/crossbar/src/modules_v1/cb_users_v1.erl
siplabs: `props:filter_empty/1` is for proplist, it doesn't filter list of binaries

applications/crossbar/priv/couchdb/schemas/accounts.json
siplabs: added to description

siplabs:
@jamesaimonetti 
using API v2

applications/crossbar/src/modules_v2/cb_users_v2.erl
```diff
@@ -614,3 +663,45 @@ username_doc_id(Username, Context, _AccountDb) ->
...
+
+-spec set_photo(wh_json:object(), cb_context:context()) -> wh_json:object().
+set_photo(JObj, Context) ->
+    UserId = wh_json:get_value(<<"id">>, cb_context:doc(Context)),
+    Attach = crossbar_doc:load_attachment(UserId, ?PHOTO, Context),
+    case cb_context:resp_status(Attach) of
+        'error' -> JObj;
+        'success' ->
+            Data = cb_context:resp_data(Attach),
+            CT = wh_json:get_value(<<"content_type">>
```
jamesaimonetti: Try using `wh_doc:attachment_content_type(cb_context:doc(Context), ?PHOTO)` instead.

siplabs: @jamesaimonetti 
done

jamesaimonetti:
Thanks! One quick request: it would be great to add some unit tests to the `to_vcard/1` function. In `core/kazoo_documents-1.0.0/test`, create a `kzd_user_test.erl` to test the various user JSON to VCARD mappings. Thanks!

PR #996, opened 2015-04-30
KAZOO-3589 check cloud registration on get

lazedo:


applications/crossbar/src/modules/cb_faxboxes.erl
```diff
@@ -367,6 +384,23 @@ is_faxbox_email_global_unique(Email, FaxBoxId) ->
...
+-spec maybe_reregister_cloud_printer(cb_context:context()) -> cb_context:context().
+-spec maybe_reregister_cloud_printer(api_binary(), cb_context:context()) -> cb_context:context().
+maybe_reregister_cloud_printer(Context) ->
+    CurrentState = wh_json:get_value(<<"pvt_cloud_state">>, cb_context:doc(Context)),
+    Ctx = maybe_reregister_cloud_printer(CurrentState, Context),
+    Ctx1 = case wh_json:get_value(<<"pvt_cloud_state">>, cb_context:doc(Ctx)) of
+               CurrentState -> Ctx;
+               _ -> crossbar_doc:save(Ctx)
+           end,
+    cb_context:set_resp_data(Ctx1, wh_doc:public_fields(leak_private_fields(cb_context:doc(Ctx1)))).
```
jamesaimonetti: I think you should check that Ctx1 is the result of a successful save (check resp_status).

PR #998, opened 2015-04-30
Kazoo 3505

macpie:


applications/crossbar/src/cb_apps_util.erl
```diff
@@ -74,6 +75,41 @@ is_authorized(AccountDoc, UserId, App) ->
...
+
+-spec filter_app(ne_binary(), wh_json:object(), ne_binary()) -> wh_json:object().
+filter_app(AccountId, App, <<"port">>) ->
+    lager:debug("filtering port application"),
+    ResellerId = wh_services:find_reseller_id(AccountId),
+    MaybeHide =
+        case kz_whitelabel:fetch(ResellerId) of
+            {'error', _R} ->
+                lager:error("failed to load whitelabel doc for ~s: ~p", [ResellerId, _R]),
+                'false';
```
jamesaimonetti: Shouldn't we hide it, as a precaution, on errors? Maybe show it on `{'error', 'not_found'}` but otherwise hide it?

applications/crossbar/src/modules/cb_port_requests.erl
```diff
@@ -636,6 +636,44 @@ read_descendants_fold(Account, Acc, Context) ->
...
+    maybe_read_descendant(Context, AccountId, wh_services:is_reseller(AccountId)).
+
+maybe_read_descendant(Context, AccountId, 'true') ->
+    Authority =
+        case kz_whitelabel:fetch(AccountId) of
+            {'error', _R} -> 'undefined';
+            {'ok', JObj} ->
+                kz_whitelabel:port_authority(JObj)
+        end,
+    case Authority of
```
jamesaimonetti: This can probably be a function, used by the other function clause's `case Authority of` too.

PR #999, opened 2015-05-01
Xref -- get excited!

fenollp:


core/kazoo_documents-1.0.0/Makefile
```diff
@@ -3,6 +3,7 @@ ROOT = ../..
 
 EBINS = $(shell find $(ROOT)/core/whistle-* -maxdepth 2 -name ebin -print) \
 	$(shell find $(ROOT)/deps/cowboy-* -maxdepth 2 -name ebin -print) \
+	$(shell find $(ROOT)/deps/lager-* -maxdepth 2 -name ebin -print) \
```
jamesaimonetti: I'm debating, but only kz_attachment has a lager call. I think I'd rather remove that than add lager into the mix. These modules should remain very simple.

PR #1000, opened 2015-05-01
KAZOO-3595: If referred-by user-part is missing, use FromUser user-part

rickying:


swysor:
This won't work depending on which party is the From in the initial call record. Closing for now so some kind soul doesn't merge it.

PR #1002, opened 2015-05-01
Xref travis

fenollp:
**Depends on #1001**

Before merging we need to sort out these last errors:
```
Xref: undefined_function_calls
               blackhole_sup:start_link,,,,,,,,,,,,,,,,,,,,/0                calls undefined                         cowboy:start_http/4
                    crossbar:maybe_start_plaintext,,,,,,,,,/1                calls undefined                         cowboy:start_http/4
                    crossbar:start_ssl,,,,,,,,,,,,,,,,,,,,,/1                calls undefined                         cowboy:start_https/4
        ecallmgr_fs_conferen:send_conference_event,,,,,,,,,/3                calls undefined                wapi_conference:publish_conference_event/2
                         fax:start_link,,,,,,,,,,,,,,,,,,,,/0                calls undefined                         cowboy:start_http/4
                     gcm_api:retry_after_from,,,,,,,,,,,,,,/1                calls undefined                          qdate:to_unixtime/1
                     gcm_api:retry_after_from,,,,,,,,,,,,,,/1                calls undefined                          qdate:unixtime/0
              wh_media_proxy:maybe_start_plaintext,,,,,,,,,/1                calls undefined                         cowboy:start_http/4
              wh_media_proxy:maybe_start_ssl,,,,,,,,,,,,,,,/1                calls undefined                         cowboy:start_https/4
```

**Are these false positives?**
With these left untouched, *every Travis CI build will fail with error 9* (there's 9 errors).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,14 @@ otp_release:
   - R15B01
   - R15B
 before_install:
-    - sudo apt-get install -qq xsltproc
+  - set -e
+  - sudo apt-get install -qq xsltproc
+  - cd $HOME
+  - git clone --depth=50 --branch=master https://github.com/manopapad/proper.git proper
+  - cd proper
+  - make fast
+  - cd $TRAVIS_BUILD_DIR
 script: make compile
 after_success:
   - make xref
-  - make test
+  - ERL_LIBS="$HOME/proper" make test
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,4 +15,4 @@ otp_release:
 before_install:
     - sudo apt-get install -qq xsltproc
 script: "make compile"
-after script: "make eunit"
+after script: "make xref eunit"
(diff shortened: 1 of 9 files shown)


jamesaimonetti:
The `cowboy:start_http(s)` functions exist. I think xref needs updating to find the right cowboy.beam perhaps?

Ignore the qdate ones for now; gcm is a third-party dep in the pusher app.

Need to fix the ecallmgr_fs_conference one though.

fenollp:
Ah. Got it.
Adding the `./deps/` didn't solve anything, but showed the reason of those `cowboy` false positives:
rabbitmq clashes with a lot of things and with this module (multiple times!).
I'll add those to the filter…

fenollp:
`make test` fails because of proper. Not sure we can fix that…

jamesaimonetti:
Let's make a new target for running proper-related tests and a new macro at compile time.

So -DTEST and `ifdef(TEST).` for eunit, and maybe -DPROPER and `-ifdef(PROPER).` for Proper-related tests.

fenollp:
@jamesaimonetti READY

PR #1004, opened 2015-05-05
Add systemwide token restrictions.

Dinkor:
Each auth type method have own restrictions and one default restriction.
"cb_api_auth_restrictions", "cb_user_auth_restrictions", ... - for each auth method.
"restrictions" - default restrictions if no match auth method.

Macroses {ACCOUNT_ID} and {USER_ID} replaced with <<"account_id">> and <<"owner_id">> from cb_context:doc(Context) when creating auth_token.

Each priv_level have own template for restrictions.
You can create as many priv_level as you wish ("operator" - wich can modify users, but cant modify account).

Example token_restrictions template
system_config / crossbar.token_restrictions
``` JSON
{
   "_id": "crossbar.token_restrictions",
   "_rev": "5-c4955b161980d140cd49ec60c3e3575d",
   "default": {
       "restrictions": {
           "*": {
               "*": [
                   "#"
               ]
           }
       },
       "cb_user_auth_restrictions": {
           "*": {
               "get": [
                   "#"
               ],
               "post": [
                   "accounts/{ACCOUNT_ID}/users/{USER_ID}",
                   "accounts/{ACCOUNT_ID}/users/{USER_ID}/*"
               ]
           },
           "admin": {
               "get": [
                   "accounts/{ACCOUNT_ID}/users",
                   "accounts/{ACCOUNT_ID}/users/*",
                   "accounts/{ACCOUNT_ID}/users/*/*"
               ],
               "put": [
                   "accounts/{ACCOUNT_ID}/users"
               ],
               "post": [
                   "accounts/{ACCOUNT_ID}/users/*"
               ],
               "delete": [
                   "accounts/{ACCOUNT_ID}/users/*"
               ]
           }
       },
       "cb_api_auth_restrictions": {
           "admin": {
               "get": [
                   "accounts/{ACCOUNT_ID}/users",
                   "accounts/{ACCOUNT_ID}/users/*",
                   "accounts/{ACCOUNT_ID}/users/*/*"
               ],
               "put": [
                   "accounts/{ACCOUNT_ID}/users"
               ],
               "post": [
                   "accounts/{ACCOUNT_ID}/users/*"
               ],
               "delete": [
                   "accounts/{ACCOUNT_ID}/users/*"
               ]
           },
           "user": {
               "get": [
                   "accounts/{ACCOUNT_ID}/#"
               ]
           },
           "*": {
           }
       }
   }
}
```

On next step i want add per account token_restriction template.
Create cb_token_restriction module and separte document in AccountDB (<<"token_restrictions">>).

### DIFF
--- a/applications/crossbar/src/crossbar_util.erl
+++ b/applications/crossbar/src/crossbar_util.erl
@@ -875,17 +875,12 @@ create_auth_token(Context, Method, JObj) ->
     AccountId = wh_json:get_value(<<"account_id">>, JObj),
     OwnerId = wh_json:get_value(<<"owner_id">>, JObj),
 
-    Restrictions = case get_forced_restrictions(Method, AccountId, OwnerId) of
-        'undefined' -> wh_json:get_ne_value(<<"restrictions">>, Data);
-        MethodRestrictions -> MethodRestrictions
-    end,
-    
     Token = props:filter_undefined(
               [{<<"account_id">>, AccountId}
                ,{<<"owner_id">>, OwnerId}
                ,{<<"as">>, wh_json:get_value(<<"as">>, Data)}
                ,{<<"api_key">>, wh_json:get_value(<<"api_key">>, Data)}
-               ,{<<"restrictions">>, Restrictions}
+               ,{<<"restrictions">>, get_token_restrictions(Method, AccountId, OwnerId)}
                ,{<<"method">>, wh_util:to_binary(Method)}
               ]),
     JObjToken = wh_doc:update_pvt_parameters(wh_json:from_list(Token)
@@ -909,30 +904,43 @@ create_auth_token(Context, Method, JObj) ->
             cb_context:add_system_error('invalid_credentials', Context)
     end.
 
--spec get_forced_restrictions(atom(), ne_binary(), ne_binary()) -> api_object().
-get_forced_restrictions(Method, AccountId, OwnerId) -> 
-    PrivLevel = get_priv_level(AccountId, OwnerId, Method),
-    case get_default_restrictions(wh_util:to_binary(Method)) of
-        'undefined' -> 'undefined';
-        Restrictions -> get_priv_level_restrictions(Restrictions, PrivLevel)
-    end.
+-spec get_token_restrictions(ne_binary(), ne_binary(), ne_binary()) -> wh_json:object().
+get_token_restrictions(Method, AccountId, OwnerId) ->
+    Restrictions = case get_account_token_restrictions(AccountId, Method) of
+        'undefined' -> get_system_token_restrictions(Method);
+        AccountRestrictions -> AccountRestrictions
+    end,
+    PrivLevel = get_priv_level(AccountId, OwnerId),
+    get_priv_level_restrictions(Restrictions, PrivLevel).
 
--spec get_priv_level(ne_binary(), ne_binary(), atom()) -> ne_binary().
+-spec get_priv_level(ne_binary(), ne_binary()) -> ne_binary().
 %%
 %% for api_auth tokens we force "admin" priv_level
 %%
-get_priv_level(_AccountId, 'undefined', 'cb_api_auth') -> <<"admin">>;
+get_priv_level(_AccountId, 'undefined') -> <<"admin">>;
 
-get_priv_level(AccountId, OwnerId, _Method) ->
+get_priv_level(AccountId, OwnerId) ->
     AccountDB = wh_util:format_account_db(AccountId),
     {'ok', Doc} = couch_mgr:open_cache_doc(AccountDB, OwnerId),
     wh_json:get_ne_value(<<"priv_level">>, Doc).
 
--spec get_default_restrictions(ne_binary()) -> api_object().
-get_default_restrictions(Method) -> 
+-spec get_system_token_restrictions(ne_binary()) -> api_object().
+get_system_token_restrictions(Method) -> 
     case whapps_config:get(<<(?CONFIG_CAT)/binary, ".token_restrictions">>, Method) of
         'undefined' ->  whapps_config:get(<<(?CONFIG_CAT)/binary, ".token_restrictions">>, <<"_">>);
         MethodRestrictions -> MethodRestrictions
+    end.
+
+-spec get_account_token_restrictions(ne_binary(), ne_binary()) -> api_object().
+get_account_token_restrictions(AccountId, Method) -> 
+    AccountDB = wh_util:format_account_db(AccountId),
+    case couch_mgr:open_cache_doc(AccountDB, ?CB_ACCOUNT_TOKEN_RESTRICTIONS) of
+        {'ok', RestrictionsDoc} -> wh_json:get_first_defined(
+                                     [[<<"pvt_restrictions">>, wh_util:to_binary(Method)] 
+                                      ,[<<"pvt_restrictions">>, <<"_">>]
+                                     ], 
+                                     RestrictionsDoc);
+        {'error', _} -> 'undefined'
     end.
 
 -spec get_priv_level_restrictions(api_object(), ne_binary()) -> api_object().
--- a/applications/crossbar/src/crossbar_util.erl
+++ b/applications/crossbar/src/crossbar_util.erl
@@ -875,12 +875,25 @@ create_auth_token(Context, Method, JObj) ->
     AccountId = wh_json:get_value(<<"account_id">>, JObj),
     OwnerId = wh_json:get_value(<<"owner_id">>, JObj),
 
+    Restrictions = case get_forced_restrictions(Method, AccountId, OwnerId) of
+        'undefined' ->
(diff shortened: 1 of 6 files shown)


jamesaimonetti:
Hi @Dinkor, how does this differ from https://github.com/2600hz/kazoo/blob/master/applications/crossbar/doc/token_auth.md#token-restrictions ?

jamesaimonetti:
Hmm, as I read the token restriction code, I wonder what its purpose actually is! I'll have to dig up the tickets related to it to see the motivation.

However, I wonder if your work here could piggy back off of it in some way? Will need to think about this some more.

jamesaimonetti:
Also, can you update `crossbar/doc` to contain the documentation above?

jamesaimonetti:
One more thought, how will this play with resellers masquerading as sub-accounts?

Dinkor:
> I wonder if your work here could piggy back off of it in some way?

Sorry, i don't understand this sentence. Could you explain?

> One more thought, how will this play with resellers masquerading as sub-accounts?

I will do more testing.

jamesaimonetti:
@Dinkor Had a little more time to review this and think on it. I like it overall but want to propose some thoughts:

1. I think the auth-method keys should be the module names and the catch-all category is '_':

```JSON
{"_id":"crossbar.token_restrictions"
 ,"default":{
   "cb_user_auth":{...}
   ,"cb_api_auth":{...}
   ,"_":{...}
 }
}
```
2. For user types (`admin`, `user`, etc), I'd use the type and "_" for a catch-all again.

3. The HTTP methods should be all-caps; that's how Crossbar processes them.

4. I go back and forth on whether the structure should be:

```JSON
"GET":["URI", "URI"]
//or
"URI":["GET","POST"]
```
My instinct is the latter. `"/account/{ACCOUNT_ID}/users":["GET"]` seems a better way to restrict the users endpoint to just the listing (so no creating via PUT). Thoughts?

I wonder, too, if we should just restrict to the endpoint itself, vs the whole URI:
`"users":["GET"]` or even `"users":{"GET":true,"PUT":false}` where false is the default for HTTP verbs if missing. So maybe:

```JSON
{"cb_user_auth":{
   "user":{
     "users":{
       "GET":true
       ,"{USER_ID}":{
         "_":true
       }
       ,"_":false
     }
   }
 }
}
```

I would read the above as saying "If authenticating via user_auth, any auth-token for a "user"-level User could access /users via GET but all other methods are forbidden. The User can also access /users/{USER_ID} where {USER_ID} is the doc id of the User. All HTTP methods are allowed.

For an admin, access to all of /users is permitted (the default without the restriction):

```JSON
{"cb_api_auth":{
   "admin":{
     "users":{
       "_":true
     }
   }
 }
}

**Comment by Dinkor** [comment]
> My instinct is the latter. "/account/{ACCOUNT_ID}/users":["GET"] seems a better way to restrict the users endpoint to just the listing (so no creating via PUT).

I like it.

> I wonder, too, if we should just restrict to the endpoint itself, vs the whole URI

When iam testing, i found a bug:
if set token restriction to "accounts/{ACCOUNT_ID}/users/#" (only one rule), i can use URI "/accounts/{ACCOUNT_ID}/users/customer/{ACCOUNT_ID}" and gain access to customer document (what definitely should not be).
So iam think function [check_restrictions in cb_token_auth.erl](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/modules/cb_token_auth.erl#L225) must be rewriten to use "endpoints".
Restriction should compare with #cb_context.req_nouns ?

**Comment by Dinkor** [comment]
I wonder, maybe do restrictions check after *.validate_resource.* (in function [validate](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/api_util.erl#L890) of api_util.erl, after run validate_resources but before validate_data)
After running validate_resource callbacks, #cb_context have all data that need (account_id, user_id, device_id etc. ) for checking restrictions.

**Comment by jamesaimonetti** [comment]
@Dinkor I've tried documenting some ideas I had on processing token restrictions [here](https://github.com/2600hz/kazoo/blob/restrict_api_clients/applications/crossbar/doc/token_auth.md#token-restrictions).

As for when these should be applied, I think in the '*.authorize` is still appropriate for this check.

**Comment by Dinkor** [comment]
> As for when these should be applied, I think in the '*.authorize` is still appropriate for this check.

I also came to this decision. And move the code to *.authorize

**Comment by Dinkor** [comment]
New version.
It understand this type of restrictions
```JSON
   "restrictions": {
       "users": [
           {
               "accounts": [
                   "{AUTH_ACCOUNT_ID}"
               ],
               "params": [
                   {
                       "*/quickcall/*": {
                           "GET": true
                       }
                   }
               ]
           },
           {
               "accounts": [
                   "{AUTH_ACCOUNT_ID}"
               ],
               "params": [
                   {
                       "*": {
                           "POST": true,
                           "GET": true
                       }
                   }
               ]
           },
           {
               "accounts": [
                   "{DESCENDANT_ACCOUNT_ID}"
               ],
               "params": [
                   {
                       "#": {
                           "GET": true
                       }
                   }
               ]
           }
       ],
       "devices": [
           {
               "params": {
                   "#": {
                       "_": true
                   }
               }
           }
       ]
   }
```

`{AUTH_ACCOUNT_ID}` match to authorized account
`{DESCENDANT_ACCOUNT_ID}` match to any descendats acount.

This is an alpha version of the code. I'll be writing more code.
I want add `{AUTH_USER_ID}` macro and per-account restrictions.

applications/crossbar/src/crossbar_util.erl
```diff
@@ -867,12 +867,21 @@ create_auth_token(Context, Method, JObj) ->
     AccountId = wh_json:get_value(<<"account_id">>, JObj),
     OwnerId = wh_json:get_value(<<"owner_id">>, JObj),
 
+    Restrictions = case get_forced_restrictions(Method, AccountId, OwnerId) of
+        'undefined' ->
+            case wh_json:get_ne_value(<<"restrictions">>, Data) of
```
fenollp: No need to `case` the output here. `wh_json:get_ne_value/2` has a defaultt of `undefined`.

To be explicit you can use `wh_json:get_ne_value/3` with `undefined` as the last argument

Dinkor: Yes. You're right.

Dinkor:
Changed restrictions template
Allowed params objects, not array.
Allowed verbs is array of values, not array of key:value

```JSON
{
  "users": [
    {
      "accounts": [
        "{AUTH_ACCOUNT_ID}"
      ],
      "params": {
        "*/quickcall/*": [
          "GET",
          "POST"
        ]
      }
    },
    {
      "accounts": [
        "{AUTH_ACCOUNT_ID}"
      ],
      "params": {
        "/": [
          "_"
        ]
      }
    },
    {
      "accounts": [
        "{DESCENDANT_ACCOUNT_ID}"
      ],
      "params": {
        "#": [
          "POST"
        ],
        "params1": [
          "POST"
        ],
        "params2": [
          "DELETE"
        ]
      }
    },
    {
      "accounts": [
        "{DESCENDANT_ACCOUNT_ID}"
      ],
      "params": {
        "#": [
          "GET"
        ]
      }
    }
  ]
}
```

Dinkor:
Maybe keys "accounts" and "params" should be renamed to some more descriptive names?

Dinkor:
Renamed "params" to "rules", "accounts" to "allowed_accounts".
First match only mode. After matching endpoint/account/rule, search is stoped and trying match verb. 
http://jsoneditoronline.org/?id=3271fdf8b3486fb3d7e55da0785af278 - template wich have all endpoints and verbs (auth_method and priv_level is wildcarded).

Dinkor:
Per account token restrictions:
Create document with name `token_restrictions`. in field with name `pvt_restrictions` put restriction rules.
For exmaple this rules allow any request, exept adding new account (no "PUT" verb in "/" and "*" rules).
```JSON
{
   "_": {
       "_": {
           "accounts": [
               {
                   "rules": {
                       "/": [
                       ],
                       "*": [
                           "GET",
                           "POST",
                           "DELETE"
                       ],
                       "#": [
                           "_"
                       ]
                   }
               }
           ],
           "_": [
               {
                   "rules": {
                       "#": [
                           "_"
                       ]
                   }
               }
           ]
       }
   }
}
```

Dinkor:
Rebased to current master.

jamesaimonetti:
@Dinkor can you update [token_auth.md](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/doc/token_auth.md#token-restrictions) to reflect the new token restriction mechanism here? Will make testing it on our side a bit easier. Thanks!

Dinkor:
Yes. I'll add documentation. Work on this module is not finished yet.

Dinkor:
Moved code to new module cb_token_restrictions.

Dinkor:
I think this module is ready fro testing.

PR #1009, opened 2015-05-06
use customized regex from carriers app

lazedo:
allow classifiers by account/reseller/global


applications/stepswitch/src/stepswitch_resources.erl
```diff
@@ -712,6 +713,12 @@ create_resource(JObj, Resources) ->
                            )
     end.
 
+-spec maybe_account_classifiers(api_binary()) -> wh_proplist().
+maybe_account_classifiers('undefined') ->
+    wh_json:to_proplist(whapps_config:get(?CONFIG_CAT, <<"classifiers">>));
```
jamesaimonetti: Why not use `wnm_util:available_classifiers()` ?

PR #1014, opened 2015-05-08
KAZOO-3596: Priority calls for all!

siplabs:


jamesaimonetti:
@siplabs This is awesome! Will make moving to 3.5 a little easier in the upcoming months. I've really only made superficial formatting changes, added a unit test for `amqp_util:trim/3`, and cleaned up a couple dialyzer issues.

PR #1017, opened 2015-05-09
Kazoo 3577

jamesaimonetti:
Flushes XML config as well as caches on update/delete to trunkstore docs

### DIFF
--- a/applications/teletype/src/templates/teletype_low_balance.erl
+++ b/applications/teletype/src/templates/teletype_low_balance.erl
@@ -71,12 +71,19 @@ get_current_balance(DataJObj) ->
     Dollars = wht_util:units_to_dollars(Units),
     wht_util:pretty_print_dollars(Dollars).
 
+-spec get_balance_threshold(wh_json:object()) -> float().
+get_balance_threshold(DataJObj) ->
+    Default = 5.00,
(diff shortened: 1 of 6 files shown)


fenollp:
Build fails because get_balance_threshold/1 got removed here https://github.com/2600hz/kazoo/commit/226fbc525a2bab7775ad09c80dbe765223064ce6

fenollp:
@jamesaimonetti welp the build didn't fail, it's xref that caught this! hehehe!

fenollp:
LGTM

applications/teletype/src/templates/teletype_low_balance.erl
```diff
@@ -76,7 +76,7 @@ handle_req(DataJObj) ->
     Macros = [{<<"system">>, teletype_util:system_params()}
               ,{<<"account">>, teletype_util:account_params(DataJObj)}
               ,{<<"current_balance">>, get_current_balance(DataJObj)}
-              ,{<<"threshold">>, teletype_util:get_balance_threshold(DataJObj)}
+              ,{<<"threshold">>, teletype_topup:get_balance(DataJObj)}
```
lazedo: teletype_topup:get_balance calls wht_util:current_balance which doesn't seem to be a threshold ?

PR #1021, opened 2015-05-12
SMS helpers

lazedo:


jamesaimonetti:
@lazedo I had just merged this when you pushed the new commit + rebase, which is why this isn't going to merge. But it is merged!

PR #1027, opened 2015-05-14
Kazoo 3563

fenollp:


applications/crossbar/src/modules/provisioner_util.erl
```diff
@@ -12,64 +12,39 @@
...
+            'undefined' -> 'undefined';
+            JObj ->
+                wh_json:get_ne_value(<<"mac_address">>, JObj)
+        end,
+    cleanse_mac_address(MACAddress).
+
+-spec cleanse_mac_address(api_binary()) -> api_binary().
+cleanse_mac_address('undefined') -> 'undefined';
+cleanse_mac_address(MACAddress) ->
+    binary:replace(MACAddress, <<":">>, <<"">>, ['global']).
```
jamesaimonetti: This doesn't seem to be the equivalent to the former regex. Stripping non-alphanumeric characters is different from only removing `:`.

applications/crossbar/doc/devices.md
```diff
@@ -39,3 +39,14 @@ esoare
 Some devices support receiving SIP NOTIFY packets with `event` = `check-sync`. This is typically used to reboot the phone if the configuration has changed. Kazoo will generate the NOTIFY packet if the device is registered.
 
     curl -v -X POST -H "X-Auth-Token:{AUTH_TOKEN}" http://{SERVER}:8000/v2/accounts/{ACCOUNT_ID}/devices/{DEVICE_ID}/sync
+
+
+One can also attempt `check-sync` while updating a device (in the same request) setting `"sync": true`:
+
+    curl -v -X POST -H "X-Auth-Token:{AUTH_TOKEN}" http://{SERVER}:8000/v{API version}/accounts/{ACCOUNT_ID}/devices/{DEVICE_ID} -d '
+    {
+      "sync": true,
```
jamesaimonetti: Put this where it belongs

fenollp:
It gets really annoying to read provisioner logs as the request id does not show up in the logs.

How can I fix that? I know I have to set some process dictionary key but which one?

PR #1029, opened 2015-05-14
KAZOO-2895 apply dialplan before restrictions

lazedo:


core/whistle_number_manager-1.0.0/src/wnm_util.erl
```diff
@@ -380,6 +383,12 @@ to_e164(Number, <<_/binary>> = Account) ->
         {'error', _} -> to_account_e164(Number, AccountId)
     end.
 
+-spec to_e164(ne_binary(), api_binary(), api_binary()) -> ne_binary().
```
jamesaimonetti: should be api_object() for the last arg

PR #1034, opened 2015-05-15
Check previous modb when determining initial agent status

danielfinke:
If the agent did not have a status update for the month, stored in the modb, the lookup would return an empty list and the agent would default to "ready" state. Search now includes the previous modb if it exists, fails to old default (ready state)

### DIFF
--- a/applications/acdc/src/acdc_agent_util.erl
+++ b/applications/acdc/src/acdc_agent_util.erl
@@ -96,6 +96,10 @@ most_recent_db_status(AccountId, AgentId) ->
 -spec prev_month_recent_db_status(ne_binary(), ne_binary()) ->
                                          {'ok', ne_binary()}.
 prev_month_recent_db_status(AccountId, AgentId) ->
+    Opts = [{'startkey', [AgentId, wh_util:current_tstamp()]}
+            ,{'limit', 1}
+            ,'descending'
+           ],
     case couch_mgr:get_results(kazoo_modb_util:prev_year_month_mod(acdc_stats_util:db_name(AccountId)), <<"agent_stats/status_log">>, Opts) of
         {'ok', [StatusJObj]} -> {'ok', wh_json:get_value(<<"value">>, StatusJObj)};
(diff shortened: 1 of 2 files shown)


applications/acdc/src/acdc_stats_util.erl
```diff
@@ -50,3 +51,14 @@ get_query_limit(JObj) ->
 -spec db_name(ne_binary()) -> ne_binary().
 db_name(Account) ->
     wh_util:format_account_mod_id(Account).
+
+-spec prev_modb(ne_binary()) -> ne_binary().
+prev_modb(Account) ->
+	{{Year, Month, _}, _} = calendar:now_to_universal_time(os:timestamp()),
+	prev_modb(Account, Year, Month-1).
```
jamesaimonetti: You could use `{PrevYear, PrevMonth} = kazoo_modb_util:prev_year_month(Year, Month)` so we don't duplicate code. You could also pass in the current MODb and get the previous month's MODb for that account too.

applications/acdc/src/acdc_agent_util.erl
```diff
@@ -81,7 +81,18 @@ most_recent_db_status(AccountId, AgentId) ->
            ],
     case couch_mgr:get_results(acdc_stats_util:db_name(AccountId), <<"agent_stats/status_log">>, Opts) of
         {'ok', [StatusJObj]} -> {'ok', wh_json:get_value(<<"value">>, StatusJObj)};
-        {'ok', []} -> {'ok', <<"unknown">>};
+        {'ok', []} ->
+        	lager:debug("Could not find a recent status for agent ~p, checking previous modb", [AgentId]),
```
jamesaimonetti: This clause should go in a new function; we like to minimize nested clauses and a nicely named function here would be more descriptive that having to parse another nested case clause.

Also, as a matter of style, we don't start log lines with capital letters, and use ~s when printing binaries.

danielfinke:
Hold that thought, I missed a change

PR #1036, opened 2015-05-15
Handle DTMF digits properly after TTS module

danielfinke:
If DTMF is entered prior to the noop being sent to the cf_tts module, they are put into a collection that is read at the start of cf_collect_dtmf. However, it did not consider that the user may have already entered MaxDigits and that the execution should not go into whapps_call_command:collect_digits. If it did, it would await a noop that would cause the default timeout to be 1 day. Thus, even if MaxDigits was entered, they would still be waiting for DTMF timeout.


jamesaimonetti:
@danielfinke I've backported the last couple of commits to 3.19/20/master.

From here on out, let's target 3.20/master (or whatever is the current production release) for any pull requests. Thanks for the patches!

PR #1037, opened 2015-05-15
Changes in KAZOO-3059 break ACDC

danielfinke:
Adding overload for maybe_owner_called_self causes whapps_call:call()s with resource_type =:= 'undefined' to crash in cf_endpoint. Added default value for the resource_type


lazedo:
we need to investigate this better on why it is undefined when it reaches cf_endpoint. setting this default could break other pieces.

lazedo:
@danielfinke check if  https://github.com/2600hz/kazoo/pull/1038 works for you. the problem is acdc using callflow modules (cf_user:get_endpoints) which will be changed in the future.

PR #1040, opened 2015-05-18
Record min sec

Dinkor:
FS by default drop record files shorten than 3 seconds. https://wiki.freeswitch.org/wiki/Variable_record_min_sec.
I add `record_min_sec` option for `record_call` module to set minmum recording time.

### DIFF
--- a/applications/callflow/doc/record_call.md
+++ b/applications/callflow/doc/record_call.md
@@ -7,6 +7,19 @@ Version: 3.20
 
 The `record_call` callflow enables you to record the audio of the call.
 
+## Mandatory fields
+
+**action** - Must be set to `start` or `stop`.
+
+## Optional fields
+
+**time_limit** - Limit, in seconds, of how long to record the call. If ommited get value from `system_config/media/max_recording_time_limit`. Default `3600`.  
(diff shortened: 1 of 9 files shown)


applications/ecallmgr/src/ecallmgr_call_command.erl
```diff
@@ -1275,10 +1275,12 @@ maybe_get_terminators(Acc, JObj) ->
 -spec start_record_call_args(atom(), ne_binary(), wh_json:object(), ne_binary()) -> ne_binary().
 start_record_call_args(Node, UUID, JObj, RecordingName) ->
     FollowTransfer = wh_json:get_binary_boolean(<<"Follow-Transfer">>, JObj, <<"true">>),
+    RecordMinSec = wh_json:get_value(<<"Record-Min-Sec">>, JObj),
```
jamesaimonetti: This should probably be `get_binary_value/2` to ensure the integer is encoded as a binary and not skipped in `ecallmgr_util:set/3`.

applications/callflow/doc/record_call.md
```diff
@@ -7,6 +7,19 @@ Version: 3.20
 
 The `record_call` callflow enables you to record the audio of the call.
 
+## Mandatory fields
```
jamesaimonetti: Thank you!!!

jamesaimonetti:
@Dinkor thanks! Works as advertised

PR #1042, opened 2015-05-18
KAZOO-3663: insert callid in provisioner logs

fenollp:


core/whistle-1.0.0/src/wh_util.erl
```diff
@@ -524,6 +525,20 @@ callid(Prop) when is_list(Prop) ->
 callid(JObj) ->
     wh_json:get_first_defined([<<"Call-ID">>, <<"Msg-ID">>], JObj, ?LOG_SYSTEM_ID).
 
+%% @public
+spawn(Module, Function, Arguments) ->
```
jamesaimonetti: specs

PR #1047, opened 2015-05-20
HELP - document type not published to wapi_conf

lazedo:
publish fields to wapi_conf
removed changed_handler in stepswitch

### DIFF
--- a/core/whistle_couch-1.0.0/src/couch_util.erl
+++ b/core/whistle_couch-1.0.0/src/couch_util.erl
@@ -732,22 +732,28 @@ do_save_docs(#db{}=Db, Docs, Options, Acc) ->
                                {'ok', wh_json:objects()} |
                                couchbeam_error().
 perform_save_docs(Db, Docs, Options) ->
-    PreparedDocs = [maybe_set_docid(D) || D <- Docs],
+    {PreparedDocs, Publish} = lists:unzip([prepare_doc_for_save(D) || D <- Docs]),
     _ = flush_cache_docs(Db, PreparedDocs),
-    case ?RETRY_504(couchbeam:save_docs(Db, maybe_tombstones(PreparedDocs), Options)) of
+    case ?RETRY_504(couchbeam:save_docs(Db, PreparedDocs, Options)) of
         {'ok', JObjs} ->
-            _ = maybe_publish_docs(Db, PreparedDocs, JObjs),
+            _ = maybe_publish_docs(Db, Publish, JObjs),
             {'ok', JObjs};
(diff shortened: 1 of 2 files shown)


core/whistle_couch-1.0.0/src/couch_util.erl
```diff
@@ -643,7 +649,7 @@ prepare_doc_for_del(Conn, #db{name=DbName}, Doc) ->
         [{<<"_id">>, Id}
          ,{<<"_rev">>, DocRev}
          ,{<<"_deleted">>, 'true'}
-         ,{<<"pvt_type">>, wh_json:get_value(<<"pvt_type">>, Doc)}
+         | publish_fields(Doc)
```
jamesaimonetti: I wonder how we can avoid this, if at all? This is the tombstone document so I was trying to minimize the number of fields we store in it to save space in the DBs.

core/whistle_couch-1.0.0/src/couch_util.erl
```diff
@@ -81,6 +81,12 @@
 -define(MAX_BULK_INSERT, 2000).
 -define(RETRY_504(F), retry504s(fun() -> F end)).
 
+-define(PUBLISH_FIELDS, [<<"pvt_type">>
+                         ,<<"pvt_account_id">>
+                         ,<<"pvt_created">>
+                         ,<<"pvt_modified">>
```
jamesaimonetti: Maybe we default these to `wh_util:current_tstamp` when building the AMQP payload?

core/whistle_couch-1.0.0/src/couch_util.erl
```diff
@@ -728,13 +734,21 @@ do_save_docs(#db{}=Db, Docs, Options, Acc) ->
...
+    case ?RETRY_504(couchbeam:save_docs(Db, maybe_tombstones(PreparedDocs), Options)) of
         {'ok', JObjs} ->
             _ = maybe_publish_docs(Db, PreparedDocs, JObjs),
             {'ok', JObjs};
         {'error', _}=E -> E
     end.
 
+-spec maybe_tombstones(wh_json:objects()) -> wh_json:objects().
+maybe_tombstones(Docs) ->
+    [ maybe_tombstone(wh_json:get_value(<<"_deleted">>, Doc), Doc) || Doc  <- Docs ].
```
jamesaimonetti: I'd use `wh_json:is_true(<<"_deleted">>, Doc)` and match 'true' and 'false' in `maybe_tombstone/2`. I'm a bit worried that this will operate on all requests.

PR #1051, opened 2015-05-21
"authz_local_resources" .. support for node specific configuration

bluesandnz:
Currently even with authz_local_resources set to true, calls placed to local number resources do not  consume an inbound trunk.  This commit addresses this issue. With "authz_local_resources" will now be checked,  if true (for individual node or default) AND the number is local, authz will process the call for authorization.

This commit is an enhancement to "https://github.com/2600hz/kazoo/pull/502" since it now validates the configuration parameters on a per node basis.


fenollp:
Could you rebase this into a single commit with a good commit message?

bluesandnz:
rebased and commit comment updated

jamesaimonetti:
@bluesandnz Thanks! I've reworked the flow a bit to hopefully increase clarity of what's happening, but the results should be the same. Let me know if I missed anything!

PR #1052, opened 2015-05-21
KAZOO-3766: Add pi to retrieve auth resp

macpie:


jamesaimonetti:
@macpie Won't this let someone query `/v2/user_auth/{ANY_RANDOM_STRING}` until they find an auth token that exists, giving them access to accounts they shouldn't be able to?

macpie:
Yes not sure how to fix this? ideas ?

jamesaimonetti:
What's the use case for this functionality? Seems unneeded?

applications/crossbar/src/modules/cb_accounts.erl
```diff
@@ -233,6 +243,26 @@ validate_account_path(Context, AccountId, ?TREE, ?HTTP_GET) ->
         _Else -> Context1
     end.
 
+-spec maybe_get_auth_token(cb_context:context(), ne_binary(), ne_binary()) -> cb_context:context().
```
jamesaimonetti: No! The request should be `/v2/accounts/{ACCOUNT_ID}/user_auth/{AUTH_TOKEN}` which means cb_user_auth will get auth_account_id and compare it to `props:get_value(<<"accounts">>, cb_context:req_nouns(Context))`.

PR #1053, opened 2015-05-21
Kazoo 3765 dialyzer -- core/

fenollp:


jamesaimonetti:
Return "\n" to "~n" and rebase please.

core/braintree-1.0.0/src/braintree_subscription.erl
```diff
@@ -23,7 +23,7 @@
 -export([update_payment_token/2]).
 -export([find/1]).
 -export([create/1, create/2]).
--export([update/1]).
+-export([update/1, update/2]).
```
jamesaimonetti: Who calls this arity/2 function?

core/webseq-1.0.0/src/webseq.erl
```diff
@@ -79,7 +79,7 @@ stop_pid([Pid, Type]) ->
     catch gproc:unreg(?GPROC_KEY(Type)),
     'ok'.
 
--spec running() -> wh_proplist().
+-spec running() -> [{pid(),_}].
```
jamesaimonetti: Should add `-type type_key() :: {?MODULE, ne_binary() | '_' | '$1'}` and use it here and in the `-spec type_key(...)`

fenollp:
@jamesaimonetti should be good now ;)

core/whistle_media-1.0.0/src/whistle_media_maintenance.erl
```diff
@@ -303,36 +303,33 @@ remove_empty_media_docs(AccountId, AccountDb) ->
...
+maybe_remove_media_doc(AccountDb, Filename, MediaJObj) ->
     case wh_doc:attachments(MediaJObj) of
         'undefined' ->
             io:format("media doc ~s has no attachments, archiving and removing~n"
-                      ,[wh_json:get_value(<<"_id">>, MediaJObj)]
+                      ,[wh_doc:id(MediaJObj)]
                      ),
-            file:write(File, wh_json:encode(MediaJObj)),
-            file:write(File, [$\n]),
+            wh_util:write_file(Filename, [wh_json:encode(MediaJObj), $\n], ['append']),
```
jamesaimonetti: Why this change? Why `write_file/3` vs passing the io_device() and writing as we go?

fenollp: It was rewriting the file every time (not sure that was intended).

Also, there was a possibility of leaking fds as the code doesn't run in a try catch.

And `wh_util:write_file/2,3` logs on error.

jamesaimonetti: The filename includes the current timestamp, so unless invocations are being run simultaneously, there shouldn't be a filename clash.

Perhaps adding a catch on the `maybe_remove_media_doc/3` would be prudent instead?

I don't know the implications of calling `file:write_file/3` X times vs `file:write/2` and closing the file afterwards. My suspicion is the latter plays nicer with the file.

core/braintree-1.0.0/src/braintree_subscription.erl
fenollp: ah only 1 thing now, was used by update_payment_token/2 before… fixing it.

core/webseq-1.0.0/src/webseq.erl
fenollp: I don't get it. The `type_key` type would have nothing to do with `[{pid(),_}]`

jamesaimonetti: It would replace the `_`, no?

jamesaimonetti: Well, something should replace it, anyway.

fenollp:
@jamesaimonetti should be good again

jamesaimonetti:
Because I like being a pain, can you add a doc somewhere (`doc/` or create `doc/engineering`) with a `dialyzer.md` to include the commands you're using on `core/`, `applications/`, etc?

fenollp:
@jamesaimonetti I will yes, except #1053 and #1057 are just pre-PRs for https://github.com/2600hz/kazoo/compare/travis-dialyzer

fenollp:
Pushed documentation to branch travis-dialyzer

PR #1054, opened 2015-05-21
KAZOO-3766: Add api to retrieve resp auth

macpie:


jamesaimonetti:
@macpie did you try `curl -X GET http://server.com/user_auth/{AUTH_TOKEN}` ? It works for me, no account id, no X-Auth-Token.

jamesaimonetti:
@macpie You don't check if account_id/auth_account_id is undefined.

applications/crossbar/src/modules/cb_user_auth.erl
```diff
@@ -132,6 +136,38 @@ put(Context, ?RECOVERY) ->
...
+%% @end
+%%--------------------------------------------------------------------
+-spec maybe_get_auth_token(cb_context:context(), ne_binary()) -> cb_context:context().
+maybe_get_auth_token(Context, Token) ->
+    Context1 = crossbar_doc:load(Token, Context),
+    case cb_context:resp_status(Context1) of
+        'success' ->
+            AuthAccountId = cb_context:auth_account_id(Context),
+            AccountId = cb_context:account_id(Context),
+            create_auth_resp(Context1, AccountId, Token, AuthAccountId);
```
jamesaimonetti: Why switch up the argument order here? Why not `create_auth_resp(Context1, Token, AccountId, AuthAccountId)`, since you came into the function with `(Context, Token)`?

jamesaimonetti:
I think `authenticate_nouns/1` should be modified. Maybe something like:

```erlang
authenticate_nouns([{<<"user_auth">>, []}]) -> 'true';
authenticate_nouns([{<<"user_auth">>, [?RECOVERY]}]) -> 'true';
authenticate_nouns([{<<"user_auth">>, [_Token]}
                    ,{<<"accounts">>, [_AccountId]}
                   ]) ->
    true';
authenticate_nouns(_Nouns) -> 'false'.
```

Only the three URIs are authenticated:
1. `/v2/user_auth`
2. `/v2/user_auth/recovery`
3. `/v2/accounts/{ACCOUNT_ID}/user_auth/{TOKEN}`

Thoughts?

jamesaimonetti:
Actually, I guess we don't need the 3rd clause, we don't want to auto-authenticate the request. It should need an auth_token header/request value.

PR #1056, opened 2015-05-22
Callflows no longer continue if channel is hung up (related to KAZOO-3250)

danielfinke:
Similar to the fix for KAZOO-3250, callflows should not continue if the a-leg is destroyed during the cf_* module. See https://groups.google.com/forum/#!topic/2600hz-dev/--DJFztGF_Q for discussion

### DIFF
--- a/applications/callflow/src/cf_exe.erl
+++ b/applications/callflow/src/cf_exe.erl
@@ -17,6 +17,7 @@
 -export([queue_name/1]).
 -export([control_queue/1, control_queue/2]).
 -export([continue/1, continue/2]).
+-export([force_continue/2]).
 -export([continue_with_flow/2]).
 -export([branch/2]).
 -export([stop/1]).
@@ -106,6 +107,13 @@ continue(Key, Srv) when is_pid(Srv) ->
 continue(Key, Call) ->
     Srv = whapps_call:kvs_fetch('consumer_pid', Call),
     continue(Key, Srv).
+
+-spec force_continue(ne_binary(), whapps_call:call() | pid()) -> 'ok'.
+force_continue(Key, Srv) when is_pid(Srv) ->
+	gen_listener:cast(Srv, {'force_continue', Key});
+force_continue(Key, Call) ->
+	Srv = whapps_call:kvs_fetch('consumer_pid', Call),
+	force_continue(Key, Srv).
 
 -spec continue_with_flow(wh_json:object(), whapps_call:call() | pid()) -> 'ok'.
 continue_with_flow(Flow, Srv) when is_pid(Srv) ->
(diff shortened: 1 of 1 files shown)


lazedo:
this will call channel status between steps  and can create some unwanted pauses between steps. please make this as optional defaulting to false. you can get value from the flow defaulting to whapps_config value

jamesaimonetti:
I agree with @lazedo and think maybe cf_exe should track whether the channel has been hungup. I would look at the `relay_amqp` function in cf_exe and maybe, after the list comprehension, look at the event and relay it into the cf_exe process (the PID is in Props) if it is `CHANNEL_DESTROY`. cf_exe can then have a flag added to its state and check it when doing the 'continue', 'branch', etc.

applications/callflow/src/cf_exe.erl
```diff
@@ -329,13 +337,20 @@ handle_call(_Request, _From, State) ->
 %%--------------------------------------------------------------------
 handle_cast({'set_call', Call}, State) ->
     {'noreply', State#state{call=Call}};
-handle_cast({'continue', Key}, #state{flow=Flow
-                                      ,cf_module_pid=OldPidRef
+handle_cast({'continue', Key}, #state{cf_module_pid=OldPidRef
+									  ,call=Call
```
jamesaimonetti: Please make sure your formatting is consistent with the project's.

jamesaimonetti:
@danielfinke Wanted to check in and see if you've made progress here. Need any help with things?

danielfinke:
@jamesaimonetti just got busy with a few things and was away from the office for the last week. Just finished the updates now

lazedo:
@danielfinke still missing the option to not use this feature by default

applications/callflow/src/cf_exe.erl
```diff
@@ -500,6 +513,10 @@ handle_event(JObj, #state{cf_module_pid=PidRef
                                ,{'cf_event_pids', Others}
                               ]}
             end;
+        {{<<"call_event">>, <<"CHANNEL_DESTROY">>}, CallId} ->
```
k-anderson: I think this case clause is unnecessary, it would be handled by line 534.  As such could you please remove this?

k-anderson:
Howdy, I added one note but otherwise the PR looks good.  I was wondering what the use case was, continuing the call flow will would result in a lot of requests to ecallmgr that could never succeeded.  I assume that perhaps this is for stopping call recording or a custom module?  The reason I ask is I wonder if there is a more efficient way to achieve the desired behavior.

applications/callflow/src/cf_exe.erl
```diff
@@ -670,6 +687,12 @@ spawn_cf_module(CFModule, Data, Call) ->
 send_amqp_message(API, PubFun, Q) ->
     PubFun(add_server_id(API, Q)).
 
+-spec maybe_flag_destroyed(pid(), {api_binary(), api_binary()}) -> any().
```
jamesaimonetti: The spec should note the return as `'ok'`.

danielfinke:
@k-anderson Do you mean what the use case for the default of disabled would be? Or the use case of the patch and the flag for the channel being destroyed?

jamesaimonetti:
@danielfinke I believe @k-anderson wants to know the latter; what does this PR solve for you that existing functionality didn't?

k-anderson:
@danielfinke Hey, yes I am wondering what the use-case of continuing the callfow after channel hangup is.

 Additionally, after looking at it again I realized the defaults should maintain the existing behavor (that is stop_on_destroy should default to true).

danielfinke:
@k-anderson @jamesaimonetti The existing behaviour is that callflow execution doesn't stop when the channel is destroyed. Like the example I gave in the Google groups post, a module that does not handle CHANNEL_DESTROY or call cf_exe:stop, but instead just executes cf_exe:continue (at some point) will allow the callflow to continue even if the call has been destroyed. Execution will end up stopping the next time it hits something like a check of whapps_call_command:b_channel_status or a cf_exe:stop in a subsequent module.
The example in my post has a cf_collect_dtmf (with timeout) followed by cf_tts. If you watch the logs with the above code disabled and hang up during the collection phase, you will see it move to cf_tts.

k-anderson:
@danielfinke hey, sorry to do this but this has encouraged an internal discussion. The result of which is a reversal of my earlier concerns....Turns out we should have listened in May and that was primarily my fault I misunderstood what you were telling me! :(

PR #1058, opened 2015-05-25
KAZOO-3774: get list_entries vcard

siplabs:
I added tests for `kzd_users.erl` to this commit.

For using `lists/crossbar_listing` couchdb view (this means "v1 API") you should add
```ini
[query_server_config]
reduce_limit = false
```
to `/etc/kazoo/bigcouch/local.ini`

And this commit brokes v1 API compatibility:
It was allowed to add/update `lists` with `enries` and their IDs in old API.
This ability allowed to have different `entries` with same IDs in different `lists`.
Now `list entry` is separate document, so I wanted prevent ability to rewrite `list entries` when updating `list`.

### DIFF
--- a/applications/crossbar/priv/couchdb/account/lists.json
+++ b/applications/crossbar/priv/couchdb/account/lists.json
@@ -6,7 +6,23 @@
     "language": "javascript",
     "views": {
         "crossbar_listing": {
-            "map": "function(doc) { if (doc.pvt_type == 'list' && !doc.pvt_deleted) emit(doc._id, {'id': doc._id, 'name': doc.name, 'entries': doc.entries}) };"
+            "map": "function(doc) {if (doc.pvt_deleted) return; doc.id = doc._id; if (doc.pvt_type == 'list_entry') emit(doc.list_id, {id: doc._id, pattern: doc.pattern, displayname: doc.displayname, firstname: doc.firstname, lastname: doc.lastname}); if (doc.pvt_type == 'list') emit(doc._id, doc);}",
+            "reduce": "function(keys, values, rereduce) {var a = {entries: {}}; var i = 0; var l = values.length; for (i; i < l; i++) {if (values[i].pvt_type == 'list') {if (!values[i].entries) values[i].entries = {}; for (j in a.entries) {values[i].entries[j] = a.entries[j]}; a = values[i];} else {if (values[i].id) a.entries[values[i].id] = values[i]; else {for (j in values[i].entries) a.entries[j] = values[i].entries[j];}}} return a;}"
+        },
+        "crossbar_listing_v2": {
+            "map": "function(doc) {if ((doc.pvt_type != 'list') || doc.pvt_deleted) return; emit(doc._id, {'id': doc._id, 'name': doc.name, 'description': doc.description});}"
+        },
+        "entries": {
+            "map": "function(doc) {if ((doc.pvt_type != 'list_entry') || doc.pvt_deleted) return; emit(doc.list_id, doc);}"
+        },
+        "match_prefix": {
+            "map": "function(doc) {if ((doc.pvt_type != 'list_entry') || doc.pvt_deleted || (doc.number == null && doc.prefix == null)) return; emit(doc.number ? doc.number : doc.prefix, null);}"
+        },
+        "match_prefix_in_list": {
+            "map": "function(doc) {if ((doc.pvt_type != 'list_entry') || doc.pvt_deleted || (doc.number == null && doc.prefix == null)) return; emit([doc.list_id, doc.number ? doc.number : doc.prefix], null);}"
+        },
+        "regexps_in_list": {
+            "map": "function(doc) {if ((doc.pvt_type != 'list_entry') || doc.pvt_deleted || !doc.regexp) return; emit(doc.list_id, doc.regexp);}"
         }
     }
 }
--- a/applications/crossbar/priv/couchdb/schemas/list_entries.json
+++ b/applications/crossbar/priv/couchdb/schemas/list_entries.json
@@ -27,11 +27,30 @@
             "required": false,
             "type": "string"
         },
+        "list_id": {
+            "description": "List id",
+            "name": "List id",
+            "required": true,
+            "type": "string"
+        },
+        "number": {
+            "description": "Phone number",
+            "name": "Number",
+            "required": false,
+            "type": "string"
+        },
         "pattern": {
             "description": "Match pattern",
             "name": "Pattern",
-            "required": true,
+            "required": false,
             "type": "string"
+        },
+        "profile": {
+            "$ref": "profile",
+            "default": {},
+            "description": "Profile data",
+            "required": false,
+            "type": "object"
         },
         "type": {
             "maxLength": 128,
--- a/applications/crossbar/src/modules_v1/cb_lists_v1.erl
+++ b/applications/crossbar/src/modules_v1/cb_lists_v1.erl
@@ -1,5 +1,5 @@
 %%%----------------------------------------------------------------------------
-%%% @copyright (C) 2011-2014, 2600Hz INC
+%%% @copyright (C) 2011-2015, 2600Hz INC
 %%% @doc
 %%% Match list module
 %%%
@@ -8,9 +8,9 @@
 %%% @end
 %%% @contributors
 %%%   Kozlov Yakov
-%%%   SIPLABS, LLC (Maksim Krzhemenevskiy)
+%%%   SIPLABS, LLC (Maksim Krzhemenevskiy, Ilya Ashchepkov)
 %%%----------------------------------------------------------------------------
--module(cb_lists).
+-module(cb_lists_v1).
 
 -export([init/0
          ,allowed_methods/0, allowed_methods/1, allowed_methods/2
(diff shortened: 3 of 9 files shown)


applications/crossbar/priv/couchdb/schemas/list_entries.json
```diff
@@ -5,12 +5,24 @@
   "required": true,
   "description": "Schema for a match list entries",
   "properties": {
-    "pattern": {
+    "list": {
```
jamesaimonetti: Should be `list_id` if it references a doc ID

applications/crossbar/src/modules_v1/cb_lists_v1.erl
```diff
@@ -95,42 +95,48 @@ validate(Context, ListId, EntryId) ->
 
 -spec validate_lists(cb_context:context(), http_method()) -> cb_context:context().
 validate_lists(Context, ?HTTP_GET) ->
-    crossbar_doc:load_view(?CB_LIST, [], Context, fun normalize_view_results/2);
+    crossbar_doc:load_view(?CB_LIST, [{group_level, 1}], Context, fun normalize_view_results/2);
```
jamesaimonetti: Please tickie your atoms.

applications/crossbar/src/modules_v1/cb_lists_v1.erl
```diff
@@ -95,42 +95,48 @@ validate(Context, ListId, EntryId) ->
...
 
 -spec check_list_schema(api_binary(), cb_context:context()) -> cb_context:context().
 check_list_schema(ListId, Context) ->
-    OnSuccess = fun(C) -> on_successful_validation(ListId, C) end,
-    NewReq = wh_json:filter(fun filter_list_req_data/1
-                            ,cb_context:req_data(Context)
-                           ),
-    NewContext = cb_context:set_req_data(Context, NewReq),
-    cb_context:validate_request_data(<<"lists">>, NewContext, OnSuccess).
+    case wh_json:get_value(<<"entries">>, cb_context:req_data(Context)) of
```
jamesaimonetti: Would `cb_context:req_value(Context, <<"entries">>)` work as well?

siplabs:
@jamesaimonetti 
fixed

applications/crossbar/priv/couchdb/account/lists.json
```diff
@@ -3,7 +3,14 @@
     ,"language": "javascript"
     ,"views": {
         "crossbar_listing": {
-            "map": "function(doc) { if (doc.pvt_type == 'list' && !doc.pvt_deleted) emit(doc._id, {'id': doc._id, 'name': doc.name, 'entries': doc.entries}) };"
+            "map": "function(doc) {if (doc.pvt_deleted) return; doc.id = doc._id; if (doc.pvt_type == 'list_entry') emit(doc.list_id, {id: doc._id, pattern: doc.pattern, displayname: doc.displayname, firstname: doc.firstname, lastname: doc.lastname}); if (doc.pvt_type == 'list') emit(doc._id, doc);}",
```
jamesaimonetti: I'd almost rather create this as a "crossbar_listing_v1" view and "crossbar_listing" view, since the newer version should be the version people are migrating to (which should be valid for v3, v4, etc, until it isn't :) )

applications/crossbar/src/modules_v2/cb_lists_v2.erl
```diff
@@ -0,0 +1,259 @@
...
+    cb_context:set_content_types_provided(Context, [{'to_binary', [{<<"text">>, <<"x-vcard">>}
+                                                                   ,{<<"text">>, <<"directory">>}]}]).
+
+-spec validate(cb_context:context()) -> cb_context:context().
+-spec validate(cb_context:context(), path_token()) -> cb_context:context().
+-spec validate(cb_context:context(), path_token(), path_token()) -> cb_context:context().
+-spec validate(cb_context:context(), path_token(), path_token(), path_token()) -> cb_context:context().
+-spec validate(cb_context:context(), path_token(), path_token(), path_token(), path_token()) -> cb_context:context().
+validate(Context) ->
+    case cb_context:req_verb(Context) of
```
jamesaimonetti: We've been putting this into its own function:

```erlang
validate(Context) ->
    validate_lists(Context, cb_context:req_verb(Context)).

validate_lists(Context, ?HTTP_GET) ->
    ...;
validate_lists(Context, ?HTTP_PUT) ->
   ...
```

applications/crossbar/src/modules_v2/cb_lists_v2.erl
```diff
@@ -0,0 +1,259 @@
...
+    validate_doc(EntryId, <<"list_entry">>, Context);
+validate_req(?HTTP_PATCH, Context, [_ListId, ?ENTRIES, EntryId] = Path) ->
+    crossbar_doc:patch_and_validate(EntryId, Context, fun(_Id, C) -> validate_req(?HTTP_POST, C, Path) end);
+validate_req(?HTTP_GET, Context, [_ListId, ?ENTRIES, EntryId, ?VCARD]) ->
+    Context1 = crossbar_doc:load(EntryId, Context),
+    JObj = cb_context:doc(Context1),
+    JProfile = wh_json:get_value(<<"profile">>, JObj),
+    JObj1 = wh_json:merge_jobjs(JObj, JProfile),
+    JObj2 = wh_json:set_values(
+              props:filter_empty([{<<"first_name">>, wh_json:get_value(<<"firstname">>, JObj1)}
```
jamesaimonetti: Tuples should be on their own lines, with wrapping list and function parens dropped a line:

```erlang
wh_json:set_values(
    props:filter_empty(
        [{key1, value1}
         ,{key2, value2}
         ,{key3, value3}
        ]
    )
).
```

applications/crossbar/src/modules_v2/cb_lists_v2.erl
```diff
@@ -0,0 +1,259 @@
...
+on_successfull_validation('undefined', Type, Context) ->
+    Doc = wh_json:set_values([{<<"pvt_type">>, Type}], cb_context:doc(Context)),
+    cb_context:set_doc(Context, Doc);
+on_successfull_validation(Id, _Type, Context) ->
+    crossbar_doc:load_merge(Id, Context).
+
+-spec set_org(wh_json:object(), cb_context:context()) -> wh_json:object().
+set_org(JObj, Context) ->
+    case wh_json:get_value(<<"org">>
+                           ,cb_context:doc(crossbar_doc:load(wh_json:get_value(<<"list_id">>, JObj)
```
jamesaimonetti: This doesn't account for if crossbar_doc:load/2 fails. Break it into steps:

```erlang
set_org(JObj, Context) ->
    Context1 = crossbar_doc:load(...),
    set_org(JObj, Context1, cb_context:resp_status(Context1)).

set_org(JObj, Context, 'success') ->
    case wh_json:get_value(<<"org">>, cb_context:doc(Context)) of
        ...
    end;
set_org(JObj, _Context, _Status) ->
    lager:debug("failed to load list: ~s", [_Status]),
    JObj.
```

What you do in the error case is up to you (maybe throw?).

applications/crossbar/priv/couchdb/account/lists.json
siplabs: @jamesaimonetti 
 I'm sorry I didn't get it: should I

1) rename `crossbar_listing` to `crossbar_listing_v1` and `crossbar_listing_v2` to `crossbar_listing`
or
2) add `crossbar_listging_v1` identical to current `crossbar_listing`
?

k-anderson:
Howdy!

This is a lot of work but the requirement to disable the reduce_limit worriers me:

[query_server_config] 
reduce_limit = false

Why is this required, seems like there may be more efficient ways to achieve your use-case?  My concern is 1) if this is accepted and the community does not add this parameter then views will fail to build because this one hits the reduce limit or 2) they add this parameter and the database loads jump up building this view.

I would really like to see this requirement removed somehow.  Let me know!

siplabs:
@k-anderson 
It's for lists compatibility. Somebody may use couchdb REST API directly. 
I tested building this list with 20000 list entries. List builded fine on a machine with 1Gb of RAM.

I can break list compatibility and remove this requirement. But this reduction and composing list will be done in the crossbar module for all requests. I think couchdb needs less resources to builds list because of cachnig. I can't add such caching to crossbar.
I can add some checks about `reduce_limit = false` and run different code with the same result.
What do you think about it?

k-anderson:
@siplabs Nice, and I love that you are testing with "large" lists ;)  My concern is more because this parameter is global, so your list may not directly cause issues but it opens up the possibility that a view could.

So when you say break backward compatibility was that with regards to the direct access to the database?  If that is the case, that is not a supported use-case so I would say feel free to change that.  If instead you are talking about backward compatibility via crossbar APIs then no we would want to do something else.

I agree that couch is preferable over doing it in the application when possible.  I would like to work with you to maybe see if we can get the view optimized in some way (perhaps even a hybrid with part in the view and some in the application).  That would be difficult to do via github, how would you feel about scheduling some pair programming time?

siplabs:
Well, we ment only backward compatibility with lists API itself, no other APIs is affected with this parameter, it's totally safe and future-proof.

siplabs:
@k-anderson @jamesaimonetti 
rebased to master and updated.
This version works without `reduce_limit = false`.

PR #1059, opened 2015-05-26
Incorrect binding declaration.

Dinkor:
Bind only to "resource".
"Numbers" and "dedicated_ip" not bounded, and not catch changes in this documents.

### DIFF
--- a/applications/stepswitch/src/stepswitch_sup.erl
+++ b/applications/stepswitch/src/stepswitch_sup.erl
@@ -25,10 +25,9 @@
                   ,'permanent', 5000, 'worker', ['poolboy']
                  }).
 
--define(ORIGIN_BINDINGS, [[{'type', <<"resource">>}
-                           ,{'type', <<"number">>}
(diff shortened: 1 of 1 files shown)


lazedo:
thank you @Dinkor , backported to 3.20

Dinkor:
I look at other *_sup modules and think found similar problems:
[applications/fax/src/fax_sup.erl](https://github.com/2600hz/kazoo/blob/master/applications/fax/src/fax_sup.erl#L28)
[applications/trunkstore/src/trunkstore_sup.erl](https://github.com/2600hz/kazoo/blob/master/applications/trunkstore/src/trunkstore_sup.erl#L21)
[applications/doodle/src/doodle_sup.erl](https://github.com/2600hz/kazoo/blob/master/applications/doodle/src/doodle_sup.erl#L17)

Dinkor:
https://github.com/2600hz/kazoo/pull/1063

PR #1063, opened 2015-05-27
Incorrect bindings

Dinkor:
This problem like in https://github.com/2600hz/kazoo/pull/1059

### DIFF
--- a/applications/trunkstore/src/trunkstore_sup.erl
+++ b/applications/trunkstore/src/trunkstore_sup.erl
@@ -18,11 +18,10 @@
 
 -include("ts.hrl").
 
--define(ORIGIN_BINDINGS, [[{'type', <<"account">>}
-                           ,{'type', <<"connectivity">>}
-                           ,{'type', <<"sys_info">>}
-                           ,{'type', <<"number">>}
-                          ]
+-define(ORIGIN_BINDINGS, [[{'type', <<"account">>}]
+                           ,[{'type', <<"connectivity">>}]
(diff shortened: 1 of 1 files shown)


applications/doodle/src/doodle_sup.erl
```diff
@@ -14,9 +14,8 @@
 
 -include("doodle.hrl").
 
--define(ORIGIN_BINDINGS, [ [{'db', ?WH_SIP_DB }
-                            ,{'type', <<"device">>}
-                           ]
+-define(ORIGIN_BINDINGS, [[{'db', ?WH_SIP_DB }]
+                          ,[{'type', <<"device">>}]
```
lazedo: the existing binding works as expected. we want devices from sip db

applications/fax/src/fax_sup.erl
```diff
@@ -25,7 +25,9 @@
                                                  ]]}
                    ,'permanent', 5000, 'worker', ['poolboy']}).
 
--define(ORIGIN_BINDINGS, [[{'db', ?WH_FAXES_DB}, {'type', <<"faxbox">>}]]).
+-define(ORIGIN_BINDINGS, [[{'db', ?WH_FAXES_DB}]
+                          ,[{'type', <<"faxbox">>}]
+                         ]).
```
lazedo: same here, we want faxbox documents from faxes db

Dinkor:
Remove changes to doodle_sup & fax_sup.

PR #1064, opened 2015-05-27
KAZOO-3777: added new voices to voicefabric tts engine

siplabs:


jamesaimonetti:
@siplabs I broke out the voice mappings into a .hrl file (along with a couple other macros). Makes getting around the .erl a little easier, I think.

PR #1067, opened 2015-05-28
Kazoo 244 - audit logs on service updates

jamesaimonetti:


jamesaimonetti:
@fenollp I think Dialyzer is coughing on the tests contained in here. Can you constrain the search for beams to ebin/ only? Test beams are put in test/ and should be ignored when dialyzing.

fenollp:
@jamesaimonetti on it

PR #1069, opened 2015-05-29
Fix for DISA setting caller id

arekfryz:
Since we always have AccountId that will point to the number owner we should use it to lookup caller id instead of caller_id_number. Old version was trying to find AccountId by caller_id_number, but that number can be anything and not always in our number inventory and assigned to any account.


jamesaimonetti:
Thanks @arekfryz; I made change to return the modified Call back up to the calling function. Also, please make sure your formatting (4 space indentation) match the project's. But otherwise, great fix!

PR #1071, opened 2015-06-01
Fixes to schema.

Dinkor:
Add presence_id to users schema. If this value set to integer (monster-ui do it at some circumstances) then `cf_attributes:presence_id/2` is crashed wen try `binary:match(PresenceId, <<"@">>)`.
Add missing coma in webhooks_attemps schema.


Dinkor:
Monster-UI issue
https://2600hz.atlassian.net/browse/UI-1597

jamesaimonetti:
@Dinkor good catch, i'm going to add some tests for the json schemas to catch these typos sooner!

PR #1072, opened 2015-06-01
KAZOO-2858: play cf-not_available prompt when trying to forward call …

fenollp:
…on DID


jamesaimonetti:
@fenollp please check out my commit and test it. Should be identical functionality though. If you're happy with it, we can merge.

fenollp:
LGTM

PR #1073, opened 2015-06-01
KAZOO-3661: handle hangups of type UNKNOWN

fenollp:


lazedo:
@fenollp cherry-picked to 3.20

PR #1074, opened 2015-06-01
KAZOO-3798: Adding service for whitelabel doc

macpie:


k-anderson:
Could you please change the category name to "branding"

PR #1079, opened 2015-06-02
Kazoo 3839: check JSON files for syntax

jamesaimonetti:


fenollp:
I like this :) Is the JSON fields order going to stay the same every time we run `make json` ?

jamesaimonetti:
The JSON structure will be untouched between runs (assuming no changes in the file).

jamesaimonetti:
@fenollp If this works for you (or @lazedo or @macpie), let's merge it and be done with JSON syntax errors!

fenollp:
LGTM

PR #1082, opened 2015-06-04
KAZOO-3846 unregister channel handlers on release

lazedo:


fenollp:
Updating the Dialyzer filter: I love it! :)

core/whistle_amqp-1.0.0/src/wh_amqp_assignments.erl
```diff
@@ -149,6 +149,11 @@ handle_call({'request_float', Consumer, Broker}, _, State) ->
     {'reply', assign_or_reserve(Consumer, Broker, 'float'), State};
 handle_call({'request_sticky', Consumer, Broker}, _, State) ->
     {'reply', assign_or_reserve(Consumer, Broker, 'sticky'), State};
+handle_call({'release_assignments', Consumer}, _, State) ->
```
k-anderson: If you are converting the cast to call please either remove the handle_cast for 'release_assignments' if it is not used or update it to behave the same way.

core/whistle_amqp-1.0.0/src/wh_amqp_assignments.erl
```diff
@@ -149,6 +149,11 @@ handle_call({'request_float', Consumer, Broker}, _, State) ->
     {'reply', assign_or_reserve(Consumer, Broker, 'float'), State};
 handle_call({'request_sticky', Consumer, Broker}, _, State) ->
     {'reply', assign_or_reserve(Consumer, Broker, 'sticky'), State};
+handle_call({'release_handlers', Consumer}, _, State) ->
+    Pattern = #wh_amqp_assignment{consumer=Consumer, _='_'},
+    Res = release_handlers(ets:match_object(?TAB, Pattern, 1)),
```
jamesaimonetti: Why build Pattern up here? I'd just pass Consumer to release_handlers/1 and do the ets magic there since that is the meat of the functionality.

PR #1087, opened 2015-06-05
KAZOO-3851 add interface info for media servers in wh_nodes

lazedo:
also fixes the media servers printing


applications/ecallmgr/src/ecallmgr_fs_node.erl
```diff
@@ -66,6 +67,7 @@
                     ,aggressive_nat
                     ,stun_enabled
                     ,stun_auto_disabled
+                    ,all = []
```
k-anderson: Please be more descriptive, perhaps 'interface_props'?

PR #1090, opened 2015-06-08
whitelabel-json-formatting: apply recent formatting changes

fenollp:
@jamesaimonetti this "went through les mailles du filet"


jamesaimonetti:
@fenollp I addressed it in #1088

PR #1091, opened 2015-06-08
KAZOO-3854: do not Dialyze test files

fenollp:
Helps with #1067

### DIFF
--- a/scripts/check-dialyzer.escript
+++ b/scripts/check-dialyzer.escript
@@ -12,9 +12,12 @@
 
 main(Args) ->
     case [Arg || Arg <- Args,
-                 is_ebin_dir(Arg)
-                     orelse is_beam(Arg)
-                     orelse is_erl(Arg)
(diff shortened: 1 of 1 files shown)


scripts/check-dialyzer.escript
```diff
@@ -12,9 +12,12 @@
...
     case [Arg || Arg <- Args,
-                 is_ebin_dir(Arg)
-                     orelse is_beam(Arg)
-                     orelse is_erl(Arg)
+                 (
+                   is_ebin_dir(Arg)
+                   orelse is_beam(Arg)
+                   orelse is_erl(Arg)
+                 ) andalso
+                     not is_test(Arg)
```
jamesaimonetti: Can this go first so we short circuit all the other conditions if it is a test beam?

fenollp:
@jamesaimonetti that's what you meant?

PR #1093, opened 2015-06-09
Issue:

aberke:
When offnet device sends to kazoo user, cannot get sms via /users/:id/sms endpoint
Issue is that view sms/listing_by_owner is not returning the SMS because the SMS lacks pvt_target_owner_id property.
Fixed with more flexible filtering fule in this PR

Drill down on issue:
Looks like code depends on pvt_target_owner_id to be set via cf_sms_user:handle .... fun whapps_call:kvs_store/3, <<"target_owner_id">>, UserId: https://github.com/2600hz/kazoo/blob/master/applications/doodle/src/module/cf_sms_user.erl#L29

Then sends to depend on doodle_exe:continue --> call end to save this property in doodle_util:save
But does doodle_exe:continue update doodle's version of the Call?


lazedo:
@aberke can you elaborate on "When offnet device sends to kazoo user" ? what does this mean ?

"is not returning the SMS because the SMS lacks pvt_target_owner_id property"
are you running an older version like 3.19 ?

"Then sends to depend on doodle_exe:continue --> call end to save this property in doodle_util:save
But does doodle_exe:continue update doodle's version of the Call?"
don't understand what you're saying here.  are saying that doodle_exe:continue does not use updated call and therefore doesn't update the pvt_target_owner_id field?

it seems to me that you are trying to address some issue in 3.19 , if that's case then you can issue a PR to 3.19 if you're not moving to 3.20, because it seems to me that this is not an issue in master/3.20.

or.., i'm missing something

aberke:
@lazedo You are right!  We are using 3.19.52 and I see now that this issue was fixed in 3.20.  Sorry!

PR #1096, opened 2015-06-13
KAZOO-3872

onnet:
Setting enabled field to 'false' in user's doc prevents user associated devices from outbound calling which is great :) 
But still user can authenticate through crossbar's user_auth. 
There could be some reasons for such a behavior which I just can't see, but from the very first look web/api access should also be restricted for blocked user. 

Please consider to accept if it make sense

Regards,


jamesaimonetti:
@onnet Makes perfect sense. Good find!

onnet:
Thank you James!

PR #1097, opened 2015-06-15
KAZOO-3780 log smtp notifications

lazedo:


applications/teletype/src/teletype_util.erl
```diff
@@ -70,8 +70,41 @@ send_email(Emails, Subject, RenderedTemplates, Attachments) ->
...
+            maybe_log_smtp(Skip, AccountId, Emails, Subject, RenderedTemplates, 'undefined', wh_util:to_binary(Reason)),
+            E
+    end.
 
+-spec maybe_log_smtp(boolean(), api_binary(), list(), ne_binary(), list(), api_binary(), api_binary()) -> 'ok'.
+maybe_log_smtp('true', _AccountId, _Emails, _Subject, _RenderedTemplates, _Receipt, _Error) ->
+    lager:debug("skipping smtp log");
+maybe_log_smtp(_, 'undefined', _Emails, _Subject, _RenderedTemplates, _Receipt, _Error) ->
+    lager:debug("skipping smtp log since account_id is 'undefined'");
+maybe_log_smtp(_, AccountId, Emails, Subject, RenderedTemplates, Receipt, Error) ->
```
jamesaimonetti: May as well match 'false' here and in the previous clause, since we know it must be that.

core/kazoo_modb-1.0.0/priv/couchdb/views/notify.json
```diff
@@ -0,0 +1,9 @@
+{
+   "_id": "_design/notify",
```
jamesaimonetti: Can we call it the `_design/notifications` view so that folks don't think it is tied to the 'notify' app?

applications/teletype/src/teletype_util.erl
```diff
@@ -644,6 +683,8 @@ fetch_templates(TemplateId, <<_/binary>> = AccountId) ->
 
     case couch_mgr:open_cache_doc(AccountDb, DocId) of
         {'ok', TemplateJObj} ->
+            put('template_id', TemplateId),
+            put('template_account_id', AccountId),
```
jamesaimonetti: Are these used anywhere?

jamesaimonetti: NM, I see them now

applications/teletype/src/teletype_util.erl
```diff
@@ -70,8 +70,41 @@ send_email(Emails, Subject, RenderedTemplates, Attachments) ->
                | add_attachments(Attachments)
               ]
             },
-    relay_email(To, From, Email).
+    Skip = wh_util:is_true(get('skip_smtp_log')),
+    AccountId = get('account_id'),
```
jamesaimonetti: You could move this into 'maybe_log_smtp/6' to reduce the arg list by one

jamesaimonetti: Then branch to log_smtp() if AccountId is not `undefined`.

applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -34,7 +34,9 @@
                                  ]).
 -define(CB_LIST, <<"notifications/crossbar_listing">>).
 -define(PREVIEW, <<"preview">>).
-
+-define(SMTP_LOG, <<"log">>).
```
jamesaimonetti: I think the URI should be `notifications/logs` since it is a collection of logs instead of just one log.

jamesaimonetti:
Also, need documentation updated for the notifications Crossbar endpoint with the `logs` query

PR #1100, opened 2015-06-17
KAZOO-3871: ignore "useless but invalid" data in whapps_maintenance

fenollp:


core/whistle-1.0.0/src/wh_json.erl
```diff
@@ -334,7 +334,7 @@ all(Pred, ?JSON_WRAPPER(Prop)) when is_function(Pred, 1) ->
 any(Pred, ?JSON_WRAPPER(Prop)) when is_function(Pred, 1) ->
     lists:any(Pred, Prop).
 
--spec foldl(fun((key(), json_term(), any()) -> any()), any(), object()) -> any().
+-spec foldl(fun((key(), json_term(), any()) -> Ret), Ret, object()) -> Ret.
```
jamesaimonetti: I don't think this is right. Dialyzer would assume all must be the same value, but since this is the accumulator param, we can't know what the caller is passing in nor how the function F will manipulate said accumulator.

core/whistle_apps-1.0.0/src/whapps_maintenance.erl
```diff
@@ -1013,9 +1013,11 @@ delete_system_media_references(DocId, CallResponsesDoc) ->
             'ok'
     end.
 
--spec remove_system_media_refs(wh_json:key(), wh_json:json_term()) ->
+-spec remove_system_media_refs(wh_json:key(), [wh_json:object()]) ->
```
jamesaimonetti: `wh_json:objects()` should do fine here

jamesaimonetti:
@k-anderson what's the punishment for merging one's own PR? Being drawn and quartered?

fenollp:
Hehe he fixed the fix so I took upon myself to merge!

PR #1101, opened 2015-06-17
KAZOO-3833: Fix presence_id to be a binary

macpie:


core/kazoo_documents-1.0.0/src/kz_device.erl
```diff
@@ -165,11 +165,15 @@ set_sip_settings(DeviceJObj, SipJObj) ->
 presence_id(DeviceJObj) ->
     presence_id(DeviceJObj, 'undefined').
 presence_id(DeviceJObj, Default) ->
-    wh_json:get_value(?PRESENCE_ID, DeviceJObj, Default).
+    wh_json:get_binary_value(?PRESENCE_ID, DeviceJObj, Default).
```
lazedo: presence_id(DeviceJObj, 'undefined') would return <<"undefined">> ?
how would this work with cf_endpoint merge ?

jamesaimonetti:
```erlang
> wh_json:get_binary_value(<<"foo">>, wh_json:new(),'undefined').
undefined
```
Should still return the atom 'undefined'

k-anderson:
@macpie 
Good start but not all the modules use this yet ;)  Please update anything extracting the presence_id to use kz_device.  There are few uses in cf_manual_presence, cf_acdc_agent, cb_users_v2, provisioner_v5 and most importantly reg_authn_req.

PR #1102, opened 2015-06-17
Kazoo 3555: white label dns testing

jamesaimonetti:


k-anderson:
@jamesaimonetti 
Why the update to deps in this PR?  Can you separate that so this is more targeted?

Also, the GET and POST is intended to work prior to saving the whitelabel document.  In my tests I am unable to fetch the initial list so the UI would not be able to show the DNS entries that need to be made until the document is saved (odd user experience).  I also briefly read the docs and think the POST will actually update the domain?  If so I may not have convayed it appropriately but the idea is to be able to preview with a domain.  If the whitelabel doc exists this will just automatically attempt the domain when loaded otherwise it will ask the user for their intended domain and test that.

Give me a call if you want to discuss :)

jamesaimonetti:
@k-anderson I updated nksip because it had DNS-related stuff in it that I was thinking to use; however, when I tried to start the nksip application it failed and a bunch of files it thought should exist, in fact, didn't. I'm not sure what happened on the initial loading of nksip, or if the they retagged 0.4.0 since nksip entered our deps/, but I felt it warranted fixing if we're going to find more places for nksip to be used.

There are two versions of POST:
1. `/v2/accounts/{ACCOUNT_ID}/whitelabel/domains`: for the reseller to test their setup domain against the system's configuration
2. `/v2/whitelabel/domains`: for the system admin to set the domains (updates the `system_config/whitelabel` document).

I'll update the first POST to allow specifying `domain=reseller.com` on the payload, in case they haven't configured whitelabeling yet. So it will check the request for `domain=`; if missing, check the whitelabel doc for domain; otherwise kick back an error. That work?

jamesaimonetti:
@k-anderson updated

core/kazoo_documents-1.0.0/test/kzd_domains_test.erl
```diff
@@ -0,0 +1,325 @@
...
+       ).
+
+-define(NAPTR
+        ,<<"{\"NAPTR\":{\"proxy-east.{{domain}}\":{\"name\":\"East NAPTR\",\"mapping\":[\"10 100 \\\"S\\\" \\\"SIP+D2U\\\" \\\"\\\" _sip._udp.proxy-east.{{domain}}.\"]},\"proxy-central.{{domain}}\":{\"name\":\"Central NAPTR\",\"mapping\":[\"10 100 \\\"S\\\" \\\"SIP+D2U\\\" \\\"\\\" _sip._udp.proxy-central.{{domain}}.\"]},\"proxy-west.{{domain}}\":{\"name\":\"West NAPTR\",\"mapping\":[\"10 100 \\\"S\\\" \\\"SIP+D2U\\\" \\\"\\\" _sip._udp.proxy-west.{{domain}}.\"]}}}">>
+       ).
+
+-define(SRV
+        ,<<"{\"SRV\":{\"_sip._udp.proxy-east.{{domain}}\":{\"name\":\"East SRV\",\"mapping\":[\"10 10 7000 us-east.{{domain}}.\",\"15 15 7000 us-central.{{domain}}.\",\"20 20 7000 us-west.{{domain}}.\"]}}}">>
+       ).
+
```
fenollp: Could you turn this JSON into `wh_json:from_list/1` form? Easier to read and update

core/kazoo_documents-1.0.0/test/kzd_domains_test.erl
```diff
@@ -0,0 +1,325 @@
...
+                      ]
+                      ,Hosts
+                     )
+      }
+     | validate_naptr_hosts(NAPTR, Hosts)
+    ].
+
+validate_naptr_hosts(NAPTR, Hosts) ->
+    lists:flatten(
+      lists:map(fun(H) -> validate_naptr_host(NAPTR, H) end
```
fenollp: I think `lists:flatmap/2` has the same semantics

core/kazoo_documents-1.0.0/test/kzd_domains_test.erl
```diff
@@ -0,0 +1,325 @@
...
+       ,?_assertEqual([<<"_sip._udp.proxy-east.{{domain}}">>]
+                      ,Hosts
+                     )
+      }
+     | validate_srv_hosts(SRV, Hosts)
+    ].
+
+validate_srv_hosts(SRV, Hosts) ->
+    lists:flatten(
+      lists:map(fun(H) -> validate_srv_host(SRV, H) end
```
fenollp: lists:flatmap

core/whistle_config-1.0.0/src/whapps_config.erl
```diff
@@ -406,6 +428,18 @@ maybe_save_category(Category, JObj, Looped) ->
             {'ok', JObj1}
     end.
 
+-spec update_pvt_fields(config_category(), wh_json:object(), api_object()) ->
+                               wh_json:object().
+update_pvt_fields(Category, JObj, 'undefined') ->
+    wh_doc:update_pvt_parameters(
+      wh_json:set_value(<<"_id">>, Category, JObj)
```
fenollp: Use wh_doc:set_id/2 (note: it flips args)

applications/crossbar/src/cb_context.erl
```diff
@@ -157,7 +157,7 @@ account_modb(Context, Year, Month) ->
     kazoo_modb:get_modb(account_id(Context), Year, Month).
 
 account_realm(Context) ->
-    wh_json:get_value(<<"pvt_realm">>, account_doc(Context)).
+    kz_account:realm(account_doc(Context)).
```
fenollp: Note: this is a change also introduced in #1135

core/kazoo_documents-1.0.0/Makefile
```diff
@@ -26,6 +29,13 @@ ebin/$(PROJECT).app: src/*.erl
 	@mkdir -p ebin/
 	erlc -v $(ERLC_OPTS) -o ebin/ -pa ebin/ $(PA) $?
 
+json: JSON = $(shell find priv/ -name *.json -print)
+json:
+	@for jobj in $(JSON); do \
+		echo checking $$jobj; \
+		python -c 'import sys,json; json.encoder.FLOAT_REPR=str; data=json.load(open("'$$jobj'")); print json.dumps(data, sort_keys=True, indent=4, separators=(",", ": "))' >$$jobj~ && mv $$jobj~ $$jobj; \
```
k-anderson: Please use python2

PR #1103, opened 2015-06-17
KAZOO-3819: skip db call when realm is an IP address in registrar

fenollp:


core/whistle-1.0.0/src/api/wapi_authn.erl
```diff
@@ -229,12 +236,14 @@ get_auth_user(ApiJObj) ->
 -spec get_auth_realm(wh_json:object() | wh_proplist()) -> ne_binary().
 get_auth_realm(ApiProp) when is_list(ApiProp) ->
```
fenollp: Note that this code does not get called anymore. I kept (and standardized it) as it is pretty similar to the above `get_auth_user`

k-anderson: If it is dead code (and you are sure) then remove it!

fenollp: I lied sorry. It's used on line 208.

PR #1108, opened 2015-06-19
KAZOO-3892: add images and openoffice compatible documents support to fax

siplabs:


lazedo:
@siplabs this sounds great, only concern is about the images (gif, jpeg) because of html  mails with image signature. thoughts ?

siplabs:
Suggestion (if approved - will be done very fast) - if the picture is smaller then IMAGE_MINIMUM_SIZE pixels (either width or height) as most signature images are.... - then ignore it ;)

applications/fax/src/fax_worker.erl
```diff
@@ -764,11 +771,69 @@ prepare_contents(JobId, RespHeaders, RespContent) ->
...
+    Cmd = binary_to_list(<<"echo -n `grep -E '^", ContentType/binary, "\\s' /etc/mime.types "
+                           "2> /dev/null "
+                           "| head -n1 "
+                           "| cut -f2`">>),
+    case os:cmd(Cmd) of
+        [] -> <<>>;
+        CT -> CT
+    end.
+
+convert_openoffice_document(CT, TmpDir, JobId, RespContent) ->
```
jamesaimonetti: Need spec please.

applications/fax/src/fax_worker.erl
```diff
@@ -764,11 +771,69 @@ prepare_contents(JobId, RespHeaders, RespContent) ->
...
+        CT when ?OPENOFFICE_COMPATIBLE(CT) ->
+            convert_openoffice_document(CT, TmpDir, JobId, RespContent);
         Else ->
             lager:debug("unsupported file type: ~p", [Else]),
             {'error', list_to_binary(["file type '", Else, "' is unsupported"])}
     end.
 
+%%% WARNING: Copied from fax_smtp
+%%% May be it should be placed somewhere else?
+-spec content_type_extension(ne_binary()) -> binary().
```
jamesaimonetti: Please move to fax_util and update any files using this function. Thanks!

lazedo:
@siplabs can you please change in a way that it can easily be enabled/disabled by group ? (openoffice, gif+jpeg+bmp).

siplabs:
@jamesaimonetti @lazedo for compatibilty `pdf` and `tiff` attachments always will be send.
`fax_smtp ` looks for `skip_openoffice_document` and `image_min_size`.
If `skip_openoffice_document` is true all openoffice compatible documents will be discarded.
If `image_min_size` is not set all images will be discarded. Image will be send only if both width and height is greater then `image_min_size`.
`image_min_size` should be in format `<WIDTH>x<HEIGHT>`

lazedo:
@siplabs i still don't agree with the images question, i want to be able to completely disable them (by default).
i would use something like 
"attachments" : { "open office" : {"Enabled" : true, "Content-Types" : {  "x1" : { Extension : "y1",  Enabled : False},  "x2" : { Extension : "y2"},  "x3" : { Extension : "y3"}}},
"images" : {"Enabled" : true, "Content-Types" : {  "i1" : { Extension : "e1",  Enabled : False},  "i2" : { Extension : "e2"},  "i3" : { Extension : "e3", "Min-Size" : {width : 88, height : 99}}},

instead of calling os shell command each time to collect the mime types, grab it from whapps:config with defaults built from proplist in code.

this can later be extended to support different converters per mime-type

Enabled field at content-type level should be considered because we may want to disable a specific content type.
thoughts ?

siplabs:
@lazedo images completely disabled by default (except `tiff`). To enable processing images you should add option `image_min_size` with a propriate value.

I think the default behaviour to accept convertable content type is good.
When new converters will be added you will need only to update system but not configuration.

I can add option `ignore-content-types`. It will be array of ignored content types.

In my opinion your configuration is too compicated.
Can you write configuration to accept `jpeg` images with `min-size` more than 100 x 100? Just for example.

siplabs:
@lazedo Please provide an example of configuration you'd like to see there (keeping in mind previous comment). We need this feature to be finished and finally merged, please guide us a bit so it fits your suggestions. Thanks!

applications/fax/src/fax_smtp.erl
```diff
@@ -670,90 +669,110 @@ process_parts([], #state{filename='undefined'
...
+    Cmd = io_lib:format(<<"identify -format \"%[fx:w]x%[fx:h]\" ~s">>, [Filename]),
+    [W, H] = re:split(os:cmd(Cmd), "x"),
+    Width = wh_util:to_integer(W),
+    Height = wh_util:to_integer(H),
+    case MinX =< Width andalso MinY =< Height of
+        true ->
+            {'ok', NewState};
+        false ->
+            lager:debug("ignoring part ~s", [CT]),
+            file:delete(Filename),
```
lazedo: dialyzer warnings, should be _ = file:delete(Filename),

applications/fax/src/fax_smtp.erl
```diff
@@ -670,90 +669,110 @@ process_parts([], #state{filename='undefined'
...
+    Extension = fax_util:content_type_to_extension(CT),
+    {'ok', Filename} = write_tmp_file(Extension, Body),
+    {'ok', State#state{filename=Filename
+                       ,content_type=CT
+                      }}.
+
+-spec maybe_process_openoffice_document(ne_binary(), binary() | mimemail:mimetuple(), state()) ->
+                                        {'ok', state()}.
+maybe_process_openoffice_document(CT, Body, State) ->
+    case whapps_config:get_is_true(?CONFIG_CAT, <<"skip_openoffice_document">>) of
```
lazedo: make this for the positive, "allow_openoffice_documents"

applications/fax/src/fax_smtp.erl
```diff
@@ -670,90 +669,110 @@ process_parts([], #state{filename='undefined'
...
-    'ok' = write_file(Filename, Body),
-    {'ok', State#state{filename=Filename
-                       ,content_type=CT
-                      }};
-process_part(<<"application/octet-stream">>, _Headers, Parameters, Body, State) ->
+maybe_process_part(CT, _Parameters, Body, State)
+  when CT =:= <<"image/tiff">> orelse CT =:= <<"application/pdf">> ->
+    process_part(CT, Body, State);
+maybe_process_part(<<"image/", _/binary>> = CT, _Parameters, Body, State) ->
+    case whapps_config:get_binary(?CONFIG_CAT, <<"image_min_size">>) of
```
lazedo: add here a check to list of allowed images.
image_min_size should come after or andalso

lazedo:
@siplabs finally got some time to look into this and agreed on keep it simple. please rebase and address the small issues. create a default list of image type allowed and use whapps_config. also, if you find a nice way to do the same for openoffice types that would be great but not required.

applications/fax/src/fax_smtp.erl
fenollp: Or use `wh_util:delete_file/1`

siplabs:
@lazedo 
I am going to add option `allowed_content_types`. Here is a sample:
```json
"allowed_content_types": [
    "application/pdf",
    {"type": "image/tiff"},
    {"prefix": "image"}
]
```
Content type can be discribed as a string or an object.
Descriptions `"image/tiff"` and `{"type": "image/tiff"}` are equal.
Also using field `prefix` may be described several content types, which is begining with this prefix. For example `{"prefix": "application/vnd.oasis.opendocument."}` to describe all OpenOffice documents.
Using objects may be useful for UI, they can be extended with `description` or some other useful fields.
Also I can add option `denied_content_types` with same syntax.
If type will be matched by both options than option with greater length will be applied.
And I want to add all processable documents as default.
What do you think about it?

lazedo:
@siplabs `If type will be matched by both options than option with greater length will be applied.` make deny override  allow if it matches.
you can add all processable to allow but put a deny for image. admin would have at least to remove that to enable image processing.

siplabs:
@lazedo rebased and updated.
Images disabled by default except tiff.

lazedo:
@siplabs great job! thanks.

PR #1114, opened 2015-06-22
fixes to compile erlang 17 and environments with python3

fenollp:


applications/acdc/src/acdc_agent_listener.erl
```diff
@@ -86,7 +86,7 @@
          ,is_thief = 'false' :: boolean()
          ,agent :: agent()
          ,agent_call_ids = [] :: api_binaries() | wh_proplist()
-         ,cdr_urls = dict:new() :: dict() %% {CallId, Url}
+         ,cdr_urls = dict:new() :: dict:dict() %% {CallId, Url}
```
fenollp: There's `nowarn_deprecated_type` over at http://www.erlang.org/doc/man/compile.html FYI.

To really switch to 17 there's a bunch of gotchas that will happen at runtime (or maybe those can get caught running xref/dialyzer using 17). One way to find most of them is to go through all CHANGELOGs in between releases.
I know there'll be some issues with the sha1 and md5 functions from the crypto module…

jamesaimonetti:
I presume this will not be merged until we shift to 17 in full?

fenollp:
yes I created the PR so we have a place to discuss and locked it to not merge it yet

fenollp:
Actually we could merge the `s/python/python2/g` changes anytime if they were part of their own commit.

fenollp:
#1143 dos the Python stuff

fenollp:
Note that compile fails because some deps (nksip) need to have the same kind of type updating.

fenollp:
Closing this in favour of PR #1239

PR #1121, opened 2015-06-23
Kazoo 3771: find vm box timezone

jamesaimonetti:


core/kazoo_documents-1.0.0/src/kz_documents.hrl
```diff
@@ -3,5 +3,7 @@
 -include_lib("whistle/include/wh_types.hrl").
 -include_lib("kazoo_documents/include/kazoo_documents.hrl").
 
+-define(DEFAULT_TIMEZONE, <<"America/Los_Angeles">>).
+
```
lazedo: @jamesaimonetti can we default this to whapps_account_config:get_global or whapps_config:get  ?

jamesaimonetti: We could, but I'm not sure what config doc to put it in; I was just preserving existing functionality.

jamesaimonetti: Maybe the `accounts` doc?

PR #1125, opened 2015-06-24
Enable PATCH for the accounts crossbar endpoint

kalda341:


applications/crossbar/src/modules/cb_accounts.erl
```diff
@@ -261,6 +265,17 @@ post(Context, AccountId, ?MOVE) ->
         _Status -> Context
     end.
 
+-spec patch(cb_context:context(), path_token()) -> cb_context:context().
+patch(Context, _AccountId) ->
+    Context1 = crossbar_doc:save(Context),
+    case cb_context:resp_status(Context1) of
+        'success' ->
+           _ = provisioner_util:maybe_update_account(Context1);
```
fenollp: This clause needs to return a `cb_context:context()`, also `provisioner_util` functions should run inside their own process, [like in cb_accounts](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/modules/cb_accounts.erl#L248)

applications/crossbar/src/modules/cb_accounts.erl
```diff
@@ -397,6 +412,10 @@ prepare_context(Context, AccountId, AccountDb) ->
                                  ,{fun cb_context:set_account_id/2, AccountId}
                                 ]).
 
+-spec validate_patch(api_binary(), cb_context:context()) -> cb_context:context().
+validate_patch(AccountId, Context) ->
+    crossbar_doc:patch_and_validate(AccountId, Context, fun validate_request/2).
```
fenollp: Given that this function is very short and will only be called from one place, why not just put its contents in the corresponding `validate_account/3` clause?

jamesaimonetti: I personally don't mind this, since we break out of `validate_account` to more specifically-named functions in general.

jamesaimonetti:
Appears you've opened a new PR instead; closing this one.

fenollp:
Other PR is #1128

PR #1132, opened 2015-06-29
Kazoo 3795: grouped CDRs

jamesaimonetti:


applications/crossbar/src/modules/cb_cdrs.erl
```diff
@@ -24,6 +24,10 @@
          ,to_csv/1
         ]).
 
+-ifdef(TEST).
+-export([group_cdrs/1]).
```
fenollp: You know what we could do? Add `+export_all` to the Makefile target that builds tests. This way it would be less painful to write tests for any function. It would also make it kind of less clear to know which functions have tests though…

applications/crossbar/src/modules/cb_cdrs.erl
```diff
@@ -594,7 +613,145 @@ load_cdr(CDRId, Context) ->
...
+-type group_values() :: [group_value(),...] | [].
 
+-type transform_fun() :: fun((wh_json:object()) -> wh_json:object()).
+
+-ifdef(TEST).
+-spec group_cdrs(wh_json:objects()) ->
+                        wh_json:object().
+group_cdrs(JObjs) ->
+    group_cdrs(JObjs, fun wh_util:identity/1).
 -endif.
```
fenollp: Ah I see. Well maybe this should be inside the `test/` and calling `group_cdrs/2` from there?

fenollp:
LGTM

k-anderson:
@jamesaimonetti just set an email but I would like to determine if we can use a view for this to better support date ranges before accepting.

k-anderson:
Deprecated by https://github.com/2600hz/kazoo/pull/1231

PR #1134, opened 2015-07-02
KAZOO-3935: using Kazoo's accessors

fenollp:
Review takes around 2 hours…

But be assured I did not break anything! :)


fenollp:
Don't look at it! Just merge it!

PR #1136, opened 2015-07-02
KAZOO-3936: fix travis for PRs

fenollp:


fenollp:
okay Travis does not recognise that this is a PR, thus lets Dialyzer check weird things.

Related Travis issue: https://github.com/travis-ci/travis-ci/issues/2881

Travis is configured to build PRs for kazoo (as well as pushes) but it seems it never built any PR for us..

lazedo:
@fenollp i don't understand what this is trying to resolve

fenollp:
@lazedo When Dialyzing, we can't check every file or Travis would die. So I'm trying to get Travis to check only files changed in the PR. That's why there's a diff against `master`.

But diffing against `master` also adds files changed on the up-to-date `master` branch, even though they were not modified in the PR. That's why sometimes there are Dialyzer errors about files just added to `master` (most of the time the check would run properly, as the files exist already).

What this PR is trying to solve is the way Travis finds changed files. It tries to find the first commit of the PR and does a diff from there.

There's 2 ways to spawn Travis builds: on pushes & on PRs. This fix is supposed to find the right modified files *when it's a PR build*; except Travis never triggers PR builds for us apparently…

lazedo:
@fenollp got it but, if we merge the pending changes of dialyze script (only analyze existing beam files) in https://github.com/2600hz/kazoo/pull/1135 our problems go away. will this keep working for other branches before PR is submitted ?

fenollp:
Closing this to only use your changes to the escript would fix the false positives we've been having yes.

Currently pending PRs would need to be rebased to get the changes to the escript. After that new PRs would be good. I'm not sure I understand your question

lazedo:
@fenollp what i'm asking is, if i push a new branch would travis build it ? or is this only to work for PRs ?
i'm confortable with the way it works now if we have the that "file_exists" merged.

fenollp:
This was supposed to be a special case for when PR builds are triggered, but since Travis is weird with kazoo…

PR #1137, opened 2015-07-03
Create onboarding.md

kalda341:
Added basic documentation for cb_onboarding.


jamesaimonetti:
Hi @kalda341, have you tested this API? It is pretty old and I'm not sure if it actually still works or not.

If it does, the only change I'd suggest making is changing all the values in the JSON to `{PLACEHOLDER}`-style variables. Too often people copy things verbatim to try and get...interesting results! :) So I'd use things like `{CITY}`, `{STATE}`, `{EMAIL}`, etc instead.

PR #1138, opened 2015-07-03
KAZOO-3942: added option "repeats" to cf_ring_group

siplabs:


applications/callflow/src/module/cf_ring_group.erl
```diff
@@ -24,14 +24,28 @@
 %%--------------------------------------------------------------------
 -spec handle(wh_json:object(), whapps_call:call()) -> 'ok'.
 handle(Data, Call) ->
-    case get_endpoints(Data, Call) of
+    Repeat = wh_json:get_value(<<"repeats">>, Data, 1),
```
jamesaimonetti: I'd use `wh_json:get_integer_value/3` to ensure Repeat is an integer (since `"repeats":"3"` could be stored).

jamesaimonetti:
@siplabs I went ahead and made the change; I also refactored the function into smaller functions, as well as added a function to constrain Repeats to positive integers. I also added `callflows.ring_group.json` schema so we can better validate the Data portion of the ring group. I tested all the options in Monster UI and was able to create/update the callflow using that validation. Let me know if you find places where the schema breaks down.

PR #1139, opened 2015-07-03
KAZOO-3943: added "weighted_random" strategy to cf_ring_group

siplabs:
This strategy to order endpoints by weight.

### DIFF
--- a/applications/callflow/doc/ring_group.md
+++ b/applications/callflow/doc/ring_group.md
@@ -1,0 +1,54 @@
+/*
+Section: Callflows
+Title: Ring Group
+Language: en-US
+Version: 3.20
+*/
+
+Ring group callflow element allows calling multiple endpoints with given strategy and timeout.
+
+## Example of `data` payload
+
+```
+"data": {
+    "endpoints": [...],
+    "strategy": "strategy",
+    "timeout": seconds,
+    "repeats": number_of_repeats
+}
+```
+
+## Mandatory fields
+**endpoints** - array of endpoints (see below)
+
+## Optional fields
+
+**strategy** - `single` | `simultaneous` | `weighted_random`, deafult is `simultaneous`
+**timeout** - time to call the endpoint before moving further, default is `20`
+**repeats** - number of repeats (rounds) this group will be called, default is `1`
+
+### Enpoint format
+#### Example
+
+```
+[...,
+    {
+        "id": "id-of-entpoint",
+        "endpoint_type": "type of endpoint",
+        "weight": weight,
+        "timeout": timeout,
+        "delay": delay
+    },
+...
+]
+```
+#### Mandatory fields
+
+**id**
+
+#### Optional fields
+
+**endpoint_type** - `device` | `user` | `group`, default is `device`
+**weight** - integer from 1 to 100, used by `weighted_random` strategy
+**timeout** - timeout to call the given endpoint, default is `20`
+**delay** - delay before this endpoint is called, default is `0`
--- a/applications/callflow/doc/ring_group.md
+++ b/applications/callflow/doc/ring_group.md
@@ -23,7 +23,7 @@ Ring group callflow element allows calling multiple endpoints with given strateg
 
 ## Optional fields
 
-**strategy** - `single` | `simultaneous` | `weighted_random`, deafult is `simultaneous`
+**strategy** - `single` | `simultaneous` | `weighted_random`, default is `simultaneous`
 **timeout** - time to call the endpoint before moving further, default is `20`
 **repeats** - number of repeats (rounds) this group will be called, default is `1`
 
--- a/applications/callflow/src/module/cf_ring_group.erl
+++ b/applications/callflow/src/module/cf_ring_group.erl
@@ -15,6 +15,9 @@
 
 -export([handle/2]).
 
+-ifdef(TEST).
+-export([weighted_random_sort/1]).
+-endif.
 %%--------------------------------------------------------------------
 %% @public
 %% @doc
--- a/applications/callflow/src/module/cf_ring_group.erl
+++ b/applications/callflow/src/module/cf_ring_group.erl
@@ -57,7 +57,10 @@ repeat(_Data, _Call, _N, 'fail') ->
     'stop' | 'fail' | 'continue'.
 attempt_endpoints(Endpoints, Data, Call) ->
     Timeout = wh_json:get_integer_value(<<"timeout">>, Data, ?DEFAULT_TIMEOUT_S),
-    Strategy = wh_json:get_binary_value(<<"strategy">>, Data, ?DIAL_METHOD_SIMUL),
+    Strategy = case wh_json:get_binary_value(<<"strategy">>, Data, ?DIAL_METHOD_SIMUL) of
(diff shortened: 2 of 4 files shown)


applications/callflow/src/module/cf_ring_group.erl
```diff
@@ -193,14 +208,50 @@ order_group_members(Member, JObj) ->
...
+                        ,{<<"weight">>, GroupWeight}
                        ]
                        ,Member
                       ).
+
+-spec weighted_random_sort(wh_proplist()) -> wh_json:objects().
+weighted_random_sort(Endpoints) ->
+    random:seed(erlang:now()),
+    WeightSortedEndpoints = lists:sort(Endpoints),
+    weighted_random_sort(WeightSortedEndpoints, []).
```
fenollp: Why not just `shuffle/1`?

applications/callflow/src/module/cf_ring_group.erl
```diff
@@ -34,7 +34,10 @@ handle(Data, Call) ->
 -spec attempt_endpoints(wh_json:objects(), wh_json:object(), whapps_call:call()) -> 'ok'.
 attempt_endpoints(Endpoints, Data, Call) ->
     Timeout = wh_json:get_integer_value(<<"timeout">>, Data, ?DEFAULT_TIMEOUT_S),
-    Strategy = wh_json:get_binary_value(<<"strategy">>, Data, ?DIAL_METHOD_SIMUL),
+    Strategy = case wh_json:get_binary_value(<<"strategy">>, Data, ?DIAL_METHOD_SIMUL) of
```
jamesaimonetti: I think if you rebase to master, now that KAZOO-3942 is in there (with ring group schema), we can go back to just fetching the value. Otherwise, let's put this in a function.

applications/callflow/doc/ring_group.md
```diff
@@ -0,0 +1,54 @@
...
+    "repeats": number_of_repeats
+}
+```
+
+## Mandatory fields
+**endpoints** - array of endpoints (see below)
+
+## Optional fields
+
+**strategy** - `single` | `simultaneous` | `weighted_random`, deafult is `simultaneous`
```
jamesaimonetti: Should be `default`

applications/callflow/src/module/cf_ring_group.erl
```diff
@@ -93,13 +96,24 @@ start_builder(EndpointId, Member, Call) ->
...
-            ,Type =:= <<"device">>
-            ,Id =/= whapps_call:authorizing_id(Call)
-    ].
+    ResolvedEndpoints = resolve_endpoint_ids(Members, [], Data, Call),
+    FilteredEndpoints = [{Weight, {Id, wh_json:set_value(<<"source">>, ?MODULE, Member)}}
+                         || {Type, Id, Weight, Member} <- ResolvedEndpoints
+                            ,Type =:= <<"device">>
+                            ,Id =/= whapps_call:authorizing_id(Call)
+                        ],
+    Strategy = wh_json:get_value(<<"strategy">>, Data, ?DIAL_METHOD_SIMUL),
```
jamesaimonetti: This would be another reason to have a `strategy(Data)` function; so that accessing it is consistent within the module.

applications/callflow/src/module/cf_ring_group.erl
```diff
@@ -193,14 +208,50 @@ order_group_members(Member, JObj) ->
...
                         ,{<<"delay">>, wh_json:get_value(<<"delay">>, Endpoint, DefaultDelay)}
                         ,{<<"timeout">>, wh_json:get_value(<<"timeout">>, Endpoint, DefaultTimeout)}
+                        ,{<<"weight">>, GroupWeight}
                        ]
                        ,Member
                       ).
+
+-spec weighted_random_sort(wh_proplist()) -> wh_json:objects().
+weighted_random_sort(Endpoints) ->
+    random:seed(erlang:now()),
```
jamesaimonetti: Please use `os:timestamp()` instead; `erlang:now()` requires a global lock.

jamesaimonetti:
@siplabs I would love to see a `callflows/test/cf_ring_group_test.erl` with unit tests for the weighted sorting.

applications/callflow/src/module/cf_ring_group.erl
siplabs: @fenollp 
I think it may be other strategies to shuffle endpoints. So this name is to know which strategy is used.

siplabs:
@jamesaimonetti , rebased to master, added test and updated after review.

applications/callflow/test/cf_ring_group_test.erl
```diff
@@ -0,0 +1,21 @@
...
+
+-include_lib("eunit/include/eunit.hrl").
+
+weighted_random_sort_test() ->
+    EndpointsInt = [
+                    {1, <<"1">>}
+                    ,{2, <<"2">>}
+                    ,{3, <<"3">>}
+                   ],
+    Endpoints = cf_ring_group:weighted_random_sort(EndpointsInt),
```
jamesaimonetti: Please test multiple endpoints at a weight:

```
EndpointsInt = [{1, <<"1a">>}
                ,{1, <<"1b">>}
                ,{2, <<"2">>}
                ,{3, <<"3">>}
               ],
```

This should return a list of `[<<"1X">>, <<"1Y">>, <<"2">>, <<"3">>]` where `X` and `Y` are either `a` or `b`. Does that make sense? Make sure the endpoints are actually sorted by weight, then randomly sorted within their weight group.

applications/crossbar/priv/couchdb/schemas/callflows.ring_group.json
```diff
@@ -66,7 +66,8 @@
             "description": "How to ring the members of the group",
             "enum": [
                 "simultaneous",
-                "single"
+                "single",
```
jamesaimonetti: Please add `weight` property to the `endpoints` list of properties.

siplabs:
@jamesaimonetti 
added `weight` property to the `endpoints` list of properties in schema `callflows.ring_group.json`

I can't create test with expected order of endpoints. When `weighted_random` strategy is used the order of all endpoints is always random but endpoints with greater weight will get to the beginning of the list with greater probability.

applications/callflow/src/module/cf_ring_group.erl
```diff
@@ -225,22 +241,68 @@ order_group_members(Member, JObj) ->
...
 -spec repeats(wh_json:object()) -> pos_integer().
 repeats(Data) ->
-    case wh_json:get_integer_value(<<"repeats">>, Data) of
-        'undefined' -> 1;
-        N when N < 1 -> 1;
-        N -> N
+    case wh_json:get_integer_value(<<"repeats">>, Data, 1) of
+        N when N >= 0 -> N;
+        _ -> 1
+    end.
```
fenollp: You can use `max(1, N)` here.

jamesaimonetti:
@siplabs thanks, i updated the docs a bit (THANK YOU for starting them) and refactored a few functions, added some types for clarity, and constrained the weight. Otherwise didn't mess with the sorting logic.

PR #1144, opened 2015-07-07
wh_media_recording terminates listener after recording starts

StevenGolemme:


fenollp:
I'm not familiar with this code. Why not apply the same commit on `maybe_start_recording` clauses?

StevenGolemme:
@fenollp I had missed those.  Thanks

lazedo:
@StevenGolemme what's the purpose of this commit ? what are you trying to solve ?

StevenGolemme:
@lazedo 
We noticed that when we enabled call-recordings to start on 'answer', wh_media_recording listener would receive a CHANNEL_ANSWER event.  The CHANNEL_ANSWER event does a handle_cast on 'maybe_start_recording_on_answer' with a #state{is_recording='false', should_store={'true', 'other', _}. This starts the call recording at line 261 but then immediatly terminates the listener by returning a {'stop', 'normal', State}. 

The call-recording was being recorded by freeswitch and stored in the /tmp directory but kazoo wasn't doing the POST because wh_media_recording was being terminated.

We were able to resolve the issue by returning a {'noreply', State}.

jamesaimonetti:
Maybe @macpie can weigh in here, but I thought this particular code path setup ecallmgr to issue the http_put when recording finished. If @macpie doesn't respond soon, I'll look more deeply at this.

StevenGolemme:
@jamesaimonetti it is an http_put i miss-spoke in the above comment, but the http_put doesn't happen because of the listener being terminated

Mike-Dunton:
Without Steven's pull request  the listener process is stopped and the CHANNEL_DESTROY event is never handled and this never happens.
```
handle_cast('stop_call', #state{is_recording='true'
                                ,media_name=MediaName
                                ,format=Format
                                ,call=Call
                                ,should_store=Store
                               }=State) ->
    lager:debug("recv stop_call event"),
    save_recording(Call, MediaName, Format, Store),
    lager:debug("sent store command"),
    {'noreply', State#state{store_attempted='true'
                            ,is_recording='false'
                           }};
```

lazedo:
@StevenGolemme i agree with this PR, however there seems to be missing the start timers.
check if this commit https://github.com/2600hz/kazoo/commit/4e940cb4367f7a3db352f2a22a9891475a3a8c91
 works for you and include it in this pull request. also rebase your branch.

lazedo:
@StevenGolemme @Mike-Dunton does this work for you ?

StevenGolemme:
@lazedo works for me

lazedo:
@StevenGolemme @Mike-Dunton

Thank You

PR #1145, opened 2015-07-08
KAZOO-3960: added next/prev actions to cf_voicemail

siplabs:


applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -755,12 +762,24 @@ play_messages([H|T]=Messages, Count, #mailbox{timezone=Timezone}=Box, Call) ->
             lager:info("caller choose to save the message"),
             _ = whapps_call_command:b_prompt(<<"vm-saved">>, Call),
             _ = set_folder(?FOLDER_SAVED, H, Box, Call),
-            play_messages(T, Count, Box, Call);
+            play_messages(T, [H|PrevMessages], Count, Box, Call);
+        {'ok', 'prev'} ->
+            lager:info("caller choose to listen previous message"),
```
jamesaimonetti: I'd like to put these into sub functions, `play_previous_message(Messages, PrevMessages, Count, Box, Call)` and handle the `case` as function clauses instead. Same for next message.

siplabs: @jamesaimonetti done

jamesaimonetti:
Thanks @siplabs !

jamesaimonetti:
@siplabs can you add MakeBusy tests for these two actions? I'll file a ticket for it

PR #1152, opened 2015-07-13
Kazoo 3915 - expose presence reset for device/user/arbitrary extensions

jamesaimonetti:


lazedo:
@jamesaimonetti why are we removing the search functions recently added/updated ?

jamesaimonetti:
@lazedo When I was asking about refactoring the module, it was partly because the module was pretty dated with the direct use of `#cb_context{}`, but also I wasn't sure that the presentity search functionality made sense from an API perspective. I wanted to see if we could make a friendlier search (since most people, I would guess, would use the API to search BLF state for some device/presence_id/extension.

So I removed the old code so we could start fresh on the Crossbar API side with the intent to revisit a GET for presence state and maybe a POST to manually set the presence state. But I didn't want to hold up the reset capabilities because support would benefit greatly from people being able to reset BLF themselves.

jamesaimonetti:
@lazedo I've re-added search, however I can't really tell how to test `type=presentity`. I don't see anything in omnipresence binding for `wapi_omnipresence:publish_search_req/1` payloads. Who should be listening for those requests? I see [here](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/modules/cb_presence.erl#L132) that you counted kamailio servers but then publish the query using [wapi_omnipresence](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/modules/cb_presence.erl#L135). I would think a Kamailio-bound query should be in `wapi_presence` instead, no?

lazedo:
@jamesaimonetti `wapi_omnipresence:publish_search_req/1` will be answered by kamailio. the reason to bind for omnipresence instead of presence is that there are other presence queries within kazoo and we don't want to flood kamailio servers with these requests.

lazedo:
@jamesaimonetti added a fix for handle_reset in omnipresence , it was issuing duplicate events to kamailio

PR #1157, opened 2015-07-17
KAZOO-3976: updated `resource_exists` callback

siplabs:


lazedo:
@siplabs what is the use case for this ?

siplabs:
@lazedo 
I want to be able to check if resource exists in `resource_exists` callback, not in `validate`.

lazedo:
i understand that, but why ?

applications/crossbar/src/api_util.erl
```diff
@@ -868,17 +868,24 @@ ensure_content_type(CT) -> CT.
...
-    Responses = crossbar_bindings:map(Event, Params),
-    crossbar_bindings:any(Responses) and 'true';
-does_resource_exist(_Context, _ReqNouns) ->
-    'false'.
+    Responses = crossbar_bindings:map(Event, [Context, Params]),
+    SuccessResponses = [X || {'true', X} <- Responses],
+    case SuccessResponses of
+        [] -> {'false', hd([X || {'false', X} <- Responses])};
+        [Context1] -> {'true', Context1};
+        [Context1 | _] ->
```
jamesaimonetti: I don't see a need to differentiate one vs many responses; I think the single item clause can be removed.

applications/crossbar/src/api_util.erl
```diff
@@ -868,17 +868,24 @@ ensure_content_type(CT) -> CT.
...
     does_resource_exist(Context, cb_context:req_nouns(Context)).
 
 does_resource_exist(Context, [{Mod, Params}|_]) ->
     Event = api_util:create_event_name(Context, <<"resource_exists.", Mod/binary>>),
-    Responses = crossbar_bindings:map(Event, Params),
-    crossbar_bindings:any(Responses) and 'true';
-does_resource_exist(_Context, _ReqNouns) ->
-    'false'.
+    Responses = crossbar_bindings:map(Event, [Context, Params]),
+    SuccessResponses = [X || {'true', X} <- Responses],
```
jamesaimonetti: I think I'd like to use (`crossbar_bindings:succeeded/1`)[https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/crossbar_bindings.erl#L97] here instead.

applications/crossbar/src/api_util.erl
```diff
@@ -868,17 +868,24 @@ ensure_content_type(CT) -> CT.
...
 does_resource_exist(Context) ->
     does_resource_exist(Context, cb_context:req_nouns(Context)).
 
 does_resource_exist(Context, [{Mod, Params}|_]) ->
     Event = api_util:create_event_name(Context, <<"resource_exists.", Mod/binary>>),
-    Responses = crossbar_bindings:map(Event, Params),
-    crossbar_bindings:any(Responses) and 'true';
-does_resource_exist(_Context, _ReqNouns) ->
-    'false'.
+    Responses = crossbar_bindings:map(Event, [Context, Params]),
```
jamesaimonetti: This should be `crossbar_bindings:map(Event, [Context | Params])` to match how we call `map/2` elsewhere. The crossbar endpoint modules must export different arity functions to handle different number of parameters from the URI.

*  `resource_exists() -> 'true'.` should become `resource_exists(Context) -> {'true', Context}.`
*  `resource_exists(_PathToken) -> 'true'.` should become `resource_exists(Context, _PathToken) -> {'true', Context}.`
*  `resource_exists(_PathToken1, _PathToken2) -> 'true'.` should become `resource_exists(Context, _PathToken1, _PathToken2) -> {'true', Context}.`

et cetera...

siplabs: @jamesaimonetti 
Sorry, haven't enough time to answer.

I called `crossbar_bindings:map` in this way to have only one function with arity 2.
It prevents spagetti of specs.
Developer can add clause to match "generic" situation.
Developer still can pattern match all needed variables.
In my opinion `resource_exists` should check if resource realy exists, and can skip checks for actions. Actions may be checked in `allowed_methods`.
Getting `/v../accounts/.../<resource_type>/<resource_id>` and `/v../accounts/.../<resource_type>/<resource_id>/<action>` will check the same resource. Do we realy need several function for this?

jamesaimonetti: Yes, to match how the rest of the callbacks operate. Having verbs on the URI isn't good practice (so `/action` shouldn't be there) anyway; should be `/things/{thing_id}/{sub_thing_or_id}` if this is accessing parts of `{thing_id}`. So `resource_exists(Context)` checks that the collection exists, `resource_exists(Context, ThingId)` checks that ThingId exists, etc... If you are acting on ThingId, should be done in the request params, imo.

siplabs:
@jamesaimonetti @lazedo rebased and updated
also added `context` to `allowed_methods` callbacks

applications/crossbar/src/api_util.erl
```diff
@@ -868,17 +870,28 @@ ensure_content_type(CT) -> CT.
...
     does_resource_exist(Context, cb_context:req_nouns(Context)).
 
 does_resource_exist(Context, [{Mod, Params}|_]) ->
-    Event = ?MODULE:create_event_name(Context, <<"resource_exists.", Mod/binary>>),
-    Responses = crossbar_bindings:map(Event, Params),
-    crossbar_bindings:any(Responses) and 'true';
-does_resource_exist(_Context, _ReqNouns) ->
-    'false'.
+    Event = api_util:create_event_name(Context, <<"resource_exists.", Mod/binary>>),
+    Responses = case catch crossbar_bindings:map(Event, [Context | Params]) of
```
jamesaimonetti: I'd prefer using try/catch here:
```erlang
Responses =
    try crossbar_bindings:map(...) of
        Rs -> Rs
    catch _:_ -> []
    end,
```

applications/crossbar/src/api_util.erl
```diff
@@ -868,17 +870,28 @@ ensure_content_type(CT) -> CT.
...
+                    {'EXIT', _Reason} -> [];
+                    Responses_  -> Responses_
+                end,
+    SuccessResponses = crossbar_bindings:succeeded(Responses),
+    case SuccessResponses of
+        [] -> {'false', hd([X || {'false', X} <- Responses])};
```
jamesaimonetti: Why not call a function to find the first {'false', Response} tuple and return it, rather than going through the whole list, deconstructing the tuple only to build it again?

jamesaimonetti:
@siplabs How does this play with CORS/OPTIONS requests? Have you tested OPTIONS followed by GET/PUT/POST/DELETE ?

applications/crossbar/src/api_util.erl
```diff
@@ -42,6 +42,8 @@
 -define(MAX_UPLOAD_SIZE, whapps_config:get_integer(?CONFIG_CAT, <<"max_upload_size">>, 8000000)).
 
 -type halt_return() :: {'halt', cowboy_req:req(), cb_context:context()}.
+-type resource_exists() :: {boolean(), cb_context:context()}.
```
jamesaimonetti: I don't like this type's name :)

I don't have anything better than `resource_exists_return()` at the moment though. Maybe `resource_existence()`, since that's what you're conceptually returning?

siplabs:
@jamesaimonetti I didn't test OPTIONS and CORS requests cause this patch doesn't change anything in crossbar modules, all changes made to handle `context` variable and to return expected value.
I can test it tomorrow. Is working `kazoo-ui` will be proper test?

jamesaimonetti:
@siplabs yes, as long as kazoo-ui is making the OPTIONS request followed by the actual request, and things work as expected, I don't see an issue with this request (other than the minor changes I've commented on already).

siplabs:
@jamesaimonetti rebased and updated. Also `kazoo-ui` works as expected

lazedo:
@siplabs i would really like if you can explain your use case here because this seems to add more pressure to crossbar without the need for that. if the intent is to validate in resource_exists, why would there be a validate ? if the intention is to query couch for the "thing" in resource_exists then when using CORS in the UI (kazoo_ui or monster) you are actually querying couchdb twice. i think resource_exists should only validate the path not the content (account/device,...) and validate will continue to do its job of making sure that the "thing" is valid. so, without a clear use case, i don't see any benefit in merging this, quite the opposite.

siplabs:
@lazedo in current implementation `resource_exists` is unusable for validating only path, path validates in `allowed_methods` before `resource_exists` callback.
I think `resource_exists`  callback should check resource existence and may load a document using `crossbar_doc:load` and put it to cache.

lazedo:
@siplabs `resource_exists` callback calls `does_resource_exist` and if 'true' it calls `does_request_validate`. so, `does_resource_exist`is used for fast validation and `does_request_validate` for more costly validate operations. if your intention is to put something in the  `cache` during `does_resource_exist`, where does the current implementation prevents you to do that ?

siplabs:
@lazedo PR closed

PR #1159, opened 2015-07-20
Fix double prepend

Dinkor:
If phone number have "prepend" attribute - this prepend apllied in "cf_route_win" and "cf_endpoint" (each time it call cf_attributes:caller_id/2).


k-anderson:
This commit was merged manually to add the ticket to the commit message.  Sometimes git figures that out, sometimes not ;)

In this case the I need to manually close the PR, but the code is in 3.21+

PR #1160, opened 2015-07-20
KAZOO-3562: correctly update api_url for already loaded apps, correct…

mark2600:
…ly load all image files (icons and screenshots)

### DIFF
--- a/applications/crossbar/src/crossbar_maintenance.erl
+++ b/applications/crossbar/src/crossbar_maintenance.erl
@@ -866,7 +866,7 @@ maybe_update_app(AppPath, MetaData, MasterAccountDb, JObj) ->
 	    end
     end,
 
-    delete_old_images(CurrentDocId, MetaData, MasterAccountDb),
+    'ok' = delete_old_images(CurrentDocId, MetaData, MasterAccountDb),
     maybe_add_images(AppPath, CurrentDocId, MetaData, MasterAccountDb).
 
 -spec find_app(ne_binary(), ne_binary()) -> {'ok', wh_json:object()} |
@@ -905,8 +905,9 @@ delete_old_images(AppId, MetaData, MasterAccountDb) ->
 	end
     end,
 
-    [Delete(X) || X <- Icons],
-    [Delete(X) || X <- Screenshots].
+    _ = [Delete(X) || X <- Icons],
+    _ = [Delete(X) || X <- Screenshots],
+    'ok'.
 
 -spec maybe_add_images(file:filename(), ne_binary(), wh_json:object(), ne_binary()) -> 'ok'.
 maybe_add_images(AppPath, AppId, MetaData, MasterAccountDb) ->
--- a/applications/crossbar/src/crossbar_maintenance.erl
+++ b/applications/crossbar/src/crossbar_maintenance.erl
@@ -896,18 +896,19 @@ create_app(AppPath, MetaData, MasterAccountDb) ->
 -spec delete_old_images(ne_binary(), wh_json:object(), ne_binary()) -> 'ok'.
 delete_old_images(AppId, MetaData, MasterAccountDb) ->
     Icons       = [wh_json:get_value(<<"icon">>, MetaData)],
-    Screenshots = wh_json:get_value(<<"screenshots">>, MetaData),
+    Screenshots = wh_json:get_value(<<"screenshots">>, MetaData, []),
 
-    Delete = fun(Image) ->
-	case couch_mgr:fetch_attachment(MasterAccountDb, AppId, Image) of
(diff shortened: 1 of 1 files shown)


applications/crossbar/src/crossbar_maintenance.erl
```diff
@@ -843,11 +843,32 @@ maybe_create_app(AppPath, MetaData) ->
...
         {'error', _E} -> io:format(" failed to find app ~s: ~p", [AppName, _E])
     end.
 
+-spec maybe_update_app(file:filename(), wh_json:object(), ne_binary(), wh_json:object()) -> 'ok'.
+maybe_update_app(AppPath, MetaData, MasterAccountDb, JObj) ->
+    CurrentDocId  = wh_doc:id(JObj),
+    CurrentApiUrl = wh_json:get_value([<<"value">>, <<"api_url">>], JObj),
+
+    case wh_json:get_value(<<"api_url">>, MetaData) of
+	undefined     -> io:format(" not updating api_url, it is undefined~n");
```
lazedo: tickie your atoms

applications/crossbar/src/crossbar_maintenance.erl
```diff
@@ -867,51 +888,74 @@ create_app(AppPath, MetaData, MasterAccountDb) ->
...
+-spec delete_old_images(ne_binary(), wh_json:object(), ne_binary()) -> 'ok'.
+delete_old_images(AppId, MetaData, MasterAccountDb) ->
+    Icons       = [wh_json:get_value(<<"icon">>, MetaData)],
+    Screenshots = wh_json:get_value(<<"screenshots">>, MetaData, []),
+
+    _ = [safe_delete_image(MasterAccountDb, AppId, X) || X <- Icons],
+    _ = [safe_delete_image(MasterAccountDb, AppId, X) || X <- Screenshots],
+    'ok'.
+
+-spec safe_delete_image(ne_binary(), ne_binary(), ne_binary()) -> 'ok'.
```
jamesaimonetti: Spec should be `api_binary()` for third arg, since it can be `'undefined' | ne_binary()`. See [wh_types.hrl](https://github.com/2600hz/kazoo/blob/master/core/whistle-1.0.0/include/wh_types.hrl#L64) for a lot of the available types.

applications/crossbar/src/crossbar_maintenance.erl
```diff
@@ -867,51 +888,74 @@ create_app(AppPath, MetaData, MasterAccountDb) ->
...
+safe_delete_image(AccountDb, AppId, Image) ->
+    case couch_mgr:fetch_attachment(AccountDb, AppId, Image) of
+	{'ok', _}    -> couch_mgr:delete_attachment(AccountDb, AppId, Image);
+	{'error', _} -> 'ok'
+    end.
+
+-spec maybe_add_images(file:filename(), ne_binary(), wh_json:object(), ne_binary()) -> 'ok'.
+maybe_add_images(AppPath, AppId, MetaData, MasterAccountDb) ->
+    Icons       = [wh_json:get_value(<<"icon">>, MetaData)],
+    Screenshots = wh_json:get_value(<<"screenshots">>, MetaData),
```
jamesaimonetti: Should this default to `[]` in case there aren't screenshots?

k-anderson:
@jamesaimonetti since you were the last to comment I will defer to your judgement but looks good to me!

applications/crossbar/src/crossbar_maintenance.erl
```diff
@@ -867,51 +888,74 @@ create_app(AppPath, MetaData, MasterAccountDb) ->
...
             io:format(" failed to save app ~s to ~s: ~p~n", [wh_json:get_value(<<"name">>, MetaData), MasterAccountDb, _E])
     end.
 
--spec maybe_add_icons(file:filename(), ne_binary(), ne_binary()) -> 'ok'.
-maybe_add_icons(AppPath, AppId, MasterAccountDb) ->
-    try find_icons(AppPath) of
-        {'ok', Icons} -> add_icons(AppId, MasterAccountDb, Icons)
+-spec delete_old_images(ne_binary(), wh_json:object(), ne_binary()) -> 'ok'.
+delete_old_images(AppId, MetaData, MasterAccountDb) ->
+    Icons       = [wh_json:get_value(<<"icon">>, MetaData)],
```
fenollp: Why build a list?

mark2600: While currently apps may only have one icon, it is not inconceivable that they may have multiple in the future.  By processing them as a list we can handle that case if / when it comes around and not have to worry about introducing a bug in this code.

PR #1163, opened 2015-07-23
KAZOO-3993: if configured include a cluster_id in the request

k-anderson:
This will need to be backported to 3.19+


jamesaimonetti:
@k-anderson please backport the commits!

PR #1165, opened 2015-07-23
KAZOO-3989: use soft_hold for konami_hold

siplabs:


jamesaimonetti:
@siplabs Manually merged; all is well. Thanks

PR #1170, opened 2015-07-27
KAZOO-4006: handle absence of the messages

siplabs:


jamesaimonetti:
If a vmbox doc doesn't have `messages` as an array of objects, it isn't a valid vmbox (see the [schema](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/priv/couchdb/schemas/vmboxes.json#L54) ). How have you come across the need for this correction? Is there something in the application layer stripping `messages` off the document?

siplabs:
Well, it popped up after version upgrade because previosly it was not that strict. Anyway, there is no chance for the API users to find this bad vmboxes because they are not listed by the crossbar. We faced it in more than one installation so I think at least for now it should be here to help people migrate to newer releases.

PR #1172, opened 2015-07-27
KAZOO-3940: hold state for webhooks

siplabs:


lazedo:
CHANNEL_EXECUTE and CHANNEL_EXECUTE_COMPLETE will put a lot of pressure in webhooks. i don't see any added value in here

siplabs:
soft_hold application can not be tracked without them.

lazedo:
maybe the app should bind to rabbit to get the events or use another way to do it.this really puts a lot of pressure in webhooks. describe your use case in jira.

jamesaimonetti:
I could maybe be on board with CHANNEL_HOLD but will not accept CHANNEL_EXECUTE(_COMPLETE); as @lazedo points out, they are too numerous and would add a disproportionate pressure on resources to the value they provide. Those events should be consumed via blackhole/websocket connections.

Also, please be aware that webhooks has received a modest refactoring in KAZOO-3912 to better support adding new webhooks. I would advise holding off on webhook PRs until that is merged, as things have changed a bit internally.

jamesaimonetti:
@siplabs #1169 is merged. Please rebase and update to use the webhooks_skel (adapted for CHANNEL_HOLD/UNHOLD) to provide a hook for these events. CHANNEL_EXECUTE(_COMPLETE) isn't going to be accepted and should be removed from here.

siplabs:
Will do in couple of days! Thanks!

fenollp:
@siplabs If after the rebase Travis still does not start, close & reopen the PR :)

jamesaimonetti:
@siplabs rebase needed.

PR #1173, opened 2015-07-27
KAZOO-4008: do not trust to cache when respond to doc_update

siplabs:


lazedo:
why are you removing the use of cache ? if there's a use case for that we will tale a look. the cache should be invalidated if the document is updated

siplabs:
because of race condition between wh_cache and webhooks_listener. They start in parallel.

jamesaimonetti:
@siplabs Which cache do you mean? The document cache in `couch_mgr` should be started as part of the `whistle_couch` core application startup. This occurs on VM startup. Only after the core applications have started are the Kazoo apps started. So there shouldn't be any issue here.

Would love to see logs and steps to recreate any issue you're seeing if this isn't the case.

siplabs:
@jamesaimonetti 
I am about `?WH_COUCH_CACHE`. It is trying to flush doc at the same time when `webhooks_listener:handle_config` trying to read hook doc. Log attached to JIRA ticket.

You can POST webhook doc with `enabled:true` and `enabled:false` to recreate issue.

jamesaimonetti:
@siplabs Please rebase and I'll see if I can reproduce a stale cache situation.

siplabs:
@jamesaimonetti 
Done! :)

lazedo:
@siplabs yes, this can happen if webhooks cache gets notified before wh_couch_cache. i believe this may happen in other places where whapps have their config handlers to test document creation/update.

i propose we change this in couch_util instead, thoughts ?
the change is to flush locally in couch instead of waiting for cache to be notified and clear the entry
```
...
maybe_publish_doc(#db{}=Db, Doc, JObj) ->
    flush_cache_doc(Db, Doc),
...
maybe_publish_docs(#db{}=Db, Docs, JObjs) ->
    flush_cache_docs(Db, Docs),
...
```

lazedo:
@siplabs so the proposed fix i made is only effective in all-in-one deployments. we need to figure out a general fix for this later.
Thank you!

PR #1174, opened 2015-07-27
Install doc changes

mark2600:


doc/installation.md
```diff
@@ -32,6 +34,15 @@ There are six major components to a Kazoo system, they can all be installed on o
 * Monster-UI
   * Provides the Kazoo web interface for configuration and monitoring of the system.
 
+## Cluster design considerations
+
+There are a few concerns that should be planned for when designing a Kazoo cluster.
+
+* Design in enough bigcouch servers from the start
+  * It is very difficult to add additional database servers once the cluster has been deployed, as such you should plan in advance and add the maximum number of bigcouch servers that you forsee needing over the lifetime of the cluster from the start.
```
jamesaimonetti: This seems too extreme for most folks. Maybe tone it down a bit to say something like "Plan to provision enough database servers to cope with your expected growth over the next year or two." Something along these lines, anyway.

doc/installation.md
```diff
@@ -180,13 +206,15 @@ listen bigcouch-data 127.0.0.1:15984
...
+    server test3.cluster1.2600hz.com 192.168.1.102:5984 check
+    server test4.cluster1.2600hz.com 192.168.1.103:5984 check backup
 
 listen bigcouch-mgr 127.0.0.1:15986
   balance roundrobin
     server test1.cluster1.2600hz.com 192.168.1.100:5986 check
     server test2.cluster1.2600hz.com 192.168.1.101:5986 check
-    server test3.cluster1.2600hz.com 192.168.1.102:5986 check backup
+    server test3.cluster1.2600hz.com 192.168.1.102:5986 check
+    server test4.cluster1.2600hz.com 192.168.1.103:5986 check backup
```
jamesaimonetti: Should we provide two configs, one for zone 1 and one for zone 2, where zone 1's config has zone 2 servers as backups and zone 2 has zone 1's servers as backup?

doc/installation.md
```diff
@@ -424,6 +461,17 @@ Verify that the status shows nodes for BOTH Server 1 and Server 3
 ```
 # sup crossbar_maintenance create_account {ACCT NAME} {REALM} {LOGIN} {PASSWORD}
 ```
+### Load applications
+
+*Server 1*
+```
+# sup crossbar_maintenance init_apps /var/www/html/monster-ui/apps/ http://192.168.1.100:8000/v2
```
jamesaimonetti: Only need to run this once; if 2 crossbar servers are in use, should probably point this to an HAProxy to round-robin between crossbar servers instead. Or not, am ambivalent but it may be confusing?

PR #1176, opened 2015-07-27
Kazoo 3880

mark2600:


applications/callflow/src/module/cf_directory.erl
```diff
@@ -262,8 +262,21 @@ play_confirm_match(Call, User) ->
...
-            {'play', <<$/, (whapps_call:account_db(Call))/binary, $/, MediaID/binary>>}
+        MediaID     -> maybe_play_media(Call, User, MediaID)
+    end.
+
+-spec maybe_play_media(whapps_call:call(), directory_user(), api_binary()) -> whapps_call_command:audio_macro_prompt().
+maybe_play_media(Call, User, MediaID) ->
+    AccountDb = whapps_call:account_db(Call),
+    MediaFile = <<$/, AccountDb/binary, $/, MediaID/binary>>,
+
+    case couch_mgr:open_doc(AccountDb, MediaID) of
```
jamesaimonetti: Use `couch_mgr:open_cache_doc/2` to cache the media doc lookup locally

applications/callflow/src/module/cf_directory.erl
```diff
@@ -262,8 +262,21 @@ play_confirm_match(Call, User) ->
...
+	{'ok', Doc}    ->
+	    case wh_json:get_value(<<"_attachments">>, Doc) of
```
jamesaimonetti: Use `wh_doc:attachments/1` to fetch the object of attachments

PR #1177, opened 2015-07-28
KAZOO-3517: prevent process crash when getting a hangup while using c…

mark2600:
…f_directory

### DIFF
--- a/applications/callflow/src/module/cf_directory.erl
+++ b/applications/callflow/src/module/cf_directory.erl
@@ -144,17 +144,20 @@ directory_start(Call, State, CurrUsers) ->
 	    lager:error("failed to collect digits: ~p", [_Error]),
 	    cf_exe:stop(Call);
 
-	{'ok', DTMF} ->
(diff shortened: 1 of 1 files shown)


applications/callflow/src/module/cf_directory.erl
```diff
@@ -139,16 +139,22 @@ handle(Data, Call) ->
...
-            directory_start(Call, State, CurrUsers);
-        {'ok', DTMFS} ->
-            maybe_match(Call, add_dtmf(add_dtmf(State, DTMF), DTMFS), CurrUsers)
+    case play_directory_instructions(Call, sort_by(State)) of
+	{'error', _Error} ->
+	    lager:error("failed to collect digits: ~p", [_Error]),
+	    cf_exe:stop(Call);
+
+	{'ok', DTMF} ->
+	    case whapps_call_command:collect_digits(100, ?TIMEOUT_DTMF, ?TIMEOUT_DTMF, Call) of
```
jamesaimonetti: I'd like to see this in a sub function, maybe something like `collect_digits(Call, State, CurrUsers, DTMF)`.

PR #1178, opened 2015-07-28
KAZOO-3954: new api to apply more than one service plan to an account

macpie:


applications/crossbar/src/modules/cb_service_plans.erl
```diff
@@ -160,7 +166,29 @@ validate_service_plan(Context, PlanId, ?HTTP_DELETE) ->
...
+-spec post(cb_context:context()) -> cb_context:context().
 -spec post(cb_context:context(), path_token()) -> cb_context:context().
+post(Context) ->
+    Plans = cb_context:fetch(Context, <<"plans">>, []),
+    Services =
+        lists:foldl(
+            fun(PlanId, Acc) ->
+                Routines = [
+                    fun(S) -> wh_services:add_service_plan(PlanId, S) end
+                    ,fun(S) -> wh_services:save(S) end
```
jamesaimonetti: Can just be `fun wh_services:save/1`

applications/crossbar/src/modules/cb_service_plans.erl
```diff
@@ -160,7 +166,29 @@ validate_service_plan(Context, PlanId, ?HTTP_DELETE) ->
 %% (after a merge perhaps).
 %% @end
 %%--------------------------------------------------------------------
+-spec post(cb_context:context()) -> cb_context:context().
 -spec post(cb_context:context(), path_token()) -> cb_context:context().
+post(Context) ->
+    Plans = cb_context:fetch(Context, <<"plans">>, []),
+    Services =
+        lists:foldl(
+            fun(PlanId, Acc) ->
```
jamesaimonetti: This should be a named function like `apply_plan(PlanId, Acc)` so we can add type/spec info

fenollp:
LGTM

PR #1184, opened 2015-07-30
KAZOO-3956: Adding minimum and activation_charge to item

macpie:


core/kazoo_documents-1.0.0/src/kzd_service_plan.erl
```diff
@@ -73,12 +73,16 @@ merge_overrides(Plan, Overrides) ->
...
 item_activation_charge(Plan, Category, Item, Default) ->
-    wh_json:get_float_value([?PLAN, Category, Item, ?ACTIVATION_CHARGE]
-                            ,Plan
-                            ,Default
-                           ).
+    kzd_item_plan:activation_charge(
+        wh_json:get_json_value(
+            [?PLAN, Category, Item]
+            ,Plan
+            ,Default
```
fenollp: Should this one be `wh_json:new()` instead of `Default`?

core/whistle_services-1.0.0/src/wh_service_item.erl
```diff
@@ -48,6 +55,8 @@
                           ,cumulative_discount = 0 :: api_integer()
                           ,cumulative_discount_rate = 0.00 :: api_float()
                           ,bookkeepers = wh_json:new() :: wh_json:object()
+                          ,activation_charge = 0.00 :: api_float()
+                          ,minimum = 0 :: api_integer()
```
fenollp: Is the `api_` part needed? They have default non-`undefined` values…

macpie: They have a default value but does not mean they could not be undefined.

fenollp: Yah

PR #1186, opened 2015-07-31
KAZOO-4037: added the rejected notification, removed the deletion

k-anderson:
and fixed the listing of templates


k-anderson:
Since I will be traveling please make sure this is backported to 3.21 and/or you let me know when this is merged :)

PR #1187, opened 2015-08-01
KAZOO-4038: fix completed transition

k-anderson:


core/whistle_number_manager-1.0.0/src/wh_port_request.erl
```diff
@@ -155,60 +158,71 @@ charge_for_port(_JObj, AccountId) ->
...
-        {'error', 'conflict'} ->
-            lager:debug("port request doc was updated before we could re-save"),
-            PortReq;
-        {'error', _E} ->
-            lager:debug("failed to clear numbers: ~p", [_E]),
-            PortReq
+-spec finalize_number_transition(ne_binary(), {wh_json:object(), wh_proplist()}) -> transition_response().
+finalize_number_transition(_, {JObj, []}) ->
+    {'ok', Updated} = couch_mgr:ensure_saved(?KZ_PORT_REQUESTS_DB, JObj),
+    {'ok', Updated};
```
jamesaimonetti: You can pattern match and leave it as the last expression; no need to re-create the tuple again to return it.

PR #1191, opened 2015-08-05
KAZOO-3963 registrar multiple listeners

lazedo:


jamesaimonetti:
@lazedo my only issue with this is that `reg_route_req:init/0` calls `whapps_maintenance:refresh(?WH_SIP_DB)` which loads all views from the `WH_SIP_DB` to compare to what's on disk. This is fine for 1 listener, but if someone wanted 5 or 10, and had several instances, that becomes a flood of requests to the `WH_SIP_DB`. Probably fine in most cases, but I wonder if we could move it to another module, like `registrar_init`, that would do it once when the app starts up, rather than each time a `registrar_shared_listener` process is started.

Thoughts?

lazedo:
@jamesaimonetti good catch. i'll add registrar_init as temporary in supervisor

PR #1192, opened 2015-08-06
KAZOO-4061: remove iolist_to_binary on json object

k-anderson:


fenollp:
@k-anderson Could you then update the function's spec for `RespHeaders`?

k-anderson:
@fenollp updated!

PR #1197, opened 2015-08-10
KAZOO-3990: properly strip non-ascii characters when converting in al…

mark2600:
…pha_to_dialplan

### DIFF
--- a/applications/callflow/src/cf_util.erl
+++ b/applications/callflow/src/cf_util.erl
@@ -339,16 +339,7 @@ vm_count_by_owner(<<_/binary>> = AccountDb, <<_/binary>> = OwnerId) ->
 %%--------------------------------------------------------------------
 -spec alpha_to_dialpad(ne_binary()) -> ne_binary().
 alpha_to_dialpad(Value) ->
-    Stripped = [strip_nonalpha(C) || <<C>> <= wh_util:to_lower_binary(Value)],
-
-    Process = fun(Char, Acc) ->
-	case Char of
-	    [] -> Acc;
-	    _  -> [dialpad_digit(Char)|Acc]
-	end
-    end,
-
-    list_to_binary(lists:foldr(Process, [], Stripped)).
+    << <<(dialpad_digit(C))>> || <<C>> <= wh_util:to_lower_binary(Value), is_alpha(C) >>.
 
 %%--------------------------------------------------------------------
 %% @public
@@ -356,9 +347,9 @@ alpha_to_dialpad(Value) ->
 %%
 %% @end
 %%--------------------------------------------------------------------
--spec strip_nonalpha(char()) -> char().
-strip_nonalpha(Value) when Value =< $z andalso Value >= $a -> Value;
-strip_nonalpha(_NotAlpha) -> "".
+-spec is_alpha(char()) -> boolean().
+is_alpha(Char) ->
+    Char =< $z andalso Char >= $a.
 
 %%--------------------------------------------------------------------
 %% @public
--- a/applications/callflow/src/cf_util.erl
+++ b/applications/callflow/src/cf_util.erl
@@ -339,7 +339,16 @@ vm_count_by_owner(<<_/binary>> = AccountDb, <<_/binary>> = OwnerId) ->
 %%--------------------------------------------------------------------
 -spec alpha_to_dialpad(ne_binary()) -> ne_binary().
 alpha_to_dialpad(Value) ->
-    << <<(dialpad_digit(C))>> || <<C>> <= strip_nonalpha(wh_util:to_lower_binary(Value))>>.
+    Stripped = [strip_nonalpha(C) || <<C>> <= wh_util:to_lower_binary(Value)],
+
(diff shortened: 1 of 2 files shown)


mark2600:
I prefer the ASCII integers as I think it is more intuitive than $` and ${.  As for the clauses, is there any benefit to doing it that way vs the case ? I generally use a case in this situation as I think it makes for better readability.

lazedo:
@mark2600 if you make inclusive comparison (>= instead of >, ...) then you can use $a and $z instead of $` and ${  , which in IMHO would be the best option for readability.

mark2600:
I like that suggestion, done.

applications/callflow/src/cf_util.erl
```diff
@@ -339,7 +339,16 @@ vm_count_by_owner(<<_/binary>> = AccountDb, <<_/binary>> = OwnerId) ->
 %%--------------------------------------------------------------------
 -spec alpha_to_dialpad(ne_binary()) -> ne_binary().
 alpha_to_dialpad(Value) ->
-    << <<(dialpad_digit(C))>> || <<C>> <= strip_nonalpha(wh_util:to_lower_binary(Value))>>.
+    Stripped = [strip_nonalpha(C) || <<C>> <= wh_util:to_lower_binary(Value)],
```
jamesaimonetti: This all seems inefficient. When I look at it, I see three things happening:

1. Iterate through each character of `Value`
2. Filter out non a-z characters
3. Convert characters to dialpad digits.

This can be solved with a binary comprehension without the need to build intermediary lists and filter empty lists and all the ruckus below.

My suggestion:

```erlang
-spec alpha_to_dialpad(ne_binary()) -> ne_binary().                                           
alpha_to_dialpad(Value) ->                                                    
    << <<(dialpad_digit(C))>>                                                                                 
       || <<C>> <= wh_util:to_lower_binary(Value),
          is_alpha(C)                                                                        
    >>.                                                                         

-spec is_alpha(char()) -> boolean().
is_alpha(Char) ->      
    Char =< $z andalso Char >= $a.
```

IMO, this more clearly shows the three stages and doesn't build intermediate representations. It also passes all the tests!

Thoughts?

applications/callflow/test/cf_util_test.erl
```diff
@@ -16,3 +16,8 @@ alpha_to_dialpad_test() ->
     ?assertEqual(<<"23456789">>, cf_util:alpha_to_dialpad(<<"behknqux">>)),
     ?assertEqual(<<"23456789">>, cf_util:alpha_to_dialpad(<<"BeHkNqUx">>)),
     ?assertEqual(<<"23456789">>, cf_util:alpha_to_dialpad(<<"1BeH@k(N$q-u+x=">>)).
+
+alpha_to_dialpad_nonascii_test() ->
+    ?assertEqual(<<"8378">>, cf_util:alpha_to_dialpad(<<"teÕst">>)),
+    ?assertEqual(<<"8378896">>, cf_util:alpha_to_dialpad(<<"testÕÀ«©§two">>)),
+    ?assertEqual(<<"">>, cf_util:alpha_to_dialpad(<<"Ì¸ûî">>)).
```
jamesaimonetti: I've been moving more tests to use the test generator/test object style:

```erlang
alpha_to_dialpad_nonascii_test_() ->
    [?_assertEqual(<<"8378">>, cf_util:alpha_to_dialpad(<<"teÕst">>))
     ,?_assertEqual(<<"8378896">>, cf_util:alpha_to_dialpad(<<"testÕÀ«©§two">>))
     ,?_assertEqual(<<"">>, cf_util:alpha_to_dialpad(<<"Ì¸ûî">>))
    ].
```

You can then wrap these in a tuple with a human-friendly description to better indicate what was tested:

```erlang
    ,{"Test that all accented characters are stripped"
      ,?_assertEqual(<<>>, cf_util:alpha_to_dialpad(<<"Ì¸ûî">>))
    }
```

Any, just a suggestion; not required.

applications/callflow/src/cf_util.erl
```diff
@@ -339,7 +339,16 @@ vm_count_by_owner(<<_/binary>> = AccountDb, <<_/binary>> = OwnerId) ->
...
+
+    Process = fun(Char, Acc) ->
+	case Char of
+	    [] -> Acc;
+	    _  -> [dialpad_digit(Char)|Acc]
+	end
+    end,
+
+    list_to_binary(lists:foldr(Process, [], Stripped)).
```
jamesaimonetti: Just a note here: when the function being used in a fold/map/etc isn't enclosing values from the outer scope, go ahead and use a named function instead. So the foldr should be `lists:foldr(fun convert_to_dialpad/2, [], Stripped)` instead of the anonymous function you created. Easier for the reader to get the intent of the fold too.

applications/callflow/src/cf_util.erl
mark2600: You are technically correct, the best kind of correct.  I have submitted an updated version.

PR #1203, opened 2015-08-13
KAZOO-4080: kill trailing whitespace

fenollp:
This should be very quick to review :)


jamesaimonetti:
@fenollp needs a rebase...wah wah

fenollp:
@jamesaimonetti But but but what about merging the actual tool updating Travis?

PR #1206, opened 2015-08-13
New features for cf_dynamic_cid

stormqloud:
Ok, first Erlang ever!

I'm sure i'm using the most simplistic language elements.

Feedback welcome on what needs to be changed in Erlang constructs etc.

This allows overwriting the caller id name and number to arbitrary values stored in couchdb..

tx


applications/callflow/doc/cf_dynamic_cid.md
```diff
@@ -0,0 +1,72 @@
...
+### Optional fields 
+
+**interdigit_timeout** - default 2000 ms
+
+Can only set the caller id number with this method.
+
+You dial the new caller id on the keypad when prompted.
+
+## List action mode
+### Manditory fields
```
jamesaimonetti: s/Manditory/Mandatory/g

applications/callflow/doc/cf_dynamic_cid.md
```diff
@@ -0,0 +1,72 @@
...
+**id** - cidlist, couchdb doc id of the document that contains the new caller
+  id name and number information.
+
+You can set the caller id number and the caller id name with this
+method.
+
+On a handset you dial `*2015149072508`
+
+It's hooked in as a part of the feature codes.  The callflow regex looks like this:
+`
```
jamesaimonetti: I think you need three tickies to start a code block

jamesaimonetti:
Quick glance, things look good. You'll want to rebase and update this branch though. I'll take a harder look once you've done that.

stormqloud:
The sup maint number addition looks ok?

tx

applications/callflow/doc/cf_dynamic_cid.md
```diff
@@ -0,0 +1,72 @@
...
+
+You can set the caller id number and the caller id name with this
+method.
+
+On a handset you dial `*2015149072508`
+
+It's hooked in as a part of the feature codes.  The callflow regex looks like this:
+```
+   "patterns": [
+       "^\\*2(|[0-9]{2,})$"
```
jamesaimonetti: Is this regex what you want? Seems there's a random `|` in there, which looks strange.

applications/callflow/src/module/cf_dynamic_cid.erl
```diff
@@ -2,13 +2,17 @@
 %%% @copyright (C) 2011-2014, 2600Hz INC
 %%% @doc
 %%% "data":{
+%%%   "action": "manual" | "list"
 %%%   "media_id":"id_of_media"
 %%%   // optional after this
 %%%   "interdigit_timeout":2000
+%%%   "id":"01fc6_change_me_294234d2324497" // the list referenced above is kept in this document
```
jamesaimonetti: We would use `{LIST_ID}` or similar as a placeholder for data instead of garbled text.

applications/callflow/src/module/cf_dynamic_cid.erl
```diff
@@ -43,9 +48,110 @@
...
+%% @end
+%%--------------------------------------------------------------------
+-spec handle_list(wh_json:object(), whapps_call:call()) -> 'ok'.
+handle_list(Data, Call) ->
+    CallerIdNumber = whapps_call:caller_id_number(Call),
+    lager:debug("callerid number before this module: ~s ", [CallerIdNumber]),
+    CaptureGroup = whapps_call:kvs_fetch('cf_capture_group', Call),
+    lager:debug("capture_group ~s ", [CaptureGroup]),
+
+    << C1:8, C2:8, Dest/binary >> = CaptureGroup,
```
jamesaimonetti: `<<CIDKey:2/binary, Dest/binary>> = CaptureGroup` should match.

However, I'm concerned that you've limited the ability for JSON keys to be defined arbitrarily. I wonder if the capture group could allow you to define two capture groups in the regex, one for the Key and one for the Destination. You'll want to look in `cf_util:test_callflow_patterns/3` for that.

However, for this initial PR, don't worry about it. Just update the docs (thanks for those!) to reflect that the CID Key *must* be 2 digits.

applications/callflow/src/module/cf_dynamic_cid.erl
```diff
@@ -43,9 +48,110 @@
...
+    CID_id = <<C1,C2>>,
+    lager:debug("cid_id ~p ", [CID_id]),    
+
+    New_CID_info = get_list_entry(Data, Call),
+
+    New_caller_id_number = wh_json:get_value(<<"number">>, New_CID_info),
```
jamesaimonetti: We use camelcase for variables; please update yours to reflect that (`New_caller_id_number` should be `NewCallerIDNumber`, for instance).

applications/callflow/src/module/cf_dynamic_cid.erl
```diff
@@ -43,9 +48,110 @@
...
+            _ = whapps_call_command:answer(Call),
+            _ = whapps_call_command:prompt(<<"cf-unauthorized_call">>, Call),
+            _ = whapps_call_command:queued_hangup(Call),
+            'ok';
+        'false' ->
+            cf_exe:branch(wh_json:get_value(<<"flow">>, Flow), Call)
+    end.
+
+
+maybe_route_to_callflow(Data, Call, Number) ->
```
jamesaimonetti: Missing spec

jamesaimonetti:
Maintenance module additions look fine

stormqloud:
Ok I refactored it to handle the variable length from json.
tx

applications/callflow/src/module/cf_dynamic_cid.erl
```diff
@@ -43,9 +48,104 @@
...
+-spec maybe_route_to_callflow(wh_json:object(), whapps_call:call(), ne_binary()) -> 'ok'.
+maybe_route_to_callflow(Data, Call, Number) ->
+    case cf_util:lookup_callflow(Number, whapps_call:account_id(Call)) of
+        {'ok', Flow, NoMatch} ->
+            lager:info("callflow ~s satisfies request", [wh_json:get_value(<<"_id">>, Flow)]),
+            Updates = [{fun whapps_call:set_request/2
+                        ,list_to_binary([Number, "@", whapps_call:request_realm(Call)])
+                       }
+                       ,{fun whapps_call:set_to/2, list_to_binary([Number, "@", whapps_call:to_realm(Call)])}
+                       ,fun(C) when NoMatch ->
```
k-anderson: @stormqloud 
Howdy,
Looks like you want to crash if the callflow found is not a NoMatch.  However, if not flow is found then you play a message to the caller. 

May I suggest that you update line 127 to:
{'ok', Flow, 'true'}
and remove this fun?

That way both conditions that this function filters for result in the same user experience?

Additionally, I would request some return to cf_exe after the prompt play for "disa-invalid_extension".  For example:
cf_exe:stop(Call).

applications/callflow/src/module/cf_dynamic_cid.erl
```diff
@@ -104,7 +204,73 @@ handle(Data, Call) ->
...
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec should_restrict_call_by_account(whapps_call:call(), ne_binary()) -> boolean().
+should_restrict_call_by_account(Call, Number) ->
+    AccountId = whapps_call:account_id(Call),
+    AccountDb = whapps_call:account_db(Call),
+    case couch_mgr:open_cache_doc(AccountDb, AccountId) of
```
k-anderson: @stormqloud 
Why just by account?  Are you sure that you do not want to check the authorization based on the endpoint again?  For example, we allow the device or user to be restricted so if it is set at this level this module could  bypass that restriction if only the account is checked....

If this was not intentionally I would recommend using:
cf_endpoint:get(Call)
this will give you a JSON object that is the result of some merge logic so that it represents the settings for the device, user and account.  Getting the call restriction from this object will be identical to the restrictions enforced when calls are initially set up:
Here: https://github.com/2600hz/kazoo/blob/master/applications/callflow/src/cf_route_win.erl#L55
the merged object is fetched (a 'endpoint' object is the merge of settings from the dialing device, user if associated, and account).  This is then passed through a series of functions until it gets here: https://github.com/2600hz/kazoo/blob/master/applications/callflow/src/cf_route_win.erl#L100 as the JObj property.

applications/callflow/src/module/cf_dynamic_cid.erl
```diff
@@ -43,9 +48,104 @@
...
 %% @doc
+%% Entry point for this module, based on the payload will either
+%% connect a caller to check_voicemail or compose_voicemail.
+%% @end
+%%--------------------------------------------------------------------
+-spec handle(wh_json:object(), whapps_call:call()) -> 'ok'.
+handle(Data, Call) ->
+    case wh_json:get_value(<<"action">>, Data) of
+        <<"manual">> ->
+            lager:info("user must manually enter on keypad the caller id for this call"),
```
k-anderson: @stormqloud 
There are some minor alignment issues, if we resolve the other comments before you get to these we can fix the formatting once accepted.  However, notice this log line is indented by one space as is the lager line below.

There are also some alignment issues on line 93-94, 174-175, 181-184, 207, and in the block starting at 258.

Sorry if it seems picky but we try to hold all committers to the these same formatting standards :)

One other nit-picky note is we attempt to use all lowercase logging that is descriptive.  Typically we would not allow:
lager:debug("cid_id2 ~p ", [CIDKey]),
and request that the following line is lowercased:
lager:info("New caller id data : ~p",  [NewCallerId])

k-anderson:
The build error was my fault, it just happened that he rebased with a commit that broke master :(

https://github.com/2600hz/kazoo/commit/8dc3748e5b6a13b57df50e403c03dfb820a3ff5b

Sorry totally my fault

stormqloud:
Y sorry about the rebase in the pull.  I'm going to use branches in the future!

I'm totally new to Erlang, so your comments are appreciated.  I'm wandering in the forrest.

Can you give me a pointer(url) on the formatting you prefer?  I'm using emacs in Erlang mode.

I'll look at the other issues also of course.

tx

applications/callflow/src/module/cf_dynamic_cid.erl
stormqloud: Let me dig at this.  I copied this bit from "DISA" without really understanding the internals.
tx

k-anderson:
@stormqloud I was apologizing your rebase was fine but include one of my commits that broke master :(

However, branches would be the better way in the future.

Regarding the formatting we rely on emacs mode as well.  I suspect that maybe some of the above lines changed miss-aligning the lower :)  If you just cursor to the indicated and hit tab on each line it should rectify!  Like I mentioned this is really minor and we can help clean that up.

If we can resolve the restriction discrepancy then I will gladly accept the PR!

k-anderson:
This PR will also fulfill https://2600hz.atlassian.net/browse/KAZOO-4063 (please close when merged)

stormqloud:
I'm not quite sure how the functionality overlaps(or not) with this one. 

https://github.com/2600hz/kazoo/blob/master/applications/callflow/src/module/cf_lookupcidname.erl

Is it generally preferred to keep things inside the same callflow module or make new ones all the time?

tx

k-anderson:
@stormqloud 
It is preferred to extended modules when possible.  This can be tricky because it needs to remain backward compatible.  As such there are a few modules that have overlapping functionality but preform the operation slightly differently.  This is going to happen when community members solve the same problem different ways, but imho it gives the admin/user options.  It is a balance, especially considering most these modules are in use when submitted :)

However, I think it is appropriate to separate these two as I can see both evolving in divergent ways.

IMHO:
cf_lookupcidname is for inbound calls.  It will find and associate a name from a list (or in the future hopefully the contacts of an account/user).  I think in the future it should also be extended to preform CNAM dips (for the US) if configured with a URL.

This module is used to set the outbound caller id from pre-configured valid options.

stormqloud:
Is this ok to go in or there other changes I should make.

tx

jamesaimonetti:
@stormqloud Thanks for working through this PR with us. I've made a handful of changes as part of the merge, almost all related to formatting or specs. Take a look at the two commits following yours in master now and let me know if you have any questions about what I did or why. Thanks for the PR!

stormqloud:
@jamesaimonetti

Thanks for taking my patch, being patient and nice about my code.

See ya at Kazoocon!

PR #1212, opened 2015-08-17
KAZOO-615: add a default system_config parameter for this setting, no…

mark2600:
…w it will default to true

### DIFF
--- a/applications/callflow/src/module/cf_disa.erl
+++ b/applications/callflow/src/module/cf_disa.erl
@@ -18,6 +18,9 @@
 
 -export([handle/2]).
 
+-define(DEFAULT_USE_ACCOUNT_CALLER_ID, whapps_config:get(?CF_CONFIG_CAT, <<"refault_use_account_caller_id">>, 'true')).
+-define(DEFAULT_PIN_LENGTH, whapps_config:get_integer(?CF_CONFIG_CAT, <<"default_pin_length">>, 10)).
+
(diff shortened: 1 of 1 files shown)


applications/callflow/src/module/cf_disa.erl
```diff
@@ -147,8 +147,9 @@ maybe_restrict_call(Data, Call, Number, Flow) ->
 
 -spec maybe_update_caller_id(wh_json:object(), whapps_call:call()) -> whapps_call:call().
 maybe_update_caller_id(Data, Call) ->
-    case wh_json:is_true(<<"use_account_caller_id">>, Data, 'false') of
-        'true' -> set_caller_id(Call);
+    DefaultCidFlag = whapps_config:get(?CF_CONFIG_CAT, <<"default_use_account_caller_id">>, 'true'),
+    case wh_json:is_true(<<"use_account_caller_id">>, Data, DefaultCidFlag) of
```
lazedo: i'm ok with this but, i would create a macro at the top and use it.
-define(DEFAULT_USE_ACCOUNT_CALLER_ID, whapps_config:get(?CF_CONFIG_CAT, <<"default_use_account_caller_id">>, 'true')).

and then..
 case wh_json:is_true(<<"use_account_caller_id">>, Data, ?DEFAULT_USE_ACCOUNT_CALLER_ID) of

PR #1217, opened 2015-08-20
KAZOO-4065: Move apps_store to its own doc and add blacklist

macpie:


applications/crossbar/src/cb_apps_maintenance.erl
```diff
@@ -0,0 +1,53 @@
...
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec save(ne_binary(), wh_json:object(), wh_json:object()) -> {'ok', wh_json:object()} | {'error', any()}.
+save(Account, Doc, AccountDoc) ->
+    AccountDb = wh_util:format_account_id(Account, 'encoded'),
+    case couch_mgr:save_doc(AccountDb, Doc) of
+        {'error', _R}=Error -> Error;
+        {'ok', _}=Ok ->
+            case couch_mgr:ensure_saved(AccountDb, wh_json:delete_key(<<"apps">>, AccountDoc)) of
```
lazedo: why `couch_mgr:ensure_saved` ? why not use the result of the saved doc ?

applications/crossbar/src/cb_apps_maintenance.erl
```diff
@@ -0,0 +1,53 @@
...
+                {'error', _R} ->
+                    lager:error("failed to save ~s : ~p", [AccountDb, _R]),
+                    Ok;
+                {'ok', JObj} ->
+                    _ = crossbar_util:replicate_account_definition(JObj),
+                    Ok
```
lazedo: you're calling `crossbar_util:replicate_account_definition` with the result of removing `apps`object but returning with the `apps` object ?

PR #1221, opened 2015-08-24
KAZOO-4111: add Diversions header 

siplabs:


applications/callflow/src/cf_endpoint.erl
```diff
@@ -1197,9 +1197,26 @@ generate_sip_headers(Endpoint, Acc, Call) ->
...
                   ,fun(J) -> maybe_add_aor(J, Endpoint, Call) end
+                  ,fun(J) -> maybe_add_diversion(J, Call, Inception, Endpoint) end
                  ],
     lists:foldr(fun(F, JObj) -> F(JObj) end, Acc, HeaderFuns).
 
+-spec maybe_add_diversion(wh_json:object(), whapps_call:call(), api_binary(), wh_json:object()) -> wh_json:object().
+maybe_add_diversion(JObj, _Call, 'undefined', _Endpoint) ->
+    JObj;
+maybe_add_diversion(JObj, Call, Inception, Endpoint) ->
+    ShouldAddDiversion = whapps_call:authorizing_id(Call) =:= 'undefined'
```
jamesaimonetti: Can we add a check to `whapps_config:get_is_true(?CONFIG_CAT, <<"should_add_call_forward_diversion">>, 'false')` to preserve existing functionality?

applications/callflow/src/cf_endpoint.erl
```diff
@@ -1197,9 +1197,26 @@ generate_sip_headers(Endpoint, Acc, Call) ->
     HeaderFuns = [fun maybe_add_sip_headers/1
                   ,fun(J) -> maybe_add_alert_info(J, Endpoint, Inception) end
                   ,fun(J) -> maybe_add_aor(J, Endpoint, Call) end
+                  ,fun(J) -> maybe_add_diversion(J, Call, Inception, Endpoint) end
```
jamesaimonetti: My sensibilities are screaming about arg order here! :)

Could we do `maybe_add_diversion(J, Endpoint, Inception, Call)` since `Endpoint` is typically the second arg in these funs?

siplabs:
@jamesaimonetti Done!

k-anderson:
@lazedo @siplabs 

I am not sure I understand the use-case either.

Authorizing id as undefined would indicate that the channel did not have a SIP auth challenge (otherwise the authorizing_id is picked up as parameters in the authn response).  So this would seem to indicate it was allowed by the trusted ACLs, typically reserved for carriers (resource documents).  Channels of this type should only be able to call to numbers in the number dbs (such as +14158867900).  These unauthenticated channels are relayed by stepswitch_inbound where the inception is set if the number is the number dbs .  Therefore, if I am understanding this correctly with an authorizing_id and inception the channel is an unauthenticated call that is dialing a number stepswitch_inbound did not find in the number dbs.  How did this call get set up?  I assume that you have custom module authenticating this?  The only other path that could have set this up would be if the IP was in the trusted ACLs for a sip_auth document, but event then registrar would have had to set a authoriziing_id (and corresponding account_id) for it to be processed by callflows....

The concept makes sense, if the original caller id is kept then add a diversion header tracking that the call was forwarded by some number.  I am just caught up on how you have calls with no inception or authorizing_id... that shouldn't exist ;)

I would also propose that the diversion header should be added to any call forward to an offnet number regardless of origination source.  For example, I propose in cf_endpoint when a call forward is built the diversions are ALWAYS updated with the request URI from whapps_call:call().  I also propose using ';reason=unconditional' in accordance with the RFC.  I assume this would also meet your use-case?  It would be closer to the RFC as well.  One downside would be the diversion header would be present on call forwarding within the account, but I feel that is acceptable.

One minor note, the logging "Add diversion as ~s" by convention should be lowercase ;)

jamesaimonetti:
@siplabs Any response/comments to @k-anderson's `cf_endpoint` modifications?

siplabs:
@jamesaimonetti 

updated.

PR #1222, opened 2015-08-24
KAZOO-4110: sup replicate!

siplabs:


stormqloud:
Curious, is this "clone tool" migrating internally?
tx

siplabs:
We created this "clone tool" to migrate/backup clusters to other clusters. Works like a charm.

stormqloud:
<<"https://github.com/2600hz/community-scripts/tree/master/CloneTools">> =:= <<"replicate">>
?

k-anderson:
@stormqloud howdy, no they preform slightly different operations.  

The clone tool recreates the documents by using HTTP request to get a chunk of documents (possible manipulate them) then push as new documents to the target db.  This allows the documents to be cleaned up, selected from within the database based on rules (such as filtering old voicemail messages) and creates it as a fresh document (no rev history and re-sharded).  It is also only in one direction so changes after the clone or on the target will not be re-transferred (sometimes preferable).

The replicate command is built into couchdb/bigcouch and will do a two-way replication of all changes in the database.  If the document is changed on either side the changes will be resolved via the couchdb/bigcouch algorithms.  Since it does not process each document it can be much faster.  Potentially you can get a conflict that requires manually intervention, but in practice this very rarely happens.

To summarize the replicate is more like a temporary master-to-master database while the clone tool is a slave-to-master copy.  Both have uses ;)

deps/couchbeam-0.8.0/src/couchbeam.erl
```diff
@@ -173,11 +173,14 @@ replicate(#server{options=IbrowseOpts}=Server, RepObj) ->
     Headers = [{"Content-Type", "application/json"}],
     JsonObj = ejson:encode(RepObj),
 
-    case couchbeam_httpc:request_stream({self(), once}, post, Url, IbrowseOpts, Headers,
-            JsonObj) of
-        {ok, ReqId} ->
-            couchbeam_changes:wait_for_change(ReqId);
-        {error, Error} -> {error, Error}
+%%   lookat couchbeam/a7ea218ef147af
+     case couchbeam_httpc:request(post, Url, ["200", "201"], IbrowseOpts,
```
k-anderson: @siplabs 
Thoughts on updating couchbeam rather than patching?  We try to not patch deps if at all possible.

core/whistle_couch-1.0.0/src/couch_replicator.erl
```diff
@@ -50,16 +66,15 @@ repl_db(Source, Target, DB) ->
...
+            {'error', {500, _}} ->
+                {'error', 'failed'}
         end
     catch
         _E:_R ->
-            lager:debug("Exception occurred, probably a gateway timeout"),
-            lager:debug("~p", [_E]),
-            lager:debug("~p", [_R]),
-            {error, failed}
+            lager:error("Exception occurred: ~p:~p", [_E, _R]),
```
k-anderson: @siplabs 
Really minor request, but since I want to discuss the patch to coucbeam, if you are going to change this line could you use a lowercase log line?   When this module was written we didn't have a standard logging policy but by convention we normalized on lowercase.  This has the result of making uppercase logging stick out like a sore thumb... might as well fix it while you are here :)

k-anderson:
@siplabs 
Other than possibly updating couchbeam rather than patching, looks good to me.  Let me know!

jamesaimonetti:
@siplabs waiting to hear about couchbeam patching...

siplabs:
@k-anderson @jamesaimonetti it was broken in couchbeam. We cherry-picked commit https://github.com/benoitc/couchbeam/commit/c9caaffe1fcdf1841ea665534d2f697afb4a9b36
We decided to keep other in current state cause other commits may break compatibility or bring new bugs.

PR #1229, opened 2015-08-25
KAZOO-3949: enable formatters for outbound caller id name and number

mark2600:


applications/stepswitch/src/stepswitch_resources.erl
```diff
@@ -541,8 +541,10 @@ gateway_to_endpoint(Number
          ,{<<"Endpoint-Options">>, EndpointOptions}
          ,{<<"Endpoint-Progress-Timeout">>, wh_util:to_binary(ProgressTimeout)}
          ,{<<"Custom-Channel-Vars">>, wh_json:from_list(CCVs)}
-         | maybe_get_t38(Gateway, JObj)
-        ])).
+        ]
+        ++ maybe_get_t38(Gateway, JObj)
+        ++ wh_json:copy(<<"Outbound-Caller-ID-Number">>, JObj)
+        ++ wh_json:copy(<<"Outbound-Caller-ID-Name">>, JObj))).
```
lazedo: i would extend wh_json:to_proplist/2 to also accept keys() and use it here

applications/stepswitch/src/stepswitch_bridge.erl
```diff
@@ -319,6 +319,17 @@ build_bridge(#state{endpoints=Endpoints
...
+    EndpointFilter = fun(Element) ->
+        case Element of
+            {<<"Outbound-Caller-ID-Number">>, Number} -> false;
+            {<<"Outbound-Caller-ID-Name">>, Name}     -> false;
+            _OtherValues                              -> true
+        end
+    end,
+
+    FmtEndpoints = format_endpoints(Endpoints, Number, JObj, AccountId, EndpointFilter),
+
```
lazedo: try not create these functions inline. create a function (remove_outbound_caller_id ?) and use it.
it also seems you could just drop these keys with delete_keys ?

core/whistle-1.0.0/src/wh_json.erl
```diff
@@ -582,6 +583,13 @@ get_first_defined([H|T], JObj, Default) ->
         V -> V
     end.
 
+-spec copy(key(), object() | objects()) -> list(object()) | list().
+copy(Key, JObj) ->
+    case get_value(Key, JObj, 'undefined') of
+        'undefined' -> [];
+        Value       -> [{Key, Value}]
+    end.
+
```
lazedo: same as above. why not extend to_proplist/2 to accept keys() ?

applications/stepswitch/src/stepswitch_bridge.erl
mark2600: When trying to move this to a separate function I realized why I used a fun here also, because we need to match against Name and Number, which are in the scope of that function.  Using a fun means that we do not have to pass these parameters all the way through to the filter function.

Given this I no longer agree with moving the fun to a function in this specific case.

lazedo: but this only happens because you are creating the fun here and applying only in line 378 right ?
if you create the name function, then in line 378 you can pass the Name, Number. or am i missing something ?

lazedo:
failing to see the formatters applied to outbound_caller_id_number. what am i missing ?

applications/stepswitch/src/stepswitch_bridge.erl
mark2600: Yes but then you have to pass those values around, and if we were to add more filters in the future, the values used there would then also have to be passed.  As opposed to just passing the fun.

mark2600:
They are applied with the rest of the formatters, the reason they were not previously is because the values were not passed, that is fixed by the wh_json:copy 's

applications/stepswitch/src/stepswitch_resources.erl
lazedo: or you could just set
{<<"Outbound-Caller-ID-Name">>, wh_json:get_value(<<"Outbound-Caller-ID-Name">>, JObj}
within the proplist above

lazedo:
yup, just saw how they get applied, amazing work done by @jamesaimonetti on the formatters!

fenollp:
@mark2600 Travis should be fixed after a rebase against master

PR #1211, opened 2015-08-17
KAZOO-1243: correctly pass outbound flags and filter ratedeck's based…

mark2600:
… on them as well as options

PR #1232, opened 2015-08-26
KAZOO-1243: j5_authz_req is not passing device Options on rate requests

lazedo:
PR https://github.com/2600hz/kazoo/pull/1211 broke inbound.

### DIFF
--- a/applications/stepswitch/src/stepswitch_inbound.erl
+++ b/applications/stepswitch/src/stepswitch_inbound.erl
@@ -104,7 +104,7 @@ maybe_find_resource(_NumberProps, JObj) ->
     case stepswitch_resources:reverse_lookup(JObj) of
         {'error', 'not_found'} -> JObj;
         {'ok', ResourceProps} ->
-            Routines = [fun maybe_add_resource_id/2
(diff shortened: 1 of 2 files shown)


applications/stepswitch/src/stepswitch_inbound.erl
```diff
@@ -115,11 +115,22 @@ maybe_find_resource(_NumberProps, JObj) ->
 
 -spec maybe_add_resource_id(wh_json:object(), wh_proplist()) -> wh_json:object().
 maybe_add_resource_id(JObj, ResourceProps) ->
+    ResourceId = props:get_value('resource_id', ResourceProps),
     case props:get_is_true('global', ResourceProps) of
-        'false' -> JObj;
+        'false' ->
```
k-anderson: @lazedo this seems like a copy paste error?  If the type is not a global resource the CCV Global-Resource is set to 'true'?  Also be very careful I think some older code assumes that if there is an authorizing-id then it was on-net as well as if there is a Resource ID then it is global....

I would prefer not modify the JObj if not global unless there is a strong need to do so.

PR #1233, opened 2015-08-26
KAZOO-4089: fix quickcall for users

fenollp:


applications/crossbar/src/modules/cb_modules_util.erl
```diff
@@ -288,7 +289,7 @@ originate_quickcall(Endpoints, Call, Context) ->
     Request = [{<<"Application-Name">>, <<"transfer">>}
                ,{<<"Application-Data">>, get_application_data(Context)}
                ,{<<"Msg-ID">>, MsgId}
-               ,{<<"Endpoints">>, maybe_auto_answer(Endpoints, AutoAnswer)}
+               ,{<<"Endpoints">>, set_outbound_call_id(CallId, maybe_auto_answer(Endpoints, AutoAnswer))}
```
jamesaimonetti: Maybe you could refactor this to not iterate over the list of endpoints twice? Convert it to a fold and run the `set_value` calls as functions to the fold?

applications/crossbar/src/modules/cb_modules_util.erl
```diff
@@ -327,11 +331,13 @@ get_application_data_from_nouns(_Nouns) ->
 
 -spec get_timeout(cb_context:context()) -> pos_integer().
 get_timeout(Context) ->
-    try wh_util:to_integer(cb_context:req_value(Context, <<"timeout">>, 30)) of
-        Timeout when Timeout > 3 -> Timeout;
-        _ -> 30
+    Default = 30,
```
jamesaimonetti: A macro may be more descriptive `?DEFAULT_TIMEOUT_S` to indicate the unit of time is seconds here.

core/kazoo_documents-1.0.0/src/kz_account.erl
```diff
@@ -156,12 +156,13 @@ is_superduper_admin(JObj) ->
 allow_number_additions(JObj) ->
     wh_json:is_true(?ALLOW_NUMBER_ADDITIONS, JObj).
 
--spec fetch(ne_binary()) -> {'ok', doc()} |
+-spec fetch(api_binary()) -> {'ok', doc()} |
                             {'error', _}.
+fetch('undefined') ->
```
jamesaimonetti: Seems like defensive code, non?

jamesaimonetti:
Hope you didn't manually sort those lists of headers :) Thanks for doing it

core/kazoo_documents-1.0.0/src/kz_account.erl
fenollp: It's actually to reflect behaviour that used to exist before. I mean the switch from open_cache_doc to fetch removed this "dont crash on undefined, but return error" behaviour.

fenollp:
@jamesaimonetti `M-x sort-lines`

core/kazoo_documents-1.0.0/src/kz_account.erl
fenollp: @k-anderson You said to backport this part, right?

fenollp: @k-anderson ping

PR #1237, opened 2015-08-27
konami_move: If owner_id not set use device owner

Dinkor:


applications/konami/src/module/konami_move.erl
```diff
@@ -93,13 +93,14 @@ get_originate_req(Data, Call) ->
...
                                 ], Data),
 
-    OwnerId = wh_json:get_value(<<"owner_id">>, Data),
     {SourceDeviceId, TargetCallId} =
         case source_leg_of_dtmf(SourceOfDTMF, Call) of
             'a' -> {whapps_call:authorizing_id(Call), whapps_call:other_leg_call_id(Call)};
             'b' -> {find_device_id_for_leg(SourceOfDTMF), whapps_call:call_id(Call)}
         end,
 
+    DeviceOwnerId = cf_attributes:owner_ids(SourceDeviceId, Call),
```
lazedo: @Dinkor cf_attributes:owner_ids belongs to callflow whapp. you are breaking the whapp abstraction here (konami may run in one server and callflow in another). can you find another way to do it ?

Dinkor: cf_attributes:owner_ids functions is pretty simple
```Erlang
owner_ids('undefined', _Call) -> [];
owner_ids(ObjectId, Call) ->
    AccountDb = whapps_call:account_db(Call),
    ViewOptions = [{'key', ObjectId}],
    case couch_mgr:get_results(AccountDb, <<"cf_attributes/owner">>, ViewOptions) of
        {'ok', []} -> [];
        {'ok', [JObj]} -> [wh_json:get_value(<<"value">>, JObj)];
        {'ok', [_|_]=JObjs} ->
            [wh_json:get_value(<<"value">>, JObj) || JObj <- JObjs];
        {'error', _R} ->
            lager:warning("unable to find owner for ~s: ~p", [ObjectId, _R]),
            []
    end.
```
As I see it dont call any callflow functions...
Another way its copy this function to konami_util module, i think...

BTW, in konamy_move:build_endpoints also used cf_attributes:owned_by

jamesaimonetti: I think this type of code it going to be moved into the forthcoming `kazoo_objects` core lib, since it is being shared among whapps more frequently now. For now, I'd say leave it as-is so when we move the functionality to core, it is easy to identify what components are using it.

Dinkor:
Just realise that i have to use cf_attributes:owner_id instead of cf_attributes:owner_ids

jamesaimonetti:
Thanks!

PR #1238, opened 2015-08-27
README.md: update list of build dependencies

fenollp:


jamesaimonetti:
@fenollp 17 builds Kazoo fine? Don't some of the specs need fixing (like where dict(), set(), etc are used)?

fenollp:
@jamesaimonetti Hum lemme try this

fenollp:
@jamesaimonetti Aye you were right. (And I'm the one that created [this branch](https://github.com/2600hz/kazoo/tree/fix-deprecated-specs))

PR #1239, opened 2015-08-27
KAZOO-4135: compile Kazoo on 17.*

fenollp:


.travis.yml
```diff
@@ -8,6 +8,13 @@ notifications:
       - "irc.freenode.org#2600hz-dev"
 
 otp_release:
+  - 17.5
+  - 17.4
+  - 17.3
+  - 17.1
+  - 17.0
+  - R16B02
+  - R16B01
```
lazedo: why are we adding all these ?
i would leave one per major release , latest of 15, 16, 17

fenollp: @lazedo It's for testing actually. Look at this build for example: https://travis-ci.org/2600hz/kazoo/builds/77743535

And it doesn't cost anything to test all versions…

fenollp:
@jamesaimonetti Could you take a look at why some PropEr tests fail sometimes?

fenollp:
@k-anderson @jamesaimonetti @lazedo Look! It compiles! It passes tests!

PR #1240, opened 2015-08-28
KAZOO-4026: set resource_db in number doc

fenollp:


k-anderson:
@fenollp Lets talk more about this before accepting.  I don't think that this is accurate enough.  For example, what would it do when there is no assigned account? Available numbers are not assigned to accounts, the fact that this is present is due to how that one API works.  Additionally, there is no relationship with between the assigned account and the resource again just a factor of the API...  We need to find a means to ensure that if the resource_db is set it is indeed were the resource document exists.

jamesaimonetti:
@fenollp please review @k-anderson's comments and adjust accordingly.

fenollp:
@jamesaimonetti Couldn't find a single place in the code where `assign_to` would not be a binary!

Agreed specs get from `ne_binary()` to `api_binary()` for some reason at some point but who reads them? (hehe)

jamesaimonetti:
@fenollp still have not addressed the concerns of @k-anderson. Y U no respond?

PR #1241, opened 2015-08-29
Change code to return c2c request which the same to quickcall

emnvn:
This change work for https://2600hz.atlassian.net/browse/KAZOO-4099


deps/gen_smtp-8f7ba01172e679e8b4262334cefc3451b4c36777/testdata/dkim-rsa-private-encrypted:12345.pem
```diff
@@ -1,18 +0,0 @@
------BEGIN RSA PRIVATE KEY-----
```
jamesaimonetti: Why is this deleted file in the PR?

applications/crossbar/src/modules/cb_clicktocall.erl
```diff
@@ -447,6 +449,7 @@ build_originate_req(Contact, Context) ->
        ,{<<"Outbound-Callee-ID-Number">>, CalleeNumber}
        ,{<<"Outbound-Caller-ID-Name">>, FriendlyName}
        ,{<<"Outbound-Caller-ID-Number">>, OutboundNumber}
+	   ,{<<"Outbound-Call-ID">>, CallId}
```
k-anderson: @emnvn 
Howdy :)

Setting the Outbound-Call-ID on the root of the originate request will cause it to fail if there are multiple endpoints (they will all attempt to use this call-id).  Instead a unique call id should be generate and set (same property name) on each of the 'Endpoints'.

fenollp: @emnvn As @k-anderson says, have a look at cb_modules_util in #1233. It sets a per-endpoint outbound call id

k-anderson:
@emnvn 
Anything we can do to help with this?

fenollp:
@emnvn CI is a bit old; please consider rebasing.
Also, build fails because there are trailing spaces in `applications/crossbar/src/modules/cb_clicktocall.erl`

deps/gen_smtp-8f7ba01172e679e8b4262334cefc3451b4c36777/testdata/dkim-rsa-private-encrypted:12345.pem
emnvn: Oh, I did not delete this file, but I don't know why it was deleted when I create PR.

jamesaimonetti:
@emnvn one more note; please respect formatting (use spaces, not tabs), so that things line up properly.

lazedo:
@emnvn please rebase and address the formatting comment.

applications/crossbar/src/modules/cb_clicktocall.erl
lazedo: @fenollp @k-anderson looking at https://github.com/2600hz/kazoo/pull/1233 it seems that it sets the outbound-call-id per endpoint but with the same call-id for all endpoints

lazedo:
closed in favor of https://github.com/2600hz/kazoo/pull/1416

PR #1243, opened 2015-09-01
KAZOO-4143: system dial plans

siplabs:


applications/callflow/src/cf_util.erl
```diff
@@ -733,6 +741,26 @@ maybe_apply_dialplan([Regex|Regexs], DialPlan, Number) ->
             <<Prefix/binary, Root/binary, Suffix/binary>>
     end.
 
+-spec load_system_dialplans(ne_binaries()) -> wh_json:object().
+load_system_dialplans(Names) ->
+    LowerNames = [wh_util:to_lower_binary(Name) || Name <- Names],
+    case whapps_config:get_all_kvs(<<"dialplans">>) of
+        Plans when is_list(Plans) ->
+            lists:foldl(
+              fun({Key, Val}, Acc) ->
```
fenollp: Move this `fun` into its own private named function, with a name starting with `fold_`

applications/crossbar/src/modules/cb_dialplans.erl
```diff
@@ -0,0 +1,68 @@
...
+%% @public
+%% @doc
+%%
+%% @end
+%%--------------------------------------------------------------------
+-spec validate(cb_context:context()) -> cb_context:context().
+validate(Context) ->
+    Doc = wh_json:from_list(whapps_config:get_all_kvs(<<"dialplans">>)),
+    cb_context:setters(Context, [{fun cb_context:set_resp_data/2, Doc}
+                                 ,{fun cb_context:set_resp_status/2, 'success'}]).
```
fenollp: Put `]).` on its own line please ( @jamesaimonetti <3 )

siplabs:
@fenollp 
updated

fenollp:
Thanks @siplabs. I wonder why CI isn't running this, now…
Trying something…

fenollp:
Yayfixedit

siplabs:
@fenollp, I realized that `dilaplans` schema may be improved. I want to try it.

siplabs:
@fenollp updated `dialplans` schema.

siplabs:
ready to commit

applications/callflow/src/cf_util.erl
```diff
@@ -6,6 +6,8 @@
 %%% @contributors
 %%%   Karl Anderson
 %%%   James Aimonetti
+%%%   Sponsored by Conversant Ltd,
```
lazedo: shouldn't this mention the jira ticket ?

lazedo:
generally i like this feature but i do see some issues with it.
i believe the intention here is to apply dialplans to all accounts, but i see its usefulness somewhat limited specially for systems that have accounts from more than one country.
lets say that a system has accounts for germany, uk, france.
in those countries when people dial 0... , it usually means that they want to apply the country specific dialplan (replace 0 with country indicator) which is hard to apply with this implementation because there's no way to differentiate that.

maybe the implementation should be done like call restrictions, or the account/user/device dialplan should have some indicator/pointer to a system dialpan.

this system dialplan could also be of two types, simple regexe as is now, and also a group/list of other dialplans, in a way that a reference to the group dialplan would apply all the others in the list.

k-anderson:
@lazedo 

I believe the use-case is slightly different @siplabs correct me if I am wrong.  I think they want to create a set of dialplans once then re-use them per account.

As in, they would create a dialplan to correct for say Germany in a common place in the system.  For all German clients they would simply apply this common dialplan.

If that is the use-case then it makes sense to me ;)  Hopefully when we finish the libphonenumber integration all this manipulation will not be necessary!

applications/callflow/src/cf_util.erl
k-anderson: @lazedo preferably but not required ;)

lazedo:
rereading it, i realize that the dialplan of the account contains the links to the system dialplan.
my previous comment makes no sense, and since the scope of this is not grouping dialplans (which can be added later), i'm ok to accept this

PR #1244, opened 2015-09-01
KAZOO-4142: added callflow element for setting caller id name from lists

siplabs:


applications/callflow/src/module/cf_lookupcidname.erl
```diff
@@ -0,0 +1,120 @@
...
+                 end,
+    case CallerName of
+        'undefined' -> cf_exe:continue(Call);
+        CallerName ->
+            lager:info("setting caller name to ~p", [CallerName]),
+            cf_exe:continue(whapps_call:set_caller_id_name(CallerName, Call))
+    end.
+
+-spec match_prefix_in_lists(ne_binary(), ne_binary(), ne_binary() | [ne_binary()]) ->
+    {'stop', api_binary()} | 'continue'.
```
fenollp: Formatting

applications/callflow/src/module/cf_lookupcidname.erl
```diff
@@ -0,0 +1,120 @@
...
+match_prefix_in_lists(AccountDb, Number, Lists) ->
+    Prefixes = build_keys(Number),
+    Keys = [[ListId, Prefix] || ListId <- Lists, Prefix <- Prefixes],
```
fenollp: That's `length(Lists) * length(Prefixes)` combinations. Is that normal?

applications/callflow/src/module/cf_lookupcidname.erl
```diff
@@ -0,0 +1,120 @@
...
+compare_prefixes(JObj1, JObj2) ->
+    byte_size(get_prefix(JObj1)) >= byte_size(get_prefix(JObj2)).
+
+-spec get_prefix(wh_json:object()) -> binary().
+get_prefix(JObj) ->
+    Doc = wh_json:get_value(<<"doc">>, JObj),
+    case wh_json:get_value(<<"number">>, Doc) of
+        'undefined' -> wh_json:get_value(<<"prefix">>, Doc, <<>>);
+        Prefix      -> Prefix
+    end.
```
fenollp: Use `get_ne_binary_value` when appropriate.
Also the `case` could be replaced using `wh_json:get_value(<<"prefix">>, Doc, <<>>)` as the default (3rd argument) in the call.

applications/callflow/src/module/cf_lookupcidname.erl
siplabs: @fenollp 
Can you tell max line length to be clear?

fenollp: I only meant that line 41 is not properly indented. Not sure on max line length.
“not too big” seems to be the only metric…

siplabs:
@fenollp 
updated

applications/callflow/src/module/cf_lookupcidname.erl
siplabs: @fenollp, could you write 40-41 lines with correct indention?
Just want to know to correct my editor.

PR #1247, opened 2015-09-01
Track whether agent or member hung up queue call

danielfinke:


lazedo:
@danielfinke sorry.., master

PR #1251, opened 2015-09-01
Added option to define increment/minimum and call direction

danielfinke:


applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -315,7 +315,7 @@ upload_csv(Context) ->
     Now = erlang:now(),
     {'ok', {Count, Rates}} = process_upload_file(Context),
     lager:debug("trying to save ~b rates (took ~b ms to process)", [Count, wh_util:elapsed_ms(Now)]),
-    _  = crossbar_doc:save(cb_context:set_doc(Context, Rates), [{'publish_doc', 'false'}]),
+    lists:foreach(fun(Rate) -> crossbar_doc:ensure_saved(cb_context:set_doc(Context, Rate), [{'publish_doc', 'false'}]) end, Rates),
```
lazedo: why the use of ensure_saved ?

danielfinke: This is related to the dev posting at https://groups.google.com/forum/#!topic/2600hz-dev/RVa2jV_kTX8 in which Lucas said that crossbar_doc:save would not update existing rates, but crossbar_doc:ensure_saved would.

lazedo: and did you see @jamesaimonetti comment ? i really hate to use `ensure_saved` unless there's a good reason for that. one example is saving account doc in account db and accounts db, ie, saving the same document in two different locations.

applications/crossbar/src/modules/cb_rates.erl
```diff
@@ -479,6 +483,24 @@ get_row_rate([_|_]=_R) ->
...
+    60.
+
+-spec get_row_increment(rate_row()) -> api_integer().
+get_row_increment([_, _, _, _, _, _, _, _, Increment | _]) -> wh_util:to_integer(Increment);
+get_row_increment([_|_]=_R) ->
+    lager:info("increment not found on row: ~p", [_R]),
+    60.
+
+-spec get_row_direction(rate_row()) -> api_integer().
+get_row_direction([_, _, _, _, _, _, _, _, _, Direction | _]) -> [wh_util:to_binary(Direction)];
```
fenollp: Hey that's not an integer :)

lazedo:
@danielfinke master please

PR #1252, opened 2015-09-01
No longer sets language again during pivot callflow exec

danielfinke:
If the call language has been changed from the default defined for the calling endpoint, and a pivot script executes, the resulting callflow will overwrite any "Set Language" modules that were called in the previous one. This patch ensures that they are not overwritten.

### DIFF
--- a/applications/callflow/src/cf_route_win.erl
+++ b/applications/callflow/src/cf_route_win.erl
@@ -214,14 +214,18 @@ store_owner_id(Call) ->
 -spec set_language(whapps_call:call()) -> whapps_call:call().
 set_language(Call) ->
     Default = wh_media_util:prompt_language(whapps_call:account_id(Call)),
-    case cf_endpoint:get(Call) of
-        {'ok', Endpoint} ->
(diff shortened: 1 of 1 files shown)


lazedo:
set_language is called by bootstrap_callflow_executer. what's the use case where this resolves the overwrite ?

danielfinke:
Here is the set of steps:
1. Start a callflow (sets the call's language to endpoint or default lang via bootstrap_callflow_executer)
2. Set the language to something else using cf_language module
3. Execute a pivot request (we are using kazoo translator) and return a valid callflow definition from the server
4. A route resume is published, cf_route_resume picks this up and re-runs maybe_restrict_call, which in turn re-runs bootstrap_callflow_executer
5. The language is once again set to endpoint or default lang

lazedo:
yes, i understand the path now. thank you.
maybe we should change how resume works because it still execs the other actions in bootstrap.
let me run some tests.

lazedo:
@danielfinke make pull request to master

PR #1254, opened 2015-09-01
Pivot sends pivot_failed if returned data cannot be translated

danielfinke:
Though a pivot HTTP request may succeed, the resulting data body may not be translated successfully. In such a case, a pivot_failed response should still be sent to the waiting cf_pivot module.

### DIFF
--- a/applications/pivot/src/pivot_call.erl
+++ b/applications/pivot/src/pivot_call.erl
@@ -214,7 +214,7 @@ handle_cast({'request', Uri, Method, Params}, #state{call=Call
                                     ,call=Call2
                                    }};
         _ ->
-            wapi_pivot:publish_failed(Q, [{<<"Call-ID">>,whapps_call:call_id(Call)}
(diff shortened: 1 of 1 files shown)


applications/pivot/src/pivot_call.erl
```diff
@@ -525,6 +527,9 @@ process_resp(Call, CT, RespBody) ->
             {'stop', Call};
         'throw':{'error', 'unrecognized_cmds'} ->
             lager:info("no translators recognize the supplied commands: ~s", [RespBody]),
+            wapi_pivot:publish_failed(RequesterQ, [{<<"Call-ID">>,whapps_call:call_id(Call)}
```
fenollp: Hey, sorry to be picky. (I have to be while @jamesaimonetti is away)
Could you add a space after your comma?

lazedo:
@danielfinke pull request to master please.

PR #1255, opened 2015-09-01
Added support to callflow modules and pivot to utilize custom KVS

danielfinke:
cf_kvs_set supports setting arbitrary key-value pairs on a call, in an isolated key in the kvs_store db. These values can be used in the modified cf_ modules to connect to a queue, play a media file, use a vmbox, etc, dynamically, based on a key's current value. This enables a fully dynamic callflow, without the need for pivot scripts. A common main callflow structure can be defined to use variables, and a small setup callflow can be created for each of a handful of numbers/extensions that initialize the variables. This avoids the need to rebuild many callflows if the structure changes.
Finally, kzt_kazoo adds in the kvs set using this module to pivot requests, if needed.

### DIFF
--- a/applications/callflow/src/module/cf_acdc_member.erl
+++ b/applications/callflow/src/module/cf_acdc_member.erl
@@ -36,7 +36,7 @@
 %%--------------------------------------------------------------------
 -spec handle(wh_json:object(), whapps_call:call()) -> 'ok'.
 handle(Data, Call) ->
-    QueueId = wh_json:get_value(<<"id">>, Data),
+    QueueId = maybe_use_variable(Data, Call),
     lager:info("sending call to queue ~s", [QueueId]),
 
     Priority = lookup_priority(Data, Call),
@@ -68,6 +68,19 @@ handle(Data, Call) ->
                                   }
                       ,is_queue_full(MaxQueueSize, CurrQueueSize)
                      ).
+
+-spec maybe_use_variable(wh_json:object(), whapps_call:call()) -> api_binary().
+maybe_use_variable(Data, Call) ->
+    case wh_json:get_value(<<"var">>, Data) of
+        'undefined' ->
+            wh_json:get_value(<<"id">>, Data);
+        Variable ->
+            Value = wh_json:get_value(<<"value">>, cf_kvs_set:get_kv(Variable, Call)),
+            case couch_mgr:open_cache_doc(whapps_call:account_db(Call), Value) of
+                {'ok', _} -> Value;
+                _ -> wh_json:get_value(<<"id">>, Data)
+            end
+    end.
 
 -spec lookup_priority(wh_json:object(), whapps_call:call()) -> api_binary().
 lookup_priority(Data, Call) ->
--- a/applications/callflow/src/module/cf_acdc_member.erl
+++ b/applications/callflow/src/module/cf_acdc_member.erl
@@ -36,7 +36,7 @@
 %%--------------------------------------------------------------------
 -spec handle(wh_json:object(), whapps_call:call()) -> 'ok'.
 handle(Data, Call) ->
-    QueueId = maybe_use_variable(Data, Call),
+    QueueId = cf_util:maybe_use_variable(Data, Call),
     lager:info("sending call to queue ~s", [QueueId]),
 
     Priority = lookup_priority(Data, Call),
@@ -68,19 +68,6 @@ handle(Data, Call) ->
                                   }
                       ,is_queue_full(MaxQueueSize, CurrQueueSize)
                      ).
-
--spec maybe_use_variable(wh_json:object(), whapps_call:call()) -> api_binary().
-maybe_use_variable(Data, Call) ->
-    case wh_json:get_value(<<"var">>, Data) of
-        'undefined' ->
-            wh_json:get_value(<<"id">>, Data);
-        Variable ->
-            Value = wh_json:get_value(<<"value">>, cf_kvs_set:get_kv(Variable, Call)),
-            case couch_mgr:open_cache_doc(whapps_call:account_db(Call), Value) of
-                {'ok', _} -> Value;
-                _ -> wh_json:get_value(<<"id">>, Data)
-            end
-    end.
 
 -spec lookup_priority(wh_json:object(), whapps_call:call()) -> api_binary().
 lookup_priority(Data, Call) ->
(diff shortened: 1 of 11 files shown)


applications/callflow/src/module/cf_acdc_member.erl
```diff
@@ -69,6 +69,19 @@ handle(Data, Call) ->
                       ,is_queue_full(MaxQueueSize, CurrQueueSize)
                      ).
 
+-spec maybe_use_variable(wh_json:object(), whapps_call:call()) -> api_binary().
+maybe_use_variable(Data, Call) ->
+    case wh_json:get_value(<<"var">>, Data) of
+        'undefined' ->
+            wh_json:get_value(<<"id">>, Data);
```
fenollp: Use `wh_doc:id/1`

applications/callflow/src/module/cf_acdc_member.erl
```diff
@@ -69,6 +69,19 @@ handle(Data, Call) ->
...
+        Variable ->
+            Value = wh_json:get_value(<<"value">>, cf_kvs_set:get_kv(Variable, Call)),
+            case couch_mgr:open_cache_doc(whapps_call:account_db(Call), Value) of
+                {'ok', _} -> Value;
+                _ -> wh_json:get_value(<<"id">>, Data)
```
fenollp: Use `wh_doc:id/1`

applications/callflow/src/module/cf_callflow.erl
```diff
@@ -39,6 +39,19 @@ maybe_branch_callflow(Data, Call) ->
             continue_if_still_active(Call, JObj)
     end.
 
+-spec maybe_use_variable(wh_json:object(), whapps_call:call()) -> api_binary().
+maybe_use_variable(Data, Call) ->
+    case wh_json:get_value(<<"var">>, Data) of
+        'undefined' ->
+            wh_json:get_value(<<"id">>, Data);
```
fenollp: Use `wh_doc:id/1`

applications/callflow/src/module/cf_callflow.erl
```diff
@@ -39,6 +39,19 @@ maybe_branch_callflow(Data, Call) ->
...
+        Variable ->
+            Value = wh_json:get_value(<<"value">>, cf_kvs_set:get_kv(Variable, Call)),
+            case couch_mgr:open_cache_doc(whapps_call:account_db(Call), Value) of
+                {'ok', _} -> Value;
+                _ -> wh_json:get_value(<<"id">>, Data)
```
fenollp: Use `wh_doc:id/1`

applications/callflow/src/module/cf_kvs_set.erl
```diff
@@ -0,0 +1,143 @@
...
+    cf_exe:continue(Call2).
+
+get_kv(Key, Call) ->
+    get_kv(?COLLECTION_KVS, Key, Call).
+
+get_kv(Collection, Key, Call) ->
+    wh_json:get_value(Key, get_collection(Collection, Call)).
+    
+set_kvs(Data, Call) ->
+    Call2 = set_kvs_mode(wh_json:get_value(<<"kvs_mode">>, Data, undefined), Call),
```
fenollp: Quote `undefined` please

applications/callflow/src/module/cf_device.erl
```diff
@@ -59,3 +59,16 @@ bridge_to_endpoints(Data, Call) ->
...
+    case wh_json:get_value(<<"var">>, Data) of
+        'undefined' ->
+            wh_json:get_value(<<"id">>, Data);
+        Variable ->
+            Value = wh_json:get_value(<<"value">>, cf_kvs_set:get_kv(Variable, Call)),
+            case couch_mgr:open_cache_doc(whapps_call:account_db(Call), Value) of
+                {'ok', _} -> Value;
+                _ -> wh_json:get_value(<<"id">>, Data)
+            end
+    end.
```
fenollp: Ok it's all the same code… Could you move it and use it from a util module?
Or maybe even the `cf_kvs_set` module?

applications/callflow/src/module/cf_kvs_set.erl
```diff
@@ -0,0 +1,143 @@
...
+    Data2 = wh_json:delete_key(<<"kvs_mode">>, Data),
+    Keys = wh_json:get_keys(Data2),
+    lists:foldl(fun(Key, Call3) ->
```
fenollp: Could you move the `fun` into a named private function? Like `fold_appropriate_name_here/`
This way if code breaks here we get a useful name instead of `#Fun<123123.123122.232>`

applications/callflow/src/module/cf_kvs_set.erl
```diff
@@ -0,0 +1,143 @@
...
+    
+kvs_to_props(Call) ->
+    case wh_json:get_value(<<"kvs_mode">>, get_collection(?COLLECTION_MODE, Call), undefined) of
+        <<"json">> ->
+            Collection = get_kvs_collection(Call),
+            Keys = wh_json:get_keys(Collection),
+            
+            lists:foldl(fun(Key, Props) ->
+                [{Key, list_to_binary(format_json(wh_json:get_value(Key, Collection)))}] ++ Props
+                end, [], Keys);
```
fenollp: Same 2 notes here (undefined & fun)

applications/callflow/src/module/cf_kvs_set.erl
```diff
@@ -0,0 +1,143 @@
...
+            [{<<"Custom-KVS">>, get_kvs_collection(Call)}]
+    end.
+    
+format_json(Data) ->
+    Proplist = case wh_json:is_json_object(Data) of
+        true ->
+            wh_json:to_proplist(Data);
+        _ ->
+            Data
+    end,
```
fenollp: Formatting is wrong

applications/callflow/src/module/cf_kvs_set.erl
```diff
@@ -0,0 +1,143 @@
...
+
+format_json_rec(V, _) ->
+    "\"" ++ wh_util:to_list(V) ++ "\"".
+    
+format_type(Data) when not is_binary(Data) ->
+    wh_util:to_list(Data);
+    
+format_type(<<Data/binary>>) ->
+    "\"" ++ binary_to_list(Data) ++ "\"".
+
```
fenollp: Hey, there's `wh_json:encode/1` :)

lazedo:
@danielfinke i like what you're tying to do here, however it seems a little restrictive. i would prefer another approach here.

your fetch for "var" element is somewhat static and it doesn't allow for 2 varables to be fetched from kv store.

if you had something like {<<"TheElementIWant">>, <<"var:kv">>} in the callflow, then we could translate the values before we pass them for execution in the module.

that would avoid changing the cf_modules to call maybe_use_variable and allow multiple dynamic elements.

thoughts ?

lazedo:
@danielfinke make pull request against master

PR #1250, opened 2015-09-01
Added find_account_by_id to crossbar_maintenance

danielfinke:

PR #1257, opened 2015-09-01
Added get_account_by_id to whapps_util

danielfinke:


core/whistle_apps-1.0.0/src/whapps_util.erl
```diff
@@ -284,6 +285,24 @@ is_account_db(Db) -> couch_util:db_classification(Db) =:= 'account'.
...
 %%--------------------------------------------------------------------
 %% @public
+%% @doc Ids are one->one with accounts.
+%% @end
+%%--------------------------------------------------------------------
+-spec get_account_by_id(ne_binary()) ->
+                                  {'ok', ne_binary()} |
+                                  {'error', 'not_found'}.
+get_account_by_id(Id) ->
+    case couch_mgr:open_doc(?WH_ACCOUNTS_DB, Id) of
```
lazedo: @danielfinke the name of the function is `get_account_by_id` but you're returning the `account db` ?

danielfinke: Take a look at the "do_get_accounts_by" function that both get_account_by_realm and get_account_by_name call. They both return the account dbs for the parameter.

lazedo: yep, but should they ? also, any reason for not using open_cache_doc ?

danielfinke: Good question. It appears that there are dependents of the existing 2 functions all over the place. No particular reason why open_cache_doc wasn't used.
However, I looked at the reason this PR was included was to actually help #1250. And as it turns out I should just be using wh_util:format_account_id to produce the account db binary string. I'm going to close this one

PR #1258, opened 2015-09-01
cf_tts no longer dies on channel_hungup

danielfinke:


danielfinke:
This resolves the issue in #1056 for this particular case. The point of #1056 was to take the onus off of the cf_ module developer for handling a channel_hungup. That PR is kind of a generic catch-all. A developer can simply call cf_exe:continue(Call) and it will only actually continue if the call is still up. Note that some modules might not have an event listening loop, which would prevent the developer from catching the channel_hungup

lazedo:
got it, that's why i'm not a big fan of #1056. there maybe additional actions not requiring that the call is still up.

danielfinke:
I believe it was discussed in the link in #1056 (https://groups.google.com/forum/#!topic/2600hz-dev/--DJFztGF_Q) that there was no reason one might still want callflows to continue after a call was down. Is there a particular case you can think of?

lazedo:
store/move related files (faxes/recordings), notifies, ..

danielfinke:
Those things aren't managed by the callflow whapps, though? The change to terminate cf_exe won't break/stop the fax/notify whapps.

lazedo:
they're not related to the whapps. one other example would be to start a callback, send a post to a http endpoint with some result. i see lots of thing we already do and others that can be done

lazedo:
@danielfinke please make pull request against master

k-anderson:
@danielfinke hey, sorry to do this but this has encouraged an internal discussion.  The result of which is we prefer #1056.  Turns out we should have listened in May and that was primarily my fault I misunderstood what you were telling me! :(

This makes this PR redundant so I will close it.  I owe you a beer.

PR #1259, opened 2015-09-01
KAZOO-3909: As a webhooks user, I would like a webhook for channel park events

macpie:


applications/webhooks/src/modules/webhooks_parking.erl
```diff
@@ -0,0 +1,91 @@
...
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec maybe_send_event(api_binary(), wh_json:object()) -> 'ok'.
+maybe_send_event('undefined', _JObj) -> 'ok';
+maybe_send_event(AccountId, JObj) ->
+    case webhooks_util:find_webhooks(?NAME, AccountId) of
+        [] -> lager:debug("no hooks to handle ~s for ~s", [AccountId]);
```
fenollp: I thought this kind of errors would show at compile time. Maybe the lager parse transform isn't applied everywhere

applications/callflow/src/module/cf_park.erl
```diff
@@ -197,6 +198,7 @@ park_call(SlotNumber, Slot, ParkedCalls, ReferredTo, Call) ->
         %% attended transfer and allowed to update the provided slot number, we are still connected to the 'parker'
         %% not the 'parkee'
         {'undefined', _} ->
+            _ = publish_parked(Call, SlotNumber),
```
lazedo: should we publish here ? call is not parked yet, its an attendant transfer

applications/callflow/src/module/cf_park.erl
```diff
@@ -116,6 +116,7 @@ retrieve(SlotNumber, ParkedCalls, Call) ->
             lager:info("the parking slot ~s currently has a parked call ~s, attempting to retrieve caller", [SlotNumber, ParkedCall]),
             case maybe_retrieve_slot(SlotNumber, Slot, ParkedCall, Call) of
                 'ok' ->
+                    _ = publish_retrieved(Call, SlotNumber),
```
lazedo: this should be ParkedCall

applications/callflow/src/module/cf_park.erl
k-anderson: @macpie 
Luis is correct, this is not yet parked.  There is a function to handle the completion of an attended transfer and swap the caller-id, that would likely be more appropriate for this.

macpie: @k-anderson what is the function?

applications/webhooks/src/webhooks_channel_util.erl
```diff
@@ -56,7 +56,12 @@ format_event(JObj, AccountId, <<"CHANNEL_DESTROY">>) ->
                       ,{<<"ringing_seconds">>, kz_call_event:ringing_seconds(JObj)}
                       ,{<<"billing_seconds">>, kz_call_event:billing_seconds(JObj)}
                      ]
-                   ).
+                   );
+format_event(JObj, AccountId, <<"CHANNEL_PARK">>) ->
```
k-anderson: @macpie 
Is this necessary?  I assume this was from an earlier attempt to use CHANNEL_PARK.  If it is you should remove the io:format ;)

k-anderson:
@macpie 
You need an event when when a parked caller hangs up (not answered but unparked).  Perhaps in the cf_park it might be better to emit a "slot_update" event?  

Thoughts?

applications/callflow/src/module/cf_park.erl
```diff
@@ -694,3 +696,41 @@ update_presence(State, Slot) ->
...
     wh_amqp_worker:cast(Command, fun wapi_presence:publish_update/1).
+
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec publish_parked(whapps_call:call(), ne_binary()) -> 'ok'.
+publish_parked(Call, Slot) ->
+    publish_event(Call, Slot, <<"PARKED">>).
```
macpie: prefix with park_

applications/callflow/src/module/cf_park.erl
```diff
@@ -694,3 +696,41 @@ update_presence(State, Slot) ->
...
+
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec publish_retrieved(whapps_call:call(), ne_binary()) -> 'ok'.
+publish_retrieved(Call, Slot) ->
+    publish_event(Call, Slot, <<"RETRIEVED">>).
```
macpie: prefix with park_

core/whistle-1.0.0/src/api/wapi_call.erl
```diff
@@ -97,6 +97,7 @@
                                       ,<<"Intercepted-By">>
                                       ,<<"Switch-Hostname">>, <<"Switch-Nodename">>
                                       ,<<"Switch-URL">>, <<"Switch-URI">>
+                                      ,<<"Slot">>, <<"Account-ID">>
```
macpie: Use ccv Account-ID instead

macpie:
ready

PR #1262, opened 2015-09-02
cf_tts no longer dies on channel_hungup

danielfinke:
#1258 against master


k-anderson:
@danielfinke hey, sorry to do this but this has encouraged an internal discussion. The result of which is we prefer #1056. Turns out we should have listened in May and that was primarily my fault I misunderstood what you were telling me! :(

This makes this PR redundant so I will close it. I owe you a beer.

danielfinke:
@k-anderson no harm done! I'll take you up on that beer some time!

PR #1256, opened 2015-09-01
Call restriction no longer missing when using ts_from_onnet

danielfinke:

PR #1263, opened 2015-09-02
Call restriction no longer missing when using ts_from_onnet

danielfinke:
#1256 against master

Conflicts:

	applications/trunkstore/src/ts_util.erl

### DIFF
--- a/applications/trunkstore/src/ts_util.erl
+++ b/applications/trunkstore/src/ts_util.erl
@@ -198,11 +198,8 @@ lookup_user_flags(Name, Realm, AccountId, _) ->
                     JObj = wh_json:set_value(<<"id">>, wh_doc:id(User), ValJObj),
 
                     {'ok', AccountJObj} = kz_account:fetch(AccountId),
-                    FlagsJObj = wh_json:from_list(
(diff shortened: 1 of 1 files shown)


applications/trunkstore/src/ts_util.erl
```diff
@@ -196,11 +196,18 @@ lookup_user_flags(Name, Realm, AccountId, _) ->
                     lager:info("cache miss, found view result for ~s@~s with id ~s", [Name, Realm, wh_doc:id(User)]),
                     ValJObj = wh_json:get_value(<<"value">>, User),
                     JObj = wh_json:set_value(<<"id">>, wh_doc:id(User), ValJObj),
+
+                    {'ok', AccountJObj} = kz_account:fetch(AccountId),
+                    FlagsJObj = wh_json:from_list(
```
k-anderson: @danielfinke 
This looks fine except I think it would be more efficient to:

...
Restrictions = wh_json:get_value(<<"call_restriction">>, AccountJObj, wh_json:new()),
FlagsJObj = wh_json:set_value(<<"call_restriction">>, Restrictions, JObj), 
...

PR #1264, opened 2015-09-02
No longer sets language again during cf_callflow or pivot cf exec

danielfinke:
#1252 against master

### DIFF
--- a/applications/callflow/src/cf_route_win.erl
+++ b/applications/callflow/src/cf_route_win.erl
@@ -237,14 +237,18 @@ store_owner_id(Call) ->
 -spec set_language(whapps_call:call()) -> whapps_call:call().
 set_language(Call) ->
     Default = wh_media_util:prompt_language(whapps_call:account_id(Call)),
-    case cf_endpoint:get(Call) of
-        {'ok', Endpoint} ->
-            Language = kz_device:language(Endpoint, Default),
-            lager:debug("setting language '~s' for this call", [Language]),
-            whapps_call:set_language(wh_util:to_lower_binary(Language), Call);
-        {'error', _E} ->
-            lager:debug("no source endpoint for this call, setting language to default ~s", [Default]),
-            whapps_call:set_language(Default, Call)
+    case whapps_call:language(Call) of
+        Default ->
+            case cf_endpoint:get(Call) of
+                {'ok', Endpoint} ->
+                    Language = kz_device:language(Endpoint, Default),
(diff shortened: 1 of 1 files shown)


applications/callflow/src/cf_route_win.erl
```diff
@@ -237,14 +237,18 @@ store_owner_id(Call) ->
...
+    case whapps_call:language(Call) of
+        Default ->
+            case cf_endpoint:get(Call) of
+                {'ok', Endpoint} ->
+                    Language = kz_device:language(Endpoint, Default),
+                    lager:debug("setting language '~s' for this call", [Language]),
+                    whapps_call:set_language(wh_util:to_lower_binary(Language), Call);
+                {'error', _E} ->
+                    lager:debug("no source endpoint for this call, setting language to default ~s", [Default]),
+                    whapps_call:set_language(Default, Call)
```
k-anderson: @danielfinke 
I don't think I understand this case clause.  If the default language is already set on the call and there is no source endpoint set the language of the call to the default (wouldn't the case clause ensure that already)?  

I suspect the intent is to only set the language if the language has not already been set but that doesn't seem like the safest means to make this determination.  It basically means if the language is set to something other than default it can not be changed during re-hunt, but if it is set to default (or just default) then it could be.  No?

Could you describe your use-case maybe it will clear things up or let me come up with some suggestions :)

lazedo: @k-anderson check #1252

k-anderson: @danielfinke 
Ah ok I understand now, thanks :)

SO I would prefer Luis's comment on the other PR:

"yes, i understand the path now. thank you.
maybe we should change how resume works because it still execs the other actions in bootstrap.
let me run some tests."

@lazedo you left it at "let me run some tests", let us know what you find :)

My other thought would be to ensure the language is not set unless explicitly provided and all functions that access this information fallback to the default if not present (perhaps even in whapps_call).  This way 'undefined' means 'not set use default', and any other value means 'this is what the user wants'.  If we did this then I would feel much better about ignoring the language if set in route_win.

jamesaimonetti:
@danielfinke where are we with this PR? Can you address @k-anderson's concerns in the last comment from him? Thanks

jamesaimonetti:
@danielfinke We're going to close this for now so we can start to get our PR queue down to those PRs being actively worked on. Please open a new PR (and reference this one!) when you have the time to address the last comments above from @k-anderson.

Thanks!

PR #1253, opened 2015-09-01
Added ability to request control queue given a call ID

danielfinke:

PR #1266, opened 2015-09-02
Added ability to request control queue given a call ID

danielfinke:
Continuation of #1253

### DIFF
--- a/applications/ecallmgr/src/ecallmgr.hrl
+++ b/applications/ecallmgr/src/ecallmgr.hrl
@@ -79,6 +79,7 @@
                   ,handling_locally = 'false' :: boolean() | '_' %% is this ecallmgr handling the call control?
                   ,to_tag :: api_binary() | '_'
                   ,from_tag :: api_binary() | '_'
+                  ,control_q :: api_binary() | '_'
                  }).
 
 -type channel() :: #channel{}.
--- a/applications/ecallmgr/src/ecallmgr_call_control.erl
+++ b/applications/ecallmgr/src/ecallmgr_call_control.erl
@@ -52,7 +52,6 @@
 -export([handle_call_command/2]).
 -export([handle_conference_command/2]).
 -export([handle_call_events/2]).
--export([handle_control_queue_req/2]).
 -export([queue_name/1]).
 -export([callid/1]).
 -export([node/1]).
@@ -113,11 +112,8 @@
                      ,{{?MODULE, 'handle_call_events'}
                        ,[{<<"call_event">>, <<"*">>}]
                       }
-                     ,{{?MODULE, 'handle_control_queue_req'}
-                       ,[{<<"call">>, <<"control_queue_req">>}]
-                      }
                     ]).
--define(QUEUE_NAME, <<>>).
+-define(QUEUE_NAME(CallId), <<"call_ctl_", CallId/binary>>).
 -define(QUEUE_OPTIONS, []).
 -define(CONSUME_OPTIONS, []).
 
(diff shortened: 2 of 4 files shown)


lazedo:
@danielfinke see if this make sense.
change start_link in ecallmgr_call_control to have a queue name based on callid (ex: call_ctl_xxxxxx), that way you would in advance the name of queue. pass that name as arg to init.
in init call ecallmgr_fs_channel/s to update the channel with the controlq.

in handle_call_events/usurp_conmtrol, call ecall_fs_channel/s to update the channel with the new controlq.

extend ecallmgr_fs_channel/s so that channel record supports the controlQ
the channels jobj resp to queries should include the controlq.

what you get with this,
1) no new api
2) controlq in channel queries response
3) doesn't change behaviour of ecallmgr_call_control to start answering questions.

thoughts ?

jamesaimonetti:
@danielfinke where are we with this PR? Any thoughts on the changes @lazedo suggested?

danielfinke:
@jamesaimonetti @lazedo made the majority of the changes requested. Do either of you know if usurp_control can pass a new control queue? Or does the control queue not matter if a usurp_control is sent?

lazedo:
@danielfinke after reviewing this again, i think the queue name as call_ctl_xxxxxx will not work if you have two ecallmgrs connected to same freeswitch.
you can use the `call_control_ready` function to update the control_q in the channel (must also make sure you clean it when receiving usurp_control.
 
one other alernative woul be to
extend the `wapi_call:publish_usurp_control` to include the control_q and maybe custom-channel-vars and listen to this message in your app.

danielfinke:
@lazedo would you say the queue name format call_ctl_xxxxxx_<node> be a reasonable alternative?

lazedo:
yes, i think it is. i think you will need to extend  `wapi_call:publish_usurp_control`anyway because you need to handle it and update the channel ?

danielfinke:
@lazedo I think that I will move the ecallmgr_fs_channels:update call into the call_control_ready function. That should suffice as the control_q will be available in the query_channels resp

lazedo:
@danielfinke i'm not sure if that's enough when you have 2 or more ecallmgrs. test it at least with 2 ecallmgrs, you may get different responses when using the query channels api.

lazedo:
@danielfinke to be clear, you can move the channel_update to call_control_ready but you need to extend wapi_call:publish_usurp_control to include control queue (and CCVs) and handle this in
handle_cast({'usurp_control', _}, State) in ecallmgr_call_control to also call channel_update. this will make sure that other nodes will update the channel so you don't get different answers when using query channels api.

lazedo:
@danielfinke another issue we will run into is that the channels api is federated, so we will get queues that are not directly accessible.

lazedo:
@danielfinke updates / comments ?

danielfinke:
@lazedo haven't got to spend much time on this. Not sure if I have a solution for the problem you pose. I'm certainly open to suggestions. By the way, I made the observation that when moving the control queue data into the usurp_control, not all ecallmgrs get it. And usually, only the original ecallmgr_call_control node will get it. Was the thought to do a pool_collect and pick out the response from query_channels_req that did contain the control queue?

lazedo:
@danielfinke looking into this again, see if this works for you. 
add `<<"Control-Queue">>` to OPTIONAL_CALL_USURP_CONTROL_HEADERS in https://github.com/2600hz/kazoo/blob/master/core/whistle-1.0.0/src/api/wapi_call.erl#L208

add `,{<<"Control-Queue">>, Q}`before https://github.com/2600hz/kazoo/blob/master/applications/ecallmgr/src/ecallmgr_call_control.erl#L542

your whapp should listen to call.usurp_control.(callid or * ) and retrieve the Control-Queue passed in payload.

thoughts ?

danielfinke:
@lazedo also looking into this again, your suggestion should suffice for what I am trying to do for now. Couple questions: first, do you know if there is guaranteed to be an ecallmgr_call_control proc (and therefore a control queue) for every single call that goes through the ecallmgr? Second, let's say I didn't want to have to capture and store the control queue from every usurp_control published, but just request it on the fly. How might that change the approach to this problem?

jamesaimonetti:
@danielfinke and @lazedo do either of you have the time and inclination to resume this conversation and get the PR into an acceptable state?

lazedo:
@jamesaimonetti @danielfinke i think this should be addressed in another way. the purpose of the control queue / controller is exactly to protect the winning app which is not being addressed here.

k-anderson:
@danielfinke This PR seems stale and the issue of going against the intention to keep the control queue from being discovered has not been resolved.  I would love to continue this in another forum but I am going to close this for now.

PR #1248, opened 2015-09-01
Added feature for excluding device from a queue

danielfinke:
When enabled, the device will be excluded from the list of devices that ring for the owning agent

PR #1267, opened 2015-09-02
Added feature for excluding device from a queue

danielfinke:
Move maybe_exclude_from_queues function calls to vars

#1248 against master

### DIFF
--- a/applications/callflow/src/cf_endpoint.erl
+++ b/applications/callflow/src/cf_endpoint.erl
@@ -666,12 +666,11 @@ maybe_do_not_disturb(Endpoint, _, _) ->
                                         'ok' |
                                         {'error', 'exclude_from_queues'}.
 maybe_exclude_from_queues(Endpoint, _, Call) ->
-    QueueId = whapps_call:custom_channel_var(<<"Queue-ID">>, Call),
-    Exclude = wh_json:is_true(<<"exclude_from_queues">>, Endpoint, 'false'),
-    case {QueueId, Exclude} of
-        {'undefined', _} -> 'ok';
-        {_, 'true'} -> {'error', 'exclude_from_queues'};
-        _ -> 'ok'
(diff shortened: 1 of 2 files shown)


lazedo:
@danielfinke can you provide a use case for this ?
i see an apent logging in to a device and not getting any calls and not being aware of that ?

danielfinke:
@lazedo we have a few clients who have a desk phone and a cell phone. They also are not dedicated agent staff; they sometimes have other roles. They can log out of acdc when they leave their desk, but when they are at their desk they want to only receive calls to their desk phone.

applications/callflow/src/cf_endpoint.erl
```diff
@@ -661,6 +662,18 @@ maybe_do_not_disturb(Endpoint, _, _) ->
             {'error', 'do_not_disturb'}
     end.
 
+-spec maybe_exclude_from_queues(wh_json:object(), wh_json:object(), whapps_call:call()) ->
+                                        'ok' |
+                                        {'error', 'exclude_from_queues'}.
+maybe_exclude_from_queues(Endpoint, _, Call) ->
+    QueueId = whapps_call:custom_channel_var(<<"Queue-ID">>, Call),
```
jamesaimonetti: Can we do something to make this short-circuit faster?

```erlang
case is_binary(whapps_call:custom_channel_var(<<"Queue-ID">>, Call))
    andalso wh_json:is_true(<<"exclude_from_queues">>, Endpoint)
of
    'false' -> 'ok';
    'true' -> {'error', 'excluded'}
end.
```

Something like this, to me, reads better as to the intention of the code (versus mashing the two toggles into a tuple).

Thoughts?

danielfinke:
@jamesaimonetti I am certainly happy with that suggestion - I've pushed a new commit

jamesaimonetti:
@danielfinke Thanks!

PR #1268, opened 2015-09-02
Track whether agent or member hung up queue call

danielfinke:
#1247 against master

### DIFF
--- a/applications/acdc/src/acdc_agent_fsm.erl
+++ b/applications/acdc/src/acdc_agent_fsm.erl
@@ -920,10 +920,10 @@ answered({'channel_hungup', CallId, <<"ATTENDED_TRANSFER">> = _Cause}, #state{me
     {'next_state', 'answered', State};
 answered({'channel_hungup', CallId, _Cause}, #state{member_call_id=CallId}=State) ->
     lager:debug("caller's channel hung up: ~s", [_Cause]),
-    {'next_state', 'wrapup', State#state{wrapup_ref=hangup_call(State)}};
+    {'next_state', 'wrapup', State#state{wrapup_ref=hangup_call(State, 'member')}};
 answered({'channel_hungup', CallId, _Cause}, #state{agent_call_id=CallId}=State) ->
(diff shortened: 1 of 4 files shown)


applications/acdc/src/acdc_agent_fsm.erl
```diff
@@ -936,10 +936,10 @@ answered({'sync_req', JObj}, #state{agent_listener=AgentListener
     {'next_state', 'answered', State};
 answered({'channel_unbridged', CallId}, #state{member_call_id=CallId}=State) ->
     lager:info("caller channel ~s unbridged", [CallId]),
-    {'next_state', 'wrapup', State#state{wrapup_ref=wrapup_timer(State)}};
+    {'next_state', 'answered', State};
 answered({'channel_unbridged', CallId}, #state{agent_call_id=CallId}=State) ->
     lager:info("agent channel unbridged"),
-    {'next_state', 'wrapup', State#state{wrapup_ref=hangup_call(State)}};
+    {'next_state', 'answered', State};
```
k-anderson: @danielfinke 
This and the diff above do not seem related to tracking the hangup unless I misunderstood something (possible as I don't know ACDc that well).  Are you sure that the transition here should not go to wrapup?  I suspect this might be a fix for something else ;)

danielfinke: We've been running this code in production for a while and haven't seen a case where this change affects the behaviour of the agent. Do you know of one? As far as the PR, it is important because the bridge always happens to unbridge from the agent side, regardless of who hangs up (which means this feature doesn't work as expected)

k-anderson: @danielfinke 
Ah, ok that makes sense.  I am not 100% sure of the impact I will need some time to follow the code path.  I suspect it is for offhook agents.  

Once I can verify that it doesn't impact existing functionality the rest of the PR looks good.

lazedo:
@danielfinke Thank You

PR #1249, opened 2015-09-01
Added participants/kick/mute/unmute to cb_conferences

danielfinke:

PR #1270, opened 2015-09-02
Added participants/kick/mute/unmute to cb_conferences

danielfinke:
Moved actions into POST to conference/id

Changed conference_participants to list comp

#1249 against master

### DIFF
--- a/applications/crossbar/src/modules/cb_conferences.erl
+++ b/applications/crossbar/src/modules/cb_conferences.erl
@@ -244,7 +244,7 @@ load_conference_details(Context, ConfId) ->
                                             ,{'ecallmgr', 'true'}
                                            ),
     case ReqResp of
-        {'error', _} -> cb_context:add_system_error('conference_search_error', Context);
+        {'error', _} -> cb_context:add_system_error('not_found', Context);
         {_, JObjs} -> participant_details(conference_participants(JObjs), Context)
     end.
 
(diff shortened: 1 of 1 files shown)


fenollp:
Check fails because you introduced trailing whitespace. Please remove all trailing whitespace :)

applications/crossbar/src/modules/cb_conferences.erl
```diff
@@ -216,6 +228,94 @@ load_conference(DocId, Context) ->
...
+    AuthDoc = cb_context:auth_doc(Context),
+    ConfDoc = cb_context:doc(load_conference(Id, Context)),
+    ConfOwnerId = wh_json:get_value(<<"owner_id">>, ConfDoc),
+    AuthOwnerId = wh_json:get_value(<<"owner_id">>, AuthDoc),
+    % Abort if the user is not room owner
+    Resp = case ConfOwnerId of
+        AuthOwnerId ->
+            Conference = whapps_conference:set_id(Id, whapps_conference:new()),
+            perform_conference_action(Conference, Action, ParticipantId);
+        _ -> <<"user_not_owner">>
```
k-anderson: @danielfinke 

This should not be a successful response ;)

Could you update it to use:
https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/cb_context.erl#L983
and only successful from the above case clause?

applications/crossbar/src/modules/cb_conferences.erl
```diff
@@ -216,6 +228,94 @@ load_conference(DocId, Context) ->
...
+    end,
+    crossbar_doc:handle_json_success(wh_json:set_value(<<"resp">>, Resp, wh_json:new()), Context).
+
+perform_conference_action(Conference, <<"mute">>, ParticipantId) ->
```
k-anderson: @danielfinke 
This function should have a spec

applications/crossbar/src/modules/cb_conferences.erl
```diff
@@ -216,6 +228,94 @@ load_conference(DocId, Context) ->
...
+           ,{<<"Fields">>, []}
+           ,{<<"Conference-ID">>, ConfId}
+           | wh_api:default_headers(?APP_NAME, ?APP_VERSION)
+          ],
+    ReqResp = whapps_util:amqp_pool_collect(Req
+                                            ,fun wapi_conference:publish_search_req/1
+                                            ,{'ecallmgr', 'true'}
+                                           ),
+    case ReqResp of
+        {'error', _} -> cb_context:add_system_error('conference_search_error', Context);
```
k-anderson: @danielfinke 
How would you feel about a more typical reply such as 
https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/cb_context.erl#L1018
or just the not_found?

PR #1271, opened 2015-09-02
KAZOO-4151: use whapps_calls own fold

fenollp:


fenollp:
@k-anderson This was for 3.21 and I just backported it to 3.20

PR #1272, opened 2015-09-02
cf_exe captures CHANNEL_DESTROY to prevent continuing after hangup

k-anderson:
This was originally from https://github.com/2600hz/kazoo/pull/1056.

However, that was issued against 3.20 and we need to target master.  Attempting merge resulted in conflicts so I corrected those and opened a new PR.  This also removes the need for https://github.com/2600hz/kazoo/pull/1258.

@danielfinke made this point way back on May 22 however it was not until PR #1258 that we internally discussed the merits of both approaches.  This resulted in a reversing our prior reluctance on #1056 and this PR was the result!

### DIFF
--- a/applications/callflow/src/cf_exe.erl
+++ b/applications/callflow/src/cf_exe.erl
@@ -386,9 +386,9 @@ handle_cast('control_usurped', State) ->
     {'stop', {'shutdown', 'control_usurped'}, State};
 handle_cast('channel_destroyed', State) ->
     {'noreply', State#state{destroyed='true'}};
-handle_cast('stop_on_destory', State) ->
+handle_cast('stop_on_destroy', State) ->
     {'noreply', State#state{stop_on_destroy='true'}};
-handle_cast('continue_on_destory', State) ->
+handle_cast('continue_on_destroy', State) ->
     {'noreply', State#state{stop_on_destroy='false'}};
 handle_cast({'continue_with_flow', NewFlow}, State) ->
     lager:info("callflow has been reset"),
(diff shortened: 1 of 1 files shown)


applications/callflow/src/cf_exe.erl
```diff
@@ -595,7 +644,6 @@ launch_cf_module(#state{flow=?EMPTY_JSON_OBJECT}=State) ->
 launch_cf_module(#state{call=Call
                         ,flow=Flow
                        }=State) ->
-    lager:debug("launching next flow module: ~p", [Flow]),
```
fenollp: Why remove this log line?

applications/callflow/src/cf_exe.erl
```diff
@@ -351,6 +384,12 @@ handle_cast('transfer', State) ->
     {'stop', {'shutdown', 'transfer'}, State};
 handle_cast('control_usurped', State) ->
     {'stop', {'shutdown', 'control_usurped'}, State};
+handle_cast('channel_destroyed', State) ->
+    {'noreply', State#state{destroyed='true'}};
+handle_cast('stop_on_destory', State) ->
```
danielfinke: typo -> stop_on_destroy

applications/callflow/src/cf_exe.erl
```diff
@@ -351,6 +384,12 @@ handle_cast('transfer', State) ->
...
+    {'noreply', State#state{stop_on_destroy='true'}};
+handle_cast('continue_on_destory', State) ->
```
danielfinke: typo -> continue_on_destroy

applications/callflow/src/cf_exe.erl
k-anderson: @fenollp because it flow is a large json object and this is not logging something specific.  Rather than fill the logs with erlang data objects we try to create logging that describes what is occurring.

PR #1275, opened 2015-09-03
KAZOO-4060: optionally record calls in stereo

fenollp:


applications/ecallmgr/src/ecallmgr_call_command.erl
```diff
@@ -1401,6 +1401,8 @@ set_record_call_vars(Node, UUID, JObj) ->
     Vars = lists:foldl(fun(F, V) -> F(V) end
                        ,[{<<"RECORD_APPEND">>, <<"true">>}
                          ,{<<"enable_file_write_buffering">>, <<"false">>}
+                         ,{<<"RECORD_STEREO">>, should_record_stereo(JObj)}
+                         ,{<<"RECORD_SOFTWARE">>, <<"2600Hz, Inc.'s Kazoo">>}
```
lazedo: i like it but, please remove it as whitelabeled accounts usually don't like this.

PR #1284, opened 2015-09-08
KAZOO-4009: optionally re-format caller id number based on regex and …

mark2600:
…classifier

### DIFF
--- a/applications/callflow/src/cf_endpoint.erl
+++ b/applications/callflow/src/cf_endpoint.erl
@@ -526,62 +526,9 @@ build(Endpoint, Properties, Call) ->
 maybe_rewrite_caller_id(Endpoint, Call) ->
     case wh_json:get_ne_value(<<"caller_id_options">>, Endpoint) of
         'undefined' -> Call;
-        CidOptions  -> maybe_class_format(Call, wh_json:get_value(<<"format">>, CidOptions))
+        CidOptions  ->
+            whapps_call:maybe_format_caller_id(Call, wh_json:get_value(<<"format">>, CidOptions))
     end.
-
--spec maybe_class_format(whapps_call:call(), api_object()) -> whapps_call:call().
-maybe_class_format(Call, 'undefined') -> Call;
-maybe_class_format(Call, Format) ->
-    Class = wnm_util:classify_number(whapps_call:caller_id_number(Call)),
-    lager:debug("checking for caller id reformating rules for ~s numbers", [Class]),
-    case wh_json:get_ne_value(Class, Format) of
-        'undefined' -> maybe_format_caller_id(Call, wh_json:get_ne_value(<<"all">>, Format));
-        UseFormat   -> maybe_format_caller_id(Call, UseFormat)
-    end.
-
--spec maybe_format_caller_id(whapps_call:call(), api_object()) -> whapps_call:call().
-maybe_format_caller_id(Call, 'undefined') -> Call;
-maybe_format_caller_id(Call, Format) ->
-    CallerId = whapps_call:caller_id_number(Call),
-    Regex = wh_json:get_ne_value(<<"regex">>, Format),
-    case maybe_regex_caller_id(CallerId, Regex, Format) of
-        CallerId -> Call;
-        Formatted ->
-            lager:info("caller id reformated from ~s to ~s", [CallerId, Formatted]),
-            whapps_call:set_caller_id_number(Formatted, Call)
-    end.
-
--spec maybe_regex_caller_id(ne_binary(), api_binary(), wh_json:object()) -> ne_binary().
-maybe_regex_caller_id(CallerId, 'undefined', _) -> CallerId;
-maybe_regex_caller_id(CallerId, Regex, Format) ->
-    Normalized = wnm_util:normalize_number(CallerId),
-    case re:run(Normalized, Regex, [{'capture', 'all_but_first', 'binary'}]) of
-        {'match', UseCid} ->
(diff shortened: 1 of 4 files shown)


applications/callflow/src/cf_endpoint.erl
```diff
@@ -516,11 +523,27 @@ build(EndpointId, Properties, Call) when is_binary(EndpointId) ->
...
+    Class   = wnm_util:classify_number(whapps_call:caller_id_number(Call)),
+    AllFmts = wh_json:get_value(<<"caller_id_format">>, Endpoint, []),
+    UseFmts = wh_json:get_value(Class, AllFmts, []),
+
+    lager:debug("possibly re-formatting caller id for ~p with regex ~p~n", [Class, UseFmts]),
+
+    case re:run(whapps_call:caller_id_number(Call), UseFmts, [{capture, all_but_first, binary}]) of
+        {match, UseCid} ->
+            lager:debug("match ~p found! re-formatting to ~p~n", [UseCid, hd(UseCid)]),
+            whapps_call:set_caller_id_number(hd(UseCid), Call);
```
k-anderson: @mark2600 what happened to prefix and suffix options?  I would also like to see the caller-id normalized prior to formatting so we would have a consistent number format to create the regex for. 

Something like:
    CallerId               = wnm_util:to_e164(whapps_call:caller_id_number(Call)),
    Classification     = wnm_util:classify_number(CallerId),
    ....
    case re:run(CallerId, Formatters, [{'capture', 'all_but_first', 'binary'}]) of
    ...

Minor formating:
* tickies
* spell out formats

applications/callflow/src/cf_endpoint.erl
```diff
@@ -265,6 +266,12 @@ merge_attributes([<<"record_call">> = Key|Keys], Account, Endpoint, Owner) ->
     OwnerAttr = get_record_call_properties(Owner),
     Merged = wh_json:merge_recursive([AccountAttr, OwnerAttr, EndpointAttr]),
     merge_attributes(Keys, Account, wh_json:set_value(Key, Merged, Endpoint), Owner);
+merge_attributes([<<"caller_id_format">> = Key|Keys], Account, Endpoint, Owner) ->
```
k-anderson: @mark2600 There is an object that exists called caller_id_options (from years ago) that was supposed to have a 'format' property.  I like the use of classifiers so maybe something like:

caller_id_options: {
    format: {
        us_did: {
            regex: "^\\+?1?([0-9]{10})$"
            prefix: "1"
        }
   }
}

thoughts?

k-anderson:
This needs to also happen in trunkstore for ts_from_offnet.  I believe the server options already has a caller_id_options object.  If we move this to a more common place we could reuse the logic.

applications/callflow/src/cf_endpoint.erl
```diff
@@ -516,11 +516,73 @@ build(EndpointId, Properties, Call) when is_binary(EndpointId) ->
...
+            lager:info("caller id reformated from ~s to ~s", [CallerId, Formatted]),
+            whapps_call:set_caller_id_number(Formatted, Call)
+    end.
+
+-spec maybe_regex_caller_id(ne_binary(), api_binary(), wh_json:object()) -> ne_binary().
+maybe_regex_caller_id(CallerId, 'undefined', _) -> CallerId;
+maybe_regex_caller_id(CallerId, Regex, Format) ->
+    Normalized = wnm_util:normalize_number(CallerId),
+    case re:run(Normalized, Regex, [{'capture', 'all_but_first', 'binary'}]) of
+        {'match', UseCid} ->
```
jamesaimonetti: Why not match the head here, rather than calling `hd/1` below?

applications/callflow/src/cf_endpoint.erl
```diff
@@ -516,11 +516,73 @@ build(EndpointId, Properties, Call) when is_binary(EndpointId) ->
...
     end.
 
+-spec maybe_rewrite_caller_id(wh_json:object(), whapps_call:call()) -> whapps_call:call().
+maybe_rewrite_caller_id(Endpoint, Call) ->
+    case wh_json:get_ne_value(<<"caller_id_options">>, Endpoint) of
+        'undefined' -> Call;
+        CidOptions  -> maybe_class_format(Call, wh_json:get_value(<<"format">>, CidOptions))
+    end.
+
+-spec maybe_class_format(whapps_call:call(), wh_json:object() | 'undefined') -> whapps_call:call().
```
fenollp: `wh_json:object() | undefined` = `api_object()`

applications/callflow/src/cf_endpoint.erl
```diff
@@ -516,11 +516,73 @@ build(EndpointId, Properties, Call) when is_binary(EndpointId) ->
...
+maybe_class_format(Call, 'undefined') -> Call;
+maybe_class_format(Call, Format) ->
+    Class = wnm_util:classify_number(whapps_call:caller_id_number(Call)),
+    lager:debug("checking for caller id reformating rules for ~s numbers", [Class]),
+    case wh_json:get_ne_value(Class, Format) of
+        'undefined' -> maybe_format_caller_id(Call, wh_json:get_ne_value(<<"all">>, Format));
+        UseFormat   -> maybe_format_caller_id(Call, UseFormat)
+    end.
+
+-spec maybe_format_caller_id(whapps_call:call(), wh_json:object() | 'undefined') -> whapps_call:call().
```
fenollp: same here

mark2600: good to know, will fix on next commit

PR #1286, opened 2015-09-09
KAZOO-4169: fixing registrar compilation

siplabs:


fenollp:
Does someone have a reproducible test case for this?
Reading the diff, it doesn't seem like anything changed…

lazedo:
i do

fenollp:
So it's unrelated to this error:

```
gcc -I/usr/lib/erlang/usr/include -I  -fpic -shared c_src/comp128.c -o priv/comp128.so
/usr/bin/ld: /tmp/ccJLs0on.o: relocation R_X86_64_32S against `.data' can not be used when making a shared object; recompile with -fPIC
/tmp/ccJLs0on.o: could not read symbols: Bad value
collect2: ld returned 1 exit status
make[2]: *** [priv/comp128.so] Error 1
```

@Macpie experienced this one

PR #1289, opened 2015-09-10
KAZOO-4107 Numbers that are 're-discovered' are not updated in the number database

lazedo:
KAZOO-4105 The number manger crawler does not appear to be crawling
KAZOO-4106 number crawler operates on chunks of documents

### DIFF
--- a/core/whistle_number_manager-1.0.0/include/wh_number_manager.hrl
+++ b/core/whistle_number_manager-1.0.0/include/wh_number_manager.hrl
@@ -73,7 +73,7 @@
                                        ,<<"failover">> ,<<"prepend">>
                                       ]).
 
--define(WNM_DB_PREFIX_L, "numbers/").
+-define(WNM_DB_PREFIX_L, "numbers/+").
 -define(WNM_DB_PREFIX, <<?WNM_DB_PREFIX_L>>).
 -define(WNM_DOC_VSN, <<"1">>).
 
--- a/core/whistle_number_manager-1.0.0/include/wh_number_manager.hrl
+++ b/core/whistle_number_manager-1.0.0/include/wh_number_manager.hrl
@@ -73,7 +73,7 @@
                                        ,<<"failover">> ,<<"prepend">>
                                       ]).
 
--define(WNM_DB_PREFIX_L, "numbers/+").
+-define(WNM_DB_PREFIX_L, "numbers/").
 -define(WNM_DB_PREFIX, <<?WNM_DB_PREFIX_L>>).
 -define(WNM_DOC_VSN, <<"1">>).
 
--- a/core/whistle_number_manager-1.0.0/src/wh_number_manager.erl
+++ b/core/whistle_number_manager-1.0.0/src/wh_number_manager.erl
@@ -1007,12 +1007,26 @@ prepare_find_results([Number|Numbers], ModuleName, ModuleResults, Found, Opts) -
                       | Found]
              end,
     case catch wnm_number:get(Number) of
-        #number{state=State} ->
+        #number{state=State
+                ,module_data=JObj
+                ,module_name = ModuleName
+               } ->
             case lists:member(State, ?WNM_AVALIABLE_STATES) of
                 'true' ->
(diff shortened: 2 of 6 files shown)


core/whistle_number_manager-1.0.0/include/wh_number_manager.hrl
```diff
@@ -73,7 +73,7 @@
                                        ,<<"failover">> ,<<"prepend">>
                                       ]).
 
--define(WNM_DB_PREFIX_L, "numbers/").
+-define(WNM_DB_PREFIX_L, "numbers/+").
```
k-anderson: @lazedo 
Why this change?  

Also, there is no guarantee that this is true.  If the number_manager e164 converters are configured differently on a cluster it may not result in a + prefix.  We have seen this before where a system had purposely been setup to result in US numbers without a '+1' prefix....

core/whistle_number_manager-1.0.0/src/wh_number_manager.erl
```diff
@@ -1007,7 +1007,10 @@ prepare_find_results([Number|Numbers], ModuleName, ModuleResults, Found, Opts) -
                       | Found]
              end,
     case catch wnm_number:get(Number) of
-        #number{state=State} ->
+        #number{state=State
+                ,module_data=JObj
+                ,module_name = ModuleName
```
k-anderson: @lazedo 
Correct me if I am wrong, but it looks like the discovery is only re-saved if the module_name does not match?  IE: if a search is preformed by the same carrier module then it can not resave the number if it already exists because it will match this case-clause (which is the old behavior).

If so lets talk about it :)

core/whistle_number_manager-1.0.0/src/wnm_number_crawler.erl
```diff
@@ -36,6 +36,10 @@
         ,whapps_config:get_integer(?WNM_CONFIG_CAT, <<"deleted_expiry_d">>, 90)
        ).
 
+-define(NUMBERS_TO_CRAWL
+        ,whapps_config:get_integer(?WNM_CONFIG_CAT, <<"number_of_documents_to_crawl">>, 1000)
```
k-anderson: @lazedo 
whistle_couch has a property default_chunk_size which I believe we have been using elsewhere.  I would prefer this as I don't see much value in changing the chunk size for individual operations.... The intent is to get chunks proportional to the size of the db cluster, so its very convenient to change it in one place...

core/whistle_number_manager-1.0.0/include/wh_number_manager.hrl
lazedo: @k-anderson there are some Dbs numbers/inum, numbers/managed numbers/local that were being crawled without reason. i understand your point here but e164 is meant to be +...

core/whistle_number_manager-1.0.0/src/wh_number_manager.erl
lazedo: @k-anderson  if it matches this clause it will not be save, ie, if its the same module and the same module_data (JObj) then we don't save it (no need to).

another clause was added to save it in case there is a difference

core/whistle_number_manager-1.0.0/src/wnm_number_crawler.erl
lazedo: @k-anderson got it, will use that one.

core/whistle_number_manager-1.0.0/src/wh_number_manager.erl
k-anderson: @lazedo Ah overlooked JObj, that makes sense.

core/whistle_number_manager-1.0.0/include/wh_number_manager.hrl
k-anderson: @lazedo true but...still makes me nervous... if it is just for an efficiency with managed numbers I would prefer to remove it.  I think that carrier module is only used on dev boxes...

PR #1290, opened 2015-09-10
KAZOO-4175: clicktocall: added option "dial_first"

siplabs:


k-anderson:
@siplabs 
I like it, any thoughts on supporting the same attribute on click-to-call?  It would need to come from c2c document rather than the query string but could be useful there as well!

PR #1291, opened 2015-09-10
KAZOO-4088: As a developer, i would like an API that generates a PDF off a directory so that we can display to user

macpie:


lazedo:
@macpie /core/kazoo-1.0.0 ? can we give it another name ? like kazoo_producer , kazoo_outputs, your suggestion here ..

/core/kazoo-1.0.0 seems more like a natural replacement for whistle-1.0.0

macpie:
Yes my point here was to replace whistle-1.0.0 in Kazoo being a util core element

macpie:
Fixing the xref issue

k-anderson:
@macpie 
After some thougth I believe core/kazoo-1.0.0 is premature.  If you put it in whistle-1.0.0 as its own module will be moved when we do conversion.  We will discuss this on the next standup ;)

fenollp:
@macpie 
```
warn_contract_types           cb_directories.erl:238: Invalid type specification for function cb_directories:pdf_users/4. The success typing is (_,_,[any()],[[{_,_},...]]) -> [[any(),...]]
warn_failing_call             cb_directories.erl:253: The call cb_directories:pdf_users(AccountDb::<<_:8,_:_*8>>,SortBy::any(),Users::any(),[[{_,_},...],...]) breaks the contract (ne_binary(),ne_binary(),wh_json:objects(),wh_proplist()) -> wh_proplist()
```

core/whistle-1.0.0/src/kz_pdf.erl
```diff
@@ -0,0 +1,168 @@
...
+%%% @doc
+%%% Builds an LOA doc using HTMLDoc (http://www.msweet.org/projects.php?Z1)
+%%% @end
+%%% @contributors:
+%%%   Peter Defebvre
+%%%   James Aimonetti
+%%%-------------------------------------------------------------------
+-module(kz_pdf).
+
+-include("kazoo.hrl").
```
lazedo: what does include add that its used by other modules ?

k-anderson:
@macpie with application/pdf why end the URL with /pdf?  seems redundant, I would prefer a GET to directories/{id} with an accepts header

macpie:
will fix

PR #1292, opened 2015-09-10
KAZOO-4149: fix credit card update if account is delinquent in braintree

fenollp:


k-anderson:
@fenollp 
I don't think you can copy ALL the parameters to a new subscription.  Things like the past due amount and dates may cause an issue...

fenollp:
@k-anderson If you like it now, I'd like to squash these commits before it gets merged.

PR #1295, opened 2015-09-11
KAZOO-3904: As a webhooks user, I would like a webhook that notifies me any time a doc changes

macpie:


fenollp:
@macpie 
```
warn_matching                 webhooks_util.erl:539: The pattern {'error', _} can never match the type boolean()
warn_matching                 webhooks_util.erl:540: The pattern {'ok', Doc} can never match the type boolean()
warn_not_called               webhooks_util.erl:541: Function will never be called
warn_not_called               webhooks_util.erl:551: Function load_metadata/2 will never be called
warn_not_called               webhooks_util.erl:555: Function will never be called
warn_not_called               webhooks_util.erl:557: Function will never be called
warn_not_called               webhooks_util.erl:559: Function will never be called
warn_not_called               webhooks_util.erl:566: Function update_metadata/2 will never be called
```

k-anderson:
@macpie I actually think I liked objects better :(  Lets discuss this with the team.

macpie:
I think objects does not make sense as it is a doc update your are subscribing for

applications/webhooks/src/webhooks_maintenance.erl
```diff
@@ -110,3 +111,20 @@ enable_account_hooks(AccountId) ->
 
 enable_descendant_hooks(AccountId) ->
     webhooks_util:reenable(AccountId, <<"descendants">>).
+
+-spec clean_webhooks_list() -> 'ok'.
+clean_webhooks_list() ->
```
jamesaimonetti: Seems like this is resetting the listing, not cleaning

macpie: Indeed renaming function

jamesaimonetti:
@macpie `doc` is a couchdb concept that we don't really extend into Kazoo. We name things `JObj` instead of `JDoc`, we are renaming `kazoo_documents` to `kazoo_objects`, etc.

macpie:
OK I get that but you are still  subscribing for couch db doc update

jamesaimonetti:
@macpie There are other "objects" that aren't tied to documents in couch. Plus, changing the module name makes it hard to see what you've actually changed! So, since there doesn't seem to be a compelling reason to rename object to doc, let's revert it back so we can better look over the PR and get it merged.

macpie:
ready

jamesaimonetti:
@k-anderson HALP? What happened here? I rebased origin/master and...it took a shit all over the PR.

jamesaimonetti:
@k-anderson I think I figured it out. Nothing to see here

PR #1297, opened 2015-09-11
KAZOO-4182: Makefile target to build bash completion for SUP

fenollp:


fenollp:
@k-anderson You could do `sup_completion_file=/etc/bash_completion.d/sup.bash make sup_completion || true`

PR #1299, opened 2015-09-14
KAZOO-4187: push only one of conflicting fields for bt_card

fenollp:


core/braintree-1.0.0/src/braintree_card.erl
```diff
@@ -225,9 +225,10 @@ record_to_xml(#bt_card{}=Card, ToString) ->
              ,{'number', Card#bt_card.number}
              ,{'cvv', Card#bt_card.cvv}
              ,{'billing-address-id', Card#bt_card.billing_address_id}],
-    Conditionals = [fun(#bt_card{billing_address='undefined'}, P) -> P;
-                       (#bt_card{billing_address=BA}, P) ->
-                            [{'billing-address', braintree_address:record_to_xml(BA)}|P]
+    Conditionals = [fun(#bt_card{billing_address=BA, billing_address_id='undefined'}, P)
+                          when BA =/= 'undefined' ->
+                            [{'billing-address', braintree_address:record_to_xml(BA)} | P];
```
k-anderson: @fenollp In this condition you should make sure that the billing_id is unset

fenollp: @k-anderson I only have one bt_card. How do I know what changed?

I think the only way to do this is to hard-wire a preference on one field (billing_address_id here) and if the user wants to change her billing address she just empties the other field.

PR #1300, opened 2015-09-14
KAZOO-4186: As a system administrator I would like to have an e164 invite format which ensures that no plus prefixed to the number (e164_without_plus - similar to 1npan).

onnet:
There was a patch done previously at trunkstore app, but it stopped working in recent versions. 
To make it work reliable it looks like it is worth to add "e164_without_plus" to wnm_util, ecallmgr_util ...

### DIFF
--- a/core/whistle_number_manager-1.0.0/src/wnm_util.erl
+++ b/core/whistle_number_manager-1.0.0/src/wnm_util.erl
@@ -26,7 +26,7 @@
          ,get_all_number_dbs/0
         ]).
 -export([normalize_number/1, normalize_number/2]).
--export([to_e164/1, to_e164/2, to_e164/3
+-export([to_e164/1, to_e164/2, to_e164/3, to_e164_without_plus/1
          ,to_npan/1, to_1npan/1
         ]).
 -export([is_e164/1, is_e164/2
@@ -431,6 +431,10 @@ maybe_convert_to_e164([Regex|Regexs], Converters, Number) ->
             Suffix = wh_json:get_binary_value([Regex, <<"suffix">>], Converters, <<>>),
             <<Prefix/binary, Root/binary, Suffix/binary>>
     end.
+
+-spec to_e164_without_plus(ne_binary()) -> ne_binary().
+to_e164_without_plus(Number) ->
+    re:replace(to_e164(Number), "[\+]", "", [{'return', 'binary'}]).
 
 %% end up with 8001234567 from 1NPAN and E.164
 -spec to_npan(ne_binary()) -> ne_binary().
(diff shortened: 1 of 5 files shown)


core/whistle_number_manager-1.0.0/src/wnm_util.erl
```diff
@@ -432,6 +432,10 @@ maybe_convert_to_e164([Regex|Regexs], Converters, Number) ->
             <<Prefix/binary, Root/binary, Suffix/binary>>
     end.
 
+-spec to_e164_without_plus(ne_binary()) -> ne_binary().
+to_e164_without_plus(Number) ->
+    re:replace(to_e164(Number), "[\+]", "", [{'return', 'binary'}]).
```
fenollp: Does this really need to be a regexp?
Looks like pattern matching on the first character, or even a binary comprehension could do the trick here.

onnet: I've chosen this way because it takes less strings :)
Please let me know if it is need to be redone in pattern matching way and I'll change it

lazedo:
@onnet doesn't formatters by itself resolve this ?

onnet:
@lazedo 

Unfortunately not. 

It is a long story. 

I've got numbers stored in numbers database with "+" by some reason (for ex. numbers/+7812) 
and reconcilation/conversion to e164 ends up with plus preffixed number.. 
So it looks like there is no another way to get rid of it 

"e164_converters": { 
"^\\+?7?(\\d{10})$": { 
"prefix": "+7" 
}, 
"^810(\\d*)$": { 
"prefix": "+" 
}, 
"^8(\\d{10})$": { 
"prefix": "+7" 
}, 
"^(\\d{7})$": { 
"prefix": "+7812" 
} 
}, 
"reconcile_regex": "^\\+\\d{10,}|^0[1-4,9]$|^10[0-4]$|^112$|^0\\d{2}$", 


On 14.09.2015 23:17, lazedo wrote: 




@onnet doesn't formatters by itself resolve this ?

lazedo:
@onnet your database is how it should be ;)
i was mentioning formatters not converters

onnet:
@lazedo 
Struggled a couple of hours with appeared problem (trunkstore stopped to strip pluses) after Sunday night upgrade to 3.21 (from 3.19)
Rolled back after that and didn't even thought about formtters while was trying to solve the problem.
My logic was the next: 
Since 1npan exists as an Invite format, and it is just a case of "e164 without plus" it should be expanded  to not only US but international usage also..
Besides that, imho, formatters is all about formatting inbound calls from defined resources but it goes here about outbound calls where I am usually use invite_format field to control format of DID I need to send to connected PBX

jamesaimonetti:
@onnet  we are moving away from using the `invite_format` field for this reason; it is too US-centric and the formatters/converters can apply across applications (trunkstore, callflow, etc). A generic solution so we don't have a bunch of special-case functions.

onnet:
@fenollp @lazedo @jamesaimonetti Many thanks for suggestions and explanations! Looks like this pull request should be abandoned.
But where formatters' rules should be placed since it goes not about resources but client's PBXs connected over trunkstore.

PR #1301, opened 2015-09-15
Delete from cache when Account DB delted/created

Dinkor:


lazedo:
@Dinkor i think you could do all this with wh_cache/wh_couch/wh_amqp without all the other modifications.

its all better to create jira tickets before actually trying to resolve the use cases.

Dinkor:
Yes. You are right. I will rewrite it and make new PR.

PR #1312, opened 2015-09-18
KAZOO-4217: simplify specs

fenollp:
usual Let's Cleanse the Codebase Friday

"Don't look at it, just merge it!"


jamesaimonetti:
@fenollp personally, I prefer `any()` over `_` in specs; helps differentiate that we're looking at a spec and not a function clause. I also read `_` as don't care versus `any()` which means anything could be returned. `any()` reads better in my mind when reading a spec. "This function returns anything" vs "This function returns don't care".

Also, I'll have to check again, but it was my thought that Dialyzer didn't like matching `[]` when just specifying `[something()]`. `[something()]` means one or more was my thought but I could be mistaken / was a long time ago that we started that convention.

jamesaimonetti:
@fenollp also, could you split this PR up into per-app/core lib dirs? We can process the PRs faster when not having to wade through so many touched files in one PR.

fenollp:
@jamesaimonetti
* I could split this PR into a commit per dir if you want…
* `[t()]` means exactly `[t(),...] | []` and Dialyzer knows that
* `_` has no pejorative connotation (seriously?). My argument pro `_` is that it is visually easier to *get* that this spec clause takes anything as its Nth argument. It a hole and visually it is too.

`-spec f(gna(), gna(), any(), gna()) -> gna().` 
reads to me much less quickly than
 `-spec f(gna(), gna(), _, gna()) -> gna().``

fenollp:
Oh and since these changes are all in the `-spec` part, if Travis says it compiles and checks, then **we can be pretty sure these changes are OK**.

jamesaimonetti:
@fenollp I know you're trying to sneak in backdoors in the commits Github doesn't show by default!

Regarding `_`, I don't have a strong preference either way. I find `any()` reads easier (even in your example); you don't and that's okay. Sounds like we need a vote from @k-anderson @lazedo @mark2600 and @macpie (and anyone else that may be following along).

lazedo:
for me anything will work as long as you change _  to any() :)

k-anderson:
@fenollp please rebase and change _ to any() as per the discussion on the standup.  Thanks!

fenollp:
NO

core/whistle_apps-1.0.0/src/whapps_sup.erl
```diff
@@ -33,15 +33,15 @@ start_link(Whapps) ->
 
 -spec start_app(atom()) ->
                        {'ok', pid() | 'undefined'} |
-                       {'ok', pid() | 'undefined', term()} |
-                       {'error', term()}.
+                       {'ok', pid() | 'undefined', _} |
```
lazedo: missed this one ?

core/whistle_apps-1.0.0/src/whapps_sup.erl
```diff
@@ -33,15 +33,15 @@ start_link(Whapps) ->
...
+                       {'error', any()}.
 start_app(App) ->
     supervisor:start_child(?MODULE, ?SUPER(App)).
 
 -spec restart_app(atom()) ->
                          {'ok', pid() | 'undefined'} |
-                         {'ok', pid() | 'undefined', term()} |
-                         {'error', term()}.
+                         {'ok', pid() | 'undefined', _} |
```
lazedo: and this

core/whistle_apps-1.0.0/src/whapps_conference.erl
```diff
@@ -451,7 +451,7 @@ kvs_append_list(Key, ValList, #whapps_conference{kvs=Dict}=Conference) ->
 kvs_erase(Key, #whapps_conference{kvs=Dict}=Conference) ->
     Conference#whapps_conference{kvs=orddict:erase(wh_util:to_binary(Key), Dict)}.
 
--spec kvs_fetch(_, whapps_conference:conference()) -> _.
+-spec kvs_fetch(_, whapps_conference:conference()) -> any().
```
lazedo: missed this

core/whistle_apps-1.0.0/src/whapps_call_command.erl
```diff
@@ -353,7 +353,7 @@ channel_status_filter([JObj|JObjs]) ->
 %%      for them in the receive blocks below.
 %% @end
 %%--------------------------------------------------------------------
--type relay_fun() :: fun((pid() | atom(), term()) -> any()).
+-type relay_fun() :: fun((pid() | atom(), _) -> _).
```
lazedo: missed

core/whistle_apps-1.0.0/src/whapps_call_command.erl
```diff
@@ -366,7 +366,7 @@ relay_event(Pid, JObj, RelayFun) ->
                            {'error', 'timeout'}.
 -spec receive_event(wh_timeout(), boolean()) ->
                            {'ok', wh_json:object()} |
-                           {'other', wh_json:object() | term()} |
+                           {'other', wh_json:object() | _} |
```
lazedo: missed

fenollp:
@lazedo yup sorry WIP

fenollp:
Found the `_` in the specs with `git grep -InE  '\-spec \w+\((.*?,\s*_|_)' -- applications/ core/`

fenollp:
Found the list stuff with `git grep -InE  '\Wlist\([^\)]+\)' -- applications/ core/`

fenollp:
Used this script to do the `?MODULE` commit:
```
#!/bin/bash

# Find modules that call themselves without using ?MODULE

dirs=.
[[ $# -ne 0 ]] && dirs=$*

function lookfor() {
    local module=$1
    local file="$2"
    grep -En --color=auto '(\W'"$module:|'$module')" "$file"
}

found=0
for file in $(find $dirs -name '*.erl' -type f); do
    module=$(grep -Eao -m 2 '\-module\([^\(\)]+\)' "$file" | sed 's/(/ /;s/)/ /' | awk '{print $2}')
    lookfor $module "$file" 2>&1 >/dev/null
    if [[ $? -ne 1 ]]; then
        echo "$file"
        grep -Fn $module: "$file"
        ((found++))
    fi
done

exit $found
```

But it's not reliable as is for it to be included in CI

PR #1314, opened 2015-09-21
KAZOO-4148: apply same validation logic for changing debit from credit

fenollp:


applications/crossbar/src/modules/cb_transactions.erl
```diff
@@ -123,9 +123,30 @@ validate(Context, PathToken) ->
...
+        AuthAccountId when AuthAccountId =/= MasterAccountId ->
+            lager:debug("allowing reseller to remove credit without invoking a bookkeeper"),
+            maybe_create_debit_tansaction(Context);
+        ResellerId ->
+            lager:debug("sub-accounts of non-master resellers must contact the reseller to change their credit"),
+            Resp = wh_json:from_list(
+                     [{<<"message">>, <<"Please contact your phone provider to remove credit.">>}
+                      ,{<<"cause">>, ResellerId}
+                     ]),
+            cb_context:add_validation_error(<<"amount">>, <<"forbidden">>, Resp, Context)
```
fenollp: Should this say `<<"amount">>`, or something different?

k-anderson: @fenollp it should match the property on the request, which in this case is 'amount' so yup!

PR #1316, opened 2015-09-22
KAZOO-4088: add missing template, fix provided content-types

k-anderson:
Also, use kz_pdf for all pdf creation


applications/crossbar/src/modules/cb_port_requests.erl
```diff
@@ -1602,3 +1498,72 @@ save_phone_numbers_doc(Context, JObj) ->
...
+
+-spec create_qr_code(api_binary(), api_binary()) -> wh_proplist() | 'undefined'.
+create_qr_code('undefined', _) -> 'undefined';
+create_qr_code(_, 'undefined') -> 'undefined';
+create_qr_code(AccountId, PortRequestId) ->
+    lager:debug("create qr code for ~s - ~s", [AccountId, PortRequestId]),
+    CHL = <<AccountId/binary, "-", PortRequestId/binary>>,
+    Url = <<"https://chart.googleapis.com/chart?chs=300x300&cht=qr&chl=", CHL/binary, "&choe=UTF-8">>,
+
+    case ibrowse:send_req(wh_util:to_list(Url)
```
lazedo: @k-anderson can we move this to a common place ? maybe wh_util ? or kz_image (manipulate and generate images) ? there is also a create_qr_code in notify whapp.

also, maybe add a type (port_request or other name in this case) to help differentiate origins ?

k-anderson: @lazedo 
Hey, I like the idea but I vote that is another ticket as this was supposed to fix a bug in the pdf for directories and already got bigger than I wanted....

PR #1318, opened 2015-09-22
KAZOO-4152: create and save a default clusterid for provisioner

fenollp:


applications/crossbar/src/modules/provisioner_v5.erl
```diff
@@ -544,9 +544,15 @@ req_headers(Token) ->
...
         ]).
 
--spec get_cluster_id() -> 'undefined' | string().
+-spec get_cluster_id() -> string().
 get_cluster_id() ->
-    whapps_config:get_string(?MOD_CONFIG_CAT, <<"cluster_id">>).
+    case whapps_config:get_string(?MOD_CONFIG_CAT, <<"cluster_id">>) of
+        'undefined' ->
+            ClusterId = wh_util:rand_hex_binary(16),
+            {'ok', _JObj} = whapps_config:set_default(?MOD_CONFIG_CAT, <<"cluster_id">>),
```
fenollp: Using `set_default` as we are creating a default value. I believe this makes more sense than using `set`.

applications/crossbar/src/modules/provisioner_v5.erl
```diff
@@ -544,9 +544,15 @@ req_headers(Token) ->
          ,{"User-Agent", wh_util:to_list(erlang:node())}
         ]).
 
--spec get_cluster_id() -> 'undefined' | string().
+-spec get_cluster_id() -> string().
 get_cluster_id() ->
-    whapps_config:get_string(?MOD_CONFIG_CAT, <<"cluster_id">>).
+    case whapps_config:get_string(?MOD_CONFIG_CAT, <<"cluster_id">>) of
```
lazedo: i would use

```
-define(DEFAULT_CLUSTER_ID, wh_util:rand_hex_binary(16)).

get_cluster_id() ->
  whapps_config:get_string(?MOD_CONFIG_CAT, <<"cluster_id">>, ?DEFAULT_CLUSTER_ID).
```

fenollp: What for? `?DEFAULT_CLUSTER_ID` would only be used here,
and if it wasn't the macro would have a different value on every call…

lazedo: you can replace the value of the macro if we decide to change the default value, no ?
the case is not needed (no extra logic needed) , its a simple line

fenollp: Then we're never setting the default and `cluster_id` is different for every provisioner request?

lazedo: check `whapps_config:get_string`. if you give a default and there is no value it sets the value.
basically it does what you are doing with the `case`

fenollp: Oh nice! Will amend, thanks

fenollp: Actually I don't want to: using `get_string/3` would compute the `rand` every time and that call is costly.
What do you think?

k-anderson:
@lazedo I agree with @fenollp regarding the use of the case clause...

lazedo:
@k-anderson so maybe its time for @fenollp to do another massive pr changing the way we get defaults in all code base ;)
the comment was strictly to maintain current pattern used. i would love to see whapps_config:get/3 accept a `fun` in the 3rd parameter. that way the costly default values would only run once.

fenollp:
@lazedo Write me a ticket ;)

k-anderson:
@lazedo not a bad idea to accept the fun ;) I am not sure if I am aware of other whapps_config usage comparable.  Personally I would have preferred function clauses but that is not a grievous style issue...

PR #1319, opened 2015-09-23
KAZOO-4108: fetch logs from notifications, consistently to the fax smtplog API

fenollp:
WIP

Also, in `cb_faxes` one can fetch `smtplog`s per-`Id`.
Should I implement the same API in `cb_notifications`?

### DIFF
--- a/applications/crossbar/src/modules/cb_notifications.erl
+++ b/applications/crossbar/src/modules/cb_notifications.erl
@@ -458,7 +458,7 @@ maybe_delete_template(Context, Id, ContentType, TemplateJObj) ->
         'undefined' ->
             lager:debug("failed to find attachment ~s", [AttachmentName]),
             JObj = wh_json:from_list([{<<"cause">>, ContentType}]),
-            cb_context:add_system_error('bad_identifier', JObj, ContentType);
+            cb_context:add_system_error('bad_identifier', JObj, Context);
         _Attachment ->
             lager:debug("attempting to delete attachment ~s", [AttachmentName]),
             crossbar_doc:delete_attachment(kz_notification:db_id(Id), AttachmentName, Context)
--- a/applications/crossbar/src/modules/cb_notifications.erl
+++ b/applications/crossbar/src/modules/cb_notifications.erl
@@ -74,14 +74,19 @@ init() ->
 -spec allowed_methods() -> http_methods().
 -spec allowed_methods(path_token()) -> http_methods().
 -spec allowed_methods(path_token(), path_token()) -> http_methods().
+
 allowed_methods() ->
     [?HTTP_GET, ?HTTP_PUT].
+
 allowed_methods(?SMTP_LOG) ->
     [?HTTP_GET];
 allowed_methods(_) ->
     [?HTTP_GET, ?HTTP_POST, ?HTTP_DELETE].
+
 allowed_methods(_, ?PREVIEW) ->
-    [?HTTP_POST].
+    [?HTTP_POST];
+allowed_methods(?SMTP_LOG, _Id) ->
+    [?HTTP_GET].
 
 %%--------------------------------------------------------------------
 %% @public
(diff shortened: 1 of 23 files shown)


core/kazoo_modb-1.0.0/priv/couchdb/views/notifications.json
```diff
@@ -3,7 +3,7 @@
     "language": "javascript",
     "views": {
         "smtp_log": {
-            "map": "function(doc) {if (doc.pvt_deleted || doc.pvt_type != 'notify_smtp_log') return; emit(doc.pvt_created);} ;"
+            "map": "function(doc) {if (doc.pvt_deleted || doc.pvt_type != 'notify_smtp_log') return; emit(doc.pvt_created); }"
```
fenollp: Should I add the same fields from the faxes view?

lazedo:
@fenollp 2/3 things missing
1) teletype creates the doc in the right modb but the document id also needs to be formatted with YYYY-MM so that we can get into the right modb.
2) cb_notifications needs endpoint for (?SMTP_LOG, Id) so that we can get a particular document
3)  optional but, the summary view for smtp in notifications could include basic fields instead of the document

fenollp:
@lazedo ready!

applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -462,11 +457,8 @@ maybe_delete_template(Context, Id, ContentType, TemplateJObj) ->
     case wh_doc:attachment(TemplateJObj, AttachmentName) of
         'undefined' ->
             lager:debug("failed to find attachment ~s", [AttachmentName]),
-            cb_context:add_system_error(
-              'bad_identifier'
-              ,wh_json:from_list([{<<"cause">>, ContentType}])
-              , Context
-             );
+            JObj = wh_json:from_list([{<<"cause">>, ContentType}]),
+            cb_context:add_system_error('bad_identifier', JObj, ContentType);
```
jamesaimonetti: I believe you wanted to use `Context` and not `ContentType`

fenollp: Nice catch!

fenollp: Thanks Dialyzer for not catching that…

jamesaimonetti: Dialyzer caught it for me. I suspect the way Travis/you run Dialyzer didn't include cb_context.beam so Dialyzer would not have known what to expect in that arg. I typically run it locally with:

```
dialyzer --plt .kazoo.plt file/i/changed.beam files/it/relies/on.beam
```

So if my changeset in `foo.erl` calls `bar:fun(X, Y)`, I'll include `foo.beam` and `bar.beam`. Plus I typically include `wh_util`, `wh_json` and `props` since most modules make some use of them.

core/kazoo_modb-1.0.0/priv/couchdb/views/notifications.json
```diff
@@ -3,7 +3,7 @@
     "language": "javascript",
     "views": {
         "smtp_log": {
-            "map": "function(doc) {if (doc.pvt_deleted || doc.pvt_type != 'notify_smtp_log') return; emit(doc.pvt_created);} ;"
+            "map": "function(doc) { if (doc.pvt_deleted || doc.pvt_type != 'notify_smtp_log') return; emit(doc.pvt_created, {'id': doc._id, 'from': doc.from, 'to': doc.to, 'created': doc.pvt_created}); }"
```
jamesaimonetti: Shouldn't you include `doc.error` in the view value? Fax's `smtp_log` view has an error property.

applications/crossbar/src/modules/cb_notifications.erl
fenollp: Then you should include these in the plt?
Would it work that way?

core/kazoo_modb-1.0.0/priv/couchdb/views/notifications.json
fenollp: I wanted to but I don't know which field that would be from the notifications doc. 
There's no errors field there

PR #1322, opened 2015-09-23
KAZOO-3896: remove chatty log line

k-anderson:


applications/crossbar/src/crossbar_doc.erl
```diff
@@ -1219,9 +1222,7 @@ filter_doc_by_querystring(Doc, QueryString) ->
 should_filter_doc(Doc, K, V) ->
     try filter_prop(Doc, K, V) of
         'undefined' -> 'true';
-        Bool ->
-            lager:debug("doc filtered by ~s(~p): ~s", [K, V, Bool]),
-            Bool
+        Bool -> Bool
```
fenollp: Maybe just log when `Bool` is `true`? that's just 0.58% of the lines emitted in total

PR #1324, opened 2015-09-24
Update to rabbitmq_server-3.5.5

k-anderson:
Update rabbitmq for bug fixes and performance improvements.

As of RabbitMQ 3.5.0 the priority queue is integrated into the broker; the plugin is no longer needed.

To start you will need to remove 'rabbitmq_priority_queue' from /etc/kazoo/rabbitmq/enabled_plugins.

@siplabs Could you please confirm that this branch does not cause issues with your priority queues?


k-anderson:
I am not sure why github only shows 2 additions but I confirmed that if I make a fresh clone and checkout this branch the new rabbit is present.

PR #1325, opened 2015-09-25
KAZOO-4131

lazedo:


fenollp:
@lazedo Is [this part](https://github.com/2600hz/kazoo/blob/KAZOO-4131/applications/blackhole/doc/index.md#wss-considerations) mergeable with the haproxy doc?

lazedo:
@fenollp no

PR #1326, opened 2015-09-26
Token restrictions tests

jamesaimonetti:
Unit tests hopefully covering the options available in token restriction rules.

### DIFF
--- a/applications/crossbar/src/crossbar_util.erl
+++ b/applications/crossbar/src/crossbar_util.erl
@@ -994,7 +994,7 @@ get_priv_level(AccountId, OwnerId) ->
 get_system_token_restrictions(AuthModule) ->
     case whapps_config:get(cb_token_restrictions:config_cat(), AuthModule) of
         'undefined' ->
-            whapps_config:get(cb_token_restrictions:config_cat(), <<"_">>);
+            whapps_config:get(cb_token_restrictions:config_cat(), ?CATCH_ALL);
         MethodRestrictions -> MethodRestrictions
     end.
 
@@ -1006,7 +1006,7 @@ get_account_token_restrictions(AccountId, AuthModule) ->
         {'ok', RestrictionsDoc} ->
             wh_json:get_first_defined(
               [[<<"restrictions">>, wh_util:to_binary(AuthModule)]
-               ,[<<"restrictions">>, <<"_">>]
+               ,[<<"restrictions">>, ?CATCH_ALL]
               ]
               ,RestrictionsDoc
              )
@@ -1018,7 +1018,7 @@ get_priv_level_restrictions(Restrictions, PrivLevel) ->
     RestrictionLevels = wh_json:get_keys(Restrictions),
     case lists:member(PrivLevel, RestrictionLevels) of
         'true' -> wh_json:get_json_value(PrivLevel, Restrictions);
-        'false' -> wh_json:get_json_value(<<"_">>, Restrictions)
+        'false' -> wh_json:get_json_value(?CATCH_ALL, Restrictions)
     end.
 
 %%--------------------------------------------------------------------
--- a/applications/crossbar/src/crossbar_util.erl
+++ b/applications/crossbar/src/crossbar_util.erl
@@ -963,7 +963,7 @@ create_auth_token(Context, AuthModule, JObj) ->
             cb_context:add_system_error('invalid_credentials', Context)
     end.
 
--spec get_token_restrictions(ne_binary(), ne_binary(), ne_binary()) ->
+-spec get_token_restrictions(atom(), ne_binary(), ne_binary()) ->
                                     api_object().
 get_token_restrictions(AuthModule, AccountId, OwnerId) ->
     case wh_util:is_system_admin(AccountId) of
@@ -990,7 +990,7 @@ get_priv_level(AccountId, OwnerId) ->
     {'ok', Doc} = couch_mgr:open_cache_doc(AccountDB, OwnerId),
     wh_json:get_ne_value(<<"priv_level">>, Doc).
 
--spec get_system_token_restrictions(ne_binary()) -> api_object().
+-spec get_system_token_restrictions(atom()) -> api_object().
 get_system_token_restrictions(AuthModule) ->
     case whapps_config:get(cb_token_restrictions:config_cat(), AuthModule) of
         'undefined' ->
@@ -998,7 +998,7 @@ get_system_token_restrictions(AuthModule) ->
         MethodRestrictions -> MethodRestrictions
     end.
 
--spec get_account_token_restrictions(ne_binary(), ne_binary()) -> api_object().
+-spec get_account_token_restrictions(ne_binary(), atom()) -> api_object().
 get_account_token_restrictions(AccountId, AuthModule) ->
     AccountDB = wh_util:format_account_db(AccountId),
     case couch_mgr:open_cache_doc(AccountDB, ?CB_ACCOUNT_TOKEN_RESTRICTIONS) of
--- a/applications/crossbar/src/crossbar_util.erl
+++ b/applications/crossbar/src/crossbar_util.erl
@@ -914,18 +914,21 @@ maybe_remove_attachments(Context) ->
              )
     end.
 
--spec create_auth_token(cb_context:context(), atom() | ne_binary()) -> cb_context:context().
-create_auth_token(Context, Method) ->
+-spec create_auth_token(cb_context:context(), atom()) ->
+                               cb_context:context().
+create_auth_token(Context, AuthModule) ->
     JObj = cb_context:doc(Context),
     case wh_json:is_empty(JObj) of
         'true' ->
             lager:debug("empty doc, no auth token created"),
             ?MODULE:response('error', <<"invalid credentials">>, 401, Context);
         'false' ->
-            create_auth_token(Context, Method, JObj)
+            create_auth_token(Context, AuthModule, JObj)
     end.
 
-create_auth_token(Context, Method, JObj) ->
+-spec create_auth_token(cb_context:context(), atom(), wh_json:object()) ->
+                               cb_context:context().
+create_auth_token(Context, AuthModule, JObj) ->
     Data = cb_context:req_data(Context),
 
     AccountId = wh_json:get_value(<<"account_id">>, JObj),
@@ -936,8 +939,8 @@ create_auth_token(Context, Method, JObj) ->
                ,{<<"owner_id">>, OwnerId}
                ,{<<"as">>, wh_json:get_value(<<"as">>, Data)}
                ,{<<"api_key">>, wh_json:get_value(<<"api_key">>, Data)}
-               ,{<<"restrictions">>, get_token_restrictions(Method, AccountId, OwnerId)}
-               ,{<<"method">>, wh_util:to_binary(Method)}
+               ,{<<"restrictions">>, get_token_restrictions(AuthModule, AccountId, OwnerId)}
+               ,{<<"method">>, wh_util:to_binary(AuthModule)}
               ]),
     JObjToken = wh_doc:update_pvt_parameters(wh_json:from_list(Token)
                                              ,wh_util:format_account_id(AccountId, 'encoded')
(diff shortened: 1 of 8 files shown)


applications/crossbar/src/crossbar_util.erl
```diff
@@ -960,51 +963,53 @@ create_auth_token(Context, Method, JObj) ->
...
 
--spec get_system_token_restrictions(ne_binary()) -> api_object().
-get_system_token_restrictions(Method) ->
-    case whapps_config:get(<<(?CONFIG_CAT)/binary, ".token_restrictions">>, Method) of
+-spec get_system_token_restrictions(atom()) -> api_object().
+get_system_token_restrictions(AuthModule) ->
+    case whapps_config:get(cb_token_restrictions:config_cat(), AuthModule) of
         'undefined' ->
-            whapps_config:get(<<(?CONFIG_CAT)/binary, ".token_restrictions">>, <<"_">>);
+            whapps_config:get(cb_token_restrictions:config_cat(), <<"_">>);
```
fenollp: Use your catchall macro?

applications/crossbar/src/crossbar_util.erl
```diff
@@ -1012,8 +1017,8 @@ get_priv_level_restrictions('undefined', _PrivLevel) -> 'undefined';
 get_priv_level_restrictions(Restrictions, PrivLevel) ->
     RestrictionLevels = wh_json:get_keys(Restrictions),
     case lists:member(PrivLevel, RestrictionLevels) of
-        'true' -> wh_json:get_ne_value(PrivLevel, Restrictions);
-        'false' -> wh_json:get_ne_value(<<"_">>, Restrictions)
+        'true' -> wh_json:get_json_value(PrivLevel, Restrictions);
+        'false' -> wh_json:get_json_value(<<"_">>, Restrictions)
```
fenollp: Here too

applications/crossbar/src/modules/cb_token_restrictions.erl
```diff
@@ -106,147 +125,148 @@ delete(Context) ->
...
-    Rules = wh_json:get_keys(R),
-    case match_rules(ReqParam, Rules) of
-        'undefined' ->
-            lager:info("no match rule for: ~p",[ReqParam]),
-                'undefined';
-        MatchRule ->
-            lager:debug("match rule: ~p",[MatchRule]),
-            wh_json:get_value(MatchRule, R)
+-spec allowed_accounts(api_binary(), api_binary()) -> ne_binaries().
+-ifdef(TEST).
```
fenollp: Why leave the spec outside the conditional macro scope?

applications/crossbar/src/modules/cb_token_restrictions.erl
```diff
@@ -106,147 +125,148 @@ delete(Context) ->
...
+                         ,ReqParams
+                        )
+    of
+        'false' -> match_rules(ReqParams, RuleKeys);
+        'true' -> RuleKey
     end.
+
+-spec does_rule_match(ne_binaries(), ne_binaries()) -> boolean().
+does_rule_match(Rule, ReqParams) ->
+    kazoo_bindings:matches(Rule, ReqParams).
```
fenollp: Not sure a new function is really needed...

applications/crossbar/src/modules/cb_token_restrictions.erl
```diff
@@ -106,147 +125,148 @@ delete(Context) ->
...
+
+-spec match_verb(cb_context:context(), http_methods()) -> boolean().
+match_verb(_Context, [?CATCH_ALL]) -> 'true';
+match_verb(Context, RuleVerbs) ->
+    lists:member(cb_context:req_verb(Context), RuleVerbs)
+        orelse lists:member(?CATCH_ALL, RuleVerbs).
```
fenollp: You could loop through `RuleVerbs` looking for `?CATCH_ALL` then return `lists:member(ReqVerb...` instead of doing the `?CATCH_ALL` 1.5 times

applications/crossbar/test/cb_token_restrictions_test.hrl
```diff
@@ -0,0 +1,129 @@
...
+                 ]
+                )
+              ]
+            }
+           ]
+          )
+       ).
+
+-define(CB_TOKEN_RESTRICTIONS_TEST_HRL, 'true').
+-endif.
```
fenollp: Actually why not put the tests in a `test/cb_token_restrictions_tests.erl` file including this hrl and the tests placed in the module? This way we have the macros and the tests in the same file and tests under `test/`.
Just conditionally export the required functions where needed.

core/kazoo_bindings-1.0.0/src/kazoo_bindings.erl
```diff
@@ -676,53 +676,79 @@ map_processor(Routing, Payload, Bindings) when not is_list(Payload) ->
...
                 ,[]
                 ,Bindings
                ).
 
+-type map_responder_fun() :: fun((kz_responder()) -> any()).
+-spec map_responder_fun(payload()) -> map_responder_fun().
+map_responder_fun(Payload) ->
+    fun(Responder) ->
+            apply_responder(Responder, Payload)
+    end.
```
fenollp: Make `map_responder_fun/1` into the actual fun (`../2`) so that a named fun is used? Otherwise you changed nothing..

core/kazoo_bindings-1.0.0/src/kazoo_bindings.erl
```diff
@@ -676,53 +676,79 @@ map_processor(Routing, Payload, Bindings) when not is_list(Payload) ->
...
+    case matches(BParts, RoutingParts) of
+        'false' -> Acc;
+        'true' ->
+            lager:debug("matched ~p to ~p", [BParts, RoutingParts]),
+            map_responders(Acc, Map, Responders)
+    end.
+
+-spec map_responders(list(), map_responder_fun(), queue()) -> list().
+map_responders(Acc, Map, Responders) ->
+    [catch Map(Responder)
```
fenollp: I believe `catch`'s weird scope is catching `Map` only, should bracket the whole function call

core/kazoo_bindings-1.0.0/src/kazoo_bindings.erl
```diff
@@ -676,53 +676,79 @@ map_processor(Routing, Payload, Bindings) when not is_list(Payload) ->
...
-                                 }, Acc) ->
-                          case Binding =:= Routing orelse matches(BParts, RoutingParts) of
-                              'true' ->
-                                  lager:debug("routing ~s matches ~s", [Routing, Binding]),
-                                  fold_bind_results(queue:to_list(Responders), Acc, Routing);
-                              'false' -> Acc
-                          end
-                  end, Payload, Bindings),
+    [Reply|_] =
+        lists:foldl(
```
fenollp: So it's folding over bindings or replies (or whatever!), why is it returning a list of replies and not just one reply? (also, why are you looking at just one reply in the output?)

applications/crossbar/src/modules/cb_token_restrictions.erl
jamesaimonetti: I think the call site reads better with `does_rule_match` than `kazoo_bindings:matches`.

applications/crossbar/src/modules/cb_token_restrictions.erl
jamesaimonetti: In theory, one would only use `["_"]` so we can short circuit `match_verb` in the first clause and avoid calls to `lists:member/2`. Otherwise we need to check that someone didn't include the catch-all in a list with other verbs.

applications/crossbar/test/cb_token_restrictions_test.hrl
jamesaimonetti: You mean like [here](https://github.com/2600hz/kazoo/blob/token_restrictions_tests/applications/crossbar/test/cb_token_restrictions_test.erl)? This macro builds the restrictions JSON to test, for the specific scenario under test.

core/kazoo_bindings-1.0.0/src/kazoo_bindings.erl
jamesaimonetti: This curries the `Payload` into an arity-1 function for easy passing later in the fold. Do you have an alternative code suggestion?

core/kazoo_bindings-1.0.0/src/kazoo_bindings.erl
jamesaimonetti: The fold processor expects the first element in the `Payload` to be the pseudo-accumulator for the fold, typically the `cb_context` record but could be anything.

applications/crossbar/src/modules/cb_token_restrictions.erl
mark2600: I agree with @fenollp on this personally, adding a function just so something reads better is just adding more code unnecessarily in my opinion.

jamesaimonetti: @fenollp @mark2600 I just checked the generated assembler from both variations (calling sub function vs direct call to `kazoo_bindings`) to be sure; the compiler optimizes the sub-function call out. The assembler for `match_rules/2` is identical, so I'm thinking readability in `match_rules/2` is improved with no impact on execution time by using the sub-function.

fenollp: I think we were talking readability, not execution time you picky you

applications/crossbar/src/modules/cb_token_restrictions.erl
fenollp: I agree

applications/crossbar/test/cb_token_restrictions_test.hrl
fenollp: Yes, why not move all the test code for this module in its own test module?

core/kazoo_bindings-1.0.0/src/kazoo_bindings.erl
fenollp: I'm saying that writing this function is no more useful than using `fun (_1) ... apply_responder... end` directly.
And it adds to the read dereferencing

applications/crossbar/src/modules/cb_token_restrictions.erl
k-anderson: @fenollp I think he was stating that this is more readable and does not impact execution time. 

@mark2600 @fenollp I agree with James, although uncommon for single function call like this it is at the discretion of the programmer to determine when readability would be improved by such a 'wrapper'.  There are also several examples of this in the existing code base so it is not  without president.  The project puts a high value on readability.  I would also point out that the the code review is not intended as forum for stylistic changes (unless grievous) as has been cropping up lately ;)  As long as it is not a drastic deviation from the typical style we should be lenient in this regards.

mark2600: I will just pose the following question then:

If instead of wrapping bindings:matches with does_rule_match you just used bindings:matches but above it put a comment that said % does the rule match ? would you be just as much in favor of it ?

applications/crossbar/test/cb_token_restrictions_test.hrl
jamesaimonetti: I put these macros in a hrl since they were big and clunky and cluttered the test module up. Easier to read `?HTTP_VERB_RESTRICTIONS(Verbs)` than the `wh_json:from_list()` monstrosity. If the macro was doing more than constructing a term, I would create a function but in this case it felt fine to put a "static" term builder in a macro instead.

core/kazoo_bindings-1.0.0/src/kazoo_bindings.erl
jamesaimonetti: I would counter that this seems like a good way to hide some clutter from the call site that would distract from reading `map_processor/3`

applications/crossbar/src/modules/cb_token_restrictions.erl
jamesaimonetti: Personally, I don't read comments unless I can't figure out what the code is telling me.

To your other point, `M-.` takes me right to the definition of `does_rule_match/2` if I care to inspect it (regardless of where that function is defined). I find jumping to/from functions like that easier than scrolling up/down to find what I want.

My thought is, reading `match_rules/2`, I can read that function body in my head and immediately know what its doing. I don't care **how** `does_rule_match/2` determines if RuleKey matches ReqParams, just that it tells me true or false. If I'm trying to determine if there's a bug in the **how**, I'll jump to `does_rule_match/2`'s definition and debug from there.

Does that peel back my thought process curtain enough? Can I invoke my best @fenollp impression with "MergeMe!"? :)

jamesaimonetti:
https://i.imgur.com/ssBZhCs.gifv

PR #1328, opened 2015-09-29
Added support to callflow modules and pivot to utilize custom KVS

danielfinke:
Continuation of #1255.
@lazedo In response to your idea to move away from changing the cf modules to use maybe_use_variable: do you have any suggestions for dealing with validation of doc existence? For example, any that are using cf_util:maybe_use_variable have the id value checked in couch. However, cf_pivot is the exception to this rule. cf_pivot just takes the voice_url value. Perhaps a third element of the tuple you suggested? ( {<<"TheElementIWant">>, {<<"kvs">>, <<"kv">>}, 'validate_exists'} )

Multiple fixes

Moved maybe_use_variable into cf_util
Formatting cleanup
Folding functions / list comps
Use wh_json:encode instead of format_json_rec

### DIFF
--- a/applications/callflow/src/cf_util.erl
+++ b/applications/callflow/src/cf_util.erl
@@ -59,6 +59,7 @@
          ,check_value_of_fields/4
          ,get_timezone/2, account_timezone/1
         ]).
+-export([maybe_use_variable/2]).
 
 -export([wait_for_noop/2]).
 -export([start_task/3]).
@@ -957,6 +958,18 @@ account_timezone(Call) ->
             kz_account:timezone(AccountJObj, ?DEFAULT_TIMEZONE);
         {'error', _E} ->
             whapps_config:get(<<"accounts">>, <<"timezone">>, ?DEFAULT_TIMEZONE)
+    end.
+
+-spec maybe_use_variable(wh_json:object(), whapps_call:call()) -> api_binary().
+maybe_use_variable(Data, Call) ->
+    case wh_json:get_value(<<"var">>, Data) of
+        'undefined' -> wh_doc:id(Data);
+        Variable ->
+            Value = wh_json:get_value(<<"value">>, cf_kvs_set:get_kv(Variable, Call)),
+            case couch_mgr:open_cache_doc(whapps_call:account_db(Call), Value) of
+                {'ok', _} -> Value;
+                _ -> wh_doc:id(Data)
+            end
     end.
 
 -spec start_task(fun(), list(), whapps_call:call()) -> 'ok'.
--- a/applications/callflow/src/module/cf_pivot.erl
+++ b/applications/callflow/src/module/cf_pivot.erl
@@ -41,7 +41,7 @@
 handle(Data, Call) ->
     Prop = props:filter_empty(
              [{<<"Call">>, whapps_call:to_json(Call)}
-              ,{<<"Voice-URI">>, wh_json:get_value(<<"voice_url">>, Data)}
+              ,{<<"Voice-URI">>, maybe_use_variable(Data, Call)}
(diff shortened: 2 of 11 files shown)


danielfinke:
(I'll fix the trailing whitespace after we finish adjustments to this)

core/kazoo_translator-1.0.0/src/convertors/kzt_kazoo.erl
```diff
@@ -65,4 +65,5 @@ req_params(Call) ->
        ,{<<"Transcription-Text">>, kzt_util:get_transcription_text(Call)}
        ,{<<"Transcription-Status">>, kzt_util:get_transcription_status(Call)}
        ,{<<"Transcription-Url">>, kzt_util:get_transcription_url(Call)}
+       | cf_kvs_set:kvs_to_props(Call)
```
jamesaimonetti: We're trying hard to not have cross-app dependencies like this. If this is "generic" functionality, it should be moved to a lib in core.

applications/callflow/src/module/cf_pivot.erl
```diff
@@ -52,6 +52,15 @@ handle(Data, Call) ->
     lager:info("published pivot request"),
     wait_for_pivot(Data, Call).
 
+-spec maybe_use_variable(wh_json:object(), whapps_call:call()) -> api_binary().
+maybe_use_variable(Data, Call) ->
```
jamesaimonetti: This function name gives me no indication of what the function actually does. I'm generally confused as to what's going on and why this functionality is needed. Perhaps you could write up some docs in `callflow/doc` to describe what you're trying to accomplish with this function and the `cf_kvs_set` module?

applications/callflow/src/cf_util.erl
```diff
@@ -959,6 +960,18 @@ account_timezone(Call) ->
             whapps_config:get(<<"accounts">>, <<"timezone">>, ?DEFAULT_TIMEZONE)
     end.
 
+-spec maybe_use_variable(wh_json:object(), whapps_call:call()) -> api_binary().
+maybe_use_variable(Data, Call) ->
```
jamesaimonetti: Same as in `cf_pivot` but I have no idea what this function is meant for. Docs describing what you're attempting to do here would be helpful.

jamesaimonetti:
@danielfinke i'm coming to this PR with a clean slate, hence the questions. I went back and read over the former PR and have an idea now what you're trying to accomplish. I still think you writing the docs will help clarify in your mind what you're accomplishing (as well as examples of how it can be used) and help others that don't have access to your thoughts know what's up.

I think this is a worthwhile addition in concept but I want to make it better in execution and need some more information to get it there!

lazedo:
@danielfinke
the proposed changed would be in https://github.com/2600hz/kazoo/blob/master/applications/callflow/src/cf_exe.erl#L646
Data = apply_dynamic_values(wh_json:get_value(<<"data">>, Flow, wh_json:new())),
do a fold over the KVs and a function clause to handle 
{Key, {<<"kvs">>, Value} which would then take the value from kvs (in this case).
i believe that 'validate_exists' is out of scope here, but we could accept a default value in case the value is not set.

the only other concern is the cross-app reference that @jamesaimonetti  referenced above.

thoughts ?

fenollp:
Travis CI is unhappy about these trailing spaces in `applications/callflow/src/module/cf_kvs_set.erl`

lazedo:
@danielfinke any updates / comments ?

jamesaimonetti:
@danielfinke similarly here, please re-open the PR when you have time to continue the conversation on getting this merged.

Thanks!

PR #1329, opened 2015-09-29
Added position/estd wait time announcements to queues

danielfinke:
This feature adds periodic audio announcements of position in queue and estimated time to acdc. In its current state it works. However, I am very open to suggestions for improvement as it feels like a bit of a hack.

The queue_manager that receives the member_call will spawn a new proc to periodically announce media to a given call ID. Estimated times are computed based on acdc_stats data. Getting position # is a synchronous request against the queue_manager. New calls are broadcast to each other queue_manager so that it does not matter which receives new calls - each queue_manager is aware of the position of each call ID.

Included is default media, amqp message functions (for adding/removing/position stats), and couch schema.


fenollp:
Travis doesn't like trailing whitespace in `applications/acdc/src/acdc_queue_manager.erl`.

lazedo:
@danielfinke Hi, can you rebase and also check trailing whitespace ?

jamesaimonetti:
@danielfinke re-open when you've got this one ready to go again.

Thanks!

PR #1347, opened 2015-10-03
KAZOO-4240: reverse grouped-by-CSeq chunks

fenollp:


applications/call_inspector/src/parsers/ci_parser_hep.erl
```diff
@@ -177,5 +177,9 @@ make_and_store_chunk(ParserId, Hep) ->
     ci_datastore:store_chunk(Chunk).
 
 -spec ip(inet:ip4_address() | inet:ip6_address()) -> ne_binary().
+ip({92,_,_,_}=IP) ->
```
fenollp: This is temporary. Right @k-anderson?

PR #1348, opened 2015-10-05
KAZOO-4258: added application to originate callback on events

siplabs:
This patch adds application to originate callbacks on time or AMQP events.

Currently it may originate call when receive `new_voicemail` or if at specified time voicemail box has unread messages.


applications/ananke/doc/index.md
```diff
@@ -0,0 +1,161 @@
...
+
+Also the following fields are checked
+
+** `notify_callback_disabled` in vmbox
+** `vm_notify_callback_disabled` in vmbox owner's document
+
+If number was determined and notifications are not disabled, the call is being originated. The second leg is voicemail check callflow number.
+A-leg custom channel variable `Caller-ID-Number` is set being equal to `mailbox` parameter in corresponding document.
+
+A call will be attempted `attempts` times every `interval` minutes. Call timeout can be adjasted by modifying `callTimeout` option (in seconds)
```
jamesaimonetti: Is it `callTimeout`? Should be `call_timeout_s` for case and so we can tell time unit at a glance.

applications/ananke/src/ananke_vm_callback.erl
```diff
@@ -0,0 +1,270 @@
...
+-module(ananke_vm_callback).
+
+-export([init/0
+         ,handle_req/2
+         ,check/2
+        ]).
+
+-include("ananke.hrl").
+
+-record(args, {account_id
```
jamesaimonetti: specs on the record would be helpful

applications/ananke/src/ananke_vm_callback.erl
```diff
@@ -0,0 +1,270 @@
...
+    {'ok', VMBox} = couch_mgr:open_cache_doc(AccountDb, VMBoxId),
+    case [X || X <- wh_json:get_value(<<"messages">>, VMBox, [])
+               , wh_json:get_value(<<"folder">>, X) =:= <<"new">>]
+    of
+        [] ->
+            lager:info("no unreaded messages"),
+            'ok';
+        _  ->
+            lager:info("found unreaded messages"),
+            handle_req(wh_json:set_values([{<<"Account-ID">>, AccountId}
```
jamesaimonetti: I typically use `wh_json:from_list([...])` instead of `wh_json:set_values([...], wh_json:new())`; I think it reads better the intent - creating a new json object.

applications/ananke/src/ananke_vm_callback.erl
```diff
@@ -0,0 +1,270 @@
...
+    _ = wh_util:put_callid(JObj),
+    AccountId = wh_json:get_value(<<"Account-ID">>, JObj),
+    AccountDb = wh_json:get_value(<<"Account-DB">>, JObj),
+    VMBoxId = wh_json:get_value(<<"Voicemail-Box">>, JObj),
+    lager:debug("handling new voicemail in ~p", [VMBoxId]),
+    {'ok', VMBoxJObj} = couch_mgr:open_cache_doc(AccountDb, VMBoxId),
+    case kzd_voicemail_box:owner_id(VMBoxJObj) of
+        'undefined' ->
+            lager:debug("no owner"),
+            'ok';
```
jamesaimonetti: `lager:Severity/{1,2}` both return 'ok'; no need for the explicit 'ok' here.

applications/ananke/src/ananke_vm_callback.erl
```diff
@@ -0,0 +1,270 @@
...
+                              ,vm_number = VMNumber
+                              ,attempts = Attempts
+                              ,interval = Interval
+                              ,call_timeout = CallTimeout
+                              ,realm = Realm
+                             },
+            maybe_start_caller(StartArgs)
+    end.
+
+-spec get_voicemail_number(ne_binary(), ne_binary()) -> ne_binary() | 'undefined'.
```
jamesaimonetti: `ne_binary() | 'undefined'` == `api_binary()`

applications/ananke/src/ananke_vm_callback.erl
```diff
@@ -0,0 +1,270 @@
...
+          ,IsFlow
+            andalso wh_json:get_value(<<"module">>, FlowJObj) =:= <<"voicemail">>
+            andalso wh_json:get_value([<<"data">>, <<"action">>], FlowJObj) =:= <<"check">>}
+    of
+        {'false', _} -> 'false';
+        {'true', 'true'} -> 'true';
+        _ -> is_voicemail_cf(FlowJObj)
+    end.
+
+-spec get_cf_flow(wh_json:object()) -> wh_json:object() | 'undefined'.
```
jamesaimonetti: `wh_json:object() | 'undefined'` == `api_object()`

applications/crossbar/priv/couchdb/schemas/accounts.json
```diff
@@ -162,6 +162,26 @@
             "name": "Default Timezone",
             "required": false,
             "type": "string"
+        },
+        "vm_notify_callback_attempts": {
```
k-anderson: @siplabs Howdy!  I would really prefer if these settings were nested and identical on all documents for a parameter controlling the same property(for example on accounts it is "vm_notify_callback_attempts" but on vmboxes it is "notify_callback_attempts").

I would propose the following object for all documents:
{
  ...
  "callback": {
    "voicemail": {
      "attempts": 1
      "disabled": false
      "interval": 300
      "timeout": 26
    }
  }
  ...
}

This helps set up for easy future expansion.  More callback modules could be added to the callback object, feature extensions of the voicemail feature are in that object.  ect...  This has been a lesson from experience which is why there are counter examples in some documents ;)

deps/leader_cron-0.1/.gitignore
```diff
@@ -0,0 +1,11 @@
+erl_crash.dump
```
k-anderson: @siplabs This is a general comment on leader_cron so I am just adding it to the first of the files.  This is using native distributed Erlang which is NOT currently part of our platform.  I am opposed to such a significant and fundamental change as to start requiring interconnected Erlang nodes for this.  I have looked at gen_leader before and am of the opinion that it would be very feasible to use AMQP (gen_listener) as the message passing.  In some places it simplifies it greatly because rather than track there is a the ability to broadcast.

So here is the dilemma if we accept this it will become messy as those that wish to use it interconnect their whapps nodes (bypassing AMQP and effectively creating two message buses).  I also have concerns that the scope of this whapps interconnect would creep into other things.  As such I would like to start discussions on how to achieve this same functionality using the existing message bus before we accept the PR.

k-anderson: On other note regarding gen_listener using AMQP.  I think that is very generic solution (for the project) that will benefit a number of applications (for example the fax worker).  If in our discussions that this the route we take that should become a core library.

siplabs:
Of course it is to be ported to AMQP. This was contributed because of Darren's requirements of projects announced on 'Con being contributed BEFORE the presentation. 
We are to port it to AMQP ASAP
We see a huge potential with this either so we are on the same page.

applications/crossbar/priv/couchdb/schemas/notify.callback.json
```diff
@@ -0,0 +1,35 @@
+{
+    "$schema": "http://json-schema.org/draft-03/schema#",
+    "_id": "notify.callback",
+    "description": "Schema for a callback options",
+    "name": "Voicemail Box",
```
jamesaimonetti: Doesn't seem like this is a Voicemail Box

applications/crossbar/priv/couchdb/schemas/users.json
```diff
@@ -474,12 +474,20 @@
             "required": false,
             "type": "boolean"
         },
-        "vm_to_email_enabled": {
-            "default": true,
-            "description": "Determines if the user would like voicemails emailed to them",
-            "name": "VM To Emain Enabled",
+        "voicemail": {
```
jamesaimonetti: Why remove the `vm_to_email_enabled` flag? That is used by notify and teletype (via `kzd_user`) to determine whether to send the email notification. I'm not opposed to seeing it added as a sub-property of the "voicemail" key here, but it shouldn't be out-right removed.

applications/ananke/src/ananke_listener.erl
```diff
@@ -0,0 +1,294 @@
...
+                                                          ,{<<"hours">>, 0}
+                                                          ,{<<"days">>, 0}
+                                                         ]),
+            {'sleeper', ?MILLISECONDS_IN_SECOND * (Seconds
+                                                   + ?SECONDS_IN_MINUTE * Minutes
+                                                   + ?SECONDS_IN_HOUR * Hours
+                                                   + ?SECONDS_IN_DAY * Days)}
+    end.
+
+-spec schedule({ne_binary(), time(), fun()}) -> {'ok', pid()} | {'error', term()}.
```
jamesaimonetti: ananke_listener.erl:229: Invalid type specification for function ananke_listener:schedule/1. The success typing is ({atom() | binary(),{'cron',{_,_,_,_,_}} | {'oneshot',pos_integer() | {_,_}} | {'sleeper',pos_integer()},{fun(),[any()]} | {atom(),atom(),[any()]}}) -> {'error',_} | {'ok',pid()}

jamesaimonetti: I think `fun()` should be `leader_cron_callback()`

jamesaimonetti:
@siplabs any progress on comments above?

applications/ananke/src/ananke_listener.erl
siplabs: @jamesaimonetti corrected to `leader_cron_task:execargs()` type.

siplabs:
@jamesaimonetti updated.

siplabs:
@jamesaimonetti I reproduced travis fail in my enviroment. It seems that
```erlang
-ifdef(OTP_AT_LEAST_18).
-type sup_start_flags() :: supervisor:sup_flags().
-else.
-type sup_start_flags() :: {supervisor:strategy(), non_neg_integer(), non_neg_integer()}.
-endif.
```
in `wh_types.hrl` doesn't work.
If I change `non_neg_integer()` to `pos_integer()` dialyzer warning disapires.
If I declare type as `-type sup_start_flags() :: supervisor:sup_flags().` it disapires too.
Also I added blank line to `ecallmgr_sup` and got the same dialyzer warning.

fenollp:
@siplabs Yes, after reading the docs it seems
```erlang
-type sup_start_flags() :: {supervisor:strategy(), non_neg_integer(), non_neg_integer()}.
```
should in fact be 
```erlang
-type sup_start_flags() :: {supervisor:strategy(), non_neg_integer(), pos_integer()}.
```
My bad, most probably!

fenollp:
We'll see how #1516 does

siplabs:
@fenollp thank you!
It seems that `-ifdef(OTP_AT_LEAST_18).` doesn't work. I got this dialyzer warning only using 18.x erlang and in this case type should be `-type sup_start_flags() :: supervisor:sup_flags().
-else.` which pass the check.

applications/crossbar/priv/couchdb/schemas/notify.callback.json
siplabs: @jamesaimonetti fixed

applications/crossbar/priv/couchdb/schemas/users.json
siplabs: @jamesaimonetti my bad, fixed.

fenollp:
Please rebase once #1516 is merged :)

siplabs:
@jamesaimonetti @k-anderson 
#1517 - amqp_leader implementation
Both PRs are quite big, I think it's better to discuss `ananke` and `amqp_leader` separetly if you don't mind.

siplabs:
@k-anderson @jamesaimonetti
this pull request was rebased to master.
I added option to schedule callbacks with different intervals.

siplabs:
@fenollp I don't understand what is wrong. Could you tell what's wrong in my code, please?

fenollp:
ah I forgot this check was on!

run `./scripts/check-app-registered.sh applications/ananke` until you have replaced `?MODULE` usage with `?SERVER`.
This is a script that finds registered modules.

siplabs:
@fenollp I replaced `?MODULE` with `?SERVER` in all registered modules. This module doesn't register anything. May it be a bug in `/scripts/check-app-registered.sh`?

fenollp:
It's saying there's one more thing to replace:
```
applications/ananke/src/ananke_callback_wrkr.erl:48:    gen_server:start_link(?MODULE, State, []).
```

siplabs:
@fenollp but it's callback module here, not name

fenollp:
I know. This is also a consistency check. It will probably be removed later on but for now please just replace that last thing

siplabs:
@fenollp ok, got that! What macros name should I use? `SERVER` may cause misunderstanding... Probably `CALLBACK_MODULE`?

fenollp:
I know `SERVER` is not the best name since alone it doesn't register the name, but we're going with this for now

siplabs:
@fenollp understood.
I'll tell other our developers about changes.

PR #1350, opened 2015-10-05
Fix typos (recieved -> received)

Esya:
See [matching PR on kazoo-configs](https://github.com/2600hz/kazoo-configs/pull/62)


jamesaimonetti:
No shortage of these to find. @k-anderson especially thanks you :)

fenollp:
[Here are some more](https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings) ;)

PR #1352, opened 2015-10-05
Fix typo, underscore misspelled as any()

Esya:


jamesaimonetti:
@fenollp are you sure you're not @Esya ?

@Esya we decided to standardize on `any()` over `_` in specs, so this PR is not going to be accepted. Sorry.

PR #1353, opened 2015-10-07
KAZOO-4231 fix fax smtp processing of images

lazedo:


fenollp:
@lazedo Was this really related to https://2600hz.atlassian.net/browse/KAZOO-4231 ?
The ticket does not look like it is related to the issue…

lazedo:
@fenollp my mistake (it was late and was without glasses)

PR #1358, opened 2015-10-13
KAZOO-4241: Adding multi search

macpie:


applications/crossbar/src/modules/cb_search.erl
```diff
@@ -87,7 +105,29 @@ validate(Context) ->
           end,
     validate_search(Ctx, Type).
 
+
+validate(Context, ?MULTI) ->
+    Type = cb_context:req_value(Context, <<"t">>),
+    Ctx = case cb_context:account_id(Context) of
+              'undefined' -> cb_context:set_account_db(Context, ?WH_ACCOUNTS_DB);
```
fenollp: Shouldn't this be setting `account_id`?

applications/crossbar/src/modules/cb_search.erl
```diff
@@ -62,12 +67,24 @@ allowed_methods() ->
...
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
 -spec authorize(cb_context:context()) -> 'true'.
+-spec authorize(cb_context:context(), path_token()) -> 'true'.
 authorize(Context) ->
     cb_context:auth_account_id(Context) =/= 'undefined'.
 
+authorize(Context, ?MULTI) ->
+    cb_context:auth_account_id(Context) =/= 'undefined'.
```
fenollp: Is this really doing auth? (I don't know, genuinely asking)

applications/crossbar/src/modules/cb_search.erl
```diff
@@ -87,7 +105,29 @@ validate(Context) ->
...
+
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
 -spec validate_search(cb_context:context(), api_binary()) -> cb_context:context().
+-spec validate_search(cb_context:context(), ne_binary(), api_binary()) ->
+                             cb_context:context().
+-spec validate_search(cb_context:context(), ne_binary(), ne_binary(), binary() | 'undefined') ->
```
fenollp: `binary() | undefined` = `api_binary()`

applications/crossbar/src/modules/cb_search.erl
```diff
@@ -125,6 +164,86 @@ validate_search(Context, Type, Query) ->
...
+    end.
+
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec validate_multi(cb_context:context(), ne_binary()) -> cb_context:context().
+-spec validate_multi(cb_context:context(), ne_binary(), wh_proplist()) -> cb_context:context().
+validate_multi(Context, 'undefined') ->
```
fenollp: Spec is missing the undefined part

applications/crossbar/src/modules/cb_search.erl
macpie: does not matter at that point

applications/crossbar/src/modules/cb_search.erl
```diff
@@ -125,6 +164,86 @@ validate_search(Context, Type, Query) ->
...
+                ,Context
+            )
+    end.
+
+validate_multi(Context, Type, Props) ->
+    QueryOptions = query_options(cb_context:account_db(Context)),
+    IsValid =
+        lists:foldl(
+            fun({Query, _}, Acc) ->
+                validate_multi_foldl(Query, QueryOptions, Acc)
```
fenollp: I'd rather you write a recursive function that stops running at the first `Invalid`, instead of carrying the `Invalid` until the end of the list. This way it would return the same result, only sooner.

macpie: ok

applications/crossbar/doc/search.md
```diff
@@ -39,3 +39,35 @@ The Search API allows queries on databases.
     }
 
 
+### _GET_ - Multi Search
+
+    t = document type
+    view_name = value
```
jamesaimonetti: How are `view_name` values discovered?

applications/crossbar/doc/search.md
```diff
@@ -39,3 +39,35 @@ The Search API allows queries on databases.
...
+
+
+    curl -v -X GET -H "X-Auth-Token: {AUTH_TOKEN}" 'http://localhost:8000/v2/search?t=account&name=test&realm=test'
```
jamesaimonetti: Wait, is is `view_name` or `name`? Or where does `view_name` come into play?

applications/crossbar/src/modules/cb_search.erl
```diff
@@ -11,18 +11,19 @@
...
         ]).
 
 -include("../crossbar.hrl").
 
 -define(QUERY_TPL, <<"search/search_by_">>).
+-define(MULTI, <<"multi">>).
 
 -define(ACCOUNT_QUERY_OPTIONS, [<<"name">>, <<"number">>, <<"name_and_number">>]).
--define(ACCOUNTS_QUERY_OPTIONS, [<<"name">>]).
+-define(ACCOUNTS_QUERY_OPTIONS, [<<"name">>, <<"realm">>, <<"id">>]).
```
jamesaimonetti: These need to be discoverable somehow (same for `ACCOUNT_QUERY_OPTIONS`). Remember, Crossbar is a black box to the clients accessing it; they can't necessarily read the source/docs to figure it out.

applications/crossbar/src/modules/cb_search.erl
```diff
@@ -125,6 +164,86 @@ validate_search(Context, Type, Query) ->
...
+        ,wh_json:from_list([{<<"message">>, <<"Search needs a document type to search on">>}])
+        ,Context
+    );
+validate_multi(Context, Type) ->
+    QS = cb_context:query_string(Context),
+    case wh_json:to_proplist(wh_json:delete_key(<<"t">>, QS)) of
+        [_|_]=Props -> validate_multi(Context, Type, Props);
+        _Other ->
+            cb_context:add_validation_error(
+                <<"multi">>
```
jamesaimonetti: It seems like this could be a place to return an `enum` error instead of `required`, using the search query options macros from above to give the client an idea of the available query string parameters they could include.

applications/crossbar/src/modules/cb_search.erl
jamesaimonetti: I agree, the overloading of `IsValid` is confusing to follow. Also, you use `IsValid` below, but in the `validate_multi_foldl/3`, you use `InValid`. Watch your variable naming for consistency.

applications/crossbar/src/modules/cb_search.erl
```diff
@@ -125,6 +164,79 @@ validate_search(Context, Type, Query) ->
...
+
+validate_multi(Context, Type, Props) ->
+    QueryOptions = query_options(cb_context:account_db(Context)),
+    case validate_multi_fold(QueryOptions, Props) of
+        'ok' -> multi_search(Context, Type, Props);
+        Invalid ->
+            cb_context:add_validation_error(
+                <<"multi">>
+                ,<<"enum">>
+                ,wh_json:from_list([
```
jamesaimonetti: Opening the list should be done on a new line:

```erlang
wh_json:from_list(
    [{foo, var}
     ,{baz, 1}
    ]
)
```

applications/crossbar/src/modules/cb_search.erl
```diff
@@ -125,6 +164,79 @@ validate_search(Context, Type, Query) ->
...
+                    ,{<<"target">>, QueryOptions}
+                    ,{<<"cause">>, Invalid}
+                 ])
+                ,Context
+            )
+    end.
+
+-spec validate_multi_fold(wh_proplist(), wh_proplist()) -> 'ok' | ne_binary().
+validate_multi_fold(_Options, []) -> 'ok';
+validate_multi_fold(Options, [{Query, _}|Props]) ->
```
jamesaimonetti: Change `Options` to `Available` to match how `search` reads.

jamesaimonetti: In fact, you should probably just call into `validate_query/3` and check the resulting Context for error to see if you should end the loop

applications/crossbar/src/modules/cb_search.erl
```diff
@@ -209,6 +315,45 @@ search(Context, Type, Query, Value) ->
...
+multi_search(Context, _Type, [], Acc) ->
+    Env = cb_context:resp_envelope(Context),
+    cb_context:set_resp_envelope(
+        cb_context:set_resp_data(Context, Acc)
+        ,wh_json:delete_keys([<<"start_key">>, <<"page_size">>], Env)
+    );
+multi_search(Context, Type, [{Query, Value}|More], Acc) ->
+    ViewName = <<?QUERY_TPL/binary, Query/binary>>,
+    ViewOptions = [
+        {'startkey', get_start_key(Context, Type, Value)}
```
jamesaimonetti: I don't think we drop the first tuple down a line; should either be put back up a line or drop the `[` down to this line.

applications/crossbar/doc/search.md
macpie: I ll add it into the doc

applications/crossbar/src/modules/cb_search.erl
macpie: I ll add it into the doc

applications/crossbar/src/modules/cb_search.erl
```diff
@@ -125,27 +165,104 @@ validate_search(Context, Type, Query) ->
...
+            Context1
+    end.
+
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
 -spec validate_query(cb_context:context(), ne_binary()) -> cb_context:context().
 -spec validate_query(cb_context:context(), ne_binary(), ne_binaries()) -> cb_context:context().
```
fenollp: i think the last arg should be `wh_proplist()`

applications/crossbar/doc/search.md
```diff
@@ -39,3 +47,35 @@ The Search API allows queries on databases.
...
+
+
+```
+{
+  "data": {
+    "realm": [],
```
jamesaimonetti: Why are `realm` and `name` on the response if they aren't populated with anything? Shouldn't `data` just be the summary listing of the search results (similar to `GET /v2/accounts/{ACCOUNT_ID}/descendants` and the like?)

core/whistle_apps-1.0.0/priv/couchdb/views/search.json
```diff
@@ -4,6 +4,12 @@
     "views": {
         "search_by_name": {
             "map": "function(doc) { if (doc.pvt_deleted) return; for (i in doc.pvt_tree) { emit([doc.pvt_tree[i], doc.pvt_type, doc.name], {'id': doc._id, 'descendants_count': doc.descendants_count || 0, 'name': doc.name, 'realm' : doc.realm}); } }"
+        },
```
jamesaimonetti: How do you compile Crossbar? The Makefile should run a JSON formatter, which puts `search_by_name` below these two new views (it sorts the keys alphabetically). When I compile this branch, I have an unstaged change in this file from the JSON formatter running. You may want to check your build process to make sure the JSON formatter is successfully running (and commit the formatted `search.json` as evidence).

applications/crossbar/doc/search.md
macpie: cause the UI needs to know which search was successful

core/whistle_apps-1.0.0/priv/couchdb/views/search.json
macpie: I ran make & make json with no difference

fenollp:
@jamesaimonetti Could you update `/.travis.yml` so that build fails when there are unstaged changes?
Or… make a ticket and I'll look into it?

applications/crossbar/doc/search.md
jamesaimonetti: Why does the query above have `name==test&realm=test` but the response has empty arrays for `name` and `realm`? The code doesn't appear to put the results of the search into the right buckets.

applications/crossbar/src/modules/cb_search.erl
```diff
@@ -62,12 +68,24 @@ allowed_methods() ->
...
 
+resource_exists(?MULTI) -> 'true'.
+
+%%--------------------------------------------------------------------
+%% @public
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
 -spec authorize(cb_context:context()) -> 'true'.
+-spec authorize(cb_context:context(), path_token()) -> 'true'.
```
jamesaimonetti: Specs should be `boolean()`, no?

applications/crossbar/src/modules/cb_search.erl
```diff
@@ -125,27 +165,104 @@ validate_search(Context, Type, Query) ->
...
         'false' ->
             cb_context:add_validation_error(
-              <<"q">>
-              ,<<"enum">>
-              ,wh_json:from_list(
-                 [{<<"message">>, <<"Value not found in enumerated list of values">>}
-                  ,{<<"target">>, Available}
+                <<"query">>
+                ,<<"enum">>
+                ,wh_json:from_list([
```
jamesaimonetti: Opening a list should be done on a new line, as the code was formatted before.

jamesaimonetti: Also, if the query string param is `q`, the validation error should be on `q`. If `q` has been changed to `query`, please make sure docs reflect that.

PR #1364, opened 2015-10-15
KAZOO-4011: add save_after_notify flag for voicemail that saves messa…

mark2600:
…ge instead of deleting after notifications are sent

### DIFF
--- a/applications/callflow/src/module/cf_voicemail.erl
+++ b/applications/callflow/src/module/cf_voicemail.erl
@@ -38,6 +38,7 @@
 -define(KEY_EXTENSION, <<"extension">>).
 -define(KEY_MAX_PIN_LENGTH, <<"max_pin_length">>).
 -define(KEY_DELETE_AFTER_NOTIFY, <<"delete_after_notify">>).
+-define(KEY_SAVE_AFTER_NOTIFY, <<"save_after_notify">>).
 
 -define(MAILBOX_DEFAULT_SIZE
         ,whapps_config:get_integer(?CF_CONFIG_CAT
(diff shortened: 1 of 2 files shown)


applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -151,7 +159,7 @@
           ,keys = #keys{} :: vm_keys()
           ,transcribe_voicemail = 'false' :: boolean()
           ,notifications :: wh_json:object()
-          ,delete_after_notify = 'false' :: boolean()
+          ,after_notify_action = 'nothing' :: atom('nothing' | 'delete' | 'save')
```
fenollp: Wrapping inside `atom/1` is superfluous

applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -1390,18 +1398,26 @@ get_caller_id_number(Call) ->
...
-            {'ok', _} = couch_mgr:del_doc(whapps_call:account_db(Call), MediaId),
-            lager:debug("attachment was sent out via notification, deleted media file");
+            case Action of
+                'delete' ->
+                    {'ok', _} = couch_mgr:del_doc(whapps_call:account_db(Call), MediaId),
+                    lager:debug("attachment was sent out via notification, deleted media file");
+                'save' ->
+                    update_folder(?FOLDER_SAVED, MediaId, Box, Call),
+                    lager:debug("attachment was sent out via notification, saved media file")
+            end;
```
fenollp: I'd personally put the log lines before the code, using ING instead of ED. So that log-reading people get an idea of what was going on.
What do you think?

mark2600: I can get onboard with that

PR #1365, opened 2015-10-15
KAZOO-4177: Use account threshold instead of topup threshold

macpie:


applications/notify/src/notify_account_crawler.erl
```diff
@@ -400,10 +400,13 @@ notify_low_balance(CurrentBalance, AccountId, AccountDb, JObj) ->
 low_balance_threshold(Account) ->
     AccountId = wh_util:format_account_id(Account, 'raw'),
     AccountDb = wh_util:format_account_id(Account, 'encoded'),
+
     ConfigCat = <<(?NOTIFY_CONFIG_CAT)/binary, ".low_balance">>,
     Default = whapps_config:get_float(ConfigCat, <<"threshold">>, 5.00),
+
     case couch_mgr:open_doc(AccountDb, AccountId) of
```
fenollp: Can't that be `kz_account:fetch(Account)` ?

PR #1367, opened 2015-10-16
Rename testdata file of gen_smtp that is incompatible with windows.

cando-1p:
It doesn't seem to be referenced anywhere.
File was dkim-rsa-private-encrypted:12345.pem. The : is the incompatable char.
Changed to dkim-rsa-private-encrypted--12345.pem


fenollp:
Is there no code that needs to be changed in order to reflect this name change?

jamesaimonetti:
Seems to be referenced only in `mimemail`'s eunit tests.

PR #1370, opened 2015-10-17
KAZOO-4292: blackhole clear_subscriptions event

onnet:
As a reseller I would like to monitor accounts' calls with blackhole. 
When I am choosing one account after another I need to remove bindings used to listen for previous account's events

### DIFF
--- a/applications/blackhole/doc/index.md
+++ b/applications/blackhole/doc/index.md
@@ -33,10 +33,10 @@ From here, you can write your own Javascript callbacks, triggered everytime a re
   <body>
     <script>
       var socket = io.connect('http://{BLACKHOLE_IP_ADDRESS}:5555');
-      socket.emit('subscribe', { account_id: '{ACCOUNT_ID}', auth_token: {AUTH_TOKEN}, binding: 'call.CHANNEL_CREATE.*' });
-      socket.emit('subscribe', { account_id: '{ACCOUNT_ID}', auth_token: {AUTH_TOKEN}, binding: 'call.CHANNEL_ANSWER.*' });
-      socket.emit('subscribe', { account_id: '{ACCOUNT_ID}', auth_token: {AUTH_TOKEN}, binding: 'call.CHANNEL_DESTROY.*' });
-      socket.emit('subscribe', { account_id: '{ACCOUNT_ID}', auth_token: {AUTH_TOKEN}, binding: 'conference.event.*' });
+      socket.emit('subscribe', { account_id: '{ACCOUNT_ID}', auth_token: '{AUTH_TOKEN}', binding: 'call.CHANNEL_CREATE.*' });
+      socket.emit('subscribe', { account_id: '{ACCOUNT_ID}', auth_token: '{AUTH_TOKEN}', binding: 'call.CHANNEL_ANSWER.*' });
+      socket.emit('subscribe', { account_id: '{ACCOUNT_ID}', auth_token: '{AUTH_TOKEN}', binding: 'call.CHANNEL_DESTROY.*' });
+      socket.emit('subscribe', { account_id: '{ACCOUNT_ID}', auth_token: '{AUTH_TOKEN}', binding: 'conference.event.*' });
 
       socket.on('participants_event', function (data) {
         console.log(data);
--- a/applications/blackhole/doc/index.md
+++ b/applications/blackhole/doc/index.md
@@ -60,7 +60,23 @@ From here, you can write your own Javascript callbacks, triggered everytime a re
     </script>
   </body>
 </html>
+
+
 ```
+To remove unnecessary bindings use 'unsubscribe' event:
+For particular subscription:
+```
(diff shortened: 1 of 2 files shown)


applications/blackhole/src/blackhole_socket_callback.erl
```diff
@@ -39,6 +39,12 @@ recv(_SessionPid, _SessionId, {'event', _Ignore, <<"subscribe">>, SubscriptionJO
     end,
     {'ok', Context1};
 
+recv(SessionPid, SessionId, {'event', _Ignore, <<"clear_subscriptions">>, _Data}, Context) ->
```
lazedo: @onnet i like this, how about changing it to "unsubscribe" and process the _Data object sent ?
the _Data object can have unsubscribe info for account_id, listof account_id, all (or empty | null _Data).
thoughts ?

onnet: @lazedo I was thinking about 'unsubscribe' of particular 'subscription' as opposite to 'clear_all' ('clear_subscriptions'). In the other words as a separate case . But if you think it is better to combine it, l'm all for.

onnet:
@lazedo "listof account_id" feels like an overkill at the moment because I just can't imagine any case it could be needed. Let's wait a bit till it could prove its necessity. At the end of the day it is just a couple of strings which could be added quickly enough..

applications/blackhole/doc/index.md
```diff
@@ -60,7 +60,24 @@ From here, you can write your own Javascript callbacks, triggered everytime a re
...
   </body>
 </html>
+
+
+```
+To remove unnecessary bindings use 'unsubscribe' event:
+
+For particular subscription:
+```
+socket.emit('unsubscribe', { account_id: '4b8c6fec4b2597882c0390202d195419', auth_token: AuthToken, binding: 'call.CHANNEL_CREATE.*'});
```
jamesaimonetti: Can you use `{ACCOUNT_ID}` and `{AUTH_TOKEN}` as placeholders here?

applications/blackhole/src/blackhole_socket_callback.erl
```diff
@@ -39,6 +39,29 @@ recv(_SessionPid, _SessionId, {'event', _Ignore, <<"subscribe">>, SubscriptionJO
     end,
     {'ok', Context1};
 
+recv(SessionPid, SessionId, {'event', _Ignore, <<"unsubscribe">>, {[]}}, Context) ->
```
jamesaimonetti: Use the macro `?EMPTY_JSON_OBJECT` here instead of `{[]}`. You may have to add `-include_lib("whistle/src/wh_json.hrl").` at the top of the file.

applications/blackhole/src/blackhole_socket_callback.erl
```diff
@@ -39,6 +39,29 @@ recv(_SessionPid, _SessionId, {'event', _Ignore, <<"subscribe">>, SubscriptionJO
...
+    lager:debug("remove all bindings for session: ~p", [SessionId]),
+    Filter = fun (_1, _2, _3, _4) -> filter_bindings(SessionPid, _1, _2, _3, _4) end,
```
jamesaimonetti: The underscore suggests these variables will be unused. Since you are obviously using them, use real variable names. `fun(A, B, C, D) -> filter_bindings(SessionPid, A, B, C, D) end` would be fine.

applications/blackhole/src/blackhole_socket_callback.erl
```diff
@@ -39,6 +39,29 @@ recv(_SessionPid, _SessionId, {'event', _Ignore, <<"subscribe">>, SubscriptionJO
...
+    blackhole_bindings:filter(Filter),
+    {'ok', Context};
+
+recv(_SessionPid, SessionId, {'event', _Ignore, <<"unsubscribe">>, SubscriptionJObj}, Context) ->
+    lager:debug("Remove binding for session: ~p. Data: ~p", [SessionId, SubscriptionJObj]),
```
jamesaimonetti: Log lines should not start with a capital letter.

applications/blackhole/src/blackhole_socket_callback.erl
```diff
@@ -39,6 +39,29 @@ recv(_SessionPid, _SessionId, {'event', _Ignore, <<"subscribe">>, SubscriptionJO
     end,
     {'ok', Context1};
 
+recv(SessionPid, SessionId, {'event', _Ignore, <<"unsubscribe">>, {[]}}, Context) ->
+    lager:debug("remove all bindings for session: ~p", [SessionId]),
+    Filter = fun (_1, _2, _3, _4) -> filter_bindings(SessionPid, _1, _2, _3, _4) end,
+    blackhole_bindings:filter(Filter),
```
jamesaimonetti: Should there be a check for `blackhole_util:is_authorized/1` before filtering?

jamesaimonetti:
@onnet lgtm!

onnet:
@jamesaimonetti @lazedo Many thanks for help! Just one more question. I see that 3.22 branch  appeared a couple of days ago. Isn't it too late to ask to chery pick this pr to it. Just afraid that 3.23 is way too long to wait for :)

PR #1371, opened 2015-10-19
Restore work allotments/consumed

Dinkor:
After 64ac39b01c3a307ce1ebc9da76981ddda6fbac15 always get 500 error on `allotments/consumed`


jamesaimonetti:
@Dinkor Would it work to call `cb_context:set_resp_status(Context, 'success')` to initialize the accumulator of the fold?

k-anderson:
@jamesaimonetti actually I think this is the best approach, the context is not successful unless the fold completes (IE: the validate function is dependent on the fold result and if it fails the context is left as non-successful).

PR #1372, opened 2015-10-19
Kazoo 4294 - authz authn-by-ip devices

jamesaimonetti:


core/whistle_apps-1.0.0/src/whapps_util.erl
```diff
@@ -288,6 +291,47 @@ get_account_by_realm(RawRealm) ->
     Realm = wh_util:to_lower_binary(RawRealm),
     get_accounts_by(Realm, ?ACCT_BY_REALM_CACHE(Realm), ?AGG_LIST_BY_REALM).
 
+-spec get_account_by_ip(ne_binary()) ->
+                               {'ok', wh_proplist()} |
+                               {'error', 'not_found'}.
+get_account_by_ip(IP) ->
```
k-anderson: @jamesaimonetti  I would argue that get_account_by_ip should return the matching document then the modules convert to CCVs.  Or maybe call it get_ccvs_by_ip?

jamesaimonetti:
@k-anderson refactored a bit, other apps using the ip lookup view now share in the sweet cached goodness.

core/whistle_apps-1.0.0/src/whapps_util.erl
```diff
@@ -288,6 +291,54 @@ get_account_by_realm(RawRealm) ->
...
+    end.
+
+-spec do_get_ccvs_by_ip(ne_binary()) ->
+                               {'ok', wh_proplist()} |
+                               {'error', 'not_found'}.
+do_get_ccvs_by_ip(IP) ->
+    case couch_mgr:get_results(?WH_SIP_DB, ?AGG_LIST_BY_IP, [{'key', IP}]) of
+        {'ok', []} ->
+            lager:debug("no entry in ~s for IP: ~s", [?WH_SIP_DB, IP]),
+            {'error', 'not_found'};
```
k-anderson: @jamesaimonetti lets cache the not_found as well please, otherwise calls from the carriers with authz enabled will always attempt a lookup.

jamesaimonetti:
@k-anderson updated

PR #1375, opened 2015-10-20
KAZOO-4069: Add macros to account notification templates

macpie:


applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -298,6 +298,23 @@ do_post(Context) ->
...
+set_system_macros(Context, 'undefined') -> Context;
+set_system_macros(Context, Id) ->
+    SysContext = read_system(Context, Id),
+    SysDoc = cb_context:doc(SysContext),
+    Macros = wh_json:get_value(?MACROS, SysDoc, wh_json:new()),
+    Updater =
+        fun(JObj) ->
+            wh_json:set_value(?MACROS, Macros, JObj)
+        end,
+    cb_context:update_doc(Context, Updater).
```
jamesaimonetti: Why not just call `cb_context:set_doc(Context, wh_json:set_value(?MACROS, Macros, JObj))` directly?

applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -298,6 +298,23 @@ do_post(Context) ->
         _Status -> Context1
     end.
 
+-spec set_system_macros(cb_context:context()) -> cb_context:context().
+-spec set_system_macros(cb_context:context(), api_binary()) -> cb_context:context().
+set_system_macros(Context) ->
+    Id = wh_doc:id(cb_context:doc(Context)),
+    set_system_macros(Context, Id).
+
+set_system_macros(Context, 'undefined') -> Context;
```
jamesaimonetti: This seems unnecessary, no? When would the doc id be missing? Would we really want to continue processing a POST with no id?

macpie: Just wanted to be defensive, want me to remove?

jamesaimonetti: In this case, yes. If there's no id, processing should not progress.

applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -298,6 +298,23 @@ do_post(Context) ->
...
+set_system_macros(Context, Id) ->
+    SysContext = read_system(Context, Id),
```
jamesaimonetti: Should probably check that `read_system/2` resulted in success too.

PR #1378, opened 2015-10-22
KAZOO-4303: cf_endpoint: fixed getting account document

siplabs:


lazedo:
@siplabs following up the call we had today, i believe that you are trying to solve an issue caused by an endpoint not having the pvt_account_id field. can you post your endpoint JObj and determine why it doesn't have this field ?

siplabs:
It's user document.
So it goes to
```erlang
merge_attributes(Keys, 'undefined', 'undefined', Doc).
```
After that it sets `Endpoint` to `wh_json:new()` and cannot get `account_id/db` from empty json document.

In this patch account document is loaded and passed to `merge_attributes/4` without trying to guess "where I can find account_db"?

siplabs:
@lazedo 
Can you take a look in `cb_clicktocall`? In click-to-call case CID also will not be set up propertly.
It provides document with type `clicktocall`... So it can't get to `merge_attributes/4`.

lazedo:
@siplabs fixed with commit https://github.com/2600hz/kazoo/commit/4e3c370276d84e7a27e580444ad6a473f51910db 
for `cb_clicktocall` check PR https://github.com/2600hz/kazoo/pull/1381

PR #1379, opened 2015-10-23
KAZOO-4228 send MWI notifications to vmbox number

lazedo:


applications/callflow/src/cf_util.erl
```diff
@@ -997,3 +1010,40 @@ account_timezone(Call) ->
 start_task(Fun, Args, Call) ->
     SpawnInfo = {'cf_task', [Fun, Args]},
     cf_exe:add_event_listener(Call, SpawnInfo).
+
+-spec mailbox(ne_binary(), ne_binary()) -> {'ok', ne_binaries()} |
```
jamesaimonetti: returns `{'ok', wh_json:object()}`, doesn't it?

applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -2313,3 +2294,25 @@ is_owner(Call, OwnerId) ->
         OwnerId -> 'true';
         _Else -> 'false'
     end.
+
+-spec send_mwi_update(mailbox(), whapps_call:call()) -> 'ok'.
+send_mwi_update(#mailbox{owner_id=OwnerId, mailbox_number=BoxNumber}=Box, Call) ->
+    AccountDb = whapps_call:account_db(Call),
+    _ = wh_util:spawn('cf_util', 'unsolicited_owner_mwi_update', [AccountDb, OwnerId]),
+    Messages = get_messages(Box, Call),
+    New = count_messages(Messages, ?FOLDER_NEW),
```
jamesaimonetti: I wonder if the `?FOLDER_*` macros should go somewhere else so `cf_util:vm_count/1` can use them instead of raw binaries? Maybe a `kzd_voicemail_box.hrl` in `core/kazoo_documents/include/` and add the appropriate `-include_lib()` in the modules?

core/whistle-1.0.0/src/api/wapi_presence.erl
```diff
@@ -108,19 +108,19 @@
...
-                                   ,<<"Messages-Urgent-Waiting">>
+                                   ,<<"Messages-Urgent-Saved">>
                                    ,<<"Call-ID">>
                                   ]).
 -define(MWI_REQ_VALUES, [{<<"Event-Category">>, <<"presence">>}
                          ,{<<"Event-Name">>, <<"mwi_update">>}
                         ]).
 -define(MWI_REQ_TYPES, [{<<"Messages-New">>, fun(I) -> is_integer(wh_util:to_integer(I)) end}
-                        ,{<<"Messages-Waiting">>, fun(I) -> is_integer(wh_util:to_integer(I)) end}
+                        ,{<<"Messages-Saved">>, fun(I) -> is_integer(wh_util:to_integer(I)) end}
```
jamesaimonetti: Personally, I think casting to integer first should be removed. The caller should put integers there! I know it was there already, but reading it now and seeing your changes elsewhere, I think just a straight `{<<"Messages-Saved">>, fun is_integer/1}` is sufficient. Thoughts?

PR #1381, opened 2015-10-23
KAZOO-4305 create a map to help merge_attributes for other pvt_type

lazedo:


applications/callflow/src/cf_endpoint.erl
```diff
@@ -116,6 +118,14 @@ maybe_have_endpoint(JObj, EndpointId, AccountDb) ->
             has_endpoint(JObj, EndpointId, AccountDb, EndpointType)
     end.
 
+-spec maybe_map_endpoint_type(api_binary()) -> api_binary().
+maybe_map_endpoint_type('undefined') -> 'undefined';
+maybe_map_endpoint_type(EndpointType) ->
```
jamesaimonetti: Was there a reason why you chose a proplist versus using pattern matching in the function clauses?

```erlang
maybe_map('undefined') -> 'undefined';
maybe_map(<<"clicktocall">>) -> kz_device:type();
maybe_map(Type) -> Type.
```

lazedo: @jamesaimonetti no particular reason. this way allows to move it to whapps_account_config/whapps_config if there is a request to support it by cluster / account_id.

jamesaimonetti: Fine by me. Can you update the binary types to use the document functions (`kz_device:type()`, `kzd_user:type()`, etc...) where available?

lazedo:
@jamesaimonetti i may have jumped into this too soon. let me check if clicktocall is actually supposed to be merged.

jamesaimonetti:
@lazedo Should we close this?

lazedo:
@jamesaimonetti thanks for the reminder :)

fenollp:
@lazedo Should we delete the branch too?

lazedo:
@fenollp thanks for the reminder

PR #1382, opened 2015-10-23
KAZOO-4247 basic auth

lazedo:


applications/crossbar/src/api_util.erl
```diff
@@ -715,24 +715,60 @@ prefer_new_context([{'halt', Context1}|_], Req, _) ->
...
+                            {cowboy_req:req(), cb_context:context()}.
+get_authorization_token(Req0, Context) ->
+    case cowboy_req:header(<<"authorization">>, Req0) of
+        {'undefined', Req1} ->
+            case cb_context:req_value(Context, <<"authorization">>) of
+                'undefined' ->
+                    lager:debug("no auth token found"),
+                    {Req1, Context};
+                Authorization ->
+                    {Token, TokenType} = get_authorization_token_type(Authorization),
```
jamesaimonetti: This clause should be its own function since its duplicated in the parent case block.

applications/crossbar/src/modules/cb_basic_auth.erl
```diff
@@ -0,0 +1,169 @@
...
+        'undefined' -> 'false';
+        JObj -> is_expired(Context, JObj)
+    end;
+check_credentials(Context, AccountId, Credentials, BasicType) ->
+    case binary:split(Credentials, <<":">>) of
+        [User, Pass] -> check_credentials(Context, AccountId, {User, Pass}, BasicType);
+        _ -> 'false'
+    end.
+
+get_credential_doc(AccountId, View, Key) ->
```
jamesaimonetti: spec

applications/crossbar/doc/basic_auth.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Password
+Should be set to md5sum of username:password
+ ```
+PASSWORD=`echo -n username:password | md5sum | awk '{print $1}'`
+```
+
+## Sample cURL Requests
+
+    curl -v -basic -user {ACCOUNT_ID}:$PASSWORD http://server.com:8000/v2/accounts/{ACCOUNT_ID}/devices
```
jamesaimonetti: We should probably clarify that the first `{ACCOUNT_ID}` is the auth account id, while the second `{ACCOUNT_ID}` is the account to perform the action against.

applications/crossbar/src/modules/cb_basic_auth.erl
```diff
@@ -0,0 +1,170 @@
...
+
+-spec check_credentials(cb_context:context(), ne_binary(), api_binary()) ->
+                              'false' |
+                              {'true' | 'halt', cb_context:context()}.
+check_credentials(Context, AccountId, Credentials) ->
+    lager:debug("checking credentials '~s' for account '~s'", [Credentials, AccountId]),
+    BasicType = whapps_account_config:get(AccountId, ?MOD_CONFIG_CAT, ?BASIC_AUTH_KEY, ?BASIC_AUTH_TYPE),
+    check_credentials(Context, AccountId, Credentials, BasicType).
+
+-spec check_credentials(cb_context:cb_context(), ne_binary(), ne_binary() | {ne_binary(), ne_binary()}, ne_binary()) ->
```
jamesaimonetti: Type should be `cb_context:context()`

jamesaimonetti: I'l fix this

PR #1383, opened 2015-10-23
KAZOO-4301: Fix pb is \

macpie:


jamesaimonetti:
@macpie Does using `cow_qs:urlencode/1` work properly for wherever you're seeing this issue? I'm trying to remove the mochiweb dep and `mochiweb_util` is really the last of the modules that we seem to use from it.

macpie:
Just tried it, does not work. It seems like it is an erlang issue.

PR #1384, opened 2015-10-23
KAZOO-4285: strip strange characters from quickcall numbers, also opt…

mark2600:
…ionally allow regex capture to format

### DIFF
--- a/applications/crossbar/src/modules/cb_modules_util.erl
+++ b/applications/crossbar/src/modules/cb_modules_util.erl
@@ -33,6 +33,8 @@
 
 -include("../crossbar.hrl").
 
+-define(QCALL_NUMBER_FILTER, [<<" ">>, <<",">>, <<".">>, <<"-">>, <<"(">>, <<")">>]).
+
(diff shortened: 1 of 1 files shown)


applications/crossbar/src/modules/cb_modules_util.erl
```diff
@@ -235,10 +237,31 @@ request_specific_extraction_funs_from_nouns(Context, ?USERS_QCALL_NOUNS(UserId,
...
+    end.
+
 -spec build_number_uri(cb_context:context(), ne_binary()) -> ne_binary().
 build_number_uri(Context, Number) ->
+    QueryStr  = cb_context:query_string(Context),
+    FilterVal = wh_json:get_value(<<"number_filter">>, QueryStr, <<"true">>),
+
+    UseNumber = case FilterVal of
+        <<"false">> -> Number;
+        <<"true">>  -> binary:replace(Number, ?QCALL_NUMBER_FILTER, <<"">>, [global]);
```
jamesaimonetti: `<<>>` and tickie global.

PR #1390, opened 2015-10-28
KAZOO-4224 allow cf_park to continue when callback is not answered

lazedo:
define children with SlotNumber to allow callflow continuation per slot
or define wildcard "_" to allow for all slots


jamesaimonetti:
@lazedo I moved the `next/2` up to be with the other API functions and removed the case clause to just reply directly with the result of the call to `wh_json`. Otherwise all looked good to me!

lazedo:
@jamesaimonetti perfect! Thanks

PR #1391, opened 2015-10-29
KAZOO-4309: recording on call forward

onnet:
CHANNEL_REPLACED + small corrections

### DIFF
--- a/core/whistle_media-1.0.0/src/wh_media_recording.erl
+++ b/core/whistle_media-1.0.0/src/wh_media_recording.erl
@@ -69,6 +69,7 @@
                                      ,{'restrict_to', ['CHANNEL_ANSWER', 'CHANNEL_DESTROY'
                                                        ,'CHANNEL_BRIDGE', 'CHANNEL_EXECUTE_COMPLETE'
                                                        ,'RECORD_START', 'RECORD_STOP'
+                                                       ,'CHANNEL_REPLACED'
                                                       ]}
                                     ]}
                            ,{'self', []}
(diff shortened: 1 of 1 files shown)


core/whistle_media-1.0.0/src/wh_media_recording.erl
```diff
@@ -355,7 +366,7 @@ handle_cast({'gen_listener',{'is_consuming', 'true'}}, #state{record_on_answer='
                                                              }=State) ->
     start_recording(Call, MediaName, TimeLimit, <<"wh_media_recording">>, SampleRate, RecordMinSec),
     lager:debug("started the recording"),
-    {'noreply', State};
+    {'noreply', State#state{is_recording='true'}};
```
jamesaimonetti: Actually, this just popped into my thought after I merged - we've only send the command to start recording here but haven't yet received the `RECORD_START` event, so why this setting of `is_recording` to `true`? Shouldn't the server wait for `RECORD_START`?

onnet: @jamesaimonetti 
My settings are:
- "record_on_answer": false;
-  "store_recordings": true,
-  "third_party_bigcouch_host": "127.0.0.1",
-  "third_party_bigcouch_port": 5984

Without changing "is_recording" to 'true' there was even no attempts to save the call.
In the other words the call was recorderd but left in /tmp untouched.
I am talking about all calls, not only forwarded/transferred.
So if it is way too early to set "is_recording" to 'true' I need to try to find out why we do not see RECORD_START later. 
Will do..

PR #1392, opened 2015-10-29
Added webhook for CHANNEL_BRIDGE

tux87:
Hi,

I created a webhook for the CHANNEL_BRIDGE callevt, and customized some variables in the webhook in order to track call transfers (to see what the original called number was).

other_leg_destination_number: shows the destination of a call transfer
original_number: shows the number on the other end of the bridge


jamesaimonetti:
I fixed the whitespace issues and a couple small formatting issues; merged manually. Thanks!!

PR #1393, opened 2015-10-29
Kazoo websockets

macpie:


applications/crossbar/src/modules/cb_websockets.erl
```diff
@@ -0,0 +1,243 @@
...
+         ,validate/1, validate/2
+        ]).
+
+-include("../crossbar.hrl").
+
+-define(CB_LIST, <<"websockets/crossbar_listing">>).
+-define(
+    AVAILABLE
+    ,wh_json:from_list([
+        {<<"call">>, [<<"call.CHANNEL_CREATE.*">>, <<"call.CHANNEL_ANSWER.*">>, <<"call.CHANNEL_DESTROY.*">>]}
```
lazedo: what about fax status ?

macpie: yep missed that thank you

macpie:
@lazedo @jamesaimonetti  any idea why?

blackhole_tracking.erl:178: Record construction #bh_context{auth_token::'_',auth_account_id::'_',bindings::'_',websocket_session_id::'_',websocket_pid::'_',req_id::'_',timestamp::'_',name::'_',metadata::'_',destination::'_',source::'_'} violates the declared type of field auth_token::'undefined' | binary() and auth_account_id::'undefined' | binary() and account_id::'undefined' | binary() and bindings::[<<_:8,_:_*8>>] and websocket_session_id::'undefined' | binary() and websocket_pid::'undefined' | pid() and req_id::<<_:8,_:_*8>> and timestamp::'undefined' | pos_integer() and name::'undefined' | binary() and destination::'undefined' | binary() and source::'undefined' | binary()

fenollp:
@macpie `#bh_context{}`'s record definition needs to be allowed to take `'_'`.

applications/crossbar/src/modules/cb_websockets.erl
```diff
@@ -0,0 +1,244 @@
...
+         ,validate/1, validate/2
+        ]).
+
+-include("../crossbar.hrl").
+
+-define(CB_LIST, <<"websockets/crossbar_listing">>).
+-define(
+    AVAILABLE
+    ,wh_json:from_list([
+        {<<"call">>, [<<"call.CHANNEL_CREATE.*">>, <<"call.CHANNEL_ANSWER.*">>, <<"call.CHANNEL_DESTROY.*">>, <<"call.RECORD_STOP.*">>]}
```
lazedo: <<"call.RECORD_STOP.*">> is this actually supported ?
also, there was a recent addition to wh_hooks_listener for CHANNEL_BRIDGE, which gets supported here automagically.

macpie: Adding it

lazedo:
@macpie did you test the RECORD_STOP event  binding ? i believe it doesn't work since call events are binded to wh_hooks_listener, or i'm missing something here.

macpie:
This was in the old doc so I just assumed it should work. Nevertheless it is not required right now, so I can just remove it.

PR #1398, opened 2015-11-02
KAZOO-4166: Removed unnecessary am/pm from notification templates

siplabs:


jamesaimonetti:
The `\\a\\t` was supposed to render as "at" in the date string. It is possible when upgrading erlydtl the last time that `\a\t` is all that is necessary now. I'd prefer to change the template to render "at" properly than remove it.

PR #1399, opened 2015-11-02
Update routing.md

kjcsb1:
sup jonny5_maintenance limits_summary doesn't accept any arguments


fenollp:
@kjcsb1 It does, though. https://github.com/2600hz/kazoo/blob/master/applications/jonny5/src/jonny5_maintenance.erl#L148

PR #1401, opened 2015-11-02
KAZOO-4200: Websockets doc

macpie:


applications/crossbar/doc/websockets.md
```diff
@@ -0,0 +1,83 @@
+/*
+Section: Crossbar
+Title: Websockets
+Language: en-US
+*/
```
jamesaimonetti: Can you add a `Version` header so readers know what version this doc is current for?

PR #1405, opened 2015-11-03
KAZOO-4166 - corrected fix

siplabs:


jamesaimonetti:
```erlang
(foo@bar.com)1> notify_util:render_template(<<"{{now|date:\"l, F j, Y \\\\a\\\\t H:i\"}}">>, vm_to_email, [{<<"now">>, calendar:universal_time()}]).
{ok,"Tuesday, November 3, 2015 at 20:08"}
(foo@bar.com)2>
```
It seems we have to escape the escapes to get "at" to render properly!

PR #1406, opened 2015-11-03
KAZOO-4328: Add account expired check

macpie:


core/whistle-1.0.0/src/wh_util.erl
```diff
@@ -21,7 +21,7 @@
          ,is_system_db/1
         ]).
 -export([get_account_realm/1, get_account_realm/2]).
--export([is_account_enabled/1]).
+-export([is_account_enabled/1, maybe_disable_account/1]).
```
jamesaimonetti: I feel like we should start moving these functions out of `wh_util` and into `kz_account`

core/whistle-1.0.0/src/wh_util.erl
```diff
@@ -420,6 +420,37 @@ is_account_expired(Account) ->
...
+
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec disable_account(ne_binary()) -> 'ok'.
+disable_account(Account) ->
+    AccountId = wh_util:format_account_id(Account, 'raw'),
+    case crossbar_maintenance:disable_account(AccountId) of
```
jamesaimonetti: This will need to be moved from `crossbar_maintenance` and `crossbar_maintenance` should call into the module that gets the code.

core/whistle-1.0.0/src/wh_util.erl
macpie: ok

core/whistle-1.0.0/src/wh_util.erl
```diff
@@ -21,8 +21,13 @@
          ,is_system_db/1
         ]).
 -export([get_account_realm/1, get_account_realm/2]).
--export([is_account_enabled/1]).
--export([is_account_expired/1]).
+-export([is_account_enabled/1, is_account_expired/1]).
+-export([maybe_disable_account/1
```
jamesaimonetti: All of these account document-related functions should be in `kz_account`; that was my point in previous comments. Since they all operate based on the account doc, they should be moved there instead of here. Slowly trying to clean up `wh_util`

core/whistle-1.0.0/src/wh_util.erl
```diff
@@ -408,13 +411,96 @@ is_account_expired(Account) ->
...
+            ({'ok', J}) ->
+                {'ok', Fun(J)}
+         end
+        ,fun({'error', _}=E) -> E;
+            ({'ok', J}) ->
+                AccountDb = wh_util:format_account_id(Account, 'encoded'),
+                couch_mgr:save_doc(AccountDb, J)
+        end
+        ,fun({'error', _}=E) -> E;
+            ({'ok', J}) ->
```
jamesaimonetti: `J` here should have the new revision; no need to do the `lookup_doc_rev` lower down

core/whistle-1.0.0/src/wh_util.erl
```diff
@@ -408,13 +411,96 @@ is_account_expired(Account) ->
...
+                AccountId = kz_account:id(J),
+                case couch_mgr:lookup_doc_rev(?WH_ACCOUNTS_DB, AccountId) of
+                    {'ok', Rev} ->
+                        couch_mgr:ensure_saved(?WH_ACCOUNTS_DB, wh_doc:set_revision(J, Rev));
+                    _Else ->
+                        couch_mgr:ensure_saved(?WH_ACCOUNTS_DB, wh_doc:delete_revision(J))
```
jamesaimonetti: This should never work anyway; when you delete the revision, it will try to create the document...but in the previous function you saved the document so there will definitely be a revision needed. I really don't see the point of this function at all.

core/whistle-1.0.0/src/wh_util.erl
macpie: kz_account only manipulate the doc but does not save it

core/whistle-1.0.0/src/wh_util.erl
macpie: this was there before in crossbar maintenance. I'll remove it

jamesaimonetti: I might write this as:

```erlang
account_update(Account, UpdateFun) ->
    case kz_account:fetch(Account) of
        {'error', _R}=E -> E;
        {'ok', AccountJObj} ->
            account_update(UpdateFun(AccountJObj))
    end.

account_update(UpdatedJObj) ->
    AccountDb = wh_doc:account_db(UpdatedJObj),
    case couch_mgr:save_doc(AccountDb, UpdatedJObj) of
        {'error', _R}=E -> E;
        {'ok', _SavedJObj} -> 'ok'
    end.
```

What the function is trying to do is apply the `UpdateFun` to the account document and save it, returning any errors or 'ok' on success. I don't think returning 'failed' is appropriate; the caller should have some indication as to why it failed.

core/kazoo_documents-1.0.0/src/kz_account.erl
```diff
@@ -52,17 +51,51 @@
 -type doc() :: wh_json:object().
 -export_type([doc/0]).
 
+%%--------------------------------------------------------------------
```
jamesaimonetti: If you aren't going to actually put docs in, why include these header comments? They serve no purpose otherwise (except to clutter the code), and any editor worth its salt should allow you to jump to a function or jump between functions easily, so I don't think they're needed visually.

core/whistle-1.0.0/src/wh_util.erl
jamesaimonetti: I think we should move save operations there, especially for account docs, since they need to be persisted to the account db and the `accounts` db. Good to have it centrally located.

I would add a save/1 to kz_account, and add an optional third param to the set functions of whether to save the set (defaulting to false). It should return the updated account document (including new _rev) or return the error tuple.

Something like:

```erlang
set_superduper_admin(AccountJObj, IsAdmin) ->
    set_superduper_admin(AccountJObj, IsAdmin, 'false').
set_superduper_admin(AccountJObj, IsAdmin, ShouldSave) ->
    Updated = wh_json:set_value(...),
    maybe_persist(Updated, ShouldSave).

maybe_persist(JObj, 'false') -> JObj;
maybe_persist(JObj, 'true') ->
    case save(JObj) of
        {'ok', Saved} -> Saved;
        {'error', _R}=E -> E
    end.
```

Seems like a good pattern to follow for the setters.

macpie: When I talked with Karl I understood that he wanted to avoid that for now.

PR #1407, opened 2015-11-03
fix depreciated code in order to compile on Debian 8

eschmidbauer:
Please apply the fixes to the depreciated code in order to compile Kazoo on Debian 8.
Thanks,
Emmanuel


jamesaimonetti:
I believe there is a branch for building against Erlang 17/18 called erlang-otp18. This may be duplicated effort; @fenollp was heading up that branch I believe?

jamesaimonetti:
@eschmidbauer just for reference, v3.x of Kazoo targets Erlang R15B03-1, which has obviously gotten a bit behind Erlang's latest and greatest. With the 4.x series of Kazoo, which master now represents, we will be dropping R15B03 in favor of 17.5 I believe. We'll also be working harder to ensure we stay in better sync with Erlang versions.

eschmidbauer:
I reviewed the branch called erlang-otp18, it is 875 commits behind master. 
@jamesaimonetti 
Is there any issue with merging my commits to master? Does it break backward compatibility? 
Seems only code touched is calls to 
"set", "queue", "dict"

Also, if master represents 4.x series, shouldn't it compile on Debian 8?

jamesaimonetti:
@eschmidbauer That is why I've asked @fenollp to weigh in. We've only just begun the 4.0 work in branches, having tagged 3.22 after the previous sprint. I suspect @fenollp will just need to rebase his work, or if he reviews this PR and sees it accomplishes the same thing, he can merge this PR instead. So yes, eventually master will build for more current versions of Erlang (irrespective of distro being used; the issue is the Erlang version installed); today, master isn't not quite there.

jamesaimonetti:
I'll just add one more thing; when 4.0 is tagged and goes into the QA cycle, that is when we'll say it should build on targeted Erlang versions. The master branch is bleeding edge and so isn't ready for use outside of development really; if you do use it for testing, its buyer beware that some things may not work yet.

eschmidbauer:
@jamesaimonetti 
I understand, not looking to deploy anything in production yet. I'm a dev and prefer compiling rather than relying on releases of rpm packages. I will look forward to the release of 4.x, in the meantime, I will move ahead with getting things running on Debian 8 with my own patches.

k-anderson:
Howdy!

Unfortunately this is not a backward compatible change.  If you see the branch:
https://github.com/2600hz/kazoo/tree/erlang-otp18
I think you will find an nearly identical change set that we are working to incorporate in a compatible way.  Likely we will need to define macros that can be changed based on the version for these incompatibilities.  

In the mean time I am able to compile on debain 8 using the following command:
ERLC_OPTS='+nowarn_deprecated_type +nowarn_deprecated_function' make compile

I appreciate the work and look forward to any future contributions but I am going to have to close this one.

fenollp:
@eschmidbauer As @k-anderson explains and as [is documented here](https://github.com/2600hz/kazoo#build-dependencies) you need to `make` with some environment variables set up for Kazoo to build on Erlang 17.

If your target is Erlang 18 and above, then adding `'-DOTP_AT_LEAST_18'` in `ERLC_OPTS` should do the work you want to commit, [as handled here](https://github.com/2600hz/kazoo/blob/master/core/whistle-1.0.0/include/wh_types.hrl#L4-L13).

https://github.com/2600hz/kazoo/tree/erlang-otp18 is far behind, superseeded, old and up for deletion (green light anyone?)
I don't remember why it is still here. Probably for me to review and note the changes to the Time API that compiling to 18.* requires.

I believe the only things in Kazoo that need to be fixed for 18 are [these 2 comments](https://github.com/2600hz/kazoo/blob/master/core/whistle-1.0.0/src/wh_util.erl#L1206-L1207).
But compilation fails due to some rabbitmq dep right now. Will look into it.

eschmidbauer:
@fenollp 
Thank you for the information. I was able to compile kazoo master branch on Debian 8 with:
  export ERLC_OPTS='+nowarn_deprecated_type +nowarn_deprecated_function'
  make

I have written systemd-scripts for kz-ecallmgr and kz-whistle_apps
I'd be happy to contribute them, please let me know. Thanks again.
Emmanuel

PR #1366, opened 2015-10-16
Fix of "Original-Number" for call rating, classification, and restrictions.

cando-1p:
The "Original" number passed into FS vars was the number match from the resource, what is put on the request uri. not the original number.
Found this because of incorrect call rating. For example (with made up number), international number was converted from 0116544351234 to +6544351234 with e164 normalization. The stepswitch_resources module in building the request uri removed the + so the call sent to the carrier was 6544351234@carrier_fqdn.com. That number was stored in the FS CCVs and pulled out on the second part of the evaluation where the "original" number is evaluated for classification and rating. Rating looked at it and said this looks like a US number. It was really a Papua New Guinea number. Call went through to the correct destination but charging, classification and restriction considerations were all wrong.

PR #1409, opened 2015-11-04
HELP-19461 clarify usage of stepswitch resource properties

lazedo:
E164-Destination precedence for authz/rating
Original-Number is set from Original Request
Add Matched-Number (value taken from former Original-Number)

### DIFF
--- a/applications/jonny5/src/j5_request.erl
+++ b/applications/jonny5/src/j5_request.erl
@@ -134,7 +134,10 @@ from_ccvs(#request{request_ccvs=ReqCCVs
 
 -spec request_number(ne_binary(), wh_json:object()) -> ne_binary().
 request_number(Number, CCVs) ->
-    case wh_json:get_value(<<"Original-Number">>, CCVs) of
+    case wh_json:get_first_defined([<<"E164-Destination">>
+                                    ,<<"Original-Number">>
+                                   ], CCVs
+                                  ) of
         'undefined' -> Number;
         Original ->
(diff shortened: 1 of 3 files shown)


cando2k6:
Looks like use of Original-Number many need to be fixed in applications/webhooks/src/webhooks_channel_util.erl and applications/fax/src/fax_smtp.erl as well.

cando2k6:
This pull request in its current state, same as pull req #1366, does have the danger of potentially breaking classifications, restrictions and rates that have been written to compensate for the problem being fixed.
Therefore, it would be advisable to augment this pull to have a system config option to enable this bug fix. It would be good to have the fix enabled on new systems by default but not on preexisting systems.
Furthermore, since the problem of miss classifying and miss rating some international calls is likely present on many systems, a bulletin and documentation should be posted clearly describing the problem, the potential financial implications and how to fix it. 
The greatest potential for the current handling being a problem is for a system primarily designed for NPANXX based countries like US and Canada.

lazedo:
@cando2k6 you are wrong about the fixes needed in other files
as for you other comment, i don't agree. people should know the risks of using regexes to `hack` the system to9 fit their needs. this PR is a clarification on the usage of stepswitch emitted properties on the amqp bus.

`since the problem of miss classifying and miss rating some international calls is likely present on many systems`, i think your assumption is incorrect.

`The greatest potential for the current handling being a problem is for a system primarily designed for NPANXX based countries like US and Canada`, give one example where this wouldn't work.

PR #1411, opened 2015-11-05
Kazoo 4196

macpie:


applications/blackhole/src/blackhole_util.erl
```diff
@@ -127,8 +127,9 @@ respond_with_authn_failure(Context) ->
 %%--------------------------------------------------------------------
 -spec get_callback_module(ne_binary()) -> atom().
 get_callback_module(Binding) ->
-    case binary:split(Binding, <<".">>) of
-        [M, _] ->
+    case binary:split(Binding, <<".">>, ['global']) of
+        [_|_]=RoutingKey ->
```
jamesaimonetti: Why not match `[M|_] -> Mod = special_bindings(M), try...`

The tail of the list, and the whole list (RoutingKey) is unused, so why bring it along unnecessarily?

applications/crossbar/src/modules/cb_websockets.erl
```diff
@@ -17,13 +17,32 @@
...
+
+-define(CALL, [<<"call.CHANNEL_CREATE.*">>
+               ,<<"call.CHANNEL_ANSWER.*">>
+               ,<<"call.CHANNEL_DESTROY.*">>
+               ,<<"call.CHANNEL_BRIDGE.*">>
+              ]).
+
+-define(
+    OBJECTS
+    ,lists:merge(
```
jamesaimonetti: No need to have nested list comprehensions + a call to `lists:merge/1`. This can be accomplished with one list comprehension. Please make that change.

applications/blackhole/src/modules/bh_object.erl
```diff
@@ -0,0 +1,83 @@
...
+
+%%--------------------------------------------------------------------
+%% @public
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec rm_amqp_binding(ne_binary(), bh_context:context()) -> 'ok'.
+rm_amqp_binding(<<"call.", _/binary>>, Context) ->
+    lager:debug("removing amqp binding....."),
+    blackhole_listener:remove_binding('call', [{'restrict_to', ['PARK_PARKED', 'PARK_RETRIEVED', 'PARK_ABANDONED']}
```
jamesaimonetti: Why is bh_object concerned with `PARK_*` events? Why would it be unbinding from those events when `add_amqp_binding` doesn't add them? Shouldn't `rm_amqp_binding` be the inverse of `add_amqp_binding`?

PR #1412, opened 2015-11-05
Kazoo 4191 - admin port request templates

jamesaimonetti:


applications/crossbar/src/api_util.erl
```diff
@@ -393,22 +396,26 @@ default_filename() ->
...
         {'error', 'badlength'} ->
             lager:debug("the request body was most likely too big"),
             ?MODULE:halt(Req0
-                         ,cb_context:add_validation_error(<<"file">>, <<"maxLength">>
+                         ,cb_context:add_validation_error(<<"file">>
+                                                          ,<<"maxLength">>
                                                           ,<<"File was larger than allowed">>
                                                           ,Context
-                                                          ));
+                                                         )
```
k-anderson: Should this not also be a 413?
cb_context:set_resp_error_code(Context, 413)

Maybe check all the modules to make sure it is consistent?

jamesaimonetti: Yeah, that should move over to use the max file size exceeded function thingy

applications/teletype/src/templates/teletype_port_request_admin.erl
```diff
@@ -0,0 +1,219 @@
...
+%%%-------------------------------------------------------------------
+-module(teletype_port_request_admin).
+
+-export([init/0
+         ,handle_req/2
+        ]).
+
+-include("../teletype.hrl").
+
+-define(TEMPLATE_ID, <<"port_request_admin">>).
```
k-anderson: This should only be exposed to the system admin (like system alerts).

applications/teletype/src/templates/teletype_port_request_admin.erl
```diff
@@ -0,0 +1,219 @@
...
+-define(MOD_CONFIG_CAT, <<(?NOTIFY_CONFIG_CAT)/binary, ".", (?TEMPLATE_ID)/binary>>).
+
+-define(TEMPLATE_MACROS
+        ,wh_json:from_list(
+           ?PORT_REQUEST_MACROS
+           ++ ?ACCOUNT_MACROS
+          )
+       ).
+
+-define(TEMPLATE_TEXT, <<"Port request submitted for {{account.name}} by {{port_request.customer_contact}}.\n\nThe account's tree:\n\n {{ account.name }} ({{ account.id }})\n{% for id, name in account_tree %} {{ name }} ({{ id }})\n{% endfor %}\n\nRequest to port numbers: {{ port_request.numbers }}.\n\nPort Details:\n\n {% for k,v in port_request %} {{ k }} : {{ v }}\n {% endfor %}\n">>).
```
k-anderson: Does this work for multiple numbers? I would expected a for...loop:

Request to port numbers: {{ port_request.numbers }}

jamesaimonetti: Yes, .numbers is converted to a binary with `wh_util:join_binary/2` towards the bottom of the file

PR #1413, opened 2015-11-05
KAZOO-4324 remove dead code and renamed route_win entry point

lazedo:


k-anderson:
@lazedo could we also remove whapps_call cache?

https://github.com/2600hz/kazoo/blob/master/core/kazoo_caches-1.0.0/src/kazoo_caches_sup.erl#L24

lazedo:
@k-anderson there are other whapps using it acdc, milliwatt, cccp, doodle, should i change them all ?

k-anderson:
@lazedo as per our IRC conversation ;) Yes please make route req/win handling consistent.  Thanks!

PR #1417, opened 2015-11-10
Compile on 18

fenollp:
* Introduces `-DOTP_AT_LEAST_18` : a flag to add to `ERLC_OPTS` for `make`
* Removes unused and annoying `deps/edown-0.0.1`
* Some deps now `include_lib("whistle/include/wh_types.hrl")`. This may not please, so I suggest that I move the `ifdef(OTP_AT_LEAST_18)...` from `wh_types.hrl` to its own file, include it in `wh_types.hrl` for `core/` and `applications/` but include just this new file for `deps/`, replacing the introduced `include_lib…` calls in `deps/`.
* `ERLC_OPTS` used to be passed along in `core/` and `applications/`. It is now also the case for `deps/`.

Note: no integration in any `Makefile` is done. So in order to build Kazoo on 18 *manually*, one needs to change his `make` summoning from `make` to
```shell
ERLC_OPTS='+nowarn_deprecated_type +nowarn_deprecated_function -DOTP_AT_LEAST_18' make
```

Travis seems pleased. No idea if it **really** works.


lazedo:
@fenollp great! i agree with the move to own file

PR #1418, opened 2015-11-10
KAZOO-4366: Get auth token after getting req data

macpie:


fenollp:
While it works (I tested it!!), I'm not sure the fix should be done here. So, can you:
* either find/fix this in `api_util:get_req_data` (I don't feel like this function should modify surrounding state. But since it does, please change its name to something that says "hey I modify state" instead of "I'm just an accessor").
* or add some comments saying why you moved the function call after `get_req_data`

@jamesaimonetti I'm starting to sound like you.. Help meee argh

macpie:
This is because without the req_data you cannot fetch the auth-token when it is used in the query string. And it is also needed in the error return.

PR #1421, opened 2015-11-11
KAZOO-4349: Adding source IP to bh_context

macpie:


applications/blackhole/src/blackhole_socket_callback.erl
```diff
@@ -74,7 +74,13 @@ close(SessionPid, SessionId, Context) ->
 %% @end
 %%--------------------------------------------------------------------
 -spec handle_info(ne_binary(), ne_binary(), any(), bh_context:context()) -> {'ok', bh_context:context()}.
+handle_info(_SessionPid, _SessionId, {'req', Req}, Context) ->
```
jamesaimonetti: Are there plans to use `Req` for something other than getting the IP? Meaning, can the change to `socketio_handler` be `{{IP, _}, Req1} = cowboy_req:peer(Req), Pid ! {'peer_ip', wh_util:to_binary(inet_parse:ntoa(IP))}` instead and this clause match the `peer_ip` tuple?

macpie: sounds good

deps/socketio-0.1.0/src/socketio_handler.erl
```diff
@@ -71,7 +71,10 @@ handle(Req, HttpState = #http_state{action = create_session, config = #config{he
     HeartbeatTimeoutBin = list_to_binary(integer_to_list(HeartbeatTimeout div 1000)),
     SessionTimeoutBin = list_to_binary(integer_to_list(SessionTimeout div 1000)),
 
-    _Pid = socketio_session:create(Sid, SessionTimeout, Callback, Opts),
+    Pid = socketio_session:create(Sid, SessionTimeout, Callback, Opts),
+
+    % Adding this to be able to get client IP in bh_context
+    Pid ! {'req', Req},
```
jamesaimonetti: Ok, which `handle_info` clause handles this payload? And where is the code the do `Pid ! {'peer_ip', IP}`?

macpie: I ignore all the deps usually this is why it wasn't committed

PR #1423, opened 2015-11-11
KAZOO-4375: For blacklist api, if apps_store doc missing create it

macpie:


fenollp:
I don't like this but since `limits` do this anyway…

applications/crossbar/src/modules/cb_apps_store.erl
```diff
@@ -599,7 +599,18 @@ get_screenshot(Context, Number) ->
...
+    Context1 = crossbar_doc:load(kz_apps_store:id(), Context),
+    case {cb_context:resp_status(Context1), cb_context:resp_error_code(Context1)} of
+        {'error', 404} ->
+            cb_context:setters(
+                Context
+                ,[{fun cb_context:set_resp_status/2, 'success'}
+                  ,{fun cb_context:set_resp_data/2, wh_json:new()}
+                ]
+            );
+        {'success', _} -> Context1;
```
jamesaimonetti: Why match other cases here? Do you really want to crash the process if `cb_context:resp_status/1` returns 'fatal' or 'halt'?

PR #1426, opened 2015-11-12
KAZOO-4384: resource's cid_rules option should honor Diversions if ex…

onnet:
…ists
While using stepswitch's cid_rules field for CID-based outbound routing, it looks like we need to check if diversions fields exists and use it to make routing decision based on account's number, but not someone's else.

### DIFF
--- a/applications/stepswitch/src/stepswitch.hrl
+++ b/applications/stepswitch/src/stepswitch.hrl
@@ -29,5 +29,9 @@
         ,wh_json:from_list([{'passive', 'true'}])
        ).
 
+-define(RULES_HONOR_DIVERSION
+        ,whapps_config:get_is_true(?SS_CONFIG_CAT, <<"cid_rules_honor_diversions">>, 'true')
+       ).
+
 -define(STEPSWITCH_HRL, 'true').
 -endif.
--- a/applications/stepswitch/src/stepswitch_resources.erl
+++ b/applications/stepswitch/src/stepswitch_resources.erl
@@ -397,7 +397,10 @@ maybe_resource_to_endpoints(#resrc{id=Id
                             ,OffnetJObj
                             ,Endpoints
                            ) ->
-    CallerIdNumber = wapi_offnet_resource:outbound_caller_id_number(OffnetJObj),
+    CallerIdNumber = case whapps_config:get_is_true(?SS_CONFIG_CAT, <<"cid_rules_honor_diversions">>, 'true') of
+        'false' -> wapi_offnet_resource:outbound_caller_id_number(OffnetJObj);
+        'true' -> check_diversion_fields(OffnetJObj)
+    end,
     case filter_resource_by_rules(Id, Number, Rules, CallerIdNumber, CallerIdRules) of
         {'error','no_match'} -> Endpoints;
         {'ok', NumberMatch} ->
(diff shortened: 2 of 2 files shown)


applications/stepswitch/src/stepswitch_resources.erl
```diff
@@ -397,7 +397,10 @@ maybe_resource_to_endpoints(#resrc{id=Id
                             ,OffnetJObj
                             ,Endpoints
                            ) ->
-    CallerIdNumber = wapi_offnet_resource:outbound_caller_id_number(OffnetJObj),
+    CallerIdNumber = case whapps_config:get_is_true(?SS_CONFIG_CAT, <<"cid_rules_honor_diversions">>, 'true') of
+        'false' -> wapi_offnet_resource:outbound_caller_id_number(OffnetJObj);
+        'true' -> check_diversion_fields(OffnetJObj)
+    end,
```
fenollp: Please put the `whapps_config` call in a macro named something like `RULES_HONOR_DIVERSION`.

Also, indent 401-3 properly

applications/stepswitch/src/stepswitch_resources.erl
```diff
@@ -416,6 +419,16 @@ maybe_resource_to_endpoints(#resrc{id=Id
             maybe_add_proxies(EndpointList, Proxies, Endpoints)
     end.
 
+-spec check_diversion_fields(wapi_offnet_resource:req()) -> ne_binary(). 
+check_diversion_fields(OffnetJObj) ->
+    case wh_json:get_value([<<"Custom-SIP-Headers">>,<<"Diversions">>], OffnetJObj) of
+        'undefined' ->
+            wapi_offnet_resource:outbound_caller_id_number(OffnetJObj);
+        [Diversion|_] ->
+            [_,_,CallerIdNumber,_] = re:split(Diversion,"(:)(\\d*)(@.*)",[{return,binary},{parts,0}]),
```
fenollp: I'd rather you use `binary:split(Diversion, [<<":">>,<<"@">>], ['global'])`.

Also, make sure that you get the right data from JSON by using `wh_json:get_binary_value/2` instead of `get_value/2`

onnet: @fenollp Thank for help!

regarding wh_json:get_binary_value/2
Diversions field is a list, so I am afraid get_binary_value doesn't pass in this case.

Just changed sequence to use catch all..

fenollp:
Also please rollback any trailing whitespace introduced in `applications/stepswitch/src/stepswitch_resources.erl`.

applications/stepswitch/src/stepswitch.hrl
```diff
@@ -29,5 +29,9 @@
         ,wh_json:from_list([{'passive', 'true'}])
        ).
 
+-define(RULES_HONOR_DIVERSION
+        ,whapps_config:get_is_true(?SS_CONFIG_CAT, <<"cid_rules_honor_diversions">>, 'true')
```
jamesaimonetti: This should default to 'false' to preserve existing functionality.

jamesaimonetti:
Made the change so we could get this in. Thanks @onnet !

onnet:
Thanks @jamesaimonetti !

PR #1427, opened 2015-11-12
KAZOO-4385 send node record on notify

lazedo:


applications/jonny5/src/j5_channels.erl
```diff
@@ -509,10 +509,14 @@ handle_cast({'rate_resp', JObj}, State) ->
 handle_cast('synchronize_channels', #state{sync_ref=SyncRef}=State) ->
     self() ! {'synchronize_channels', SyncRef},
     {'noreply', State};
-handle_cast({'wh_nodes', {'expire', _Node}}, #state{sync_ref=SyncRef}=State) ->
-    lager:debug("notifed that node ~s is no longer reachable, synchronizing channels", [_Node]),
-    self() ! {'synchronize_channels', SyncRef},
-    {'noreply', State};
+handle_cast({'wh_nodes', {'expire', #wh_node{node=NodeName, whapps=Whapps}}}, #state{sync_ref=SyncRef}=State) ->
```
jamesaimonetti: We've been moving away from having records leak out across lots of modules/applications. I wonder if this would be better as `wh_node:node_name/1` and `wh_node:whapps/1`?

lazedo: i tend to agree but since this is part of core and whapp_info may include specific values for each whapp (besides `startup`, ex: leader), i would prefer to allow this.

PR #1429, opened 2015-11-12
Kazoo 4376 - updates to lager_syslog/erlang_syslog

jamesaimonetti:


core/whistle-1.0.0/src/wh_util.erl
```diff
@@ -1126,8 +1126,9 @@ write_pid(FileName) ->
 
 -spec ensure_started(atom()) -> 'ok' | {'error', any()}.
 ensure_started(App) when is_atom(App) ->
-    case application:start(App) of
+    case application:ensure_all_started(App) of
```
fenollp: `application:ensure_all_started/1` isn't available until R16B02; meaning: it fails at runtime (or in `make xref`) on R15B03.

Code such as [this](https://github.com/2600hz/kazoo/blob/master/applications/pusher/lib/apns/src/apns.erl#L53-L62) could go in `wh_util:ensure_all_started/1`.
Note that a bunch of `deps/` also call this function.

deps/lager_syslog-2.0/src/lager_syslog.app.src
```diff
@@ -3,7 +3,7 @@
 {application, lager_syslog,
  [
   {description, "Syslog backend for Lager"},
-  {vsn, "1.2.2"},
+  {vsn, "2.0.3"},
```
fenollp: Why isn't this version number reflected in the name of the deps directory?

deps/syslog-1.0.3/Makefile
```diff
@@ -0,0 +1,59 @@
...
+
+ERL_INTERFACE := $(find $USR_INCLUDE/../../ -type d -name "erl_interface-*" -print)
+ifeq ($(ERL_INTERFACE),)
+# 3.7.9 is R15B03-1
+# Change to 3.7.20 for 17.5
+	ERL_INTERFACE := $(USR_INCLUDE)/../../lib/erl_interface-3.7.20
+endif
+
+priv/syslog_drv.so: c_src/syslog_drv.c
+	gcc  -I$(USR_INCLUDE) -I$(ERL_INTERFACE)/include -L$(ERL_INTERFACE)/lib -fpic -shared c_src/syslog_drv.c -o priv/syslog_drv.so -lerl_interface -lei
```
fenollp:
```
gcc  -I/home/travis/otp/17.5/bin/../erts-6.4/include -I/home/travis/otp/17.5/bin/../erts-6.4/include/../../lib/erl_interface-3.7.20/include -L/home/travis/otp/17.5/bin/../erts-6.4/include/../../lib/erl_interface-3.7.20/lib -fpic -shared c_src/syslog_drv.c -o priv/syslog_drv.so -lerl_interface -lei
/usr/bin/ld: cannot open output file priv/syslog_drv.so: No such file or directory
collect2: ld returned 1 exit status
```

I suspect that `deps/syslog-1.0.3/priv` does not exist. (needs a `.placeholder` file)

fenollp:
@jamesaimonetti This PR includes 2 rebar binaries that should be removed and make use of the one already in `/utils`.

Also, could you create a file (of which format you get to pick) named `/deps/deps.‹something›` describing the repository URL and a tag or commit of the dependency you introduce?
This would allow for easier bumping of deps later on. And I would gladly add entries to this file for other dependencies we have! (or write a ticket, you lazy you)

deps/lager_syslog-2.0/src/lager_syslog.app.src
jamesaimonetti: I downloaded a zip, didn't check the app file for what version it was. Will update

jamesaimonetti:
Since we're dropping R15B03 support in Kazoo 4, this build is working in relevant versions of Erlang. Ready to merge!

lazedo:
@jamesaimonetti yep, but if you merge it, every build is going to fail.
can we try to use something like https://github.com/2600hz/kazoo/blob/master/core/whistle-1.0.0/include/otp_compatibility.hrl#L3

can we try to have 2 versions of ensure_all_started in wh_util that is compiled depending of the otp version ?

jamesaimonetti:
@lazedo done!

PR #1432, opened 2015-11-13
KAZOO-4315: rm mochiweb

mark2600:


core/kazoo_web-1.0.0/src/kz_http.erl
```diff
@@ -0,0 +1,262 @@
...
+-spec urlencode(binary(), binary()) -> binary().
+urlencode(<<>>, Acc) ->
+    Acc;
+
+urlencode(<<$\s, R/binary>>, Acc) ->
+    urlencode(R, <<Acc/binary, $+>>);
+
+urlencode(<<C, R/binary>>, Acc) ->
+    case C of
+        $\. -> urlencode(R, <<Acc/binary, C>>);
```
fenollp: No need for backslash here

core/kazoo_web-1.0.0/src/kz_http.erl
```diff
@@ -0,0 +1,262 @@
...
+%% @doc Splits out the scheme portion of the URL (if present)
+%%--------------------------------------------------------------------
+-spec urlsplit_s(binary()) -> {binary(), binary()}.
+urlsplit_s(Source) ->
+    case urlsplit_s(Source, <<>>) of
+        'no_scheme' -> {<<>>, Source};
+        ValidScheme -> ValidScheme
+    end.
+
+-spec urlsplit_s(binary(), binary()) -> {binary(), binary()}.
```
fenollp: Spec needs to say `-> … | 'no_scheme'.`

fenollp:
@mark2600 Do you think you can remove mochiweb from the deps?

mark2600:
We will be able to ditch the mochinum stuff when we finally drop support for R15 (as there are much better BIF's for float manipulation in R16). I will remove the mochi ebins from call_inspector's Makefile.  Not sure on the mochiweb-git situation, either way I don't think that will allow us to remove the mochiweb files from deps just yet.

fenollp:
Note that `mochinum.{erl,beam}` can also be found in `deps/rabbitmq_server-3.5.5`

fenollp:
@mark2600 https://github.com/2600hz/kazoo/compare/KAZOO-4315-FTW See? No harm possible here

PR #1435, opened 2015-11-16
KAZOO-4347 park presence type by slot/account/system

lazedo:


jamesaimonetti:
@lazedo this all looks good to me but probably need @k-anderson to take a once-over as well. We can discuss on the call.

jamesaimonetti:
Too slow @k-anderson

PR #1436, opened 2015-11-16
KAZOO-4324 route req/win for conferences

lazedo:


lazedo:
@jamesaimonetti thanks. backported to 3.22

PR #1441, opened 2015-11-17
KAZOO-4390

fenollp:


applications/webhooks/src/webhooks_shared_listener.erl
```diff
@@ -258,8 +258,7 @@ handle_cast(_Msg, State) ->
 %% @end
 %%--------------------------------------------------------------------
 handle_info(?HOOK_EVT(AccountId, EventType, JObj), State) ->
-    _ = wh_util:spawn(?MODULE
-                      ,'maybe_handle_channel_event'
+    _ = wh_util:spawn(fun webhooks_channel_util:maybe_handle_channel_event/3
```
fenollp: This was actually a bug!

lazedo: i believe this is dead code or a result of a copy/paste. i don't see the registration for the webhooks.

PR #1442, opened 2015-11-17
compile-on-centos: hopefuly fix syslog compilation

fenollp:


jamesaimonetti:
Works for me on ubuntu

PR #1445, opened 2015-11-17
KAZOO-4383: Change port_request status update to patch

macpie:


applications/crossbar/src/modules/cb_port_requests.erl
```diff
@@ -1291,6 +1278,60 @@ has_new_comment(OldComments, NewComments) ->
...
+%% @end
+%%--------------------------------------------------------------------
+-spec revert_patch(cb_context:context()) -> cb_context:context().
+revert_patch(Context) ->
+    Doc = cb_context:doc(Context),
+    DBDoc = cb_context:fetch(Context, 'db_doc'),
+
+    Rev = wh_json:get_value(<<"_rev">>, Doc),
+
+    RevertedDoc = wh_json:set_value(<<"_rev">>, Rev, DBDoc),
```
fenollp: There's accessors in wh_doc for _rev

applications/crossbar/src/cb_context.erl
```diff
@@ -1040,8 +1040,7 @@ add_system_error('invalid_method'=Error, JObj, Context) ->
     J = wh_json:set_value(<<"message">>, <<"invalid method">>, JObj),
     build_system_error(405, Error, J, Context);
 add_system_error('bad_gateway'=Error, JObj, Context) ->
-    J = wh_json:set_value(<<"message">>, <<"bad gateway">>, JObj),
```
fenollp: Why remove the message?

macpie: needs to be custom and not forced

PR #1446, opened 2015-11-17
KAZOO-4392: If is_trial_account is set it will add the proper pvt_tri…

macpie:
…al_expires flag to account


jamesaimonetti:
@macpie I assume this was discussed with @k-anderson; the last I had heard, screwdriver was the app setting the pvt value.

PR #1447, opened 2015-11-18
KAZOO-4325: Re order check for user auth, now: creds -> expired -> en…

macpie:
…abled


jamesaimonetti:
I would have liked to see this converted to something like using `lists:all/2`:

```erlang
Fs = [fun() -> is_user_authenticated(...) end
      ,fun() -> is_account_active(...) end
      ,fun() -> is_account_enabled(...) end
     ],
lists:all(fun(F) -> F() end, Fs).
```

Now it is clear, from the code, the order of operations for determining authn-ness. Just something to keep in mind when you have this logical ANDing of conditions.

PR #1450, opened 2015-11-18
KAZOO-4293: distinguish local vs remote when reconnecting

jamesaimonetti:
allow timeout to grow larger for remote brokers vs local brokers

### DIFF
--- a/core/whistle_amqp-1.0.0/src/wh_amqp_connection.erl
+++ b/core/whistle_amqp-1.0.0/src/wh_amqp_connection.erl
@@ -27,8 +27,9 @@
 -include("amqp_util.hrl").
 
 -define(SERVER, ?MODULE).
--define(START_TIMEOUT, 200).
--define(MAX_TIMEOUT, ?MILLISECONDS_IN_SECOND).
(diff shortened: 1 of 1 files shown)


core/whistle_amqp-1.0.0/src/wh_amqp_connection.erl
```diff
@@ -451,8 +470,10 @@ maybe_add_all_exchanges(Connection, 'true') ->
 declare_exchanges(#wh_amqp_connection{}=Connection, []) ->
     Connection#wh_amqp_connection{exchanges_initialized='true'};
 declare_exchanges(#wh_amqp_connection{channel=Channel
-                                      ,broker=_Broker}=Connection
-                  ,[Exchange|Exchanges])
+                                      ,broker=_Broker
```
jamesaimonetti: Couldn't tell you; was there already and didn't feel like changing it.

fenollp: Ah, just saw they're used by `lager:debug`. Nevermind.

PR #1451, opened 2015-11-19
KAZOO-3669: replace reset password behaviour

fenollp:
Notes
* needs testing
* waiting on a UI ticket: ~~[UI-1925](https://2600hz.atlassian.net/browse/UI-1925)~~ [UI-1819](https://2600hz.atlassian.net/browse/UI-1819)
* not sure about the way to `crossbar_cleanup`
* why does my vm not send emailsdammit

### DIFF
--- a/applications/crossbar/src/modules/cb_user_auth.erl
+++ b/applications/crossbar/src/modules/cb_user_auth.erl
@@ -16,29 +16,37 @@
          ,authenticate/1
          ,validate/1, validate/2
          ,put/1, put/2
+         ,post/2
+         ,cleanup_reset_ids/1
         ]).
 
 -include("crossbar.hrl").
 
 -define(ACCT_MD5_LIST, <<"users/creds_by_md5">>).
 -define(ACCT_SHA1_LIST, <<"users/creds_by_sha">>).
--define(USERNAME_LIST, <<"users/list_by_username">>).
+-define(LIST_BY_RESET_ID, <<"users/list_by_reset_id">>).
+-define(LIST_BY_MTIME, <<"users/list_by_mtime">>).
 -define(DEFAULT_LANGUAGE, <<"en-us">>).
 -define(USER_AUTH_TOKENS, whapps_config:get_integer(?CONFIG_CAT, <<"user_auth_tokens">>, 35)).
 
 -define(RECOVERY, <<"recovery">>).
+-define(RESET_ID, <<"reset_id">>).
+-define(RESET_ID_SIZE, 256).
+-define(RESET_PVT_TYPE, <<"password_reset">>).
 
 %%%===================================================================
 %%% API
 %%%===================================================================
 init() ->
     kz_datamgr:db_create(?KZ_TOKEN_DB),
+    _ = crossbar_bindings:bind(crossbar_cleanup:binding_account(), ?MODULE, 'cleanup_reset_ids'),
     _ = crossbar_bindings:bind(<<"*.authenticate">>, ?MODULE, 'authenticate'),
     _ = crossbar_bindings:bind(<<"*.authorize">>, ?MODULE, 'authorize'),
     _ = crossbar_bindings:bind(<<"*.allowed_methods.user_auth">>, ?MODULE, 'allowed_methods'),
     _ = crossbar_bindings:bind(<<"*.resource_exists.user_auth">>, ?MODULE, 'resource_exists'),
     _ = crossbar_bindings:bind(<<"*.validate.user_auth">>, ?MODULE, 'validate'),
-    _ = crossbar_bindings:bind(<<"*.execute.put.user_auth">>, ?MODULE, 'put').
+    _ = crossbar_bindings:bind(<<"*.execute.put.user_auth">>, ?MODULE, 'put'),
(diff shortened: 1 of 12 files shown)


applications/crossbar/src/modules/cb_user_auth.erl
```diff
@@ -314,50 +341,72 @@ load_md5_results(Context, JObj) ->
...
+            lager:debug("no expired reset_ids found");
+        {'ok', UserDocs} ->
+            lager:debug("checking ~b user documents", [length(UserDocs)]),
+            couch_mgr:suppress_change_notice(),
+            ensure_reset_id_deleted(UserDocs),
+            couch_mgr:enable_change_notice(),
+            lager:debug("removed yesterday's expired reset_ids")
+    end.
+
+ensure_reset_id_deleted([Doc|Docs]) ->
```
jamesaimonetti: spec

applications/crossbar/src/modules/cb_user_auth.erl
```diff
@@ -314,50 +341,72 @@ load_md5_results(Context, JObj) ->
...
+    {'ok', _} = couch_mgr:save_doc(wh_doc:account_db(Doc)
+                                  ,wh_json:delete_key(?RESET_ID, Doc)),
```
jamesaimonetti: Curious why you chose not to use `couch_mgr:save_docs/2` instead? Bulk saves are more efficient for Couch to process.

fenollp:
@jamesaimonetti On using `couch_mgr:save_docs/2` instead: I don't think I can save every user's doc doc `?WH_ACCOUNTS_DB`, right? Because I need to save each `Doc` to its own `wh_doc:account_db(Doc)`, right?

applications/crossbar/src/modules/cb_user_auth.erl
```diff
@@ -314,50 +341,73 @@ load_md5_results(Context, JObj) ->
...
+-spec cleanup_reset_ids() -> 'ok'.
+cleanup_reset_ids() ->
+    CreatedBefore = wh_util:current_tstamp() - 2 * ?SECONDS_IN_DAY,
+    ViewOptions = [{'startkey', 0}
+                   ,{'endkey', CreatedBefore}
+                   %% TODO: find a limit that matches the number of pwd resets per day
+                   ,{'limit', couch_util:max_bulk_insert()}
+                   ,'include_docs'
+                  ],
+    case couch_mgr:get_results(?WH_ACCOUNTS_DB, ?LIST_BY_MTIME, ViewOptions) of
```
jamesaimonetti: I'm confused why the view is being looked up in `accounts`? AFAICT you are saving the reset docs back to the account database (and loading the user doc via `reset_id` pulls out the account id and uses that account database). Why would cleanup occur by querying the `accounts` database then? Those are account docs, not user docs.

applications/crossbar/src/modules/cb_user_auth.erl
```diff
@@ -314,50 +341,73 @@ load_md5_results(Context, JObj) ->
...
+            lager:debug("checking ~b user documents", [length(UserDocs)]),
+            couch_mgr:suppress_change_notice(),
+            ensure_reset_id_deleted(UserDocs),
+            couch_mgr:enable_change_notice(),
+            lager:debug("removed yesterday's expired reset_ids")
+    end.
+
+-spec ensure_reset_id_deleted(wh_json:objects()) -> 'ok'.
+ensure_reset_id_deleted([Doc|Docs]) ->
+    {'ok', _} = couch_mgr:save_doc(wh_doc:account_db(Doc)
```
jamesaimonetti: I don't think you should save the doc if it doesn't have `?RESET_ID` set on it.

applications/crossbar/src/modules/cb_user_auth.erl
fenollp: You are totally right. I just have no idea how I can fetch user documents that were modified yesterday and that have `?RESET_ID` ≠ `undefined`?

jamesaimonetti: You set `reset_id` on the user though; what happens if their doc is updated by an admin? What if they have a script that sets some property every day, bumping the `pvt_modified` each time (and thus never cleaning up the `reset_id`).

Could it simplify life if you tracked the reset attempt as its own document?
```json
{"_id":"reset_id_256"
 ,"user_id":"user_id_32"
 ,"pvt_created":timestamp
 ,"pvt_type":"password_reset"
}
```

Now you can directly fetch the doc instead of using a view, and looking up the user doc is a `open_cache_doc/2` call away as well.

Cleanup then becomes binding for `crossbar_cleanup:binding_account()` and `cleanup_reset_ids(AccountDb)` can query `maintenance/listing_by_type` with `{'key', <<"password_reset">>}`. I would wager this result set will typically be small enough that iterating over it and checking pvt_modified won't be overly expensive in memory/compute power.

Thoughts?

jamesaimonetti:
@fenollp you have conflicts here now. Resolve so I can merge it!

fenollp:
@jamesaimonetti please make sure this works on your side too

PR #1452, opened 2015-11-19
KAZOO-4350: Send delete req to mobile manager when you delete an account

macpie:


applications/crossbar/src/cb_mobile_manager.erl
```diff
@@ -0,0 +1,94 @@
...
+
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%%
+%% @end
+%%--------------------------------------------------------------------
+-spec req_uri(ne_binaries()) -> list().
+req_uri(ExplodedPath) ->
+    Url = whapps_config:get_binary(?MOD_CONFIG_CAT, <<"url">>),
```
fenollp: No default value, or otherwise handling if undefined?

applications/crossbar/src/cb_mobile_manager.erl
```diff
@@ -0,0 +1,94 @@
...
+    AccountId = cb_context:account_id(Context),
+    AuthToken = cb_context:auth_token(Context),
+
+    UrlString = req_uri([<<"accounts">>, AccountId]),
+    Headers = req_headers(AuthToken),
+
+    lager:debug("mobile_manager delete via ~s", [UrlString]),
+
+    Resp = ibrowse:send_req(UrlString, Headers, 'delete'),
+    handle_resp(Resp).
```
fenollp: I really like this API. Could you create a whistle "rest" module that exports your private functions
* `req_uri/1`
* `req_headers/1`
* `send_req/3`
* `handle_resp/1`
* `decode/1`
?

This way we'd have an adapter and we could switch to whatever httpc library we'd like (hackney, httpc, whatevs).
Also, I know a lot of places where this same code is duplicated (or should be used anyway), so I vote for a new REST utils module.

applications/crossbar/src/cb_mobile_manager.erl
```diff
@@ -0,0 +1,94 @@
...
+
+-define(MOD_CONFIG_CAT, <<"mobile_manager">>).
+
+%%--------------------------------------------------------------------
+%% @public
+%% @doc
+%%
+%% @end
+%%--------------------------------------------------------------------
+delete_account(Context) ->
```
jamesaimonetti: spec

applications/crossbar/src/cb_mobile_manager.erl
```diff
@@ -0,0 +1,94 @@
...
+
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%%
+%% @end
+%%--------------------------------------------------------------------
+-spec decode(string()) -> ne_binary().
+decode(JSON) ->
+    try wh_json:encode(JSON) of
```
jamesaimonetti: I truly don't understand the point of this function. In `handle_resp`, the last element in the tuple will be the response body as a string (I believe that's the default in ibrowse). So the first thing tried with that string is to encode it to JSON? Why not just log the response body directly above? The response is unused aside from logging it so why bother doing anything to it?

applications/crossbar/src/cb_mobile_manager.erl
```diff
@@ -0,0 +1,94 @@
...
+%% @private
+%% @doc
+%%
+%% @end
+%%--------------------------------------------------------------------
+-spec req_uri(ne_binaries()) -> list().
+req_uri(ExplodedPath) ->
+    Url = whapps_config:get_binary(?MOD_CONFIG_CAT, <<"url">>),
+    Uri = wh_util:uri(Url, ExplodedPath),
+    binary:bin_to_list(Uri).
```
jamesaimonetti: Why use the `binary` module vs `wh_util:to_binary/1` which will wrap that up? Seems non-standard with the rest of kazoo.

fenollp: My guess is that this way you know it turns a binary into a list. 
Whereas with a random call to to_binary you have no idea of the type you're turning into a list. 
Easier to read / refactor, sometimes. 
I'm sure that this was the OTP team's idea when the designed this API. 
(Actually not exactly, as I feel like they never really had a discussion for some places in the stdlib)

applications/crossbar/src/cb_mobile_manager.erl
```diff
@@ -0,0 +1,87 @@
...
+    lager:error("mobile_manager fatal error ~p", [_Error]).
+
+
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%%
+%% @end
+%%--------------------------------------------------------------------
+-spec req_uri(ne_binaries()) -> list() |  'undefined'.
```
fenollp: That's `api_list()`

PR #1454, opened 2015-11-21
KAZOO-4231: tollfree number search

fenollp:
Needs testing (@k-anderson knows the actual login:pwd info)

### DIFF
--- a/core/whistle-1.0.0/src/wh_util.erl
+++ b/core/whistle-1.0.0/src/wh_util.erl
@@ -556,26 +556,19 @@ pad_binary_left(Bin, _Size, _Value) -> Bin.
 %% @public
 %% @doc
 %% Join a binary together with a seperator.
-%% Changed to Accumulator from the binary-contruction for speed reasons:
-%%
-%% Bins = [to_binary(N) || N <- lists:seq(1,10000)]
-%% Old join_binary(Bins): 171.1ms fastest, 221.9ms slowest
-%% New join_binary(Bins):   1.1ms fastest,   2.6ms slowest
-%% Obvious winner
 %%
 %% @end
 %%--------------------------------------------------------------------
 -spec join_binary([text() | atom(),...]) -> binary().
 -spec join_binary([text() | atom(),...], binary()) -> binary().
 
-join_binary(Bins) -> join_binary(Bins, <<", ">>, []).
-join_binary(Bins, Sep) -> join_binary(Bins, Sep, []).
-
-join_binary([], _, Acc) -> iolist_to_binary(lists:reverse(Acc));
-join_binary([Bin], _, Acc) ->
-    iolist_to_binary(lists:reverse([to_binary(Bin) | Acc]));
-join_binary([Bin|Bins], Sep, Acc) ->
-    join_binary(Bins, Sep, [Sep, to_binary(Bin) |Acc]).
+join_binary(Bins) -> join_binary(Bins, <<", ">>).
+join_binary([], _) -> <<>>;
+join_binary([Bin], _) -> to_binary(Bin);
+join_binary([Bin|Bins], Sep) ->
+    iolist_to_binary(
+      [to_binary(Bin)] ++ [[Sep, to_binary(B)] || B <- Bins]
+     ).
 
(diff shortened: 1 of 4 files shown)


core/whistle-1.0.0/src/wh_util.erl
```diff
@@ -695,6 +683,32 @@ get_xml_value(Paths, Xml) ->
             'undefined'
     end.
 
+%% @private
```
jamesaimonetti: This seems unrelated to the ticket and differs from the previous implementation.

The LCs have filters to only process `#xmlText{}` OR `#xmlAttribute{}` records; below will process both together. Now, in practice it is probably safe, but it is important to note the differences and check call sites to see if there's a case where this needs differentiation.

core/whistle_number_manager-1.0.0/src/carriers/wnm_bandwidth.erl
```diff
@@ -21,19 +21,63 @@
...
+        Prefix == <<"883">> orelse
+        Prefix == <<"884">> orelse
+        Prefix == <<"885">> orelse
+        Prefix == <<"886">> orelse
+        Prefix == <<"887">> orelse
+        Prefix == <<"888">> orelse
+        Prefix == <<"889">>
+       ).
+-define(TOLLFREE_LOGIN,
+        whapps_config:get_string(?WNM_BW_CONFIG_CAT, <<"tollfree_login">>, "darren@2600hz.com")).
```
jamesaimonetti: probably don't want the CEOs email here :)

core/whistle-1.0.0/src/wh_util.erl
jamesaimonetti: This seems unrelated to the ticket and differs from the previous implementation.

The LCs have filters to only process `#xmlText{}` OR `#xmlAttribute{}` records; below will process both together. Now, in practice it is probably safe, but it is important to note the differences and check call sites to see if there's a case where this needs differentiation.

fenollp:
https://gist.github.com/fenollp/32380c9d263570bc1c8a
So as long as the first argument of get_xml_value/2 ends with `/text()` it will only return xmlText elements
```
0 kazoo.git KAZOO-4231 λ git grep -InE 'get_xml_value\(' -- core/ applications/ | \grep -Fv 'text()'
applications/ecallmgr/src/ecallmgr_fs_xml.erl:1017:    Key = wh_util:get_xml_value("/variable/@name", Var),
applications/ecallmgr/src/ecallmgr_fs_xml.erl:1018:    Value = wh_util:get_xml_value("/variable/@value", Var),
core/whistle-1.0.0/src/wh_util.erl:667:-spec get_xml_value(wh_deeplist(), xml_el() | string()) -> api_binary().
core/whistle-1.0.0/src/wh_util.erl:668:get_xml_value(Paths, Xml) ->
```
…we're good here.

The rest is fetching attributes, and that only returns xmlAttribute elements :)

core/whistle-1.0.0/src/wh_util.erl
```diff
@@ -556,26 +556,19 @@ pad_binary_left(Bin, _Size, _Value) -> Bin.
...
-join_binary([Bin], _, Acc) ->
-    iolist_to_binary(lists:reverse([to_binary(Bin) | Acc]));
-join_binary([Bin|Bins], Sep, Acc) ->
-    join_binary(Bins, Sep, [Sep, to_binary(Bin) |Acc]).
+join_binary(Bins) -> join_binary(Bins, <<", ">>).
+join_binary([], _) -> <<>>;
+join_binary([Bin], _) -> to_binary(Bin);
+join_binary([Bin|Bins], Sep) ->
+    iolist_to_binary(
+      [to_binary(Bin)] ++ [[Sep, to_binary(B)] || B <- Bins]
```
jamesaimonetti: Do you like `[to_binary(Bin) | [[Sep, to_binary(B)] || B <- Bins]]` better?

fenollp: `[bla] ++ […]` gets optimised into `[bla|…]` at compile time

PR #1458, opened 2015-11-23
KAZOO-4403: use lists:keyreplace/4 in props.erl

fenollp:
And I just saw that lists:keyreplace/4 was already used by wh_json to set_value/3


fenollp:
Actually, I just saw that keyreplace does not add the tuple if it Key is not already a key in Props…

jamesaimonetti:
The props version also strips all instances of Key from the proplist.

PR #1459, opened 2015-11-23
KAZOO-4170: Filter feature codes when checking if number already exis…

macpie:
…t when adding/updating a  callflow

### DIFF
--- a/applications/crossbar/src/modules/cb_callflows.erl
+++ b/applications/crossbar/src/modules/cb_callflows.erl
@@ -249,6 +249,7 @@ validate_unique_numbers(CallflowId, Numbers, Context, AccountDb) ->
             check_callflow_schema(CallflowId, Context);
         {'ok', JObjs} ->
             FilteredJObjs = filter_callflow_list(CallflowId, JObjs),
+            io:format("MARKER:cb_callflows.erl:252 ~p~n", [[wh_json:get_value([<<"value">>, <<"featurecode">>], J) || J <- FilteredJObjs ]]),
             C = check_uniqueness(Numbers, FilteredJObjs, Context),
             check_callflow_schema(CallflowId, C)
     end.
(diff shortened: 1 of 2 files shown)


core/kazoo_documents-1.0.0/src/kzd_callflow.erl
```diff
@@ -17,11 +18,39 @@
...
+
+%%--------------------------------------------------------------------
+%% @public
+%% @doc
+%%
+%% @end
+%%--------------------------------------------------------------------
+-spec is_feature_code(wh_json:object()) -> boolean().
+is_feature_code(JObj) ->
+    case wh_json:get_value(?FEATURE_CODE, JObj) of
```
fenollp: `wh_json:get_is_true(?FEATURE_CODE, JObj, 'true')`

applications/crossbar/src/modules/cb_callflows.erl
```diff
@@ -442,12 +443,14 @@ check_uniqueness(Numbers, JObjs, Context) ->
...
 check_numbers_uniqueness([], _, Context) -> Context;
 check_numbers_uniqueness([Number|Numbers], JObjs, Context) ->
+    io:format("MARKER:cb_callflows.erl:445 ~p~n", [Number]),
     case lists:dropwhile(fun(J) -> is_number_unique(J, Number) end
                          ,JObjs
                         )
     of
         [] -> check_numbers_uniqueness(Numbers, JObjs, Context);
         [JObj|_] ->
+            io:format("MARKER:cb_callflows.erl:451 ~p~n", [JObj]),
```
fenollp: rm these

core/kazoo_documents-1.0.0/src/kzd_callflow.erl
macpie: Can't do that but I can add a default 'false' value

fenollp:
@macpie Can you add a comment in the code where you can't do get_is_true ?

macpie:
@fenollp Pb is that a wh_json:is_true on a json object does not return true

fenollp:
@macpie No I get it, it's just that it's not something I get from just the code, so a comment'd be nice.

PR #1461, opened 2015-11-25
KAZOO-4405 allow konami offnet and prevent start multiple metaflows

lazedo:


applications/callflow/src/cf_util.erl
```diff
@@ -825,10 +825,16 @@ encryption_method_map(JObj, Endpoint) ->
                          ).
 
 -spec maybe_start_metaflows(whapps_call:call(), wh_json:objects()) -> 'ok'.
+-spec maybe_start_metaflows(whapps_call:call(), wh_json:object(), api_binary()) -> 'ok'.
 -spec maybe_start_metaflow(whapps_call:call(), wh_json:object()) -> 'ok'.
+
 maybe_start_metaflows(Call, Endpoints) ->
+    maybe_start_metaflows(Call, Endpoints, whapps_call:kvs_fetch(<<"Metaflow-App">>, Call)).
```
fenollp: I'm curious: why is that a `kvs_fetch` call and the previous a `custom_channel_var` one?

lazedo: because its a bug :) nice catch

PR #1453, opened 2015-11-20
KAZOO-4398 - Generate recording_url

pramsky:
Generates recording url using media_transfer_destination and media_file
from the channel_vars in the cdr.
This displays the recording_url when called via API, or exported to csv.

PR #1463, opened 2015-11-27
KAZOO-4398 - Generate recording_url

pramsky:
Generates recording url using media_transfer_destination and media_file
  from the channel_vars in the cdr.
 This displays the recording_url when called via API, or exported to csv.
 Thanks to @fenollp for the suggestions

### DIFF
--- a/applications/crossbar/src/modules/cb_cdrs.erl
+++ b/applications/crossbar/src/modules/cb_cdrs.erl
@@ -520,11 +520,23 @@ col_account_call_type(JObj, _Timestamp) -> wh_json:get_value([<<"custom_channel_
 col_rate(JObj, _Timestamp) -> wh_util:to_binary(wht_util:units_to_dollars(wh_json:get_value([<<"custom_channel_vars">>, <<"rate">>], JObj, 0))).
 col_rate_name(JObj, _Timestamp) -> wh_json:get_value([<<"custom_channel_vars">>, <<"rate_name">>], JObj, <<>>).
 col_bridge_id(JObj, _Timestamp) -> wh_json:get_value([<<"custom_channel_vars">>, <<"bridge_id">>], JObj, <<>>).
-col_recording_url(JObj, _Timestamp) -> wh_json:get_value([<<"recording_url">>], JObj, <<>>).
+col_recording_url(JObj, _Timestamp) -> recording_url(wh_json:get_value([<<"custom_channel_vars">>, <<"media_transfer_destination">>], JObj, <<>>), wh_json:get_value([<<"custom_channel_vars">>, <<"media_name">>], JObj, <<>>)).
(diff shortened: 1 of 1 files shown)


fenollp:
Follow up of #1453

applications/crossbar/src/modules/cb_cdrs.erl
```diff
@@ -520,12 +520,24 @@ col_account_call_type(JObj, _Timestamp) -> wh_json:get_value([<<"custom_channel_
...
 col_reseller_cost(JObj, _Timestamp) -> wh_util:to_binary(reseller_cost(JObj)).
 col_reseller_call_type(JObj, _Timestamp) -> wh_json:get_value([<<"custom_channel_vars">>, <<"reseller_billing">>], JObj, <<>>).
 
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%% generates a recording_url entry from media_transfer_destination and media_file
+%% @end
+%%--------------------------------------------------------------------
+-spec recording_url(api_binary(), api_binary()) -> api_binary().
```
fenollp: Spec should be `-spec recording_url(ne_binary(), ne_binary()) -> api_binary().`

pramsky: Got it. I'll updated and issue another pull request.

fenollp:
wait sorry not `ne_binary()` nor `api_binary()` but `binary()`. I was very tired…

k-anderson:
@fenollp what was recording_url? Please ensure this is a backward compatible change, if not we will need to make it a new parameter in the result set so as to not disrupt those who might have already coded to the current value of recording_url.

fenollp:
https://github.com/2600hz/kazoo/commit/9925c4c39291d7b55f054a4aa611587640de2610
@k-anderson not sure how to make sure this doesn't break something the same way above commit might have broken stuff too..

@pramsky Did you try using the code in the commit I mention? What is your experience with it?

pramsky:
@fenollp 
With the earlier commit, in the cdrs the recording_url was always blank. 

I could go into the details of the cdr and see the media_transfer_destination and media_file for recording in there. The recording_url was not populated in the API result or in the exported csv.

k-anderson:
@pramsky howdy!  Unfortunately as we progress into 4.0 this is no longer a valid issue as media_transfer_destination does not exist anymore.  However, as this is existing functionality I have asked @lazedo to ensure this will be available with the new recording process.  Please watch the announcements because the new recording process allows multiple recordings per call and as such we may change the recording_url to an array!

https://github.com/2600hz/kazoo/blob/master/ANNOUNCEMENTS.md

pramsky:
@k-anderson 

Thanks for the update =) Looking forward to 4 =)

PR #1465, opened 2015-12-01
KAZOO-4321: Truncate prepend caller id to not exceed characters long limit

icehess:


applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -1366,9 +1367,14 @@ get_caller_id_number(Call) ->
...
     case whapps_call:kvs_fetch('prepend_cid_number', Call) of
         'undefined' -> CallerIdNumber;
-        Prepend -> <<(wh_util:to_binary(Prepend))/binary, CallerIdNumber/binary>>
+        Prepend -> Pre = <<(wh_util:to_binary(Prepend))/binary, CallerIdNumber/binary>>,
+                   truncate_caller_id(Pre)
     end.
 
+-spec truncate_caller_id(ne_binary()) -> ne_binary().
+truncate_caller_id(Prepend) ->
+    <<Prepend:35/binary>>.
```
jamesaimonetti: I'm confused on how this is supposed to work:

```erlang
5> P = <<"abc">>.
<<"abc">>
6> <<P:35/binary>>.
** exception error: bad argument
```

So what happens when the prepended caller ID isn't 35 characters? Or am I missing a step earlier?

icehess: It was my mistake, just pushed an update which checks the boundry

applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -1366,7 +1367,8 @@ get_caller_id_number(Call) ->
     CallerIdNumber = whapps_call:caller_id_number(Call),
     case whapps_call:kvs_fetch('prepend_cid_number', Call) of
         'undefined' -> CallerIdNumber;
-        Prepend -> <<(wh_util:to_binary(Prepend))/binary, CallerIdNumber/binary>>
+        Prepend -> Pre = <<(wh_util:to_binary(Prepend))/binary, CallerIdNumber/binary>>,
+                   wh_util:truncate_right_binary(Pre, 35)
```
fenollp: Yeah how about just a local private function `truncate_right/1` which would just be your `wh_util:truncate_right_binary/2`? Let's not add more useless stuff to wh_util…

icehess: We could and I did it in a local function first, but I thought it should be nice to have it as a generic function so other modules can using it if they need it.

fenollp: Well, I'm happy with it if you also find more than one other place in master to use it, and call the function there too.

icehess: How about these places:

applications/stepswitch/src/stepswitch_util.erl: Line 129
core/whistle-1.0.0/src/wh_json.erl: Line 123
applications/call_inspector/src/parsers/ci_parser_freeswitch.erl: Line 340
applications/call_inspector/src/parsers/ci_parsers_util.erl: Line 108

Using a meaningful function name helps code reading and makes coder’s intention clear.

fenollp: Could you make the changes there as part of this PR?

core/kazoo_documents-1.0.0/src/kzd_schema.erl
```diff
@@ -0,0 +1,58 @@
...
+		'undefined' ->
+			'undefined';
+		SchemaJObj ->
+			get_path(Key, SchemaJObj)
+	end.
+
+%%% Keywords
+%%% ==================================
+%% String
+-spec max_length(wh_json:key(), api_object()) -> wh_json:object() | 'undefined'.
```
fenollp: BTW `wh_json:object() | 'undefined'` is `api_object()`.

core/kazoo_documents-1.0.0/src/kzd_schema.erl
```diff
@@ -0,0 +1,58 @@
...
+max_length(Key, Schema) ->
+	Properties = properties(Key, Schema),
+	wh_json:get_value(?SCHEMA_KEYWORDS_MAXLENGTH, Properties).
```
fenollp: `properties/1` can return `undefined`, and given the spec for `max_length` I think you want the `get_value` call to return either an object or `undefined`, not crash.
My thought is that `properties` should return an object, be it an empty one and `get_value` will give `api_object()`.

core/kazoo_documents-1.0.0/src/kzd_schema.erl
```diff
@@ -0,0 +1,58 @@
...
+
+%%% Internal function
+%%% =================
+get_path(K, P) when not is_list(K) ->
+	get_path([K], P);
+get_path([], P) ->
+	P;
+get_path([K | Ks], P) ->
+	get_path(Ks, get_value(K, P)).
```
fenollp: This code looks awfully like what `wh_json:get_value/2` already does (supporting a list of keys forming a path).
Do you think you could use it instead?

core/kazoo_documents-1.0.0/src/kzd_schema_caller_id.erl
```diff
@@ -0,0 +1,19 @@
...
+%%%-------------------------------------------------------------------
+-module(kzd_schema_caller_id).
+
+-define (SCHEMA_NAME, <<"caller_id">>).
+-define (SCHEMA_PATH_PATTERN(K1, K2), [<<"properties">>, K1, <<"properties">>, K2]).
+-define (SCHEMA_EXTERNAL_NAME, ?SCHEMA_PATH_PATTERN(<<"external">>, <<"name">>)).
+
+-export([external_name_max_length/0]).
+
+-spec external_name_max_length() -> wh_json:object() | 'undefined'.
```
fenollp: spec

fenollp:
Note that `binary:part(Bin, {0, Len})` [is the same as](http://erldocs.com/current/stdlib/binary.html?i=0&search=binary:part/2#part/2) `binary:part(Bin, 0, Len)`, thus there are more places in the codebase that could benefit from the functions you introduce.

fenollp:
Forgot to mention that Travis is unhappy because you need to add lager to `PA` in `kazoo_documents`

icehess:
Weird, lager is in PA already.

lazedo:
@icehess to use lager, you need to include wh_log.hrl
`-include_lib("whistle/include/wh_log.hrl").`
or .. include kz_documents.hrl which already has it.

PR #1466, opened 2015-12-02
KAZOO-4278: fix database auto cache

macpie:
…ase', Db} to remove changes

### DIFF
--- a/core/whistle-1.0.0/src/wh_cache.erl
+++ b/core/whistle-1.0.0/src/wh_cache.erl
@@ -577,36 +577,23 @@ get_props_origin(Props) -> maybe_add_db_origin(props:get_value('origin', Props))
 -spec maybe_add_db_origin(wh_proplist()) -> 'undefined' | origin_tuple() | origin_tuples().
 maybe_add_db_origin(Props) when is_list(Props) -> maybe_add_db_origin(Props, []);
 maybe_add_db_origin({'db', Db}) ->
-    [{'db',Db}, {'database', Db}];
+    [{'db',Db}, {'type', <<"database">>, Db}];
 maybe_add_db_origin({'db', Db, Id}) ->
-    [{'db',Db,Id}, {'database', Db}];
+    [{'db',Db,Id}, {'type', <<"database">>, Db}];
 maybe_add_db_origin(Props) -> Props.
 
 -spec maybe_add_db_origin(wh_proplist(), wh_proplist()) -> 'undefined' | origin_tuple() | origin_tuples().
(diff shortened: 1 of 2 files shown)


core/whistle-1.0.0/src/wh_cache.erl
```diff
@@ -602,7 +602,11 @@ maybe_erase_changed(Db, 'database', _Id, Tab) ->
     erase_changed(Objects, [], Tab);
 
 maybe_erase_changed(Db, Type, Id, Tab) ->
-    MatchSpec = [{#cache_obj{origin = {'db', Db}, _ = '_'}
+    MatchSpec = [{#cache_obj{origin = {'database', Db}, _ = '_'}
```
lazedo: @macpie why do you think this should be done here ?

macpie: Don't have a specific reason for that spot other than this is fixing the issue

lazedo: @macpie well, if you clear all the cache, you'll never have any cache problems, right ? check the origin bindings that were created. this changed will invalidate every cache entry for the same database.

for instance, 2 users with simple binding
{mydb, user, 1}
{mydb, user, 2}
would create 4 entries in cache 
{mydb, user, 1} + pointer{database, mydb}
{mydb, user, 2} + pointer{database, mydb}

your change will invalidate {mydb, user, 2} if {mydb, user, 1} is deleted.

macpie: Ok, so you are saying that I should fix the "data" when it is enter in the cache?

lazedo: yes, for instance, https://github.com/2600hz/kazoo/blob/master/core/whistle_apps-1.0.0/src/whapps_util.erl#L400-L405

seems wrong to me, we are querying accounts db but caching the accountdb retrieved.
in this case a simple {'db', ?WH_ACCOUNTS, <<"account">>} is more correct because it would also invalidate when a new account is created which it doesn't right now

lazedo: and there's actually a missed cache happening on database delete. i'll add that do the branch.

lazedo:
@macpie the function clause (checking type for 'database') removed was wrong. i changed the database auto cache and removed the clause, i think its simpler this way. let me know if you need any explanation on how caches work. i also fixed the `whapps_util:cache/2`

macpie:
Thanks you @lazedo! I will test that today.

macpie:
OK working thanks again @lazedo

PR #1468, opened 2015-12-02
KAZOO-4417 xmpp new library and some fixes

lazedo:
and some fax fixes


.travis.yml
```diff
@@ -33,7 +33,7 @@ before_install:
 script: ERLC_OPTS="$ERLC_OPTS" make compile
 
 after_success:
-  - if [[ 0 -ne `git status --porcelain -- applications/ core/ deps/ | wc -l` ]]; then echo Unstaged changes! && git status --porcelain && exit 1; fi
```
fenollp: Vetoooooo!

fenollp: Actually no veto. That should have been removed after `.gitignore` sanitization.

applications/fax/src/wapi_xmpp.erl
```diff
@@ -94,3 +99,19 @@ publish_event(Event, ContentType) ->
...
+    regexp_get(JID, <<"^([^/]*)">>).
+
+jid_username(JID) ->
+    regexp_get(JID, <<"^([^@]*)">>).
+
+jid_server(JID) ->
+    regexp_get(JID, <<"^[^@]*[@]([^/]*)">>).
+
+jid_resource(JID) ->
+    regexp_get(JID, <<"^[^/]*[/](.*)">>).
```
fenollp: Do you have functional tests for these?

lazedo: JID is always user@domain/resource

fenollp: Why use regexps then?
```erlang
[User,Domain,Resource] = binary:split(JID, [<<"@">>, <<"/">>], ['global'])
```

lazedo: its an option, does it not work ?

fenollp:
@lazedo What's the repos you got the deps from? I'd like to keep tabs on these for easier updates in the future

lazedo:
@fenollp https://github.com/esl/escalus as base

PR #1471, opened 2015-12-03
KAZOO-4413 upgrade poolboy to 1.5.1

lazedo:


deps/poolboy-1.5.1/rebar.config
```diff
@@ -0,0 +1,7 @@
+{erl_opts, [
+  debug_info,
+  {platform_define, "^R", pre17}
```
fenollp: Does this mean we can't use it with OTP ≥ 17?

deps/poolboy-1.5.1/Makefile
```diff
@@ -0,0 +1,34 @@
+REBAR = $(shell command -v rebar || echo ./rebar)
```
fenollp: Use the rebar bin we already have in /utils

fenollp: Same for #1468 : remove the rebar binaries

lazedo: @fenollp rebar came with the repo, i'll remove it if you update our rebar in utils which is quite old :)

fenollp: I'm fine with that

PR #1474, opened 2015-12-04
KAZOO-4415: Add subscribe/unsubscribe to multiple bindings

macpie:


applications/blackhole/src/bh_context.erl
```diff
@@ -197,16 +170,38 @@ set_account_id(#bh_context{}=Context, AcctId) ->
...
 %%--------------------------------------------------------------------
 -spec bindings(context()) -> ne_binaries().
-bindings(#bh_context{bindings=Bindings}) ->
-    Bindings.
+bindings(#bh_context{bindings=Bds}) ->
+    Bds.
+
+-spec bindings_from_json(wh_json:object()) -> ne_binary() | ne_binaries().
+bindings_from_json(JObj) ->
+    wh_json:get_first_defined([<<"binding">>, <<"bindings">>], JObj).
```
fenollp: Why not both?

applications/blackhole/src/blackhole_socket_callback.erl
```diff
@@ -142,13 +161,13 @@ unsubscribe(Context, JObj, SessionPid, SessionId, 'true') ->
 %% @end
 %%--------------------------------------------------------------------
 -spec unsubscribe_for_all(bh_context:context(), wh_json:object(), ne_binary(), ne_binary()) -> cb_return().
-unsubscribe_for_all(Context, JObj, SessionPid, SessionId) ->
-    Context1 = bh_context:unsubscribe(Context, JObj),
+unsubscribe_for_all(Context, _JObj, SessionPid, SessionId) ->
+    Context1 = bh_context:set_bindings(Context, []),
```
fenollp: This doesn't unsubscribe from the bindings, does it?

applications/blackhole/src/bh_context.erl
```diff
@@ -197,16 +170,38 @@ set_account_id(#bh_context{}=Context, AcctId) ->
...
+remove_binding(#bh_context{bindings=Bds}=Context, Binding) ->
+    Context#bh_context{bindings=lists:delete(Binding, Bds)}.
+
+%%--------------------------------------------------------------------
+%% @public
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec already_binded(context(), ne_binary()) -> context().
+already_binded(#bh_context{bindings=Bds}, Binding) ->
```
fenollp: `is_bound`

applications/blackhole/src/blackhole_socket_callback.erl
```diff
@@ -91,28 +91,47 @@ handle_info(_SessionPid, _SessionId, _Msg, Context) ->
...
+    Bindings = bh_context:bindings_from_json(JObj),
+    check_bindings(Context, JObj, Bindings).
 
+-spec check_bindings(bh_context:context(), wh_json:object(), ne_binaries() | ne_binary()) -> cb_return().
+check_bindings(Context, _JObj, []) ->
+    {'ok', Context};
+check_bindings(Context, JObj, [Binding|Bds]) ->
+    {'ok', Context1} = check_bindings(Context, JObj, Binding),
+    check_bindings(Context1, JObj, Bds);
+check_bindings(Context, JObj, Binding) ->
```
fenollp: Maybe a `Binding = ?NE_BINARY` will help Dialyzer; here or in `bh_context:already_bindeded/2`.

applications/blackhole/src/bh_context.erl
macpie: There is no point of using binding in you are already using bindings

applications/blackhole/src/blackhole_socket_callback.erl
macpie: Yes it does remove all of them

applications/blackhole/src/bh_context.erl
macpie: ok

applications/blackhole/src/bh_context.erl
fenollp: Could we only support `bindings`? Allowing both when the difference is that one takes an array of what the other takes, I find that very weird.

applications/blackhole/doc/index.md
```diff
@@ -32,37 +34,90 @@ From here, you can write your own Javascript callbacks, triggered everytime a re
...
+                console.log('=ERROR REPORT==== ' + Events[idx]);
+            });
+        }
     </script>
   </body>
 </html>
 ```
-You can also add a friendly name and some metadata to any subscribe command.
 
+You can add one or multiple bindings by using:
```
jamesaimonetti: You should add a note that `binding` is taken before `bindings` in case someone sets both.

applications/blackhole/src/bh_context.erl
jamesaimonetti: This function is named `bindings_from_json` and yet can return a single binding. Why not check `binding` first and wrap it in a list; otherwise check `bindings`, returning the empty list as a default (as is, later usage will crash if neither key is included).

applications/blackhole/src/blackhole_socket_callback.erl
```diff
@@ -91,28 +91,47 @@ handle_info(_SessionPid, _SessionId, _Msg, Context) ->
...
-    subscribe(Context, JObj, Binding, Module).
+maybe_subscribe(Context, JObj, 'true') ->
+    Bindings = bh_context:bindings_from_json(JObj),
+    check_bindings(Context, JObj, Bindings).
 
+-spec check_bindings(bh_context:context(), wh_json:object(), ne_binaries() | ne_binary()) -> cb_return().
+check_bindings(Context, _JObj, []) ->
+    {'ok', Context};
+check_bindings(Context, JObj, [Binding|Bds]) ->
+    {'ok', Context1} = check_bindings(Context, JObj, Binding),
```
jamesaimonetti: This function is called `check_bindings/3` but only checks a single binding. This is bad naming. If you only allow `Bindings` to be a list of binaries, you don't need to overload `check_bindings/3` to handle the single binary case. Move the last clause into a `check_binding/3`, or better yet, use a fold!

applications/blackhole/src/blackhole_socket_callback.erl
```diff
@@ -91,28 +91,47 @@ handle_info(_SessionPid, _SessionId, _Msg, Context) ->
...
+    check_bindings(Context1, JObj, Bds);
+check_bindings(Context, JObj, Binding) ->
+    case bh_context:is_bound(Context, Binding) of
+        'true' ->
+            ErrorJObj = wh_json:from_list([
+                {<<"message">>, <<"binding already in use">>}
+                ,{<<"cause">>, Binding}
+            ]),
+            {'ok', blackhole_util:respond_with_error(Context, <<"error">>, ErrorJObj)};
```
jamesaimonetti: I'm confused why you emit an error to the websocket and return an unmodified Context record? Why doesn't `respond_with_error/3` do just that, write to the websocket?

fenollp:
@macpie needs rebase plz

PR #1478, opened 2015-12-08
KAZOO-4427 skip already hangup

lazedo:


fenollp:
What about [this one](https://github.com/2600hz/kazoo/blob/KAZOO-4427/applications/ecallmgr/src/ecallmgr_call_control.erl#L1126) and [this one](https://github.com/2600hz/kazoo/blob/KAZOO-4427/applications/ecallmgr/src/ecallmgr_call_control.erl#L1140)?

lazedo:
what about ? what were you expecting to see ?

fenollp:
Ah ok, so the two calls I linked are sending an error back but we don't know if channel is already hangup, right? so they rightly shouldn't be `maybe_*` calls then?

PR #1479, opened 2015-12-08
KAZOO-4426 log changes & utilities

lazedo:


core/whistle_apps-1.0.0/src/whapps_maintenance.erl
```diff
@@ -1104,6 +1104,13 @@ maybe_delete_db(Database) ->
...
                             'ok'.
+purge_doc_type(Type, Account)
+  when is_binary(Type);
+       is_binary(Account) ->
+    purge_doc_type(Type
+                   ,Account
+                   ,whapps_config:get_integer(?SYSCONFIG_COUCH, <<"default_chunk_size">>, ?MILLISECONDS_IN_SECOND)
+                  );
 purge_doc_type([], _Account) -> 'ok';
 purge_doc_type([Type|Types], Account) ->
```
fenollp: I would add a when `is_binary(Account)` here, to make more sense of the code below.

PR #1481, opened 2015-12-08
Fix "account_update" to check revs in case it got out of sync

macpie:


core/whistle-1.0.0/src/wh_util.erl
```diff
@@ -478,9 +478,12 @@ account_update(JObj) ->
     case couch_mgr:save_doc(AccountDb, JObj) of
         {'error', _R}=E -> E;
         {'ok', SavedJObj} ->
-            case couch_mgr:save_doc(?WH_ACCOUNTS_DB, SavedJObj) of
-                {'error', _R}=E -> E;
-                {'ok', _} -> 'ok'
+            AccountId = wh_doc:id(SavedJObj),
+            case couch_mgr:lookup_doc_rev(?WH_ACCOUNTS_DB, AccountId) of
```
jamesaimonetti: I wonder if you read the code for [`couch_util:do_ensure_saved`](https://github.com/2600hz/kazoo/blob/master/core/whistle_couch-1.0.0/src/couch_util.erl#L681-L692)? It does exactly this:

```erlang
do_ensure_saved(#db{}=Db, Doc, Opts) ->
    case do_save_doc(Db, Doc, Opts) of
        {'ok', _}=Ok -> Ok;
        {'error', 'conflict'} ->
            case do_fetch_rev(Db, wh_doc:id(Doc)) of
                {'error', 'not_found'} ->
                    do_ensure_saved(Db, wh_doc:delete_revision(Doc), Opts);
                Rev ->
                    do_ensure_saved(Db, wh_doc:set_revision(Doc, Rev), Opts)
            end;
        {'error', _}=E -> E
    end.
```

So why not just call `couch_mgr:ensure_saved/2` from the outset?

fenollp: NB: `git grep -C4 wh_doc:set_revision` reveals a couple potential uses too

PR #1482, opened 2015-12-10
KAZOO-4431 use kazoo dptools

lazedo:


applications/ecallmgr/src/freeswitch.erl
```diff
@@ -188,6 +189,39 @@ bgapi(Node, Cmd, Args, Fun) when is_function(Fun, 2) ->
         {'api', Result} -> Result
     end.
 
+-spec bgapi(atom(), atom(), string() | binary(), fun(), list()) ->
+                   {'ok', binary()} |
+                   {'error', 'timeout' | 'exception' | binary()}.
+bgapi(Node, Cmd, Args, Fun, CallBackParams) when is_function(Fun, 3) ->
+    Self = self(),
+    _ = wh_util:spawn(
```
fenollp: Why do you `spawn` here, since you wait for it anyway?
Also, no timeouts on the receives?

lazedo: @fenollp if the bgapi start is ok, we don't wait. the timeout is on the `gen_server:call`

PR #1484, opened 2015-12-11
KAZOO-4424: Try to polling ACL modification before flushing

icehess:


applications/ecallmgr/src/ecallmgr_maintenance.erl
```diff
@@ -328,6 +328,36 @@ remove_acl(Name, 'false') ->
...
+maybe_reload_acls(_Name, _Action, 0) ->
+    io:format("Timeout during updating ACLs, try reload ACLs and flush config later:~n"),
+    io:format("Use these commands:~n"),
+    io:format("# sup whapps_config flush~n"),
+    io:format("# sup -necallmgr ecallmgr_maintenance reload_acls~n"),
+    'no_return';
+maybe_reload_acls(Name, Action, Tries) ->
+    case has_acl(Name, Action, get_acls()) of
+        'true' ->
+            io:format("flushing"),
```
jamesaimonetti: This doesn't seem particularly user-friendly. Either remove it or make it more useful to sysadmins (they ultimately consume these log messages and should be our target audience).

PR #1486, opened 2015-12-14
Add AccountId to debug on non-200 responses

neurosys-zero:
One less step when debugging account webhooks if receiving non-200
responses.

### DIFF
--- a/applications/webhooks/src/webhooks_util.erl
+++ b/applications/webhooks/src/webhooks_util.erl
@@ -211,7 +211,7 @@ fire_hook(_JObj, Hook, _URI, _Method, _Retries, {'ok', "200", _, _RespBody}) ->
     successful_hook(Hook);
 fire_hook(_JObj, Hook, _URI, _Method, Retries, {'ok', RespCode, _, RespBody}) ->
     _ = failed_hook(Hook, Retries, RespCode, RespBody),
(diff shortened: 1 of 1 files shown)


applications/webhooks/src/webhooks_util.erl
```diff
@@ -211,7 +211,7 @@ fire_hook(_JObj, Hook, _URI, _Method, _Retries, {'ok', "200", _, _RespBody}) ->
     successful_hook(Hook);
 fire_hook(_JObj, Hook, _URI, _Method, Retries, {'ok', RespCode, _, RespBody}) ->
     _ = failed_hook(Hook, Retries, RespCode, RespBody),
-    lager:debug("non-200 response code: ~s", [RespCode]);
+    lager:debug("non-200 response code: ~s on account ~s", [RespCode, AccountId]);
```
jamesaimonetti: You need to pattern match the account id out of the Hook variable. Something like:

```erlang
fire_hook(_JObj, #webhook{account_id=AccountId}=Hook, ...) ->
```

Otherwise, you're referring to an unbound variable (no value assigned to that name). Make sense?

lazedo:
@neurosys-zero merged in https://github.com/2600hz/kazoo/issues/1554

PR #1487, opened 2015-12-14
KAZOO 4436

icehess:


applications/ecallmgr/src/ecallmgr_fs_authz.erl
```diff
@@ -83,6 +84,13 @@ maybe_authorize_channel(Props, Node) ->
             end
     end.
 
+-spec is_referred(wh_proplist()) -> 'true' | 'false'.
+is_referred(Props) ->
+  case kzd_freeswitch:ccv(Props, [<<"Referred-By">>, <<"Referred-To">>]) of
```
jamesaimonetti: I'm confused how this can ever return a two-tuple? Did you mean to put those in separate calls to `kzd_freeswitch:ccv/2`?

applications/ecallmgr/src/ecallmgr_fs_authz.erl
```diff
@@ -83,6 +84,13 @@ maybe_authorize_channel(Props, Node) ->
...
+    {'undefined', 'undefined'} -> 'false';
```
jamesaimonetti: Also, please watch your formatting. 4 spaces is the proper indentation.

applications/ecallmgr/src/ecallmgr_fs_authz.erl
icehess: Aah I see, I thought kzd_freeswitch:ccv/2 is passing Key as is. I’ll fix it.

applications/ecallmgr/src/ecallmgr_fs_authz.erl
```diff
@@ -83,6 +84,15 @@ maybe_authorize_channel(Props, Node) ->
             end
     end.
 
+-spec is_referred(wh_proplist()) -> 'true' | 'false'.
```
fenollp: `true | false` is `boolean()`

k-anderson: @fenollp Incorrect, see http://erlang.org/doc/reference_manual/typespec.html

boolean() is defined as 'false' | 'true'

fenollp: Well

applications/callflow/src/cf_route_req.erl
```diff
@@ -277,14 +277,37 @@ update_call(Flow, NoMatch, ControllerQ, Call) ->
...
+-spec fix_rerouted_call(api_binary(), whapps_call:call()) -> whapps_call:call().
+fix_rerouted_call(ReroutedBy, Call) ->
+    [Username|_] = binary:split(ReroutedBy, <<"@">>),
+    case cf_util:endpoint_id_by_sip_username(whapps_call:account_db(Call), Username) of
+        {'ok', EndpointId} ->
+            C = whapps_call:set_authorization(<<"device">>, EndpointId, Call),
+            OwnerId = cf_attributes:owner_id(C),
+            whapps_call:set_owner_id(OwnerId, C);
+        {'error', 'not_found'} ->
+            Routines = [fun(C) -> whapps_call:set_authorization(<<"undefined">>, <<"undefined">>, C) end
```
jamesaimonetti: I don't think we want to set these values to "undefined"; perhaps your intent was to unset them from the CCVs?

applications/callflow/src/cf_route_req.erl
```diff
@@ -277,14 +277,37 @@ update_call(Flow, NoMatch, ControllerQ, Call) ->
...
+                        ,fun(C) -> whapps_call:set_owner_id(<<"undefined">>, C) end
+                        ,fun(C) -> whapps_call:set_account_id(<<"undefined">>, C) end
+                       ],
+            lists:foldl(fun(F, C) -> F(C) end, Call, Routines)
```
jamesaimonetti: While you're in here, look at `whapps_call:exec/2` and consider using tuples instead of anonymous functions:

```erlang
whapps_call:exec(
    [{fun whapps_call:set_authorization/3, 'undefined', 'undefined'}
     ,{fun whapps_call:set_owner_id/2, 'undefined}
     ,{fun whapps_call:set_account_id/2, 'undefined'}
    ]
    ,Call
)
```

Something along those lines.

applications/callflow/src/cf_route_req.erl
icehess: Do you mean set them to <<>> or remove them from CCVs(and if so how can I removed from whapps_call:call record?) ? My intention was to unset them somehow.

fenollp: Yes it appears `whapps_call_command` does not provide an `unset/3` whereas it has `set/3`. @jamesaimonetti ?

jamesaimonetti: Sounds like adding unset, or setting to `'undefined'` is in order!

icehess: So for unsetting what should I do? Setting to an empty binary or setting to <<‘undefined'>>?

fenollp: > this or that?

Yes.

(Hehe. I don't know. Try it?)

jamesaimonetti: @icehess `'undefined'` as an atom, when calling the setter, should effectively be an unset.

You'll need to check the `set_*` functions to make sure they'll take the atom.

applications/ecallmgr/src/ecallmgr_fs_authz.erl
```diff
@@ -83,6 +84,15 @@ maybe_authorize_channel(Props, Node) ->
             end
     end.
 
+-spec is_referred(wh_proplist()) -> boolean().
+is_referred(Props) ->
+    ReferredBy = kzd_freeswitch:ccv(Props, <<"Referred-By">>),
```
jamesaimonetti: Could do:

```erlang
kzd_freeswitch:ccv(Props, <<"Referred-By">>) =/= 'undefined'
    orelse kzd_freeswitch:ccv(Props, <<"Referred-To">>) =/= 'undefined'
```

Save yourself some bindings

jamesaimonetti: Or any variation of boolean logic you wish...

core/whistle_apps-1.0.0/src/whapps_call.erl
```diff
@@ -868,6 +870,25 @@ set_from_tag(FromTag, #whapps_call{}=Call) when is_binary(FromTag) ->
...
+remove_custom_channel_vars(Keys, #whapps_call{}=Call) ->
+    whapps_call_command:set(wh_json:from_list([{Key, <<>>} || Key <- Keys]), 'undefined', Call),
+    handle_ccvs_remove(Keys, Call).
+
+-spec handle_ccvs_remove(wh_json:object(), whapps_call:call()) -> whapps_call:call().
+handle_ccvs_remove(Keys, #whapps_call{ccvs=CCVs}=Call) ->
+    lists:foldl(fun(Key, C) ->
+        case props:get_value(Key, ?SPECIAL_VARS) of
+          'undefined' -> C;
+          Index -> setelement(Index, C, 'undefined')
```
jamesaimonetti: I don't like using setelement on a record (breaks the abstraction), but I'm not thinking of a better way to do this...so, just stating it for the record.

core/whistle_apps-1.0.0/src/whapps_call.erl
```diff
@@ -868,6 +870,25 @@ set_from_tag(FromTag, #whapps_call{}=Call) when is_binary(FromTag) ->
 from_tag(#whapps_call{from_tag=FromTag}) ->
     FromTag.
 
+-spec remove_custom_channel_var(ne_binary(), whapps_call:call()) -> whapps_call:call().
+remove_custom_channel_var(Key, #whapps_call{}=Call) ->
+    whapps_call_command:set(wh_json:from_list([{Key, <<>>}]), 'undefined', Call),
```
jamesaimonetti: Basically duplicated code, I'd just call `remove_custom_channel_vars([Key], Call).`

PR #1489, opened 2015-12-14
KAZOO-3635: be notified if auto-provisioning fails

macpie:


core/whistle_apps-1.0.0/src/whapps_alert.erl
```diff
@@ -0,0 +1,130 @@
...
+    kzd_alert:fetch(AlertId).
+
+%%--------------------------------------------------------------------
+%% @public
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec create(ne_binary(), ne_binary(), wh_json:object(), wh_json:object()) -> {'ok', wh_json:object()} |
+                                                                              {'error', any()}.
+-spec create(ne_binary(), ne_binary(), wh_json:object()
```
fenollp: Why expect `ne_binary()`es when you match `undefined`?

core/whistle_apps-1.0.0/src/whapps_alert.erl
```diff
@@ -0,0 +1,130 @@
...
+        'true' ->
+            couch_mgr:save_doc(?WH_ALERTS_DB, JObj)
+    end.
+
+%%--------------------------------------------------------------------
+%% @public
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec delete(wh_json:object()) -> {'ok', wh_json:object()} | {'error', any()}.
```
fenollp: what about the binary that you match?

core/whistle_apps-1.0.0/src/whapps_alert.erl
```diff
@@ -0,0 +1,130 @@
...
+delete(AlertId) when is_binary(AlertId) ->
+    case ?MODULE:fetch(AlertId) of
```
fenollp: Why `?MODULE`? Do you really intend for this code to use only the latest version of fcall?

core/kazoo_documents-1.0.0/src/kzd_alert.erl
```diff
@@ -0,0 +1,254 @@
...
+fetch(<<_/binary>> = AlertId) ->
+    couch_mgr:open_cache_doc(?WH_ALERTS_DB, AlertId).
+
+%%--------------------------------------------------------------------
+%% @public
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec title() -> ne_binary().
+-spec title(doc()) -> api_binary().
```
jamesaimonetti: How can we help you write proper specs? I feel like your PRs too often require us to go through and tell you which ones are incorrect, which is unfair to our time. You have the tools and intelligence to do it yourself.

The first thing I do when I feel "done" is to dialyze the code I've made changes to:
```bash
kazoo$> dialyzer --plt .kazoo.plt core/kazoo_documents-1.0.0/ebin/kzd_alert.beam
```

Once I feel like its internally consistent, I dialyze with the modules referenced:

```bash
kazoo$> dialyzer --plt .kazoo.plt core/kazoo_documents-1.0.0/ebin/kzd_alert.beam core/whistle-1.0.0/ebin/wh_json.beam ...
```

If both passes go successfully, I'm feeling good. Dialyzer thinks things are both internally consistent and consistent with the modules I'm calling off to, so I can be reasonably confident I've wired things up right.

Does that make sense? Do you have questions about how all this works? Please ask if so; we all get to learn more that way.

Same goes for @lazedo @lfurrea @icehess @k-anderson @fenollp; if anyone has questions about how others work, how to do something more efficiently, how to take advantage of tooling to make life easier, etc, lets share those questions and ideas. Write engineering@ so we can all participate in getting better.

macpie: I agree and I asked multiple time for these commands. Just keep me in the loop on these tools so that I do not waste anyone's time.

PR #1491, opened 2015-12-15
Allow "X-Forward-For" only from allowed proxies

skorobkov:
Any client can set "X-Forward-For" header, and Crossbar accept this request, and set #context.client_ip to value of this header.
This allow avoiding token buckets and authenticate with cb_ip_auth from any address.


jamesaimonetti:
Good check to have! Thanks for the PR.

PR #1493, opened 2015-12-17
KAZOO-4451 - check_bookkeeper incorrect

bluesandnz:
Balance - Amount should be TRUE if greater than or equal to Zero


bluesandnz:
I will close and reopen with corrected code.

PR #1494, opened 2015-12-17
KAZOO-4451 - check_bookkeeper incorrect

bluesandnz:
Correctly typo in previous pull request.

Should be
Balance - Amount >= 0;

Now the function will return true when deducting the charge from the balance and the resulting balance is greater than or equal to zero


lazedo:
@bluesandnz thanks.

as a note, this implementation should defer to the book keeper (new method?) instead of testing a particular book keeper.

PR #1497, opened 2015-12-17
Update hon_util.erl to fix priorities

kalda341:
I believe what was there before was wrong, but I'm not certain - I have a really hard time getting my head around comparison functions.
In rates.json the description field says that lower weights are preferred over higher ones, but this implies otherwise.

### DIFF
--- a/applications/hotornot/src/hon_util.erl
+++ b/applications/hotornot/src/hon_util.erl
@@ -119,7 +119,7 @@ sort_rate(RateA, RateB) ->
 
     case PrefixA =:= PrefixB of
         'true' ->
-            wh_json:get_integer_value(<<"weight">>, RateA, 100) >
+            wh_json:get_integer_value(<<"weight">>, RateA, 100) <
                 wh_json:get_integer_value(<<"weight">>, RateB, 100);
         'false' ->
             PrefixA > PrefixB


applications/hotornot/src/hon_util.erl
```diff
@@ -119,7 +119,7 @@ sort_rate(RateA, RateB) ->
 
     case PrefixA =:= PrefixB of
         'true' ->
-            wh_json:get_integer_value(<<"weight">>, RateA, 100) >
+            wh_json:get_integer_value(<<"weight">>, RateA, 100) <
                 wh_json:get_integer_value(<<"weight">>, RateB, 100);
         'false' ->
             PrefixA > PrefixB
```
fenollp: What about this one?

[Here](https://github.com/kalda341/kazoo/blob/patch-6/applications/hotornot/src/hon_util.erl#L114) it speaks of the weight, but how should the prefix be handled?
Can you give us a complete little table of in/outs so that we can reflect together on what code to write?

kalda341: The reason I was looking at this code is to allow rates that only reply to specific resellers. This is currently in testing.
The code I actually have is:

````
%% Return true if RateA is more applicable than RateB
-spec sort_rate(wh_json:object(), wh_json:object()) -> boolean().
sort_rate(RateA, RateB) ->
    PrefixA = byte_size(wh_json:get_binary_value(<<"prefix">>, RateA)),
    PrefixB = byte_size(wh_json:get_binary_value(<<"prefix">>, RateB)),

    %% Do these rates only apply to a reseller?
    ResellerA = wh_json:get_value(?EXCLUDE_TO_RESELLERS_KEY, RateA) =/= 'undefined',
    ResellerB = wh_json:get_value(?EXCLUDE_TO_RESELLERS_KEY, RateB) =/= 'undefined',

    case {{ResellerA, ResellerB}, PrefixA =:= PrefixB} of
        %% Prefer a rate specific to the reseller
        {{'true', 'false'}, _} ->
            'true';
        {{'false', 'true'}, _} ->
            'false';
        {_, 'true'} ->
            %% If prefix length is equal then prefer rates with lower weights
            wh_json:get_integer_value(<<"weight">>, RateA, 100) <
                wh_json:get_integer_value(<<"weight">>, RateB, 100);
        {_, 'false'} ->
            %% Prefer longer prefixes
            PrefixA > PrefixB
    end.
````

Rates that don't apply to this account will have already be filtered out by this time.
So, what I have will always prefer a rate that is exclusive to the reseller, if both or neither are exclusive:

A prefix length, B prefix length | weight A, weight B | result
------------ | ------------- | -------------
A < B  | * | B
A > B  | * | A
A = B  | A < B | A
A = B  | A > B | B
A = B  | A = B | A?

jamesaimonetti:
In talking with @k-anderson about this today, I think we believe this was merged in erroneously. Code should always take precedence over docs! So the documentation in the rates.json and elsewhere should have been updated to reflect what the code says, versus changing the code to match the docs.

@kalda341 would you want to issue a PR with doc changes and this code change reverted? Or should we handle it here?

Sorry for the confusion on how rates work!

PR #1499, opened 2015-12-19
Support for serializing registrations to and from file

vladimirralev:
You may be interested in this. Useful to save your ecallmgr registrations to some temp file before restart and then reload them after restart with no downtime. It won't overwrite new registrations and is safe for production. Especially good for single machine installs. It is also useful if you want to manipulate registrations manually for tests since the file is editable text. If you want this feature I will update the docs properly in another pull request.

### DIFF
--- a/applications/ecallmgr/src/ecallmgr_maintenance.erl
+++ b/applications/ecallmgr/src/ecallmgr_maintenance.erl
@@ -88,6 +88,7 @@
          ,enable_authz/0, enable_local_resource_authz/0
          ,disable_authz/0, disable_local_resource_authz/0
         ]).
+-export([export_registrations/1,import_registrations/1]).
 
 -export([show_channels/0]).
 -export([show_calls/0]).
@@ -696,3 +697,15 @@ hangup_long_running_channels(MaxAge) ->
     io:format("hanging up channels older than ~p seconds~n", [MaxAge]),
     N = ecallmgr_fs_channels:cleanup_old_channels(wh_util:to_integer(MaxAge)),
     io:format("hungup ~p channels~n", [N]).
+
+-spec export_registrations(ne_binary()) -> integer().
+export_registrations(Filename) ->
+    Exported = ecallmgr_registrar:export_registrations(Filename),
+    io:format("exported ~p registrations", [Exported]),
+    Exported.
+
+-spec import_registrations(ne_binary()) -> integer().
+import_registrations(Filename) ->
+    Imported = ecallmgr_registrar:import_registrations(Filename),
+    io:format("imported ~p registrations", [Imported]),
+    Imported.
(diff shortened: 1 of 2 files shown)


applications/ecallmgr/src/ecallmgr_maintenance.erl
```diff
@@ -696,3 +697,15 @@ hangup_long_running_channels(MaxAge) ->
     io:format("hanging up channels older than ~p seconds~n", [MaxAge]),
     N = ecallmgr_fs_channels:cleanup_old_channels(wh_util:to_integer(MaxAge)),
     io:format("hungup ~p channels~n", [N]).
+
+-spec export_registrations(ne_binary()) -> integer().
+export_registrations(Filename) ->
+    Exported = ecallmgr_registrar:export_registrations(Filename),
+    io:format("exported ~p registrations", [Exported]),
```
fenollp: Add a `\n` at the end of the format string

applications/ecallmgr/src/ecallmgr_maintenance.erl
```diff
@@ -696,3 +697,15 @@ hangup_long_running_channels(MaxAge) ->
...
+    Exported.
+
+-spec import_registrations(ne_binary()) -> integer().
+import_registrations(Filename) ->
+    Imported = ecallmgr_registrar:import_registrations(Filename),
+    io:format("imported ~p registrations", [Imported]),
```
fenollp: Add a `\n` at the end of the format string

applications/ecallmgr/src/ecallmgr_registrar.erl
```diff
@@ -381,6 +382,12 @@ init([]) ->
 handle_call('registrar_age', _, #state{started=Started}=State) ->
     wh_util:put_callid(?LOG_SYSTEM_ID),
     {'reply', wh_util:current_tstamp() - Started, State};
+handle_call({'import_registrations', Filename}, _, State) ->
+    wh_util:put_callid(?LOG_SYSTEM_ID),
+    {'reply', import_registrations_from_file(Filename), State};
```
fenollp: The registrar will be blocked by the import/export action and will be unable to respond during the completion of this job.

Please spawn a separate process to actually do the work. For example here the answer can be `{reply, Ref, State}`, this way the spawned worker process can send back `{Ref, Imported}` and this reply should be caught by the caller using something akin to `catch_my_request(Ref) -> receive {Ref, Result} -> Result end.`

vladimirralev: How about moving the file IO to the ```ecallmgr_maintenace.erl``` which would run in the sup process and not block the ecallmgr_registrar process? I will just pass the list in-memory between the ecallmgr proc and the sup proc. This will also allow to add support for stdin/out export and import from shell.

lazedo:
@vladimirralev although this could be useful, did you tried the kamailio registrar-sync role ?
when a ecallmgr starts it will ask every kamailio server to sync its registrations.

vladimirralev:
Haven't tried the registrar sync. Does it work for single machine installations?

lazedo:
@vladimirralev yep

vladimirralev:
I think I will use the registrar sync. We can lose this.

PR #1501, opened 2015-12-21
KAZOO-4456 allow tagged couch connections

lazedo:
supporting connections other than local

### DIFF
--- a/core/whistle_couch-1.0.0/src/couch_mgr.erl
+++ b/core/whistle_couch-1.0.0/src/couch_mgr.erl
@@ -79,6 +79,7 @@
          ,enable_change_notice/0
          ,change_notice/0
         ]).
+-export([server_tag/0, server_tag/1]).
 
 %% Types
 -export_type([get_results_return/0
@@ -263,7 +264,7 @@ do_load_fixtures_from_folder(DbName, [F|Fs]) ->
 %%--------------------------------------------------------------------
 -spec db_exists(text()) -> boolean().
 db_exists(DbName) when ?VALID_DBNAME ->
-    couch_util:db_exists(wh_couch_connections:get_server(), DbName);
+    couch_util:db_exists(get_server(), DbName);
 db_exists(DbName) ->
     case maybe_convert_dbname(DbName) of
         {'ok', Db} -> db_exists(Db);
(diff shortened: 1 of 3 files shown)


core/whistle_couch-1.0.0/src/couch_mgr.erl
```diff
@@ -1090,10 +1091,25 @@ move_doc(FromDB, FromId, Options) ->
...
+
+-spec server_tag() -> term().
+server_tag() ->
+    case get('$wh_couch_server_tag') of
+        'undefined' -> 'local';
+        Tag -> Tag
+    end.
+
+-spec server_tag(term()) -> 'ok'.
+server_tag(Tag) ->
```
fenollp: Nothing uses it yet? There are no calls to `server_tag/1`

lazedo: @fenollp export([server_tag/0, server_tag/1]). where are you looking ?

fenollp: Yes it's exported. But nothing uses it. I was asking if that was normal

lazedo: @fenollp yes, its normal. following pr will use it.

core/whistle_couch-1.0.0/src/wh_couch_connections.erl
```diff
@@ -71,14 +75,36 @@ update(#wh_couch_connection{}=Connection) ->
...
+-spec wait_for_connection() -> 'ok' | 'no_connection'.
+-spec wait_for_connection(term()) -> 'ok' | 'no_connection'.
+-spec wait_for_connection(term(), wh_timeout()) -> 'ok' | 'no_connection'.
+
 wait_for_connection() ->
-    try test_conn() of
+    wait_for_connection('local').
+
+wait_for_connection(Tag) ->
+    wait_for_connection(Tag, 'infinity').
```
jamesaimonetti: Can we make this one day or one month instead?

core/whistle_couch-1.0.0/src/wh_couch.hrl
```diff
@@ -66,8 +66,10 @@
                               ,ready = 'false'
                               ,admin = 'false'
                               ,server = #server{}
+                              ,tag = 'local'
```
jamesaimonetti: This should be typed as `'local' | ne_binary()` or `atom()` or something, since I doubt we want to tag connections arbitrarily from a type perspective.

jamesaimonetti:
@lazedo merged so you can continue to move forward but can you address the two concerns in the next related PR? Thanks!

PR #1505, opened 2015-12-22
KAZOO-4465 loopback changes

lazedo:


applications/callflow/src/cf_endpoint.erl
```diff
@@ -1367,9 +1367,19 @@ maybe_set_call_forward({Endpoint, Call, CallFwd, JObj}) ->
     {Endpoint, Call, CallFwd
      ,wh_json:set_values([{<<"Call-Forward">>, <<"true">>}
                           ,{<<"Authorizing-Type">>, <<"device">>}
+                          | bowout_settings(whapps_call:call_id_direct(Call) =:= 'undefined')
                          ], JObj)
     }.
 
+bowout_settings('true') ->
```
jamesaimonetti: spec! :)

jamesaimonetti:
@lazedo can you write up a little doc in `doc/eng/loopback.md` about what is changing with how loopbacks are processed? How hairpining calls to stay on-net works? And any other relevant details related to these changes.

applications/ecallmgr/src/ecallmgr.hrl
```diff
@@ -178,6 +178,11 @@
                                ,{<<"Caller-ID-Number">>, <<"effective_caller_id_number">>}
                                ,{<<"Callee-ID-Name">>, <<"effective_callee_id_name">>}
                                ,{<<"Callee-ID-Number">>, <<"effective_callee_id_number">>}
+                               ,{<<"effective_callee_id_number">>, <<"effective_callee_id_number">>}
```
k-anderson: @lazedo these should not be required, if they are we are leaking!  I suspect this was due to the CCV issue patched earlier.

applications/stepswitch/src/stepswitch.hrl
```diff
@@ -33,5 +33,7 @@
         ,whapps_config:get_is_true(?SS_CONFIG_CAT, <<"cid_rules_honor_diversions">>, 'false')
        ).
 
+-define(CHANNEL_LOOPBACK_HEADER_PREFIX, "FSLoopBack-").
```
k-anderson: @lazedo Hmm, lets talk about this too, feels like it breaks the FS abstraction... but I like the idea...

PR #1507, opened 2015-12-23
KAZOO-4462: While mute/unmute conference participant over API, conf_partic…

onnet:
…ipant's state remains intact. This results in incorrect behavior if next mute/unmute action is taken over participant's phone. 

Made conf_participant listen to the conference's commands which lets keep mute option synced between conf_participant and ecallmgr_fs_conference.

In case mute mute/unmute action is taken over dialpad, participant's state updated twice which is an overkill. Don't see how to avoid it yet.

### DIFF
--- a/applications/conference/src/conf_participant.erl
+++ b/applications/conference/src/conf_participant.erl
@@ -188,10 +188,10 @@ handle_participants_commands(JObj, Props) ->
 -spec handle_commands(wh_json:object(), wh_proplist()) -> 'ok'.
 handle_commands(JObj, Props) ->
     case wh_json:get_value(<<"Application-Name">>, JObj) of
-        <<"mute_participant">> -> 
+        <<"mute_participant">> ->
             Srv = props:get_value('server', Props),
             gen_listener:cast(Srv, 'mute_participant_update');
-        <<"unmute_participant">> -> 
+        <<"unmute_participant">> ->
             Srv = props:get_value('server', Props),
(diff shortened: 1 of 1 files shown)


fenollp:
Trying to have CI to run…

fenollp:
Please remove any trailing whitespace that this PR introduced :)

onnet:
@fenollp sorry, I thought everything is OK with this PR checks. Should be good good now.

k-anderson:
@onnet Howdy,

Looking at this I am not a fan of using the command event to change state.  A) it might fail B) conceptually that is not what it is intended for.

Currently we publish two events from ecallmgr "participants_event" and "participant_event".  The plural is what is currently used but I would like to remove it completely from the system and use only the participant_event.  The participants event is a single event with an array of (basically) participant event objects so it should be easy to convert.  The issue is the plural can become very large and chatty for conferences with lots of members which makes it not very efficient.

I believe that when this change is made the participant will remain in sync regardless of what mutes it.  I am actually surprised it does not currently!

I am not sure if you would be willing to work on removing participants event for participant but I would prefer that approach over this one.

onnet:
@k-anderson Many thanks for a comment! Will look at what I can do with this.

lazedo:
@onnet any update on this ?

onnet:
@lazedo Not now. Busy with another stuff. Will close it and reopen when get back to it.

PR #1510, opened 2015-12-25
KAZOO-4470: Run start_dtmf in case of "101 telephone-event" absence

onnet:
<<"One gateway that is connecting to Kazoo is using inband DTMF (changing to RFC2833 or INFO is not possible for them). On a standard freeswitch server, this is solved by using <action application="start_dtmf"> in the extension . Is the equivalent possible with Kazoo?">> 

Please check 2600-dev for details 

https://groups.google.com/forum/#!searchin/2600hz-dev/kiril/2600hz-dev/jV7UWPnMyKc/2qbM4nKmAwAJ

### DIFF
--- a/applications/callflow/src/cf_route_req.erl
+++ b/applications/callflow/src/cf_route_req.erl
@@ -240,10 +240,11 @@ get_ringback_media(Flow, JObj) ->
 pre_park_action(Call,JObj) ->
     case whapps_config:get_is_true(<<"callflow">>, <<"ring_ready_offnet">>, 'true')
         andalso whapps_call:inception(Call) =/= 'undefined'
-        andalso (whapps_call:authorizing_type(Call) =:= 'undefined' orelse whapps_call:authorizing_type(Call) =:= <<"resource">>)
+        andalso (whapps_call:authorizing_type(Call) =:= 'undefined'
+                 orelse whapps_call:authorizing_type(Call) =:= <<"resource">>)
     of
         'false' -> <<"none">>;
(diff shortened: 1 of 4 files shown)


applications/callflow/src/cf_route_req.erl
```diff
@@ -236,14 +236,22 @@ get_ringback_media(Flow, JObj) ->
...
         andalso whapps_call:authorizing_type(Call) =:= 'undefined'
     of
         'false' -> <<"none">>;
-        'true' -> <<"ring_ready">>
+        'true' -> check_dtmf_type(JObj) 
+    end.
+
+-spec check_dtmf_type(wh_json:object()) -> ne_binary().
+check_dtmf_type(JObj) ->
+    SDPRemote = wh_json:get_value(<<"Remote-SDP">>, JObj),
```
fenollp: No default value?

applications/callflow/src/cf_route_req.erl
```diff
@@ -236,14 +236,22 @@ get_ringback_media(Flow, JObj) ->
...
+    case re:run(SDPRemote, <<"101 telephone-event">>, []) of
```
fenollp: Use `binary:match(es)`

jamesaimonetti:
@onnet I think moving the SDP check to ecallmgr_fs_route (once the `route_resp` is received) would work better. Then trunkstore and other apps get it for free!

It should also be behind an `ecallmgr_config:is_true(<<"should_detect_inband_dtmf">>)` or similar, so people can opt in to the addition. 

Thoughts?

onnet:
@jamesaimonetti Thanks for a hint. This way it collapsed to a couple of lines. Will close/reopen this PR to make it more transparent.

PR #1511, opened 2015-12-25
KAZOO-4445: handling loopbacks in ecallmgr

siplabs:
This patch is to prevent loopbacks legs created by freeswitch.
Modified only clicktocall to use it.

### DIFF
--- a/applications/crossbar/src/modules/cb_clicktocall.erl
+++ b/applications/crossbar/src/modules/cb_clicktocall.erl
@@ -414,7 +414,7 @@ build_originate_req(Contact, Context) ->
     CalleeNumber = knm_converters:normalize(wh_json:get_binary_value(<<"outbound_callee_id_number">>, JObj, Exten)),
     FriendlyName = wh_json:get_ne_value(<<"name">>, JObj, <<>>),
     OutboundNumber = wh_json:get_value(<<"caller_id_number">>, JObj, Contact),
-    AutoAnswer = wh_json:is_true(<<"auto_answer">>, cb_context:query_string(Context), 'true'),
+    AutoAnswer = wh_json:is_true(<<"auto_answer">>, cb_context:query_string(Context), 'false'),
     {Caller, Callee} = get_caller_callee(wh_json:get_value(<<"dial_first">>, JObj, <<"extension">>)
                                          ,#contact{number = OutboundNumber
                                                    ,name = FriendlyName
@@ -462,7 +462,7 @@ build_originate_req(Contact, Context) ->
        ,{<<"Outbound-Caller-ID-Number">>, Caller#contact.number}
        ,{<<"Outbound-Call-ID">>, CallId}
        ,{<<"Ringback">>, wh_json:get_value(<<"ringback">>, JObj)}
-       ,{<<"Dial-Endpoint-Method">>, <<"single">>}
+       ,{<<"Loopback-Handler">>, <<"ecallmgr">>}
        ,{<<"Continue-On-Fail">>, 'true'}
        ,{<<"Custom-SIP-Headers">>, wh_json:get_value(<<"custom_sip_headers">>, JObj)}
        ,{<<"Custom-Channel-Vars">>, wh_json:from_list(CCVs)}
--- a/applications/ecallmgr/src/ecallmgr_originate.erl
+++ b/applications/ecallmgr/src/ecallmgr_originate.erl
@@ -186,8 +186,27 @@ handle_call(_Request, _From, State) ->
 %%--------------------------------------------------------------------
 handle_cast({'gen_listener', {'created_queue', Q}}, State) ->
     lager:debug("starting originate request"),
-    gen_listener:cast(self(), {'get_originate_action'}),
-    {'noreply', State#state{queue=Q}};
+    Routines = [fun create_uuid/1
+                ,fun set_app/1
+                ,fun maybe_update_endpoints/1
+                ,fun build_originate_args/1
+                ,fun get_originate_action/1
+                ,fun update_dialstrings/1
+               ],
+    Return = lists:foldl(fun(F, #state{} = S) -> F(S);
+                            (_, {'break', #state{} = NState}) -> NState;
+                            (_, Ret) -> Ret
+                         end
+                         ,State#state{queue=Q}
+                         ,Routines),
+    case Return of
+        #state{} = NewState ->
+            {'noreply', NewState};
+        {'stop', #state{} = NewState} ->
+            {'stop', 'normal', NewState};
+        Return ->
+            Return
+    end;
 handle_cast({'update_server_id', ServerId}, State) ->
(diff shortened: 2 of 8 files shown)


applications/crossbar/src/modules/cb_clicktocall.erl
```diff
@@ -414,7 +414,7 @@ build_originate_req(Contact, Context) ->
     CalleeNumber = wnm_util:to_e164(wh_json:get_binary_value(<<"outbound_callee_id_number">>, JObj, Exten)),
     FriendlyName = wh_json:get_ne_value(<<"name">>, JObj, <<>>),
     OutboundNumber = wh_json:get_value(<<"caller_id_number">>, JObj, Contact),
-    AutoAnswer = wh_json:is_true(<<"auto_answer">>, cb_context:query_string(Context), 'true'),
+    AutoAnswer = wh_json:is_true(<<"auto_answer">>, cb_context:query_string(Context), 'false'),
```
siplabs: Auto answer doesn't work, and default value is `false`.

applications/crossbar/src/modules/cb_clicktocall.erl
```diff
@@ -460,7 +460,7 @@ build_originate_req(Contact, Context) ->
        ,{<<"Outbound-Caller-ID-Number">>, Caller#contact.number}
        ,{<<"Outbound-Call-ID">>, CallId}
        ,{<<"Ringback">>, wh_json:get_value(<<"ringback">>, JObj)}
-       ,{<<"Dial-Endpoint-Method">>, <<"single">>}
```
siplabs: If we are requesting `loopback` endpoint, a responder will determinate `dial-method`

applications/crossbar/src/modules/cb_clicktocall.erl
```diff
@@ -460,7 +460,7 @@ build_originate_req(Contact, Context) ->
...
+       ,{<<"Loopback-Handler">>, <<"ecallmgr">>}
```
siplabs: Without this option loopback will be done in old way.

applications/ecallmgr/src/ecallmgr_originate.erl
```diff
@@ -188,8 +188,25 @@ handle_call(_Request, _From, State) ->
 %%--------------------------------------------------------------------
 handle_cast({'gen_listener', {'created_queue', Q}}, State) ->
     lager:debug("starting originate request"),
-    gen_listener:cast(self(), {'get_originate_action'}),
```
siplabs: I converted all related casts to routine-functions

applications/ecallmgr/src/ecallmgr_originate.erl
```diff
@@ -197,73 +214,6 @@ handle_cast({'maybe_update_node', Node}, #state{node=Node}=State) ->
 handle_cast({'maybe_update_node', Node}, #state{node=_OldNode}=State) ->
     lager:debug("updating node from ~s to ~s", [_OldNode, Node]),
     {'noreply', State#state{node=Node}, 'hibernate'};
-handle_cast({'create_uuid'}, #state{node=Node
```
siplabs: routine `create_uuid/1`

applications/ecallmgr/src/ecallmgr_originate.erl
```diff
@@ -197,73 +214,6 @@ handle_cast({'maybe_update_node', Node}, #state{node=Node}=State) ->
...
-        {'ok', #state{control_pid=Pid}=State1} ->
-            lager:debug("started control proc ~p uuid ~p", [Pid, UUID]),
-            maybe_send_originate_uuid(UUID, Pid, State),
-            gen_listener:cast(self(), {'build_originate_args'}),
-            {'noreply', State1, 'hibernate'};
-        {'error', _E} ->
-            lager:debug("failed to start control proc for ~p: ~p", [UUID, _E]),
-            {'stop', 'normal', State}
-    end;
-handle_cast({'get_originate_action'}, #state{originate_req=JObj
```
siplabs: routine `get_originate_action/1`

applications/ecallmgr/src/ecallmgr_originate.erl
```diff
@@ -197,73 +214,6 @@ handle_cast({'maybe_update_node', Node}, #state{node=Node}=State) ->
...
-    UseNode = maybe_update_node(JObj, Node),
-    lager:debug("maybe updating node from ~s to ~s", [Node, UseNode]),
-    lager:debug("originate action: ~s", [Action]),
-    {'noreply', State#state{action=Action
-                            ,app=ApplicationName
-                            ,node=UseNode
-                           }
-    ,'hibernate'
-    };
-handle_cast({'build_originate_args'}, #state{uuid='undefined'}=State) ->
```
siplabs: routine `build_originate_args/1`

lazedo:
@siplabs this needs rebasing. also, make sure that what you're proposing here works not only for click2call but also for quickcall, faxing, voice (with parking, transfer, .. ) and it creates the proper cdr documents for each of the accounts involved.

lazedo:
@siplabs any updates one this ?

siplabs:
@lazedo I'm working with this PR. Rebased it to master and trying to check. I am going to pull new updates on Monday.

siplabs:
@lazedo updated. I checked cdrs.
There are three documents in database, one about clicktocall and two for a legs. Is it what expected? When I tried clicktocall from master, it generates 4 documents. May it cause side effects in other parts?

lazedo:
@siplabs is this ready for account to account calling ?

lazedo:
@siplabs crossbar (or any otther app) shouldn't be aware of the inner implementation details of ecallmgr.
`,{<<"Loopback-Handler">>, <<"ecallmgr">>}` should be fetched from ecallmgr configuration. in this case `cb_clicktocall` is asking for loopback and its up to `ecallmgr` to decide how to deliver that.

as stated previously, i'm ok with this implementation if it also handles bridge commands, ie, a regular call comes in, bridges to some endpoint that may be a loopback.

in the case where  `account A` makes a call to `account B`, when the call is fully established you should have 4 channels (two in each account) so that parking, pickup, transfers do work.

i'm not fan of this implementation just to serve `cb_clicktocall` possible issues.

siplabs:
@lazedo this implementation is not for serving clicktocall issues. Clicktocall was modified to use this loopback resolving as an example.
Property `Loopback-Handler` should be set in application, cause only originator application may decide, is it ready for this loopback resolving. I didn't test `konami` but I think it may have issues using this resolving.

lazedo:
@siplabs we disagree, you are making the originating application aware of inner implementation details in ecallmgr. imho, this is not the correct to handle this. if you want to replace the current working implementation of loopback, you'll have to go to every possible application, every possible use case, and modify it. as i said before, the setting should be handled at ecallmgr that should handle loopbacks the way it does today or with your new proposed way. for this to be accepted, it must handle all the current use cases.

PR #1513, opened 2015-12-29
KAZOO-4355: update the default templates

k-anderson:


fenollp:
What about `master`?

jamesaimonetti:
I'm sure @k-anderson was just about to do that :)

PR #1515, opened 2015-12-29
KAZOO-4473: handle when devices are authd by IP

k-anderson:


fenollp:
What about merging to `master`?

k-anderson:
@fenollp I will do so if accepted but it is a bug on 3.22 (same with the other PR I just opened)

PR #1517, opened 2015-12-29
KAZOO-4258: amqp leader core application

siplabs:
This is implementation of `amqp_leader`, analog of `gen_leader` uses AMQP to deliver messages between processes
It's dependency for #1348


deps/leader_cron-0.1/rebar.config
```diff
@@ -0,0 +1,9 @@
+{erl_opts, [debug_info, fail_on_warning]}.
+{deps, [{gen_leader, ".*",
+	 {git, "git://github.com/jeraymond/gen_leader_revival.git",
```
fenollp: If this is indeed a dependency of `leader_cron` then it should be part of this PR.
Deps should never be downloaded at compile time.

deps/leader_cron-0.1/src/leader_cron.app.src
```diff
@@ -0,0 +1,12 @@
+{application, leader_cron,
+ [
+  {description, ""},
+  {vsn, "0.2.1"},
+  {registered, []},
```
fenollp: add `{modules, []},` in there so that `make` can add `leader_cron`'s modules in the `.app` file please

deps/leader_cron-0.1/rebar.config
siplabs: @fenollp removed rebar.config, it has no dependencies

deps/leader_cron-0.1/src/leader_cron.app.src
siplabs: @fenollp added `modules`

core/whistle-1.0.0/src/wh_nodes.erl
```diff
@@ -415,10 +415,8 @@ handle_cast({'advertise', JObj}, #state{tab=Tab}=State) ->
 handle_cast({'gen_listener', {'is_consuming', 'true'}}
             ,#state{heartbeat_ref='undefined'}=State
            ) ->
-    Reference = erlang:make_ref(),
-    erlang:send_after(?HEARTBEAT, self(), {'heartbeat', Reference}),
     lager:debug("start heartbeat"),
-    {'noreply', State#state{heartbeat_ref=Reference, is_consuming='true'}};
+    handle_info({'heartbeat', 'undefined'}, State#state{is_consuming='true'});
```
k-anderson: @siplabs I am ready to accept this PR (thanks what a tremendous amount of work, I look forward to playing with it!)  However, this change to wh_nodes bypasses the saftey check to make sure there are not more than one timer running at a time.  If there is a bug or something broken here lets talk but I don't think it is related to this PR.  If you remove the change to wh_nodes we could accept this!

siplabs: @k-anderson `wh_nodes` announces that it alive after timer with random timeout (5-15 seconds, as I can remember).
This change only for announcing that this node is alive (amqp_leader needs it) after start. It is a bit tricky, but I prefer to use existing code even usage is tricky.
`Amqp_leader` will work without this change, but it will fail to start several times (1-3) and will be restarted with supervisor.

core/amqp_leader-1.0.0/src/amqp_leader.app.src
```diff
@@ -0,0 +1,13 @@
+{application, amqp_leader,
+ [
+  {description, ""},
+  {vsn, "1.0.0"},
+  {modules, []},
+  {registered, []},
+  {applications, [
+                  kernel,
+                  stdlib
```
fenollp: Make sure to add the apps this app depends on

siplabs: @fenollp updated

core/whistle-1.0.0/src/wh_nodes.erl
lazedo: @siplabs maybe `amqp_leader` needs a different start strategy ?

siplabs: @k-anderson @lazedo removed `wh_nodes` change.

lazedo:
@siplabs one last thought :)
we are including leader_cron in deps but it has significant changes from the source repo which we don't like to do.

i would suggest to, move the leader_cron to `core` renamed as `amqp_cron or [insert ideas here]`, edit the README and add a statement like "this library is based on the work of ... from base repo ... with changes to use amqp_leader instead of gen_leader".

siplabs:
@lazedo there are 3 commits about `leader_cron`:
1) makefile and something about building it
2) changes to use `amqp_leader` instead of `gen_leader`. May be someday `amqp_leader` will change `gen_leader` in kazoo, but I think it's too early. :)
3) several changes to `leader_cron`'s logic (it's broken in my opinion). I want to create PR with this changes but haven't had enough time. :(

2 and 3 may be done using parse transform...

lazedo:
will the `deps/leader_cron` run without `core/amqp_leader` ? i think not.
you can't have a `deps` application depending on `core`

siplabs:
@lazedo ogh, missed it! Thank you!
I will move it to `core` and will rename it to `amqp_leader`.

lazedo:
that too , just noticed `amqp_leader` was both in `core` and `deps`.
but i'm talking about about `leader_cron`  which will not work without `amqp_leader`.
`leader_cron` as is, should also be moved to `core` under `amqp_cron` and mention in the README that it is based on `leader_cron` and changed to work with `amqp_leader`

siplabs:
@lazedo on the other side, if all changes will be done with parse_transform after loading code... it will run without `core/amqp_leader` using `gen_leader`. But it should be compiled with debug information.

siplabs:
But `parse_transform`ed code is hard to support.
I will move it to `core` tommorow.

core/amqp_leader-1.0.0/src/amqp_leader_listener.erl
```diff
@@ -0,0 +1,241 @@
...
+%% @doc
+%% Allows listener to pass options to handlers
+%%
+%% @spec handle_event(JObj, State) -> {reply, Options}
+%% @end
+%%--------------------------------------------------------------------
+-spec handle_event(wh_json:object(), state()) -> {'reply', []}.
+handle_event(JObj, #state{name = Name}) ->
+    wh_util:put_callid(wapi_leader:queue(Name, node())),
+    NodeBin = iolist_to_binary(io_lib:format("~s", [node()])),
```
k-anderson: @siplabs
To get the node as a binary you can utilize the wh_util functions ;)
NodeBin = wh_util:to_binary(node()),

core/amqp_leader-1.0.0/src/amqp_leader_listener.erl
```diff
@@ -0,0 +1,241 @@
...
+        {Ref, 'ready'} -> ready_when_node_is_up()
+    after
+        1 * ?MILLISECONDS_IN_SECOND ->
+            lager:debug("listener not ready yet"),
+            recv_ready(Ref)
+    end.
+
+-spec ready_when_node_is_up() -> 'true'.
+ready_when_node_is_up() ->
+    case ets:match('wh_nodes', #wh_node{node = node(), expires = '$2', _ = '_'}) of
```
k-anderson: @siplabs This is the right idea from the wh_node modification but you should not access the ets directly, it breaks the abstraction.  I would suggest moving this as a new function in wh_nodes, perhaps wh_nodes:is_up(node())?

core/amqp_leader-1.0.0/src/amqp_leader_listener.erl
siplabs: Definitly I was tired :)
Fixed.

core/amqp_leader-1.0.0/src/amqp_leader_listener.erl
siplabs: @k-anderson updated.

siplabs:
@lazedo moved to `core`, changed `README.md`

fenollp:
Build fails because of trailing whitespace in `amqp_cron_task.erl`

core/amqp_leader-1.0.0/src/amqp_leader_listener.erl
k-anderson: @siplabs looks like its still using wh_nodes ets table, and I didn't see the update to wh_nodes module.  Perhaps there is a push to this branch that is missing?

siplabs: @k-anderson added forgoten commit

PR #1522, opened 2015-12-30
KAZOO-4478: Fix app store when not apps is installed

macpie:


macpie:
This needs to go in 3.22

PR #1523, opened 2015-12-30
KAZOO-4479: federate bind for destroy/disconnect

k-anderson:


applications/jonny5/src/j5_channels.erl
```diff
@@ -91,6 +96,12 @@
                      ,{{?MODULE, 'handle_rate_resp'}
                        ,[{<<"rate">>, <<"resp">>}]
                       }
+                     ,{{?MODULE, 'handle_channel_destroy'}
+                       ,[{<<"call_event">>, <<"CHANNEL_DESTROY">>}]
+                      }
+                     ,{{?MODULE, 'handle_channel_destroy'}
+                       ,[{<<"call_event">>, <<"CHANNEL_DISCONNECTED">>}]
```
jamesaimonetti: I think you can put this tuple in the list above this one to only have one tuple with two events routed to the callback.

applications/jonny5/src/j5_channels.erl
```diff
@@ -438,6 +446,13 @@ handle_rate_resp(JObj, Props) ->
     Srv = props:get_value('server', Props),
     gen_server:cast(Srv, {'rate_resp', JObj}).
 
+-spec handle_channel_destroy(wh_json:object(), wh_proplist()) -> 'ok'.
+handle_channel_destroy(JObj, Props) ->
+    'true' = wapi_call:event_v(JObj),
+    CallId = wh_json:get_ne_value(<<"Call-ID">>, JObj),
```
jamesaimonetti: Use `kz_call_event:call_id/1`

PR #1526, opened 2015-12-31
Kazoo 4481 - only add caller id if resource changes it

jamesaimonetti:


jamesaimonetti:
Move stepswitch_formatters:apply/3 calls to stepswitch_bridge/local/originate so caller ids are picked up and formatted properly.

PR #1533, opened 2016-01-05
crossbar presence: simplify query

lazedo:


jamesaimonetti:
Has anything changed here that needs updating in the [presence docs](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/doc/presence.md) ?

lazedo:
@jamesaimonetti yes, can we make it part of the doc reviews you're going to start ?

PR #1535, opened 2016-01-07
omnipresence : add user agent to subscription

lazedo:


applications/omnipresence/src/omnip_subscriptions.erl
```diff
@@ -407,6 +407,7 @@ subscribe_to_record(JObj) ->
                         ,subscription_id=wh_json:get_value(<<"Subscription-ID">>, JObj)
                         ,proxy_route= wh_json:get_value(<<"Proxy-Route">>, JObj)
                         ,version=Version
+                        ,user_agent=wh_json:get_value(<<"User-Agent">>, JObj)
```
fenollp: Sorry I never really understood it: when should we call the `get_ne_binary_value` or equivalent, given that we also have just `get_value`?

jamesaimonetti: In theory, `get_value/2` could return `<<>>` as well. `get_ne_binary/2` promises to only return `'undefined' | ne_binary()`

fenollp: Ok. And what about `get_value` vs. `get_binary_value`? (here `user_agent :: api_binary()`)

PR #1536, opened 2016-01-07
Added cdrtofile app

kalda341:
Added application to dump CDRs from AMQP to filesystem. Primarily for use with https://github.com/2600hz/community-scripts/pull/56


jamesaimonetti:
Hi @kalda341, thanks for the PR! 

Since this PR will be included only to master, and master representing 4.0 work, and with Kazoo no longer supporting R15 (targeting 18+ but I suspect 17 will work for most of that too), could you update the specs using `set()` to read `sets:set()`?

Once travis is happy with at least the 18 build, I'll take a closer inspection of the app and see what feedback I can give to get this included.

Thanks again!

k-anderson:
@kalda341 see James comment, but I was also thinking it might be a good idea to include this as part of the CDR application.  The reason being if it is a seperate app then it will uniquely bind for channel destroy events.  If a node was running cdr and this app then each message would be sent from AMQP twice.  If the apps were combined then one AMQP delievery of the channel destroy could internally be distributed to 'modules'.  One cdr module would store in bigcouch as we typically do, and another on disk.  This also provides a great place for additional modules that push to everyone's various preferred CDR data stores!

What do you think?

kalda341:
I never quite got around to responding, sorry!
Sounds great to me. I'll move it over when I get a chance.

PR #1538, opened 2016-01-11
KAZOO-4470: docs update plus small adjustment

onnet:


jamesaimonetti:
No muss, no fuss! Thanks @onnet !

PR #1539, opened 2016-01-11
KAZOO-4481: move formatting to bridge

jamesaimonetti:
caller id settings should be applied to the endpoints once calculated,
formatters given a chance to format the caller id, and endpoint-specific
caller id stripped if it doesn't differ from the root bridge's caller id.

KAZOO-4481: fix spec

KAZOO-4481: set caller id per endpoint

KAZOO-4481: add spec and use macros for keys

KAZOO-4481: cleanup usage of offnet req

KAZOO-4481: move formatting to util for reuse

KAZOO-4481: repoint call for default_realm/1

KAZOO-4481: update specs with more descriptive type

use wapi_offnet_resource accessors too

KAZOO-4481: format endpoints when originating

KAZOO-4481: update specs with descriptive types

KAZOO-4481: update spec

KAZOO-4481: typo ;)

Conflicts:
	applications/stepswitch/src/stepswitch_resources.erl

### DIFF
--- a/applications/stepswitch/src/stepswitch_sms.erl
+++ b/applications/stepswitch/src/stepswitch_sms.erl
@@ -31,7 +31,7 @@
                 ,response_queue :: api_binary()
                 ,queue :: api_binary()
                 ,message = [] :: wh_proplist()
-                ,messages = queue:new() :: queue()
+                ,messages = queue:new() :: queue:queue()
                }).
 -type state() :: #state{}.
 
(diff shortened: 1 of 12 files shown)


jamesaimonetti:
@k-anderson resolved conflict, plus changed it so we `insert_values` in `stepswitch_util` to not overwrite CID if set in `stepswitch_resources`

applications/stepswitch/src/stepswitch_sms.erl
```diff
@@ -31,7 +31,7 @@
                 ,response_queue :: api_binary()
                 ,queue :: api_binary()
                 ,message = [] :: wh_proplist()
-                ,messages = queue:new() :: queue()
+                ,messages = queue:new() :: queue:queue()
```
fenollp: Please revert this: hinders compilation on R15.

jamesaimonetti: Master no longer cares for R15's types. Look forward to the future!

fenollp: What about failing builds? and not rebasing every PR?

jamesaimonetti:
@k-anderson if you can review

k-anderson:
@jamesaimonetti I think that you were working on this with CanDo in sandbox, am I mistaken or did all that get pushed into this already?

PR #1540, opened 2016-01-13
Trunkstore REFER support

jamesaimonetti:
Set authz info on endpoint's CCVs so REFERs come in as authz'd and trunkstore processes the request.


jamesaimonetti:
@k-anderson check it, may need to push to sandbox for testing, as I am compiling FreeSWITCH atm

PR #1543, opened 2016-01-14
KAZOO-4448: Use msec and round it up for billing

icehess:


applications/ecallmgr/src/ecallmgr_call_events.erl
```diff
@@ -998,6 +998,13 @@ get_disposition(Props) ->
 get_hangup_code(Props) ->
     kzd_freeswitch:hangup_code(Props).
 
+-spec get_billing_seconds(wh_proplist()) -> api_binary().
+get_billing_seconds(Props) ->
+    case props:get_integer_value(<<"variable_billmsec">>, Props) of
+        'undefined' -> props:get_value(<<"variable_billsec">>, Props);
+        Billmsec -> wh_util:to_binary(wh_util:ceiling(Billmsec / 1000))
```
fenollp: Don't you want to apply ceiling to the clause just above too?

icehess: We can always add a second to `variable_billsec` but I thought since we don;t have enough time resolution it's better leave seconds as is(what if it's round exactly, e.g. 4). If someone needs more precise billing based on millisecond then he can enable `variable_billmsec` on FreeSWITCH.

fenollp: Ah, read to quickly. Never mind then

lazedo:
@icehess i would prefer to just collect the billing_ms and add it as is to the payload with a new name.
whapps (cdr or others) can then process one or the other or whatever rule they implement based on availability and also business rules.

k-anderson:
@lazedo I would be ok if we added it in-addition, but it has not be requested so in the interest of keeping the payloads small maybe we wait?  However, the PR addresses the specific complaint so we still need to preform this mapping to seconds as well :)

lazedo:
@k-anderson i believe the rounding should be done in `cdr` app based on system/account setting. what if someone wants to use `floor` instead of `ceiling` ?

lazedo:
@icehess we will add the msec to payload if requested by enough number of people :)

PR #1548, opened 2016-01-18
convert mailbox to integer in mwi query

lazedo:


applications/callflow/src/cf_util.erl
```diff
@@ -1016,13 +1016,19 @@ start_task(Fun, Args, Call) ->
...
-        {'error', 'not_found'} -> get_mailbox(AccountDb, VMNumber)
+    case filter_numbers(VMNumber) of
+        <<>> -> {'error', 'not_found'};
+        VMNumber ->
+            Number = wh_util:to_integer(VMNumber),
+            case wh_cache:peek_local(?CALLFLOW_CACHE, ?VM_CACHE_KEY(AccountDb, Number)) of
+                {'ok', _}=Ok -> Ok;
+                {'error', 'not_found'} -> get_mailbox(AccountDb, Number)
+            end;
+        _Filtered -> {'error', 'not_found'}
```
k-anderson: @lazedo I don't think this can ever match....

lazedo: user_756 => 756

PR #1553, opened 2016-01-20
ts_refer: added authz id and type to ccvs for endpoint

jamesaimonetti:
updated specs/types and var naming a bit too

### DIFF
--- a/applications/trunkstore/src/ts_from_offnet.erl
+++ b/applications/trunkstore/src/ts_from_offnet.erl
@@ -21,15 +21,19 @@ init(Parent, RouteReqJObj) ->
     proc_lib:init_ack(Parent, {'ok', self()}),
     start_amqp(ts_callflow:init(RouteReqJObj, ['undefined', <<"resource">>])).
 
+-spec start_amqp(ts_callflow:state() |
+                 {'error', 'not_ts_account'}
+                ) -> 'ok'.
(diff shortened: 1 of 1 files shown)


jamesaimonetti:
@k-anderson over here

applications/trunkstore/src/ts_from_offnet.erl
```diff
@@ -306,6 +317,7 @@ routing_data(ToDID, AcctID, Settings) ->
                          ,{<<"To-IP">>, build_ip(ToIP, ToPort)}
                          ,{<<"Route-Options">>, RouteOpts}
                          ,{<<"Hunt-Account-ID">>, HuntAccountId}
+                         ,{<<"Authorizing-ID">>, wh_doc:id(Settings)} % connectivity doc id
```
lazedo: @jamesaimonetti shouldn't this be going in `<<"Custom-Channel-Vars">>` ? if so, maybe also add `  ,{<<"Authorizing-Type">>, <<"sys_info">>}`

jamesaimonetti: Is added to endpoint's CCVs [here](https://github.com/2600hz/kazoo/pull/1553/files#diff-4b15a7d704417b061680d835865cc059R205)

PR #1555, opened 2016-01-20
KAZOO-4509 use cowboy compression

lazedo:


fenollp:
What about blackhole?

fenollp:
And fax and wh_media_proxy ?

All results of `git grep -InE cowboy:start -- core/ applications/`

lazedo:
@fenollp `blackhole` is another story. `fax` and `wh_media_proxy` are used by infrastructure not by users. this ticket is to address the rest api. users of rest api like `monster-ui` users will benefit from this.

fenollp:
[Since content is compressed iff the client says it supports it](https://github.com/ninenines/cowboy/blob/master/src/cowboy_req.erl#L766-L775), it should be safe to merge this

PR #1559, opened 2016-01-25
rewrite transactions.md

colecut:


applications/crossbar/src/modules/cb_transactions.erl
```diff
@@ -79,7 +79,9 @@ allowed_methods(?CURRENT_BALANCE) ->
     [?HTTP_GET];
 allowed_methods(?DEBIT) ->
     [?HTTP_DELETE];
-allowed_methods(_) ->
+allowed_methods(?MONTHLY) ->
+    [?HTTP_GET];
```
fenollp: omagad thank you so much for this stuff! Reading crossbar modules was pain because of blackholes like these.

Thank you for taking the time to do this!

jamesaimonetti:
@colecut thanks! Made a few updates in the code to generate the swagger json file, added the missing paths that weren't auto-detected.

For now, for query string parameters, just include them like you did with a little header above it. We'll think of something better later.

PR #1562, opened 2016-01-27
Bump kazoo apps that got missed

fenollp:


lazedo:
@fenollp i do not agree with the change to versions. each app has its own version. kazoo is composed of multiple apps, each with their own version

fenollp:
well, a major version bump can bump every app of the project. rabbitmq does it. this way one knows which app is compatible with which kazoo version

lazedo:
@fenollp tha's what i'm saying, it can, i don't agree with it as exposed above. my believe is that each app has its own versioning. kazoo its like a kernel where apps run, kernel has version and so do apps.

lazedo:
@fenollp also check that dev-start-apps works as it seems its failing since the releases merge. do a clean install, build it, and make sure it runs. it complains about several things one is failing to find rabbit_common.app.

jamesaimonetti:
@fenollp @lazedo I think we've been pretty lax in our bumping of versions for individual apps. In fact, I doubt many have been bumped in years!

I'd be fine bumping all the apps/core to 4.0.0 for the 4.x release and then being more consistent with bumping versions as we make changes. I've added a todo to the docs effort to create a guide to [semantic versioning](http://semver.org/) so we can all be on the same page for when to bump and which segment to bump.

We can discuss more with the team...

k-anderson:
@fenollp Hey, I agree with @lazedo regarding the dev scripts.  We don't want to change the development workflow to rely on releases.  I would say revert those changes and this looks good otherwise.

PR #1563, opened 2016-01-27
KAZOO-4516: introduce type api/1, i.e. "maybe undefined"

fenollp:
Found the corrected culprits using

    git grep -InE '\|\s*.undefined|undefined.\s*\|' -- applications/ core/


jamesaimonetti:
@fenollp rebase, but also the `api(...)` should be consistently applied across the codebase. So `api_binary()` should become `api(binary())`, `api_object()` becomes `api(wh_json:object())`, etc.

fenollp:
merge it nao!

lazedo:
i wouldn't merge this before 5.0

PR #1574, opened 2016-01-29
KAZOO-4522: core/AppName-* --> core/AppName

fenollp:


lazedo:
@fenollp please rebase it so it can be merged

PR #1579, opened 2016-01-30
KAZOO-4525 add direction to call

lazedo:


applications/ecallmgr/src/ecallmgr_fs_route.erl
```diff
@@ -377,6 +377,7 @@ route_req(CallId, FetchId, Props, Node) ->
     SwitchURI = <<"sip:", SwitchURIHost/binary>>,
     [{<<"Msg-ID">>, FetchId}
      ,{<<"Call-ID">>, CallId}
+     ,{<<"Call-Direction">>, kzd_freeswitch:call_direction(Props)}
```
fenollp: Why did Call Direction not exist in whapps_call before?

PR #1580, opened 2016-01-30
KAZOO-4526 add reset functions to omnipresence_maintenance

lazedo:
reset_subscription (user@realm or user realm)
   resets the subscription and notifies every subscriber
reset_subscriber (user@realm or user realm)
   gets all subscriptions of user@realm
   calls reset_subscription


fenollp:
LGTM

PR #1581, opened 2016-02-01
Local render of system_alerts

jamesaimonetti:
rather than using the pool, where a process has to build the render
module then render the template, create fixed renderer modules, based on
the content-types found in the system_alert notification, to render the
payloads.

This should significantly reduce the processing needed for
system_alerts, as well as free the render farm up for other types of
notifications.

It could be a good idea to extend this thought to all the templates,
build their system_config versions statically, and only use the render
farm for account-overridden templates


jamesaimonetti:
@k-anderson barring travis, this PR is ready. The `teletype_system_alert` module creates render modules based on content types of templates available for system alerts, and calls those modules when processing system_alert messages.

One thing to note, the Subject is still going through the render farm, so we're not completely out of the woods, but the heavy renderings of the message bodies is out of the farm.

PR #1586, opened 2016-02-02
KAZOO-4501: Don't crawl disabled accounts

icehess:


applications/notify/src/notify_account_crawler.erl
```diff
@@ -148,7 +148,8 @@ handle_info(_Info, State) ->
 
 -spec check_then_process_account(ne_binary(), {'ok', kz_account:doc()} | {'error',any()}) -> 'ok'.
 check_then_process_account(AccountId, {'ok', AccountJObj}) ->
-    case wh_doc:is_soft_deleted(AccountJObj) of
+    case wh_doc:is_soft_deleted(AccountJObj)
+            andalso kz_account:is_enabled(AccountJObj) of
```
lazedo: @icehess account can be disabled but not soft deleted. maybe use `orelse not  kz_account:is_enabled` or add `is_disabled` to kz_account (return `not is_enabled(..)`) and use `orelse is_disabled(..`

PR #1587, opened 2016-02-03
Fixed some ACDC Bugs which were making it unusable

tsromana:
acdc_agent_fsm.erl
AMQP Queue was not getting created on time …

bug fix for ACDC Queue, Queue was not getting created on time causing to send undefined queue name (Server-ID), making the response never reach the publishing Agent

acdc_agent_listener.erl
Insted of assigning the value was being compared …

The Value of state was not being initalised and sent but was being compared making the agent crash and giving out: no match of right hand value

### DIFF
--- a/applications/acdc/src/acdc_agent_listener.erl
+++ b/applications/acdc/src/acdc_agent_listener.erl
@@ -728,8 +728,7 @@ handle_cast({'send_sync_req'}, #state{my_id=MyId
     case MyQ of
         'undefined' ->
             lager:debug("queue not ready yet, waiting for sync request"),
-            timer:sleep(100),
-            gen_listener:cast(self(), {'send_sync_req'});
+            timer:apply_after(100 , gen_listener, cast, [self(), {'send_sync_req'}]);
          _ ->
             lager:debug("queue retrieved: ~p , sending sync request", [MyQ]),
             send_sync_request(AcctId, AgentId, MyId, MyQ)
--- a/applications/acdc/src/acdc_agent_listener.erl
+++ b/applications/acdc/src/acdc_agent_listener.erl
@@ -725,8 +725,15 @@ handle_cast({'send_sync_req'}, #state{my_id=MyId
                                       ,acct_id=AcctId
                                       ,agent_id=AgentId
                                      }=State) ->
-    lager:debug("sending sync request"),
-    send_sync_request(AcctId, AgentId, MyId, MyQ),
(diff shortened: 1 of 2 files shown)


applications/acdc/src/acdc_agent_listener.erl
```diff
@@ -725,8 +725,15 @@ handle_cast({'send_sync_req'}, #state{my_id=MyId
                                       ,acct_id=AcctId
                                       ,agent_id=AgentId
                                      }=State) ->
-    lager:debug("sending sync request"),
-    send_sync_request(AcctId, AgentId, MyId, MyQ),
+    case MyQ of
+        'undefined' ->
+            lager:debug("queue not ready yet, waiting for sync request"),
+            timer:sleep(100),
+            gen_listener:cast(self(), {'send_sync_req'});
```
fenollp: What about `erlang:send_after`?

fenollp:
Nice catches!
In `applications/acdc/src/acdc_agent_fsm.erl`:
1. I think you missed one more occurence of this matching-instead-of-binding at line `1518` (`outbound_hungup/1`)
1. `1176` (`{'next_state', 'ready', apply_state_updates(ReadyState)}`) looks wrong to me, when all the other clauses return `{next_state, atom(), fsm_state()}` and `apply_state_updates/1` returns a 3-tuple. 

@lazedo

fenollp:
Oh and Travis is mad because you introduced trailing whitespace in `applications/acdc/src/acdc_agent_listener.erl`

tsromana:
thank you for your response will look into it right now

tsromana:
@fenollp i am not sure how to use erlang:send_after as we need to write handle_info for it to work as will not cast should i write handle cast. could you guide me how we could use it, or is there other proper way to do it. i am trying to use wait(time) function if anything doesn't work. 
@lazedo done !

tsromana:
@fenollp  i added timer:apply_after which is guess is non blocking

fenollp:
LGTM

tsromana:
tested it, its working fine for the time will add new request if i came across another bug. BTW do you want me to backport it to 3.22 ??

jamesaimonetti:
Thanks @tsromana. If you want to issue a PR for 3.22 as well, feel free. I'm sure the community would be grateful to have it there!

tsromana:
@jamesaimonetti sending a pull request for 3.22 right now!

PR #1589, opened 2016-02-05
KAZOO-4530: sanity checks for dns and timezone

jamesaimonetti:
Was concerned the time check might be weird, but seems pretty reliable. Not sure if we want to make the dns time check configurable (5ms seemed reasonable on my box, as I was getting an occasional outlier around 3ms).

```erlang
19> lists:all(fun(_) -> whistle_apps_init:sanity_checks() end, lists:seq(1,1000)).
true
```

For now, just logs it at critical and continues on its merry way.

### DIFF
--- a/core/whistle_apps-1.0.0/src/whistle_apps_init.erl
+++ b/core/whistle_apps-1.0.0/src/whistle_apps_init.erl
@@ -1,5 +1,5 @@
 %%%-------------------------------------------------------------------
-%%% @copyright (C) 2013, 2600Hz
+%%% @copyright (C) 2013-2016, 2600Hz
 %%% @doc
 %%% Init to be done
 %%% @end
@@ -9,17 +9,20 @@
 -module(whistle_apps_init).
 
 -export([start_link/0
+         ,sanity_checks/0
          ,init/0
         ]).
 
 -include("whistle_apps.hrl").
 
 start_link() ->
+    _ = sanity_checks(), %% one day make this true
     _ = wh_util:spawn(?MODULE, 'init', []),
     'ignore'.
 
 init() ->
     wh_util:put_callid(?MODULE),
+
     case wh_config:get_atom('whistle_apps', 'cookie') of
         [] ->
             lager:warning("failed to set whistle_apps cookie trying node ~s", [node()]),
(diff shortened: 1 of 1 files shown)


core/whistle_apps-1.0.0/src/whistle_apps_init.erl
```diff
@@ -45,3 +48,57 @@ set_loglevel() ->
...
+        'true' -> 'true';
+        'false' ->
+            lager:warning("hostname results (in us): ~p < ~p < ~p"
+                         ,[Min, (Total div (Tests+1)), Max]
+                         ),
+            lager:critical("hostname resolution is painfully slow!!! This will cause "),
+            'false'
+    end.
+
+-spec time_hostname_resolution(any(), {pos_integer(), pos_integer(), pos_integer()}) ->
```
fenollp: This tuple looks like `wh_now()`

core/whistle_apps-1.0.0/src/whistle_apps_init.erl
```diff
@@ -9,17 +9,20 @@
 -module(whistle_apps_init).
 
 -export([start_link/0
+         ,sanity_checks/0
          ,init/0
```
fenollp: Why export `init/0` BTW?

core/whistle_apps-1.0.0/src/whistle_apps_init.erl
```diff
@@ -9,17 +9,20 @@
...
         ]).
 
 -include("whistle_apps.hrl").
 
 start_link() ->
+    _ = sanity_checks(), %% one day make this true
```
fenollp: What do you mean make this true?

core/whistle_apps-1.0.0/src/whistle_apps_init.erl
jamesaimonetti: Nah, `wh_now()` is for `erlang:timestamp()`'s return. These are the min, max, and total microseconds it took to resolve the hostname.

core/whistle_apps-1.0.0/src/whistle_apps_init.erl
jamesaimonetti: Well, the debate was whether to crash the process (and ultimately the app/VM) if the sanity checks didn't hold. For now, we just log critically but we could make this `'true' = sanity_checks()` to really enforce the fixing of these checks.

PR #1590, opened 2016-02-05
Fix j5 balance crawler

skorobkov:
After changes in konami application, j5_ballance_crawler stop working (don't disconnect calls).
So I decided to rewrite it.

- Added to wapi_route support new routing keys `route.win.broadcast.{{APPLICATION}}` (where {{APPLICATION}} some name). Developers now can receive copy of route_win messages to find out `Control-Queue`. For this developers add application name to list `system_config/ecallmgr/broadcast_route_win_to_queues` (example: add `jonny5` and bind to `route.win.broadcast.jonny5` in `callmgr` exchange).
- j5_channels on route_win message store `Control-Queue` to channel table, and send disconnect commands to this queue.


lazedo:
@skorobkov the purpose of route win is to provide the control-queue to a single process.

skorobkov:
Ok. How this should be made?
- Use another exchange/routing_key?
- Use another way to obtain Control-Queue?
- Use another way to disconnect calls?
- ... ?

jamesaimonetti:
@skorobkov We see requests for this every so often (the ability for arbitrary processes to send commands to the `ecallmgr_call_control` process for a call). I'm not sure we have a good solution that doesn't descend into chaos! I think Konami is a good middle ground; jonny5 could publish using `wapi_metaflow` to hangup the call and let Konami relay that into the control process.

skorobkov:
But konami changed his behaviour.
Nowi ts work only if you create metaflow in account/user/device.
Withou metaflow it wont work and can't disconnect calls.

jamesaimonetti:
I would propose, then, to make metaflows implicitly started if there's no specific flow available. Maybe only bind for `CHANNEL_DESTROY` and `wapi_metaflow` commands in that case.

Just brainstorming here though; open to other suggestions.

skorobkov:
Maybe add this functionality to ecallmgr?

k-anderson:
@skorobkov I agree with james regarding the call control queue, and just closed another PR that was attempting to leak this to other apps as well.  The concept is to have just one application controlling the call at a time and explicitly hand off control if necessary (transferring the authority/responsibility).

I will close this PR while the feature is developed using a different approach.  I have made a recommendation in the the following Jira ticket I created to support the development of this feature:
https://2600hz.atlassian.net/browse/KAZOO-4599

For ease this is the description from that ticket:

What I would recommend is the following:
In ecallmgr_call_control (https://github.com/2600hz/kazoo/blob/master/applications/ecallmgr/src/ecallmgr_call_control.erl#L202) init check the CCVs.  If either the account_billing or reseller_billing is per_minute then pull a timer period from system_config.  If this value does not exist or is less than 30 seconds then do not start the timer.  If it is greater than 30 seconds start a timer.  Something sorta like this:

case should_nibble_credit(CCVs) of
'true' ->
  case ecallmgr_config:get_integer(<<"per_minute_nibble_period">>, 0) of
    NibblePeriod when NibblePeriod > 30 ->
      send_after(NibblePeriod, self(), 'nibble');
    NibblePeriod ->
      lager:debug("invalid or disabled nibble period ~p", [_Else])
  end
'false' -> ok
end.

When the timer expires it will send an info message to itself which in ecallmgr_call_control could spawn a wh_amqp_worker:call to request jonny5 to nibble the credit.  The reply would be affirmative if credit remained or negative if not.  If negative the spawned worker could then issue a request to the call control process to kill the call.  Something like this:

handle_info('nibble', State) ->
   Srv = self(),
   wh_util:spawn(fun() ->
       Response = wh_amqp_worker:call(create_nibble_request(State)
                                                   ,fun wapi_authz:publish_nibble_request/1
                                                   ,fun wapi_authz:nibble_response_v/1
                                                 ),
         case is_negative_nibble(Response) of
              'true' -> gen_server:cast(Srv, 'kill_call');
              'false' ->
                   NibblePeriod = ecallmgr_config:get_integer(<<"per_minute_nibble_period">>),
                   send_after(NibblePeriod, Srv, 'nibble')
       end),
   {'noreply', State}

You would have to create the wapi_authz nibble messages.

In Jonny5 you would then bind via a shared queue (so it round-robbins) and add a debit doc each time the nibble comes in.  The transaction ID should be the call-id then "-" then the duration so if multiple attempts are made it conflicts.  It will also need to debit the account and/or the reseller based on the authz type in the CCVs (passed along on the nibble request).  After the transaction is written check the balance then send back a response if it is now less than 0.

That is how I would approach it.  Some notes:

* Without writing the transaction other per-minute calls going on at the same time will not be accounted for unless you track all of them in Jonny5.  This would be ideal since there is the ets table of authz already, it could be used to calculate the cost if it had rate information for all active calls belonging to an account.  This value could then be subtracted from the current balance to determine availability without hammering the database.
* For backward compatibility it woiuld need to default to a nibble period of 0 (or disabled)
* We had this a while back (I think before 3.x) but it used FreeSWITCH session hearbeats on EVERY call and was very inefficient.  The approach above is similar but should resolve the original issue that forced us to remove this capability.
* Its important to debit the correct account ;) Be it reseller or account or both!
* All the code in this comment should be expected to fail :)  I referenced functions that would be expected to be created as part of this work.


If you have any questions or comments please feel free to reach out to us via the ticket or otherwise!

PR #1591, opened 2016-02-05
read access to system_media for everyone!

siplabs:


lazedo:
@siplabs what use case is this commit is trying to solve ?

jamesaimonetti:
@siplabs thanks!

PR #1592, opened 2016-02-05
Fixed some ACDC Bugs which were making it unusable

tsromana:
fixed acdc bugs which were causing unexpected behaviour and crashes


jamesaimonetti:
@tsromana w00t! Thanks

tsromana:
:)

PR #1593, opened 2016-02-07
KAZOO-4539: Customers update over teletype

onnet:
Discussion could be seen over here: https://groups.google.com/forum/?fromgroups#!topic/2600hz-dev/iMj0oa5V5kA

### DIFF
--- a/applications/crossbar/src/modules/cb_notifications.erl
+++ b/applications/crossbar/src/modules/cb_notifications.erl
@@ -275,10 +275,7 @@ may_be_validate_recipient_id(Context) ->
 
 -spec validate_recipient_id(ne_binary(), cb_context:context()) -> cb_context:context().
 validate_recipient_id(RecipientId, Context) ->
-    SenderId = case cb_context:account_id(Context) of
-                   'undefined' -> cb_context:auth_account_id(Context);
-                   AccountId -> AccountId
-               end,
+    SenderId = sender_account_id(Context),
     ViewOpts = [{'startkey', [SenderId]}
                ,{'endkey', [SenderId, wh_json:new()]}
                ],
@@ -293,6 +290,16 @@ validate_recipient_id(RecipientId, Context) ->
             lager:info("failed to load children. error: ~p", [E]),
             Context
     end.
+
+-spec sender_account_id(cb_context:context()) -> ne_binary().
+sender_account_id(Context) ->
+    sender_account_id(Context, cb_context:account_id(Context)).
+
+-spec sender_account_id(cb_context:context(), ne_binary()|'undefined') -> ne_binary().
+sender_account_id(Context, 'undefined') ->
+    cb_context:auth_account_id(Context);
+sender_account_id(_Context, AccountId) ->
+    AccountId.
 
 %%--------------------------------------------------------------------
 %% @public
@@ -389,10 +396,7 @@ post(Context, Id, ?PREVIEW) ->
 
 -spec build_customer_update_payload(cb_context:context()) -> wh_proplist().
 build_customer_update_payload(Context) ->
-    SenderId = case cb_context:account_id(Context) of
-                   'undefined' -> cb_context:auth_account_id(Context);
-                   AccountId -> AccountId
-               end,
+    SenderId = sender_account_id(Context),
     props:filter_empty(
       [{<<"Account-ID">>, SenderId}
        ,{<<"Recipient-ID">>, cb_context:req_value(Context, <<"recipient_id">>)}
@@ -494,8 +498,10 @@ publish_fun(<<"webhook_disabled">>) ->
     fun wapi_notifications:publish_webhook_disabled/1;
 publish_fun(<<"denied_emergency_bridge">>) ->
     fun wapi_notifications:publish_denied_emergency_bridge/1;
+publish_fun(<<"customer_update">>) ->
+    fun wapi_notifications:publish_customer_update/1;
 publish_fun(_Id) ->
-    lager:debug("no wapi_notification:publish_~s/1 defined", [_Id]),
+    lager:debug("no wapi_notifications:publish_~s/1 defined", [_Id]),
     fun(_Any) -> 'ok' end.
 
 -spec preview_fold(ne_binary(), {wh_proplist(), wh_json:object()}) ->
@@ -901,7 +907,8 @@ update_template(Context, Id, FileJObj) ->
     Contents = wh_json:get_value(<<"contents">>, FileJObj),
     CT = wh_json:get_value([<<"headers">>, <<"content_type">>], FileJObj),
     lager:debug("file content type for ~s: ~s", [Id, CT]),
-    Opts = [{'content_type', CT}],
+ %   Opts = [{'content_type', CT}],
+    Opts = [{'content_type', wh_util:to_list(CT)}], % Temporary until couchbeam update
 
     AttachmentName = attachment_name_by_content_type(CT),
 
--- a/applications/crossbar/src/modules/cb_notifications.erl
+++ b/applications/crossbar/src/modules/cb_notifications.erl
@@ -37,6 +37,7 @@
 -define(CUSTOMER_UPDATE, <<"customer_update">>).
 -define(MESSAGE, <<"message">>).
 -define(CB_LIST_SMTP_LOG, <<"notifications/smtp_log">>).
+-define(ACC_CHILDREN_LIST, <<"accounts/listing_by_children">>).
 
 -define(MACROS, <<"macros">>).
 
@@ -279,7 +280,7 @@ validate_recipient_id(RecipientId, Context) ->
     ViewOpts = [{'startkey', [SenderId]}
                ,{'endkey', [SenderId, wh_json:new()]}
                ],
-    case couch_mgr:get_results(?WH_ACCOUNTS_DB, <<"accounts/listing_by_children">>, ViewOpts) of
+    case couch_mgr:get_results(?WH_ACCOUNTS_DB, ?ACC_CHILDREN_LIST, ViewOpts) of
         {'ok', Accounts} ->
             AccountIds = lists:map(fun(Account) -> wh_json:get_value(<<"id">>, Account) end, Accounts),
             case lists:member(RecipientId, AccountIds) of
--- a/applications/crossbar/src/modules/cb_notifications.erl
+++ b/applications/crossbar/src/modules/cb_notifications.erl
@@ -329,7 +329,6 @@ set_system_macros(Context) ->
             Context
     end.
 
-  
 post(Context, ?CUSTOMER_UPDATE, ?SEND_MESSAGE) ->
     case
         whapps_util:amqp_pool_request(
--- a/applications/crossbar/src/modules/cb_notifications.erl
+++ b/applications/crossbar/src/modules/cb_notifications.erl
@@ -10,13 +10,13 @@
 -module(cb_notifications).
 
 -export([init/0
-         ,allowed_methods/0, allowed_methods/1, allowed_methods/2
-         ,resource_exists/0, resource_exists/1, resource_exists/2
+         ,allowed_methods/0, allowed_methods/1, allowed_methods/2, allowed_methods/3
+         ,resource_exists/0, resource_exists/1, resource_exists/2, resource_exists/3
          ,content_types_provided/2
          ,content_types_accepted/2
-         ,validate/1, validate/2, validate/3
+         ,validate/1, validate/2, validate/3, validate/4
          ,put/1
-         ,post/2, post/3
+         ,post/2, post/3, post/4
          ,delete/2
 
          ,flush/0
@@ -34,6 +34,8 @@
 -define(CB_LIST, <<"notifications/crossbar_listing">>).
 -define(PREVIEW, <<"preview">>).
 -define(SMTP_LOG, <<"smtplog">>).
+-define(CUSTOMER_UPDATE, <<"customer_update">>).
+-define(SEND_MESSAGE, <<"send_message">>).
 -define(CB_LIST_SMTP_LOG, <<"notifications/smtp_log">>).
 
 -define(MACROS, <<"macros">>).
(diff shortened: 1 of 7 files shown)


jamesaimonetti:
Hi @onnet 

First, this is most awesome!

My only concern is on the URL structure. You are essentially overloading the `preview` token but calling it `send_message`, and then offering another token for the recipient account id.

Could we rework this a bit?

I think we could do `POST /notificiations/customer_update/message` since we want to POST the message to the account(s). I think the `recipient_id` should be included in the request body `-d '{"data":{"recipient_id":"..."}}'`.

applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -313,6 +329,21 @@ set_system_macros(Context) ->
             Context
     end.
 
+post(Context, ?CUSTOMER_UPDATE, ?SEND_MESSAGE) ->
```
jamesaimonetti: Don't forget your `-spec`s!

applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -216,7 +226,12 @@ validate(Context, Id, ?PREVIEW) ->
     DbId = kz_notification:db_id(Id),
     update_notification(maybe_update_db(Context), DbId);
 validate(Context, ?SMTP_LOG, Id) ->
-    load_smtp_log_doc(Id, Context).
+    load_smtp_log_doc(Id, Context);
+validate(Context, ?CUSTOMER_UPDATE, ?SEND_MESSAGE) ->
+    cb_context:set_resp_status(Context, 'success').
+
+validate(Context, ?CUSTOMER_UPDATE, ?SEND_MESSAGE, _Id) ->
```
jamesaimonetti: You need to validate that the `recipient_id` is a child of the account making the request.

applications/teletype/src/templates/teletype_customer_update.erl
```diff
@@ -0,0 +1,157 @@
...
+        'true' -> process_req(DataJObj)
+    end.
+
+-spec process_req(wh_json:object()) -> 'ok'.
+process_req(DataJObj) ->
+    SenderId = wh_json:get_value(<<"account_id">>, DataJObj),
+    ViewOpts = [{'startkey', [SenderId]}
+               ,{'endkey', [SenderId, wh_json:new()]}
+               ],
+    case couch_mgr:get_results(<<"accounts">>, <<"accounts/listing_by_children">>, ViewOpts) of
```
jamesaimonetti: I think the `accounts` db has a macro `?WH_ACCOUNTS_DB`

bazemoreb:
Awesome work, Kirill - this rocks!  Nothing useful to add other than thanks, but reviewed video and yes, this is exactly what I was thinking of. Should be a great add for clients and resellers and a huge time saver.

onnet:
@jamesaimonetti 
"../notifications/customer_update/send_message/{RECIPIENT_ACCOUNT_ID}" was a kind of attempt to make this feature bullet/foolproof for case someone will accidentally forget to add "recipient_id" field to its payload, and the particular account addressed message will be send to all the world.. 
But your suggestion makes code look more compact and transparent and I am aware of mentioned danger anyway :)

applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -255,6 +262,38 @@ disallow_delete(Context, Id) ->
...
+        'undefined' ->
+            cb_context:set_resp_status(Context, 'success');
+        <<RecipientId:32/binary>> ->
+            validate_recipient_id(RecipientId, Context);
+        _ ->
+            Context
+    end.
+
+-spec validate_recipient_id(ne_binary(), cb_context:context()) -> cb_context:context().
+validate_recipient_id(RecipientId, Context) ->
```
jamesaimonetti: I think you can use `wh_util:is_in_account_hierarchy(SenderId, RecipientId)` here. It check's `RecipientId`'s account tree (`pvt_tree`) for SenderId.

applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -313,6 +352,21 @@ set_system_macros(Context) ->
             Context
     end.
 
+post(Context, ?CUSTOMER_UPDATE, ?MESSAGE) ->
```
jamesaimonetti: Missing spec?

applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -313,6 +352,21 @@ set_system_macros(Context) ->
...
+    case
+        whapps_util:amqp_pool_request(
```
jamesaimonetti: Please use `wh_amqp_worker:call/3` instead. I'd like to phase out the whapps_call amqp pool stuff as its not doing anything of note that `wh_amqp_worker` doesn't already do.

applications/teletype/src/templates/teletype_customer_update.erl
```diff
@@ -0,0 +1,157 @@
...
+        'true' -> process_req(DataJObj)
+    end.
+
+-spec process_req(wh_json:object()) -> 'ok'.
+process_req(DataJObj) ->
+    SenderId = wh_json:get_value(<<"account_id">>, DataJObj),
+    ViewOpts = [{'startkey', [SenderId]}
+               ,{'endkey', [SenderId, wh_json:new()]}
+               ],
+    case couch_mgr:get_results(?WH_ACCOUNTS_DB, <<"accounts/listing_by_children">>, ViewOpts) of
```
jamesaimonetti: 1. You create a macro for the view at the top of the file, but don't use it here? :)
2. Do you want to use `accounts/listing_by_descendants` instead? That should load all the descendant accounts of the sender, while `accounts/listing_by_children` returns only the immediate child accounts?

applications/teletype/src/templates/teletype_customer_update.erl
```diff
@@ -0,0 +1,157 @@
...
+        {'ok', Accounts} ->
+            process_accounts(Accounts, DataJObj);
+        {'error', _Reason} = E ->
+            lager:info("failed to load children. error: ~p", [E]),
+            E
+    end.
+
+process_accounts(Accounts, DataJObj) ->
```
jamesaimonetti: Please add specs to all the functions missing them.

applications/teletype/src/templates/teletype_customer_update.erl
```diff
@@ -0,0 +1,157 @@
...
+    case wh_json:get_value(<<"recipient_id">>, DataJObj) of
+        'undefined' ->
+            [process_account(wh_json:get_value(<<"value">>, Account), DataJObj) || Account <- Accounts];
+        RecipientId ->
```
jamesaimonetti: I think you should check for `recipient_id` first, and if undefined, load descendant accounts to process. Why traverse the list of accounts when you know exactly which account you want to process?

applications/teletype/src/templates/teletype_customer_update.erl
```diff
@@ -0,0 +1,157 @@
...
+        {'ok', Users} ->
+            Templates = teletype_templates:fetch(?TEMPLATE_ID, DataJObj),
+            select_users_to_update(lists:map(fun(User) -> wh_json:get_value(<<"value">>, User) end, Users), DataJObj, Templates);
+        {'error', _Reason} = E ->
+            lager:info("failed to load users from: ~p", [AccountDb]),
+            E
+    end.
+
+select_users_to_update(Users, DataJObj, Templates) ->
+    case wh_json:get_value(<<"user_type">>, DataJObj) of
```
jamesaimonetti: Similarly to `recipient_id`, if you know a particular `user_id` to use, go for it; otherwise load the list of users and go from there.

applications/teletype/src/templates/teletype_customer_update.erl
```diff
@@ -0,0 +1,157 @@
...
+            [send_update_to_user(User, DataJObj, Templates) || User <- Users, wh_json:get_value(<<"id">>, User) == UserId];
+        <<"all_users">> ->
+            [send_update_to_user(User, DataJObj, Templates) || User <- Users];
+        _ ->
+            [send_update_to_user(User, DataJObj, Templates) || User <- Users, wh_json:get_value(<<"priv_level">>, User) == <<"admin">>]
+    end.
+
+send_update_to_user(UserJObj, DataJObj, Templates) ->
+    Macros = [{<<"system">>, teletype_util:system_params()}
+              ,{<<"system">>, teletype_util:account_params(DataJObj)}
```
jamesaimonetti: Wrong key here? Should be "account" no?

applications/teletype/src/templates/teletype_customer_update.erl
```diff
@@ -0,0 +1,157 @@
...
+              | build_macro_data(UserJObj)
+             ],
+    RenderedTemplates = [{ContentType, teletype_util:render(?TEMPLATE_ID, Template, Macros)}
```
jamesaimonetti: How you render has changed; please refer to a module like `teletype_low_balance` for how to render your templates now. Quickly,

```erlang
    RenderedTemplates = teletype_templates:render(?TEMPLATE_ID, Macros, DataJObj),
    {'ok', TemplateMetaJObj} = teletype_templates:fetch_notification(?TEMPLATE_ID, teletype_util:find_account_id(DataJObj)),
```

This should be easier for everyone. It will also bypass the render farm when using the system's template, speeding up processing.

core/whistle/src/api/wapi_notifications.erl
```diff
@@ -445,6 +448,23 @@
...
+%% Customer update
+-define(CUSTOMER_UPDATE_HEADERS, [<<"Account-ID">>]).
+-define(OPTIONAL_CUSTOMER_UPDATE_HEADERS, [<<"Recipient-ID">>
+                                           ,<<"User-Type">>
+                                           ,<<"Subject">>
+                                           ,<<"From">>
+                                           ,<<"Reply-To">>
+                                           ,<<"To">>,<<"CC">>,<<"BCC">>
+                                           ,<<"HTML">>
+                                           ,<<"Plain">>
```
jamesaimonetti: Please look at the `-define(...)` of `DEFAULT_OPTIONAL_HEADERS`; you have duplicated most of them here, yet included the defaults below. I'd just remove them from this listing and let the macro include them for you!

applications/teletype/src/templates/teletype_customer_update.erl
onnet: @jamesaimonetti 

Regarding comment #2 here and this one above:
"I think you can use wh_util:is_in_account_hierarchy(SenderId, RecipientId) here. It check's RecipientId's account tree (pvt_tree) for SenderId."


I'm sending updates to direct children only by intention.
Let's imagine that I'm sending new rates update. 
Such a message shouldn't be delivered to my reseller's customers because resellers could have different rates and may be even don't want their customers be aware of reseller's hierarchy.

I believe that such cases could appear more often than the case when you need to update all the descendants and it should be reseller's responsibility to decide if it's customers should recieve this update and to retranslate it.

In case you sure we also need a possibility to send message to all the descendants an additional key should be introduced.

Please let me know what you think.

applications/crossbar/src/modules/cb_notifications.erl
onnet: No, post/3 spec is a bit upstairs, introduced with post/2.
https://github.com/onnet/kazoo/blob/e0b1365578444a14e2788ca512cdc602a299f78a/applications/crossbar/src/modules/cb_notifications.erl#L319

Will move it closer to not to be confused

applications/teletype/src/templates/teletype_customer_update.erl
onnet: @jamesaimonetti I am doing that to be sure Recipient belongs to children list. And yes, we've got this check also in cb_notifications, but looking to the future, customer updates could be sent not only over API, but by another whapps also (or even by third party apps connected to rabbitmq).
Do you think I should remove it anyway, or remove it in cb_notifications, or may be leave it as is?

jamesaimonetti: I would leave the check in `cb_notifications` and not here. Since teletype can only receive this via AMQP, and AMQP is a trusted source of messages, teletype assumes all is well with the request. So, if these payloads are to be sent in by other whapps, those whapps are responsible for ensuring the data is correct before sending to teletype. Same for 3rd party apps you allow to connect to RabbitMQ, though that is a scary proposition!

core/kazoo_documents/src/kzd_user.erl
```diff
@@ -254,3 +255,10 @@ devices(UserJObj) ->
             lager:warning("unable to find documents owned by ~s: ~p", [UserId, _R]),
             []
     end.
+
+-spec fetch(ne_binary(), api_binary()) -> {'ok', doc()} | {'error', any()}.
+fetch(<<_/binary>> = UserId, <<_/binary>> = Account) ->
```
jamesaimonetti: I would prefer `fetch(Account, UserId)` if we put this here. Also, it should be `invalid_parameters` below ;)

applications/teletype/src/templates/teletype_customer_update.erl
onnet: @jamesaimonetti Thanks for a comment. Will remove it.

applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -334,6 +387,27 @@ post(Context, Id, ?PREVIEW) ->
             crossbar_util:response('error', <<"Failed to process notification preview">>, Context)
     end.
 
+-spec build_customer_update_payload(cb_context:context()) -> wh_proplist().
+build_customer_update_payload(Context) ->
+    SenderId = case cb_context:account_id(Context) of
```
jamesaimonetti: Can we put this into a function `sender_account_id(Context)`?

```erlang
sender_account_id(Context) ->
    sender_account_id(Context, cb_context:account_id(Context)).

sender_account_id(Context, 'undefined') ->
    cb_context:auth_account_id(Context);
sender_account_id(_Context, AccountId) ->
    AccountId.
```

With appropriate spec, of course! :)

applications/teletype/src/templates/teletype_customer_update.erl
onnet: @jamesaimonetti regarding changes to render. Did you remove load_templates from teletype_templates:fetch_notification/2 (fetch/2 previously) and left it at preview only (load_preview_templates) by intention?
It looked very tempting to be able to make changes to the message you want to send on the fly not editing existing template.

onnet: @jamesaimonetti sorry for bothering. Looks like I can export render/4 and use it with preview flag ('true').

jamesaimonetti:
@onnet Thanks! This all looks good for now. Let's see how it gets used whether the children/descendants distinction becomes an issue.

onnet:
@jamesaimonetti Yep. If anyone will say even one word about descendants, we'll ask him to provide PR and NOT FORGET ABOUT SPECS :)))
PS Thanks for help!

PR #1594, opened 2016-02-07
HELP-20903 back port fax changes to 3.21

lazedo:


applications/fax/src/fax_smtp.erl
```diff
@@ -38,7 +38,7 @@
           options = [] :: list()
           ,from :: binary()
           ,to :: binary()
-          ,docs = [] :: wh_json:objects()
+          ,doc :: api_object()
```
fenollp: Shouldn't there be a default value here?

applications/fax/src/fax_smtp.erl
```diff
@@ -218,7 +228,7 @@ terminate(Reason, State) ->
 -spec handle_message(state()) -> 'ok'.
 handle_message(#state{errors=[_Error | _Errors]}=State) ->
     maybe_faxbox_log(State);
-handle_message(#state{docs=[]}=State) ->
+handle_message(#state{doc='undefined'}=State) ->
```
fenollp: ah then should be ok

fenollp:
LGTM

PR #1598, opened 2016-02-10
SUP commands for reload/set/unset app env

icehess:
- Add SUP commands to reload/set/unset app environment variables through `wh_config`
- Add SUP command to `whapps_config` to lock db writes on `whapps_config` doc


lazedo:
@icehess port forward to  `master`

PR #1599, opened 2016-02-10
KAZOO-4543: System dialplan. Multiple dialplans per each regexp.

onnet:
As stated in internationalization/numbers.md system dialplan looks like this: 

'{"data":{"^(2\\d{6}$":{"prefix":"+7383","name":"Novosibirsk"}}}' 

This obliges each dialplan to have unique regexp/key which is wrong, 
because different dialplans could have same dial patterns. 

So, it is good to make cf_util to be able to also parse multiple dialplans for each key: 

'{"data":{"^(\\d{7})$":[{"prefix":"+7495","name":"Moscow"},{"prefix":"+7812","name":"Saint Petersburg"}]}}'

### DIFF
--- a/applications/callflow/src/cf_util.erl
+++ b/applications/callflow/src/cf_util.erl
@@ -804,12 +804,20 @@ load_system_dialplans(Names) ->
 -spec fold_system_dialplans(ne_binaries()) ->
                                    fun(({ne_binary(), wh_json:object()}, wh_json:object()) -> wh_json:object()).
 fold_system_dialplans(Names) ->
-    fun({Key, Val}, Acc) ->
-            Name = wh_util:to_lower_binary(wh_json:get_value(<<"name">>, Val)),
-            case lists:member(Name, Names) of
-                'true' -> wh_json:set_value(Key, Val, Acc);
-                'false' -> Acc
-            end
+    fun({Key, Val}, Acc) when is_list(Val) ->
+        lists:foldl(fun(ValElem, A) -> may_be_dialplan_suits({Key, ValElem}, A, Names) end, Acc, Val);
+       ({Key, Val}, Acc) ->
+        may_be_dialplan_suits({Key, Val}, Acc, Names)
+    end.
+
+-spec may_be_dialplan_suits({ne_binary(), wh_json:object()|wh_json:objects()}
+                            ,wh_json:object(), ne_binaries()
+                           ) -> wh_json:object().
+may_be_dialplan_suits({Key, Val}, Acc, Names) ->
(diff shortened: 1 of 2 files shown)


applications/callflow/src/cf_util.erl
```diff
@@ -804,12 +804,20 @@ load_system_dialplans(Names) ->
...
-                'true' -> wh_json:set_value(Key, Val, Acc);
-                'false' -> Acc
-            end
+    fun({Key, Val}, Acc) when is_list(Val) ->
+        lists:foldl(fun(ValElem, A) -> may_be_dialplan_suits({Key, ValElem}, A, Names) end, Acc, Val);
+       ({Key, Val}, Acc) ->
+        may_be_dialplan_suits({Key, Val}, Acc, Names)
+    end.
+
+-spec may_be_dialplan_suits({ne_binary(), wh_json:object()|wh_json:objects()}
```
fenollp: From your changes `Val` can't be a list anymore. Which means `s/|wh_json:objects()//`.

onnet:
@fenollp Done.

jamesaimonetti:
@onnet looks good. Thanks for expanding the use case.

lazedo:
@onnet do you have problems using ?
```
data : { 
"^(\d{7})$":{"prefix":"+7495","name":"Moscow"},
"^(\d{7})$":{"prefix":"+7812","name":"Saint Petersburg"}
}
```
also, you should provide backward compatibility, it seems its not doing this.

the `system_dialplans` should have been designed to be indexed by name like this
```
data : {
  "Moscow" : { dialpan : "^(\d{7})$", "prefix":"+7495"},
  "Saint Petersburg" : { dialpan : "^(\d{7})$", "prefix":"+7812"}
}
```
which i think its the best way to evolve this, we could match the `location` and auto assign a dialplan, but it would require some `migrate` work.

so for now, i would be ok if you provide backward compatibility.

lazedo:
@onnet indentation tricked me about compat. looks good.

onnet:
@lazedo @jamesaimonetti Hi and many thanks for accepting my PR.

Luis, I also tried to use the same construction firstly, but because of the same key one line was overwriting another, which is correct. I believe know that for sure, but for me it was a surprise  :)))
data : { 
"^(\d{7})$":{"prefix":"+7495","name":"Moscow"},
"^(\d{7})$":{"prefix":"+7812","name":"Saint Petersburg"}
}

No thanks needed, I've better had a favour :))  :
Can I ask you to comment the next ticket: https://2600hz.atlassian.net/browse/KAZOO-4544
It looks to me as a next step to make this feature usable.
Will it be accepted, what is the best approach, any imaginable threats?

PR #1602, opened 2016-02-11
KAZOO-4545: handle lookup errors to clean up logs

k-anderson:
This will need to be pick'd into master :)


lazedo:
cherry-picked to master

PR #1603, opened 2016-02-11
KAZOO-4547: dont force stop, conditionally continue for the sanity of…

k-anderson:
This will need to be pick'd into master :)


lazedo:
cherry-picked to master

PR #1604, opened 2016-02-11
KAZOO-4548: handle when the bucket is shutting down

k-anderson:
This will need to be pick'd into 3.22 and master :)


fenollp:
@lazedo you cherry picked it?

lazedo:
cherry-picked, to 3.22 and master

PR #1606, opened 2016-02-15
Undergoing fixes to master

lazedo:


applications/crossbar/src/modules/cb_clicktocall.erl
```diff
@@ -424,21 +424,23 @@ build_originate_req(Contact, Context) ->
                                                    ,route = Exten}),
 
     lager:debug("attempting clicktocall ~s in account ~s", [FriendlyName, AccountId]),
+    {'ok', AccountDoc} = couch_mgr:open_cache_doc(?WH_ACCOUNTS_DB, AccountId),
```
jamesaimonetti: This should have been `kz_account:fetch(AccountId)` as the account's version is considered authoritative and not the `accounts` version.

lazedo: @jamesaimonetti thanks. i'll fix it.

applications/ecallmgr/src/ecallmgr_originate.erl
```diff
@@ -817,8 +824,10 @@ start_control_process(#state{control_pid=_Pid
     {'ok', State}.
 
 -spec maybe_start_call_handlers(created_uuid(), state()) -> 'ok'.
-maybe_start_call_handlers(UUID, State) ->
-    case start_control_process(State#state{uuid=UUID}) of
+maybe_start_call_handlers(UUID, #state{originate_req=JObj}=State) ->
+case wh_json:is_true(<<"Start-Control-Process">>, JObj, 'true')
```
jamesaimonetti: Formatting seems off here?

applications/fax/src/fax_maintenance.erl
```diff
@@ -409,3 +410,19 @@ next_key(<<>>) ->
     <<"\ufff0">>;
 next_key(Bin) ->
     <<Bin/binary, "\ufff0">>.
+
+load_smtp_attachment(DocId, Filename) ->
```
jamesaimonetti: Missing spec

PR #1618, opened 2016-02-17
HELP-21054 do not set authz on redirects

lazedo:


2600hz-admin:
Whoops, we added auth stuff I think because people could redirect and commit fraud. Did you just remove that work? I'm dumb so I don’t fully understand what this commit resolves

lazedo:
the commit removes authz from redirects set on route_req. still applies restrictions based on redirects.
inbound calls to a device that have a redirect will still be treated as a inbound call without setting the authz to the target device

k-anderson:
@lazedo As per our discussion I believe that the authorizing id belongs on the forwarded leg of the call.  Conceptually the call is initiated by the user, just automatically.

We also discussed that the in this condition since there is a code path we could pull the merged endpoint initiating the forward and use the call-forwarding settings to control the caller id so it matches server side forwarding.

lazedo:
@k-anderson i would agree if the call was forwarded but it isn't, its redirected.
```
A -- INVITE -> B
B -> 302 (call C) -> A
A -- INVITE --> C

A is freeswitch inbound, it doesn't send the 302 back to the carrier, it auto redirects.
```
i even have doubts on removing the  `channel_vars` because that would remove `authorization-type` set by `stepswitch_inbound`. i think we should only remove `authorization-id` as a safety check.

lazedo:
current behaviour http://i.imgur.com/QDQpISl.png
proposed behaviour http://i.imgur.com/hBODIBj.png

lazedo:
closing for move to 4.0

PR #1619, opened 2016-02-17
KAZOO-4567 allow delete if announced

lazedo:
faxes api announces the inbox/outbox delete api but does not honor it.

backport to 3.22 please


jamesaimonetti:
Backport away @lazedo

PR #1620, opened 2016-02-18
KAZOO-4434: created an API to promote/demote reseller

k-anderson:


core/whistle_services-1.0.0/src/whistle_services_maintenance.erl
```diff
@@ -140,41 +140,19 @@ sync(Account) when not is_binary(Account) ->
...
-    _ = maybe_update_services(AccountId, <<"pvt_reseller">>, 'true'),
-    io:format("promoting account ~s to reseller status, updating sub accounts~n", [AccountId]),
-    cascade_reseller_id(AccountId, AccountId).
-
 %%--------------------------------------------------------------------
 %% @public
 %% @doc
-%% Remove reseller status from an account and set all its sub accounts
-%% to the next higher reseller
+%% Set teh reseller_id to the provided value on the provided account
```
jamesaimonetti: s/teh/the/

core/whistle_services-1.0.0/src/whs_account_conversion.erl
```diff
@@ -0,0 +1,174 @@
...
+                {'error', _R}=Error ->
+                    io:format("unable to set pvt_reseller on account ~s defintion: ~p~n", [AccountId, _R]),
+                    Error
+            end
+    end.
+
+-spec has_reseller_descendants(ne_binary()) -> boolean().
+has_reseller_descendants(AccountId) ->
+    %% its very important that this check not operate against stale data!
+    _ = couch_mgr:flush_cache_docs(),
```
jamesaimonetti: This is very heavy-handed...surely `couch_mgr:flush_cache_docs(?WH_ACCOUNTS_DB)` would be specific enough?

Also, the cache isn't involved in view lookups, so not sure why you're flushing here anyway?

core/whistle_services-1.0.0/src/whs_account_conversion.erl
```diff
@@ -0,0 +1,174 @@
...
+                {'error', _R}=Error ->
+                    io:format("unable to set ~s on services doc ~s: ~p~n", [Key, AccountId, _R]),
+                    Error
+            end
+    end.
+
+-spec update_account_definition(ne_binary(), ne_binary(), any()) -> {'error', _} | 'ok'.
+update_account_definition(AccountId, Key, Value) ->
+    AccountDb = wh_util:format_account_id(AccountId, 'encoded'),
+    case couch_mgr:open_doc(AccountDb, AccountId) of
```
jamesaimonetti: There is a `wh_util:account_update/1` which takes the modified account object and saves it to both the account db and the accounts db

core/whistle_services-1.0.0/src/whs_account_conversion.erl
```diff
@@ -0,0 +1,174 @@
...
+-spec force_make_reseller(ne_binary()) -> {'error', _} | 'ok'.
+force_make_reseller(Account) ->
+    AccountId = wh_util:format_account_id(Account, 'raw'),
+    do_make_reseller(AccountId).
+
+-spec do_make_reseller(ne_binary()) -> {'error', _} | 'ok'.
+do_make_reseller(AccountId) ->
+    _ = update_account_definition(AccountId, <<"pvt_reseller">>, 'true'),
+    _ = maybe_update_services(AccountId, <<"pvt_reseller">>, 'true'),
+    io:format("promoting account ~s to reseller status, updating sub accounts~n", [AccountId]),
```
jamesaimonetti: Since these functions can be called from Crossbar's `cb_accounts`, `lager` should be use instead of `io`

core/whistle_services-1.0.0/src/whs_account_conversion.erl
k-anderson: @jamesaimonetti  It is for wh_services:is_reseller(kz_account:id(JObj)) but I could flush just the services db from the cache, thats not a bad idea ;)

core/whistle_services-1.0.0/src/whs_account_conversion.erl
k-anderson: @jamesaimonetti yup, copy paste from the maintenance module but I will update

core/whistle_services-1.0.0/src/whs_account_conversion.erl
k-anderson: @jamesaimonetti they can also be called from SUP, they used to be maintenance commands but I didn't like calling a maintenance module from crossbar.  So I left them for people who are used to it.

core/kazoo_documents-1.0.0/src/kz_account.erl
```diff
@@ -397,6 +401,52 @@ is_expired(JObj) ->
...
+    wh_json:set_value(?RESELLER, 'true', JObj).
+
+%%--------------------------------------------------------------------
+%% @public
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec demote(doc()) -> doc().
+demote(JObj) ->
+    io:format("demote~n", []),
```
siplabs: io:format?

core/whistle_services-1.0.0/src/whs_account_conversion.erl
```diff
@@ -0,0 +1,156 @@
...
+-spec force_promote(ne_binary()) -> {'error', _} | 'ok'.
+force_promote(Account) ->
+    AccountId = wh_util:format_account_id(Account, 'raw'),
+    do_promote(AccountId).
+
+-spec do_promote(ne_binary()) -> {'error', _} | 'ok'.
+do_promote(AccountId) ->
+    _ = wh_util:account_update(AccountId, fun kz_account:promote/1),
+    _ = maybe_update_services(AccountId, <<"pvt_reseller">>, 'true'),
+    io:format("promoting account ~s to reseller status, updating sub accounts~n", [AccountId]),
```
siplabs: io:format?

PR #1623, opened 2016-02-18
KAZOO-4569: set PAs to …/test instead of …/ebin when running test suite

fenollp:
Ticket explains a bit what this does.

Tests rely on other apps' code off course, and we've always been running the test suite app by app.

BUT I'm stuck in KNM as stepswitch relies on KNM code, and KNM's calls to whapps_config are replaced at test-time. Thus stepswitch doesn't know about them and directly calls whapps_config…

Now, maybe whapps_config should provide mock functions for when we're running the test suite,
but that would be unreliable until now:

This patch compiles all the code in "test mode" (setting `-DTEST`) before running apps tests app-by-app.

(Providing mock functions and testing against them is probably terrible BTW).

BUT since it's easier for development to not pollute ebin/ we're generating these BEAMs to test/ then `-pa`ing apps with `-pa my/app/test` (instead of `-pa my/app/ebin`).

So, either
* we stay inside `test/` and we write our own `lib_dir/1` JUST to please the tests suite
* we pollute `ebin/` with test modules & normal modules *that were recompiled in test mode* (ie: a module can have changed completely)
    * but that's not too bad: one just needs to run `make clean` before anything else

so here I'm stuck

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,5 +36,6 @@ after_success:
   - make build-plt
   - ./scripts/check-dialyzer.escript .kazoo.plt $files
   - make build-release
+  - make compile-test
   - ERL_LIBS="$HOME/proper" make eunit
   - ./scripts/check-whitespace.sh core/ applications/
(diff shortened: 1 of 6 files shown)


fenollp:
It's ready now

PR #1624, opened 2016-02-18
cache_perf-3.22: serialize certain event processing

k-anderson:


jamesaimonetti:
@k-anderson do you intend for this to go to 3.21 too (since you know who is on there as well and has fast computers)?

PR #1626, opened 2016-02-19
HELP-21030: on-board tools use a series of requests give them time to…

k-anderson:
… finish before searching for admins

### DIFF
--- a/applications/teletype/src/templates/teletype_new_account.erl
+++ b/applications/teletype/src/templates/teletype_new_account.erl
@@ -71,8 +71,6 @@ handle_new_account(JObj, _Props) ->
 -spec process_req(wh_json:object()) -> 'ok'.
 -spec process_req(wh_json:object(), wh_proplist()) -> 'ok'.
 process_req(DataJObj) ->
-    %% Give onboarding tools time to add initial users...
-    timer:sleep(30000),
     %% Load templates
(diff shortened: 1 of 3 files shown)


applications/teletype/src/templates/teletype_new_account.erl
```diff
@@ -71,6 +71,8 @@ handle_new_account(JObj, _Props) ->
 -spec process_req(wh_json:object()) -> 'ok'.
 -spec process_req(wh_json:object(), wh_proplist()) -> 'ok'.
 process_req(DataJObj) ->
+    %% Give onboarding tools time to add initial users...
+    timer:sleep(30000),
```
jamesaimonetti: I would put this on the `cb_accounts` side personally; maybe `timer:apply_after/4` on the [publish](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/modules/cb_accounts.erl#L1465) (which should probably use a call to wh_amqp_worker as well).

My 2 cents

k-anderson:
@jamesaimonetti I am not opposed to that, I will make that change today
@lazedo sorry when I say on-board I am referring to external systems (like our PHP demo) that call a sequence of crossbar APIs.  We don't have a way to determine when it is complete but 30 seconds seems like a reasonable assumption ;)

PR #1631, opened 2016-02-20
correct attachment content_type type

onnet:
Attachment couldn't be retrieved because of wrong content_type:
"content_type": "<<\"text/html\">>" instead of "content_type": "text/html"


onnet:
Closed due to Luis comment about updated couchbeam

PR #1632, opened 2016-02-20
KAZOO-4572: mock functions for whapps_config

fenollp:


core/whistle_config/src/whapps_config.erl
```diff
@@ -223,6 +260,12 @@ get_ne_binary(Category, Key, Default, Node) ->
 -spec get(config_category(), config_key(), Default) -> any() | Default.
 -spec get(config_category(), config_key(), Default, ne_binary() | atom()) -> any() | Default.
 
+-ifdef(TEST).
```
jamesaimonetti: Shouldn't this be the only macro test? All the other accessors (`get_*`) call get/2 or get/4, no?

fenollp: Eh. Very good idea!

PR #1636, opened 2016-02-24
KAZOO-4574 track dependencies better

lazedo:
@fenollp please checkout , test and update it.

check the logic of `ERL_LIBS` inclusion in `core/Makefile` and  `applications/Makefile`
i think we should create a kz.mk with only that logic (ERL_LIBS) and another include to `../make/kz.mk` in core and applications and have the core apps and whapps include the `kz.mk`from the level above.
with this change i believe we can remove the `EBINS` logic from all Makefiles.

thoughts ?


lazedo:
@fenollp i went through the changes to `kz.mk` and remove of `EBINS` and include of it to set `ERL_LIBS`. looks good :)

PR #1637, opened 2016-02-25
Prompt flushing

jamesaimonetti:
We accomplish several improvements here:

1. Tombstoning of documents is handled in `couch_util`, so we remove the code in `crossbar_doc` as we lost valuable information for the publishing of the doc change AMQP payload.
2. `ecallmgr_util` was failing to set prompt-specific cache bindings to process the doc changes for media. Currently, we remove all account-specific media on changes, as we can't easily reconstruct the prompt name that would be in the doc change routing key.
3. Checks have been added to either directly execute ETS operations or call into the `wh_media_map` process to do the ETS work. This alleviates some deadlock conditions found.
4. Check for undefined when building the prompt path and use `system_media`

### DIFF
--- a/core/whistle_media-1.0.0/src/wh_media_map.erl
+++ b/core/whistle_media-1.0.0/src/wh_media_map.erl
@@ -357,6 +357,8 @@ maybe_add_prompt(AccountId, JObj, PromptId) ->
 
     insert_map(UpdatedMap).
 
+-spec insert_map(media_map()) -> 'ok' | 'true'.
+-spec insert_map(media_map(), pid()) -> 'ok' | 'true'.
 insert_map(Map) ->
     insert_map(Map, whereis(?MODULE)).
 insert_map(Map, Srv) when Srv =:= self() ->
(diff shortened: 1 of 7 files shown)


jamesaimonetti:
@k-anderson barring travis, specs are in place and PR is ready for review/merge.

core/whistle_media-1.0.0/src/wh_media_map.erl
```diff
@@ -306,9 +306,17 @@ init_map(Db, View, _StartKey, Limit, SendFun, ViewResults) ->
     end.
 
 -spec add_mapping(ne_binary(), fun(), wh_json:objects()) -> 'ok'.
+-spec add_mapping(ne_binary(), fun(), wh_json:objects(), pid()) -> 'ok'.
 add_mapping(Db, SendFun, JObjs) ->
+    add_mapping(Db, SendFun, JObjs, whereis(?MODULE)).
+
+add_mapping(Db, _SendFun, JObjs, Srv) when Srv =:= self() ->
```
fenollp: Since start_link registers `?MODULE` (locally), isn't `Srv` always `self()`?

Note: `whereis/1` "fails" http://erldocs.com/18.2/erts/erlang.html#whereis/1

PR #1638, opened 2016-02-25
Syntax correction

neurosys-zero:
Syntax correction of function name in user notice from “channel_detail”
to “channel_details”

### DIFF
--- a/applications/webhooks/src/webhooks_util.erl
+++ b/applications/webhooks/src/webhooks_util.erl
@@ -211,7 +211,7 @@ fire_hook(_JObj, Hook, _URI, _Method, _Retries, {'ok', "200", _, _RespBody}) ->
     successful_hook(Hook);
 fire_hook(_JObj, Hook, _URI, _Method, Retries, {'ok', RespCode, _, RespBody}) ->
     _ = failed_hook(Hook, Retries, RespCode, RespBody),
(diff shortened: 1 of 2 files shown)


fenollp:
Please rebase onto master

applications/webhooks/src/webhooks_util.erl
```diff
@@ -211,7 +211,7 @@ fire_hook(_JObj, Hook, _URI, _Method, _Retries, {'ok', "200", _, _RespBody}) ->
     successful_hook(Hook);
 fire_hook(_JObj, Hook, _URI, _Method, Retries, {'ok', RespCode, _, RespBody}) ->
     _ = failed_hook(Hook, Retries, RespCode, RespBody),
-    lager:debug("non-200 response code: ~s", [RespCode]);
+    lager:debug("non-200 response code: ~s on account ~s", [RespCode, AccountId]);
```
jamesaimonetti: This will never compile as `AccountId` isn't defined anywhere in this function clause. I think you need to change the function header to read:

```erlang
fire_hook(_JObj, #webhook{account_id=AccountId}=Hook, _URI, _Method, Retries, {'ok', RespCode, _RespHeaders, RespBody}) ->
```

Then your log line should compile nicely.

k-anderson:
Updated manually to resolve conflict.

PR #1640, opened 2016-02-25
KAZOO-4488

icehess:


applications/crossbar/src/crossbar_doc.erl
```diff
@@ -156,10 +158,65 @@ load([_|_]=IDs, Context, Opts, _RespStatus) ->
...
+    case props:get_value(Name, ?SPECIAL_EXPECTED_TYPE) of
+        'undefined' -> depluralize_resource_name(Name);
+        Special -> Special
+    end.
+
+-spec depluralize_resource_name(ne_binary()) -> ne_binary().
+depluralize_resource_name(Name) ->
+    Size = byte_size(Name) - 1,
+    case Name of
+        <<Bin:Size/binary, L:8>> when L == $s -> Bin;
```
fenollp: I believe you can match `<<Bin:Size, "s">>` now that BEAM knows the size.

(yes it's pretty dumb, binaries "know" their size, so why doesn't BEAM allow rhs matching…)

k-anderson:
@icehess please don't kill me but this should be against master.  I know that when we talked I said 3.22 but I misspoke.  If you would like I can move the PR, let me know.

PR #1643, opened 2016-02-26
KAZOO-4220: set custom-sip-headers on outbound legs

icehess:
Set custom-sip-headers on outbound legs for calls from the device.
It reads CSH from the old key-value format and then merge it with
the new format which is a nest of two objects, e.g.:

custom_sip_headers: {x-device-header: value}
custom_sip_headers: {out:{x-outbound-header: value},in:{x-device-header: value}}

### DIFF
--- a/applications/callflow/src/cf_endpoint.erl
+++ b/applications/callflow/src/cf_endpoint.erl
@@ -1226,8 +1226,7 @@ generate_sip_headers(Endpoint, Acc, Call) ->
 maybe_add_sip_headers(JObj, Endpoint, Call) ->
     case ?MODULE:get(Call) of
         {'ok', AuthorizingEndpoint} ->
-            MergeFuns = [fun() -> kz_device:custom_sip_headers(Endpoint) end
-                         ,fun() -> kz_device:custom_sip_headers_inbound(Endpoint) end
+            MergeFuns = [fun() -> kz_device:custom_sip_headers_inbound(Endpoint) end
                          ,fun() -> kz_device:custom_sip_headers_outbound(AuthorizingEndpoint) end
                         ],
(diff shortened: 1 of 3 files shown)


applications/callflow/src/cf_endpoint.erl
```diff
@@ -1216,19 +1216,33 @@ generate_sip_headers(Endpoint, Call) ->
...
-        'undefined' -> JObj;
-        CustomHeaders -> wh_json:merge_jobjs(CustomHeaders, JObj)
+-spec maybe_add_sip_headers(wh_json:object(), wh_json:object(), whapps_call:call()) -> wh_json:object().
+maybe_add_sip_headers(JObj, Endpoint, Call) ->
+    case ?MODULE:get(Call) of
+        {'ok', AuthorizingEndpoint} ->
+            MergeFuns = [fun() -> kz_device:custom_sip_headers(Endpoint) end
+                         ,fun() -> kz_device:custom_sip_headers_inbound(Endpoint) end
+                         ,fun() -> kz_device:custom_sip_headers_outbound(AuthorizingEndpoint) end
+                        ],
```
lazedo: @icehess i think it would be cleaner if you search/replace every call to kz_device:custom_sip_headers to kz_device:custom_sip_headers_inbound, and have kz_device:custom_sip_headers_inbound handle both legacy and new format.

k-anderson:
@icehess confirmed with requester this can go to 3.22 and we should update as such.

applications/callflow/src/cf_endpoint.erl
icehess: @lazedo 
done

icehess:
@k-anderson ok, waiting for review and CI pass.

PR #1644, opened 2016-02-27
KAZOO-4562 Part I

lazedo:
still a lot to be done but basic refactoring done.
needs consolidating on abstraction specs


fenollp:
Can you remove your changes to /Makefile? these should be part of a separate PR
Also, there's a bunch of indentation messed up due to `sed`, but I'm being James-y :)

fenollp:
I think your changes to `sip_headers` stuff is the cause of the `conflicts that need to be resolved`

fenollp:
Also, why leave out commented code in fax_request?

lazedo:
@fenollp there's a lot of code commented before full removal

fenollp:
kazoo_couch.app.src should be
```erlang
{application, kazoo_couch,
 [
   {description, "Simplified interface to CouchDB"}
  ,{vsn, "4.0.0"}
  ,{modules, []}
  ,{registered, [kazoo_couch_app, kazoo_couch_sup]}
  ,{applications, [ kernel
                  , stdlib
                  , crypto
                  , public_key

                  , couchbeam
                  , lager
                  ]}
  ,{mod, {kazoo_couch_app, []}}
 ]}.
```

(thus moving the application behaviour implementation to kazoo_couch_app, following the way of rest of our apps)
Same for kazoo_data

lazedo:
i don't agree with this naming

fenollp:
okay there's a lot of commented code and missing things, I'll trust you on fixing that in Part 1+n :)

What's that WITH_JIFFY about in Makefile.deps?

lazedo:
what's missing ?
`WITH_JIFFY` is for couchbeam compilation

fenollp:
Okay LGTM. I would have put wh_doc in kazoo_documents but whatever.

Could you add a check to .travis.yml that makes sure there are no calls to the abstracted couch modules in apps other than kazoo_couch? That would make it easier to maintain this abstraction.
(Have a look at scripts/check-whitespace.sh)

lazedo:
not sure what you're asking. `kz_dataconfig` will check if module claiming to be a `driver` implements the abstraction layer. other apps may implement kz_data, `kazoo_couch` is only the first to ensure stuff keeps working.

lazedo:
@fenollp need your help here. the `-DWITH_JIFFY` is not propagated down to `couchbeam`. `erlang.mk` is explicitely setting `ERLC_OPTS =` instead of adding it, in the generated `Makefile`. `erlang.mk` bug ?

fenollp:
Regarding the `..._app` naming scheme, what's wrong with it? also, consistency

lazedo:
what's wrong with not using it ? i don't want to start whistle_apps_app i want to start whistle_apps. do you use  `-s lager_app` or do you use `-s lager` as with other apps (except kazoo apps) ?

fenollp:
haha but that name isn't the actual application name. The app's name is the one after `application, ` in the .app

When you `application:start(App)` it fetches `App.app` then the `{mod, {Mod,...}}` there and calls `Mod:start/2`. 

Now, you're talking about `start/0` (`-s` option) and that's a different issue. All I'm saying is: let's keep it consistent and implement the `application` behaviour in the usual `..._app` module.

fenollp:
I asked Loic: there is no way to add flags to ERLC "on the fly" while compiling deps (he says edit the generated Makefile at compile time).
So vendor it is, I guess.

lazedo:
i know what it does, but IIRC this changed recently when you renamed all the stuff around  `_app`

lazedo:
@fenollp (erlamg.mk) , that or change the way we maintain dependencies. i don't want to go back on deps matter. i think there are some options in erlang.mk to not generate the makefile, not sure if that«s by dep or it works as global. how far did you go with rebar3 ?

fenollp:
I'm saying you can put your start/0 in kazoo_couch and use `-s kazoo_couch`, but put the app callbacks in the _app. Then we're both happy?

My changes only removed start/0 from App.erl and I guess we should maybe revert that. Anyway, apps kinda have to be started by whapps_controller

fenollp:
For ERLC_OPTS to take effect, did you try putting [this line](https://github.com/2600hz/kazoo/blob/KAZOO-4562/make/Makefile.deps#L4) at the very end of the file?

lazedo:
sure

PR #1645, opened 2016-02-29
KAZOO-4220: set custom-sip-headers on outbound legs

icehess:
Supersedes #1643


lazedo:
@icehess please close this and make the changes you want to merge against 3.22

PR #1650, opened 2016-03-01
KAZOO-4582: As an admin I would like cb_transactions to be able not o…

onnet:
…nly to debit but also to credit account's balance

### DIFF
--- a/applications/crossbar/src/modules/cb_transactions.erl
+++ b/applications/crossbar/src/modules/cb_transactions.erl
@@ -13,6 +13,7 @@
          ,resource_exists/0, resource_exists/1
          ,validate/1, validate/2
          ,to_csv/1
+         ,put/2
          ,delete/2
         ]).
 
@@ -22,6 +23,7 @@
 -define(CURRENT_BALANCE, <<"current_balance">>).
 -define(MONTHLY, <<"monthly_recurring">>).
 -define(SUBSCRIPTIONS, <<"subscriptions">>).
+-define(CREDIT, <<"credit">>).
 -define(DEBIT, <<"debit">>).
 
 -type payload() :: {cowboy_req:req(), cb_context:context()}.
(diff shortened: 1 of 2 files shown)


applications/crossbar/src/modules/cb_transactions.erl
```diff
@@ -77,6 +80,8 @@ allowed_methods() ->
 
 allowed_methods(?CURRENT_BALANCE) ->
     [?HTTP_GET];
+allowed_methods(?CREDIT) ->
+    [?HTTP_PUT];
 allowed_methods(?DEBIT) ->
     [?HTTP_DELETE];
```
skorobkov: Maybe add PUT for debit to?
And DELETE for credit.

onnet: I, actually do not care which verbs to use, but it looks like PUT is the better choice for both cases.
I think PUT is better because it reflects the point of creating of transaction - debit or credit.

DELETE, in its turn, reflects an attempt to delete already created transaction. And at this moment I believe we've better avoid this feature until it is widely discussed and proven as a must.    

As of me, this PR is an attempt to understand whether core developers are open at least to touch this area. After we'll find out current position, we'll define our TODO/Feature list as discussed at 2600hz-dev list.

At the same time, since we've already got Darren's comments regarding possible billing improvements/development I'm thinking  very positive about both ways - step by step PRs or separate addon development (crossbar module/bookkeeper/etc..), because at the end of the day it should be converted into different flavors billing plugins ... dreams, dreams ... :)

k-anderson:
@onnet Howdy!

This needs a rebase but more importantly if I am reading this correctly it will allow direct accounts of the master reseller to add to their credit... Unless I missed it then it was only adding to transactions which does not invoke the bookkeeper.

I think there are a few use-cases that need to be addressed by such an API.

1) The super-admin should be able to add credit to any account without invoking the bookkeeper to provide a "free" credit to the account.
2) The super-admin should be able to add credit to an account invoking the bookkeeper to provide "normal" credit
3) Accounts that are belong to the master reseller should invoke the bookkeeper.
4) If the account belongs to a sub-reseller and the auth user is that reseller then the credit should be added without invoking the bookkeeper.
5) All other requests should be denied.

Also, one possible approach would be to add the transaction to "transactions" on the services doc.  This is how top-up and other transactions are batched to be processed by the appropriate bookkeeper.  This would avoid having to preform bookkeeper actions in the module.  For example, you could create the transaction then write it to "transactions" on the service doc unless it was the super-admin "free" credit in which you write the transaction to the appropriate MODb. 

Let me know if you have questions! :)

onnet:
@k-anderson Hi Karl. Thanks for the comment. Everything looks pretty straightforward. Will get back to this PR next weekend. Regards,

lazedo:
@onnet can you please squash your commit ? thanks

onnet:
@lazedo Hi! Squashed

PR #1653, opened 2016-03-03
Correct content length for crossbar CSV response

kalda341:
The Content-Length header was incorrect before - though I'm not sure why. This works as intended.
I know this is potentially inefficient, but if the CSV response is expected to be large then it will probably be chunked anyway.

### DIFF
--- a/applications/crossbar/src/api_resource.erl
+++ b/applications/crossbar/src/api_resource.erl
@@ -738,7 +738,7 @@ to_csv(Req, Context) ->
         _ ->
             RespBody = maybe_flatten_jobj(Context1),
             RespHeaders1 = [{<<"Content-Type">>, <<"application/octet-stream">>}
-                            ,{<<"Content-Length">>, iolist_size(RespBody)}
+                            ,{<<"Content-Length">>, size(wh_util:to_binary(RespBody))/8}
                             ,{<<"Content-Disposition">>, <<"attachment; filename=\"data.csv\"">>}
                             | cb_context:resp_headers(Context1)
                            ],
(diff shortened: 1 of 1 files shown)


jamesaimonetti:
Do you have an example where this wasn't computed properly? I'd hate to convert an iolist to binary just to get the size. Much more efficient when we can just send the iolist to the socket for sending than doing the conversion to binary first.

kalda341:
I played around a bit and was unable to reproduce it!
...Until I went back to curl. The content length is wrong with the --compressed option. The gzip encoding is being used.

With compression only 2534 bytes are received out of 18321.
Without compression all 18321 bytes are received.

For some reason calculating the length the other way fixes this (though I don't know why.)

kalda341:
But then I realised I hadn't tested my fix without compression - it gives the same length as it did when compressed. So it's just a coincidence that my fix worked, and the problem is elsewhere?

jamesaimonetti:
I suspect its the introduction of the [compress](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/crossbar_init.erl#L141) flag that does it. I wonder if we remove calculating the content length here and see if cowboy will add the header with appropriate size depending on compression options of the client? Care to test that @kalda341 ?

kalda341:
That does indeed fix the issue. I wonder why it was there in the first place?

jamesaimonetti:
@kalda341 I suspect when we were using mochiweb initially, or early versions of Cowboy, we didn't get the right header values? Hard to say; lost to the mists of time I guess. Thanks for looking into this though; good to have it do the right thing.

PR #1655, opened 2016-03-03
Event changes

mark2600:


applications/callflow/src/cf_exe.erl
```diff
@@ -544,78 +544,92 @@ handle_info(_Msg, State) ->
     lager:debug("unhandled message: ~p", [_Msg]),
     {'noreply', State}.
 
+handle_channel_destroyed(Self, NotifyPids, JObj) ->
```
jamesaimonetti: Pulling a @fenollp to ask for specs!

PR #1660, opened 2016-03-04
Only allow super duper admins to view sibling accounts

kalda341:
It doesn't seem right to allow any account to view their siblings


k-anderson:
@kalda341 

Howdy,

Conceptually if you are above an account in the hierarchy it denotes ownership/responsibility over that account.  The intermediary tiers are usually organizational.  I think this makes sense but everyone has a different use-cases :)

We would be unable to accept this PR as is because it removes existing functionality!  I know of a few clients that utilize this, including accounts on our hosted offering.

I can think of two approaches to achieve your desired behavior.  The first is to reorganize your accounts which would be painful.  The second would be to make this a system configuration parameter, which by default kept the existing functionality.  On your cluster you would be able to change this to opt out of this feature.

If you have questions on how to make this a system configuration parameter let us know!

k-anderson:
@kalda341 I am going to close this PR but if you would like to re-open it with a system config parameter to change the behavior feel free!

kalda341:
I don't seem to be able to reopen this, and my new commit isn't showing up. Can you please reopen the PR?
Thanks

k-anderson:
@kalda341 after further discussion among the team we have decided to change the default value of the configuration parameter to 'false' and update the ANNOUNCEMENTS file.  The system configuration will allow those using this feature to continue doing so but after polling some community members as well as internally the decision was made to create a non-backward compatible change.

PR #1661, opened 2016-03-04
Add reloader to dep for scripts/dev-start-apps.sh

skorobkov:
It was part of mochiweb.


jamesaimonetti:
I'd rather look at using the [reload](http://erlang.mk/guide/plugins_list.html#_reload_mk) plugin for erlang.mk since that's what we've been moving the build system to. Would like to compare it to reloader.erl at the least.

fenollp:
@jamesaimonetti erlang.mk only builds deps.

For now this path LGTM. We should be careful though that `reloader` isn't started in non-dev mode, but I understand this is already taken care of as `rel/relx.config.src` does not start it, only loads it.

jamesaimonetti:
Good to know. Moving along...

PR #1670, opened 2016-03-05
decode to {} and log input when failing hard

fenollp:
cherry pick of #1668


lazedo:
i don't agree on this. If a payload is malformed, I would prefer to throw the exception

jamesaimonetti:
I agree, too many chances of defaults being applied to an empty object rather than letting the client know their request was malformed. Let it crash!

fenollp:
I'll rework it. What should I do about this patch in 3.22? (it's there already)

fenollp:
updated as discussed on standup

PR #1675, opened 2016-03-05
ci-proper: run proper tests too

fenollp:


jamesaimonetti:
@fenollp not sure this worked. I don't see the [token bucket](https://github.com/2600hz/kazoo/blob/master/core/kazoo_token_buckets/test/kz_token_bucket_pqc.erl) tests run (unless I missed it in the logs).

jamesaimonetti:
[kazoo bindings](https://github.com/2600hz/kazoo/blob/master/core/kazoo_bindings/test/kazoo_bindings_test.erl) too

jamesaimonetti:
Sweet!

PR #1676, opened 2016-03-07
add documentation for the custom_sip_headers new format

icehess:


fenollp:
@jamesaimonetti is there a way to have these test pass all the time?

applications/callflow/doc/custom_sip_headers.md
```diff
@@ -0,0 +1,41 @@
...
+> **For compatibility reason**, the pervious flat object format (basic key value) is still supported by merging its contents with the contents of `in` object of the new format. The result is set for SIP headers on the `INVITE` to a device and not on other calls.
+
+#### Configuring custom SIP headers
+To set custom SIP headers for calls to/from a device specify desire headers in `custom_sip_headers` in the device document as follow:
+
+* New nested format(`in` will be set on calls to a device **and** `out` will be set on calls from a device):
+
+```
+"custom_sip_headers": {
+    "in": {
```
jamesaimonetti: This is more for @k-anderson but why was `in` and `out` chosen instead of `inbound` and `outbound`?

jamesaimonetti:
@fenollp The problem is that doing property-based testing when time is important is *hard*!. I'd like to refine the tests but don't really have the time (har). [This paper](https://www.researchgate.net/publication/259785220_paper_LNCS13_authorcopy) seems interesting though. Maybe we can work on it during EF/training.

fenollp:
In the meanwhile I just restarted the build.
Maybe this test should be a Common Test instead? I'm wondering, because otherwise we'd have to disable these tests again, or have them run just for builds on the master branch (that wouldn't be too bad I guess)

applications/callflow/doc/custom_sip_headers.md
k-anderson: @jamesaimonetti no particular reason the ticket stated in and out, I think that inbound and outbound was only discussed.  I think I would prefer it but it would need to be backward compatible since its released.

icehess: @jamesaimonetti that was discuss on the ticket. 
@k-anderson Do you mean change it to inbound and outbound?

PR #1677, opened 2016-03-07
Qubicle

mark2600:

PR #1678, opened 2016-03-07
merge qubicle branch changes, the hard way

mark2600:


fenollp:
@mark2600 Should #1677 be closed?

mark2600:
@fenollp yes, it should be destroyed, never to be seen again

PR #1682, opened 2016-03-08
HELP-21287

lazedo:
port forward

### DIFF
--- a/applications/stepswitch/src/stepswitch_inbound.erl
+++ b/applications/stepswitch/src/stepswitch_inbound.erl
@@ -18,11 +18,12 @@
 %%--------------------------------------------------------------------
 -spec handle_req(wh_json:object(), wh_proplist()) -> 'ok'.
 handle_req(JObj, _Props) ->
-    'true' = wapi_route:req_v(JObj),
     _ = wh_util:put_callid(JObj),
+    'true' = wapi_route:req_v(JObj),
     case wh_json:get_ne_value(?CCV(<<"Account-ID">>), JObj) of
         'undefined' -> maybe_relay_request(JObj);
-        _AcctID -> 'ok'
+        AccountId -> lager:debug("fetch-id ~s already has account-id ~s, skipping.",
+                                 [wapi_route:fetch_id(JObj), AccountId])
     end.
 
 %%--------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


applications/trunkstore/src/ts_route_req.erl
```diff
@@ -18,22 +18,22 @@ init() -> 'ok'.
...
     of
         {AcctID, 'undefined'} when is_binary(AcctID) ->
             %% Coming from carrier (off-net)
-            lager:info("call began from outside the network"),
+            lager:info("call with fetch-id ~s began from outside the network", [wapi_route:fetch_id(ApiJObj)]),
             ts_offnet_sup:start_handler(<<"offnet-", CallID/binary>>, ApiJObj);
         {AcctID, AuthID} when is_binary(AcctID) andalso is_binary(AuthID) ->
             %% Coming from PBX (on-net); authed by Registrar or ts_auth
-            lager:info("call began on the network"),
+            lager:info("call with fetch-id began on the network", [wapi_route:fetch_id(ApiJObj)]),
```
fenollp: missing `~s`

applications/stepswitch/src/stepswitch_inbound.erl
```diff
@@ -18,11 +18,12 @@
 %%--------------------------------------------------------------------
 -spec handle_req(wh_json:object(), wh_proplist()) -> 'ok'.
 handle_req(JObj, _Props) ->
-    'true' = wapi_route:req_v(JObj),
     _ = wh_util:put_callid(JObj),
+    'true' = wapi_route:req_v(JObj),
     case wh_json:get_ne_value(?CCV(<<"Account-ID">>), JObj) of
         'undefined' -> maybe_relay_request(JObj);
-        _AcctID -> 'ok'
+        AccountId -> lager:debug("fetch-id ~s already has account-id ~s, skipping.",
```
jamesaimonetti: Typically we keep variables prefixed with _ when only used in logging

jamesaimonetti: Additionally, I don't think the app should log if its not handling the request. My personal feeling but others can chime in.

k-anderson: @jamesaimonetti actually that was the point ;) we were missing the information during a debug session and we needed to know which apps ignored  a route_req.  Since it was a commit to 3.22 it needs to be on master as well (by policy) and I am behind that.  We can discuss removal if the logs are too chatty after we have collected the data we need :)

PR #1683, opened 2016-03-08
KAZOO-4591: set custom tags read from endpoint doc to ccvs

link89:
This PR enable reading custom tags from device documents and set them to CCVS, so that whenever you make or receive a phone call from this device, channel events generated by Kazoo will carry those tags (and can also be found in CDR and Webhooks events).

The custom tags looks like this in couchdb:
"custom_tags" : {
    "k1" : "v1",
    "k2" : "v2"
}

### DIFF
--- a/applications/callflow/src/cf_endpoint.erl
+++ b/applications/callflow/src/cf_endpoint.erl
@@ -1307,6 +1307,7 @@ generate_ccvs(Endpoint, Call, CallFwd) ->
                ,fun set_sip_invite_domain/1
                ,fun maybe_set_call_waiting/1
                ,fun maybe_auto_answer/1
+               ,fun maybe_set_custom_tags/1
               ],
     Acc0 = {Endpoint, Call, CallFwd, wh_json:new()},
     {_Endpoint, _Call, _CallFwd, JObj} = lists:foldr(fun(F, Acc) -> F(Acc) end, Acc0, CCVFuns),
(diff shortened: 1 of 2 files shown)


lazedo:
@link89 what's the difference between this and custom_sip_headers ?

link89:
@lazedo I didn't go into the details of custom_sip_headers. I guess by its name that it may refer to something like "alert-info" which would appears in sip headers. What I want is just label channel events generated by specific devices and leave sip or other things undisturb.

lazedo:
@link89 but what's the end goal? to have these tags in cdrs ? you can get them from device by using the authorizing-id in cdr to fetch the device properties you want ? i would like not to mix these tags with custom_channvel_vars

link89:
@lazedo We use Kazoo as an outbound call center and devices are shared by employees  who work shifts. Since we have an ERP system, what we need is an API to put our employee info to CDRs and Webhooks. With this patch, this requirement can be accomplished by simply issue a http request to crossbar to add those custom tags to device profile whenever a staff start to work. 

I create this PR just in case someone else may also need it:)

lazedo:
@link89 and can't you get this information already by correlating the info in cdr ? IIRIC, if you use agent login/logout you'll have the owner set to the agent.

link89:
@lazedo Well, by assign a device to a kazoo user (no matter by crossbar, hotdesk, or agent login/logout as you had mentioned) does set an owner in CDRs.  If doing this way,  we have to maintain an relation between kazoo users and our employee in ERP. What this patch does is hard code this relation into CCVS. It just creates a short cut to make kazoo easily to integrate with other systems :)

lazedo:
@link89 but don't you already have it when you call the crossbar to update the device doc ? anyhow, my point is that if you have this information Authorizing-Id, you can easily fetch the rest of the data from wherever you put it without the need to flow this info thru the system.

jamesaimonetti:
@link89 Also, we're getting into a bad place where we use the custom channel vars as a key/value dumping ground. I wonder if you could instead extend the cdr app's `CHANNEL_DESTROY` handler to lookup the Authorizing-ID device/resource, and follow to the `owner_id` if any, looking for `"custom_cdr_vars":{"your-key":"your-value","your-key-2":"your-value-2"}` and would append that object to the CDR stored in the MODB. Would that work for your use-case?

link89:
@jamesaimonetti @lazedo  I would try. I think I may use channel vars a wrong way. Thanks for your kindly replys :)

PR #1688, opened 2016-03-08
KAZOO-4594: catch possible crashes in fax_workers

icehess:


lazedo:
@icehess please backport to 3.21

PR #1689, opened 2016-03-09
Crossbar: create MODB before reconciliation when add account

skorobkov:
Call `wh_services:reconcile/1` before creating MODB lead to crash:
> |0000000000|Undefined:Undefined (<0.1771.0>) CRASH REPORT Process <0.1771.0> with 0 neighbours exited with reason: timeout in couchbeam_view_stream:do_init_stream/2 line 120

because `wh_service_ledgers` trying to get data from nonexistent MODB, but `couchbeam_view_stream:do_init_stream/2` is not handle responce with 404 code and died with `timeout` reason.

### DIFF
--- a/applications/crossbar/src/modules/cb_accounts.erl
+++ b/applications/crossbar/src/modules/cb_accounts.erl
@@ -1246,11 +1246,11 @@ create_new_account_db(Context) ->
             _ = crossbar_bindings:map(<<"account.created">>, C),
             lager:debug("alerted listeners of new account"),
 
-            _ = wh_services:reconcile(AccountDb),
-            lager:debug("performed initial services reconcile"),
-
             _ = create_account_mod(cb_context:account_id(C)),
             lager:debug("created this month's MODb for account"),
+
+            _ = wh_services:reconcile(AccountDb),
+            lager:debug("performed initial services reconcile"),
 
             _ = create_first_transaction(cb_context:account_id(C)),
             lager:debug("created first transaction for account"),


k-anderson:
@skorobkov wow, nice catch and good debugging!  Just a heads up it might be related to your environment as we create accounts daily without seeing this and I don't believe this order has changed since 3.22 ;)  However, your logic and reasoning is valid and the reordering seems appropriate!

skorobkov:
I dont think the problem is in my envirement. It was fresh DB, created from scratch. Creating second (after super-admin) account already gave me that error.
And for example if i execute kz_datamgr/get_result/2 with nonexistant DB or view i got 10 seconds of delay and after that CRASH REPORT in console.
> kz_datamgr:get_results(<<"account%2Fc3%2F37%2F3f1bf158b002d4583767cdf7d7d8-201701">>, <<"ledgers/listing_by_service">>).
{error,timeout}
18:54:58.782 [error] CRASH REPORT Process <0.1456.0> with 0 neighbours exited with reason: timeout in couchbeam_view_stream:do_init_stream/2 line 120

Isn't it must return {error, not_found} (or something like that) imideately?
It not problem with DB (its answer 404 without delay), the problem in `couchbeam_view_stream:do_init_stream/2`, it dont handle answer from hackney like {status, 404, .<<"Not found">>}, only {status, 200, ...} or {error, ...}.

lazedo:
@skorobkov @k-anderson the change makes sense, but we need to solve the 404 upstream or at least in kazoo_couch and/or kz_datamgr

PR #1360, opened 2015-10-14
Access to "pvt_" fields in "limits" document for user/reseller.

skorobkov:
User can view some pvt_ fileds (`system_config/crossbar.limits/leak_pvt_fields`)
Reseller can view/edit some pvt_ fileds (`system_config/crossbar.limits/reseller_pvt_fields`)
By default it `allow_prepay`, `allow_postpay` and `max_postpay_amount` fields.

If reseller dont set some field (listed in reseller_pvt_fields), ten this "pvt_" field will be removed from "limits" document".

PR #1691, opened 2016-03-09
Dinkor cb limits more pvt

k-anderson:
@skorobkov 

This is my proposal to fix #1360 since we would rather not leak the pvt_ fields directly like that.  It will make our intentions difficult later since the pvt_ fields are not explicitly set or fetched so its ambiguous as to what is being changed.

As such I have created this PR which will expose the pvt_ fields in an object "overrides" so the end user could know if they have hard limits imposed by the system/reseller admins.  

Additionally, I created a two new system_config parameters for crossbar.limits which will allow you to control which accounts can change either the root level limits or the pvt_ (overrides).  These are lists of "types" of which the valid values are:
* master 
* master-client
* reseller
* reseller-client

I have set the defaults to maintain the existing behavior.

Also please be aware since this is changing the API payload (which has caused issues in the past with some systems, despite just being an addition) I have left v1 unchanged.

Let us know if this allows you to fulfill your use-case or if not what is missing.  If you are happy with it we will merge it in!

### DIFF
--- a/applications/crossbar/priv/couchdb/schemas/limits.json
+++ b/applications/crossbar/priv/couchdb/schemas/limits.json
@@ -39,6 +39,62 @@
             "required": false,
             "type": "integer"
         },
+        "overrides": {
+            "additionalProperties": false,
+            "properties": {
+                "allow_prepay": {
+                    "default": true,
+                    "description": "Determines if the account would like to allow per-minute calls if they have credit",
+                    "name": "Allow Prepay",
+                    "required": false,
+                    "type": "boolean"
+                },
+                "burst_trunks": {
+                    "description": "The number of two-way, flat-rate trunks used only if no other trunks are available",
+                    "minimum": 0,
+                    "name": "Burst Trunks",
+                    "required": false,
+                    "type": "integer"
+                },
+                "calls": {
+                    "description": "A hard limit for the total number calls",
+                    "minimum": 0,
+                    "name": "Calls",
+                    "required": false,
+                    "type": "integer"
+                },
+                "inbound_trunks": {
+                    "description": "The number of inbound, flat-rate trunks",
+                    "minimum": 0,
+                    "name": "Inbound Trunks",
+                    "required": false,
+                    "type": "integer"
+                },
+                "outbound_trunks": {
+                    "description": "The number of outbound, flat-rate trunks",
+                    "minimum": 0,
+                    "name": "Outbound Trunks",
+                    "required": false,
+                    "type": "integer"
+                },
+                "resource_consuming_calls": {
+                    "description": "A hard limit for the number of resource consuming calls",
+                    "minimum": 0,
+                    "name": "Resource Consuming Calls",
+                    "required": false,
+                    "type": "integer"
+                },
+                "twoway_trunks": {
+                    "description": "The number of two-way, flat-rate trunks",
+                    "minimum": 0,
+                    "name": "Twoway Trunks",
+                    "required": false,
+                    "type": "integer"
+                }
+            },
+            "required": false,
+            "type": "object"
+        },
         "resource_consuming_calls": {
             "description": "A hard limit for the number of resource consuming calls",
             "minimum": 0,
--- a/applications/crossbar/src/modules_v1/cb_limits_v1.erl
+++ b/applications/crossbar/src/modules_v1/cb_limits_v1.erl
@@ -21,9 +21,6 @@
 
 -define(CB_LIST, <<"limits/crossbar_listing">>).
 -define(PVT_TYPE, <<"limits">>).
--define(LIMITS_CONFIG_CAT, <<(?CONFIG_CAT)/binary, ".", (?PVT_TYPE)/binary>>).
--define(DEFAULT_LEAK_PVT_FIELDS, [<<"allow_postpay">>, <<"max_postpay_amount">>]).
--define(DEFAULT_RESELLER_PVT_FIELDS, [<<"allow_postpay">>, <<"max_postpay_amount">>]).
 
 %%%===================================================================
 %%% API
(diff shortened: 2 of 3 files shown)


skorobkov:
Main reason for my PR was code of `j5_limits`, where many limits exists only in "pvt_" form. These limits not showed in crossbar and can be modifed only by hands in DB.
- allow_postpay
- max_postpay_amount
- reserve_amount
- soft_limit_inbound
- soft_limit_outbound
- disconnect_active_calls

https://github.com/2600hz/kazoo/blob/master/applications/jonny5/src/j5_limits.erl#L79

Your PR is not solve the situation, This limits must be added into "override" schema and `LIMIT_PROPERTIES` macro. Then I can view/set them through "override" object. 
Another issues with this PR:
- there is no way I can remove "pvt_" limit, only add or change
- no need for default value of `allow_prepay` in "override" object.

Can you tel why this "pvt_" limits so special?
Maybe j5_limits should be changed? Make this limits not "pvt_" and behave like "standard" limits.

k-anderson:
@skorobkov 

Howdy, the original intent was for the pvt_ to only be modified by admins with access to the db directly, as a means of overriding the end-users.

I did not understand the following statement, perhaps you could rephrase it?
"Your PR is not solve the situation, This limits must be added into 'override' schema and LIMIT_PROPERTIES macro. Then I can view/set them through "override" object. "
As I understand this statement you are saying that my PR does not solve the problem because to set the pvt_ limits you would need to view/set them in the override object.  That is what I am proposing we change about your PR, access to the pvt_ settings SHOULD be done outside the normal limit properties so as to be more restrictive as well as disambiguate which data set is being manipulated.

The inability to delete was an oversight, I can correct that once we agree here ;)

This PR did not add a default value for `allow_prepay` so I am a bit unclear what you mean there.

Sure!  The pvt_ limits are parameters that the system operators can use to FORCE end-users to maximum values.  The pvt is short for "private", originally the concept was no pvt parameter should ever be set/viewed by the API as it would be sensitive internal data.

For example:
An end-user might attempt to set their "twoway_trunks" to a value like 5000.  If the system operator has set "pvt_twoway_trunks" to 100 then limits will use the system operators value as the accounts limits.  This protected the system operator from being abused.  However, if the end-user wanted to only pay for 50 twoway trunks then that would be used as the value because it is less than the pvt_twoway_limits, so the end-user still has control (just up to the pvt_ value).

Basically the end-user adds the values WITHOUT the pvt_ and the system operators add the values WITH the pvt_ to conditionally override the end-user.  If the end-user has access the the pvt_ values then this concept doesn't work at all ;)

The only parameters (iirc) that cannot exist without pvt_ are those associated with post-pay because those parameters control how much risk a system operator is willing to take on for a given account.  If a person looking to defraud an operator could set their account to use prepay with no maximum and unlimited trunks they could cost the operator a LOT of money!!!  The pvt_ fields are there to help control that.  If we leave it were only people who have access to the db can set them then there is NO risk for an exploit in the API that allows someone to set them that shouldn't.  At the same time I can see why an operator would want to access this via API so I am working with you to figure out how to safely do that ;)

Maybe you could tell me what you are trying to do, what is not working like you want, and what you want it to do?  That would help me code to your use-case closer!

skorobkov:
Ok. I will try explain from other side on example of postpay.

We have postapay clients. This clients must see information about postpay status through UI (API) - "You have balance XXX, you have postapy anabled, you have postpay limit YYY". And reseller must be able to change this postpay setting (why not?).

**Problem:** Current version cb_limits dont show `allow_postpay` & `max_postpay_amount`
And your version of cb_limits_v2 dont show it, because `allow_postpay` and `max_postpay_amount` not listed in macro LIMIT_PROPERTIES. And your version dosesn't allow to set this parameters, because`allow_postpay` and `max_postpay_amount` not listed in schema (in "override" object).

>For example:
An end-user might attempt to set their "twoway_trunks" to a value like 5000.
...

AFAIK [cb_limits](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/modules_v2/cb_limits_v2.erl#L75) doesn't allow end-user to change limits settings. Only reseller or master.  
So AccountDB-limits already protected, and maybe we can work with `allow_postpay` and `max_postpay_amount` as other "standard" paraamters? Current version `j5_limits` looks only `pvt_allow_postpay` and `pvt_max_postpay_amount` parameters.

>This PR did not add a default value for allow_prepay so I am a bit unclear what you mean there.

In schema, property `allow_prepay` in `override` object have default value:
```JSON
        "overrides": {
            "additionalProperties": false,
            "properties": {
                "allow_prepay": {
                    "default": true,
                    "description": "Determines if the account would like to allow per-minute calls if they have credit",
                    "name": "Allow Prepay",
                    "required": false,
                    "type": "boolean"
                },
```
But this is not critical...

There another issue with your PR.  
When update limit with same values (`have_overrides_changed/1` return `false`), "overrides": object from request writes into document.
Example:
POST data
```JSON
{
    "data": {
      	"outbound_trunks": 1,
      	"inbound_trunks": 5,
      	"twoway_trunks": 5,
      	"overrides": {
           	"inbound_trunks": 3
      	}
    }
}
```
document saved as
```JSON
{
   "_id": "limits",
   "_rev": "43-44a45d2e816a450d3d1c85fcfda0862e",
   "outbound_trunks": 1,
   "inbound_trunks": 5,
   "twoway_trunks": 5,
   "allow_prepay": true,
   "pvt_inbound_trunks": 3,
   "pvt_type": "limits",
...
}
```

If repeat same POST data, saved documents will look like this:
```JSON
{
   "_id": "limits",
   "_rev": "44-1c14b14c2dbe82c73d6cf321b547d084",
   "outbound_trunks": 1,
   "inbound_trunks": 5,
   "twoway_trunks": 5,
   "overrides": {
       "inbound_trunks": 3,
       "allow_prepay": true
   },
   "allow_prepay": true,
   "pvt_inbound_trunks": 3,
   "pvt_type": "limits",
...
}
```

PR #1692, opened 2016-03-09
KAZOO-4588: exctract userid properly in cb_apps_util:is_authorized

icehess:


applications/crossbar/src/cb_apps_util.erl
```diff
@@ -63,11 +63,12 @@ is_authorized(AccountId, UserId, AppId) ->
             AppJObj = wh_json:get_value(AppId, kz_apps_store:apps(Doc)),
             Allowed = wh_json:get_value(<<"allowed_users">>, AppJObj, <<"specific">>),
             Users = wh_json:get_value(<<"users">>, AppJObj, []),
-            case {Allowed, Users} of
+            UserIds = [wh_json:get_value(<<"id">>, U, U) || U <- Users],
```
fenollp: Default value is the jobj `U`?

icehess: @fenollp typo, I'll fix it.

applications/crossbar/src/cb_apps_util.erl
```diff
@@ -63,11 +63,12 @@ is_authorized(AccountId, UserId, AppId) ->
...
             Allowed = wh_json:get_value(<<"allowed_users">>, AppJObj, <<"specific">>),
             Users = wh_json:get_value(<<"users">>, AppJObj, []),
-            case {Allowed, Users} of
+            UserIds = [wh_json:get_value(<<"id">>, U, []) || U <- Users],
+            case {Allowed, UserIds} of
                 {<<"all">>, _} -> 'true';
                 {<<"specific">>, []} -> 'false';
-                {<<"specific">>, Users} when is_list(Users) ->
-                    lists:member(UserId, Users);
+                {<<"specific">>, UserIds} when is_list(UserIds) ->
```
lazedo: `when is_list(UserIds)` i believe this is not needed ?

applications/crossbar/src/cb_apps_util.erl
```diff
@@ -63,11 +63,12 @@ is_authorized(AccountId, UserId, AppId) ->
             AppJObj = wh_json:get_value(AppId, kz_apps_store:apps(Doc)),
             Allowed = wh_json:get_value(<<"allowed_users">>, AppJObj, <<"specific">>),
             Users = wh_json:get_value(<<"users">>, AppJObj, []),
-            case {Allowed, Users} of
+            UserIds = [wh_json:get_value(<<"id">>, U, []) || U <- Users],
```
fenollp: Why default to `""`?

Also, shouldn't all occurrences of that default value be removed from the list? Matching on `[]` and `lists:member` look like they could be harmed by `undefined` being where it shouldn't be

PR #1693, opened 2016-03-10
KAZOO-4595: when adding a mobile add its MDN to the number DB

fenollp:


k-anderson:
Dont merge this yet ;)

fenollp:
@fenollp plz port this to master

PR #1704, opened 2016-03-15
Fix _design/numbers

skorobkov:
In knm_number_manager property "pvt_number_state" renamed to "pvt_state".
PS: maybe [core/whistle_apps/priv/couchdb/views/inum.json](https://github.com/2600hz/kazoo/blob/master/core/whistle_apps/priv/couchdb/views/inum.json) need change too?


fenollp:
Closing this in favour of https://github.com/2600hz/kazoo/pull/1718

PR #1709, opened 2016-03-17
KAZOO-4603: handle uncaught exception when decoding JSON

fenollp:
This was missed by https://github.com/2600hz/kazoo/pull/1668
However, I was able to successfully wh_json:decode the JSON string
logged in the ticket.

Need more data!

(needs cherry picking to master)


lazedo:
i really don't like returning new() on errors

fenollp:
@icehess why the ref?

PR #1712, opened 2016-03-17
KAZOO-4613: read a release`s cookie from conf, for services or SUP co…

fenollp:
…nnection

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ ELVIS = $(ROOT)/utils/elvis/elvis
 
 KAZOODIRS = core/Makefile applications/Makefile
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -92,6 +92,10 @@ else
 	@RELX_REPLACE_OS_VARS=true KZname='-name $(REL)' _rel/kazoo/bin/kazoo $(ACT) "$$@"
 endif
 
+read-release-cookie: REL ?= whistle_apps
+read-release-cookie:
+	@RELX_REPLACE_OS_VARS=true KZname='-name $(REL)' _rel/kazoo/bin/kazoo escript lib/whistle_config-*/priv/read-cookie.escript "$$@"
```
fenollp: This is the piece that will go into the Cookie part of SUP

fenollp: @k-anderson

PR #1713, opened 2016-03-17
Request envelope requires "data"

jamesaimonetti:
Had an issue in training where user sent POST -d '{"real":"stuff"}' and was getting schema validation errors saying "real" is required. That is a confusing because the real error is that the user didn't create the request envelope properly. This PR checks that and returns a more appropriate error payload.

### DIFF
--- a/applications/crossbar/src/api_util.erl
+++ b/applications/crossbar/src/api_util.erl
@@ -238,7 +238,7 @@ set_request_data_in_context(Context, Req, JObj, QS) ->
     case is_valid_request_envelope(JObj) of
         'false' ->
             lager:info("failed to find 'data' in envelope, invalid request"),
-            ?MODULE:halt(Req, Context);
+            halt_on_invalid_envelope(Req, Context);
         'true' ->
             Setters = [{fun cb_context:set_req_json/2, JObj}
                       ,{fun cb_context:set_req_data/2, wh_json:get_value(<<"data">>, JObj)}
@@ -275,7 +275,7 @@ halt_on_invalid_envelope(Req, Context) ->
                                                     ,{<<"target">>, <<"data">>}
                                                     ]
                                                    )
-                                                 ,Context
+                                                 ,cb_context:set_resp_error_code(Context, 400)
                                                  )
                 ).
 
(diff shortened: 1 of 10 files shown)


applications/crossbar/src/api_util.erl
```diff
@@ -245,14 +226,25 @@ handle_url_encoded_body(Context, Req, QS, ReqBody, JObj) ->
...
+set_request_data_in_context(Context, Req, JObj, QS) ->
+    case is_valid_request_envelope(JObj) of
+        'false' ->
+            lager:info("failed to find 'data' in envelope, invalid request"),
+            halt_on_invalid_envelope(Req, Context);
+        'true' ->
             Setters = [{fun cb_context:set_req_json/2, JObj}
-                       ,{fun cb_context:set_req_data/2, wh_json:get_value(<<"data">>, JObj, wh_json:new())}
-                       ,{fun cb_context:set_query_string/2, QS}
+                      ,{fun cb_context:set_req_data/2, wh_json:get_value(<<"data">>, JObj)}
```
fenollp: Don't cb_ modules depend on req_data being a JObj? (instead of undefined)

jamesaimonetti: If we get here, we've verified `data` exists in the request data.

lazedo:
@jamesaimonetti can you check if OPTIONS verb does not require `data`. it seems we missed this https://2600hz.atlassian.net/browse/KAZOO-4626

fenollp:
This breaks backwards compatibility of KNM @jamesaimonetti @k-anderson

jamesaimonetti:
@fenollp how so? PUT/POST has always required a "data" object, even an empty one.

jamesaimonetti:
Perhaps we could amend it so that if there is no request body, we continue processing with `wh_json:new()` as req_data?

fenollp:
Wrong. Checkout 3.22 and PUT a number to a `wnm_allow_additions` `true` account.

fenollp:
I believe this would fix it, yes :)

jamesaimonetti:
@fenollp see if #1773 addresses the issue

PR #1714, opened 2016-03-18
patch-9: list siblings only by admin or if it is for child account

icehess:
(#1660)
KAZOO-4603 (#1709)


k-anderson:
@kalda341 Howdy,

We had to revert the patch that you submitted as it was too restrictive.  However, we did make a change so that no user (except the super-admin) can list the siblings for themself.  Additionally, we kept a system configuration pararmeter so if you would like to completely disable it you can.  Unfortuntetly, since we are reverting the default and it was only on one build we changed that parameter name to "allow_sibling_listing".

PR #1715, opened 2016-03-18
Basic audio levels modification support

vladimirralev:
The basic groundwork for audio levels, mute, unmute in konami and standalone commands. If you accept it I will add more docs and channels HTTP API.

### DIFF
--- a/applications/ecallmgr/src/ecallmgr_call_command.erl
+++ b/applications/ecallmgr/src/ecallmgr_call_command.erl
@@ -110,6 +110,13 @@ get_fs_app(_Node, _UUID, JObj, <<"hangup">>) ->
         'true' ->  {<<"unbridge">>, <<>>}
     end;
 
+get_fs_app(_Node, UUID, JObj, <<"audio_level">>) ->
+    Action = wh_json:get_value(<<"Action">>, JObj),
+    Level = wh_json:get_value(<<"Level">>, JObj),
+    Mode = wh_json:get_value(<<"Mode">>, JObj),
+    Data = <<UUID/binary, " ", Action/binary, " ", Mode/binary, " mute ", Level/binary>>,
+    {<<"audio_level">>, Data};
+
 get_fs_app(_Node, UUID, JObj, <<"play_and_collect_digits">>) ->
     case wapi_dialplan:play_and_collect_digits_v(JObj) of
         'false' -> {'error', <<"play_and_collect_digits failed to execute as JObj did not validate">>};
--- a/applications/ecallmgr/src/ecallmgr_util.erl
+++ b/applications/ecallmgr/src/ecallmgr_util.erl
@@ -106,7 +106,7 @@ send_cmd(Node, UUID, "call_pickup", Target) ->
 send_cmd(Node, UUID, "hangup", _) ->
     lager:debug("terminate call on node ~s", [Node]),
     freeswitch:api(Node, 'uuid_kill', wh_util:to_list(UUID));
-send_cmd(Node, UUID, "audio_level", Args) ->
+send_cmd(Node, _UUID, "audio_level", Args) ->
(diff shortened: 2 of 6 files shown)


applications/ecallmgr/src/ecallmgr_call_command.erl
```diff
@@ -110,6 +110,13 @@ get_fs_app(_Node, _UUID, JObj, <<"hangup">>) ->
         'true' ->  {<<"unbridge">>, <<>>}
     end;
 
+get_fs_app(_Node, UUID, JObj, <<"audio_level">>) ->
+    Action = wh_json:get_value(<<"Action">>, JObj),
+    Level = wh_json:get_value(<<"Level">>, JObj),
+    Mode = wh_json:get_value(<<"Mode">>, JObj),
+    Data = <<UUID/binary, " ", Action/binary, " ", Mode/binary, " mute ", Level/binary>>,
```
k-anderson: @vladimirralev 
Howdy,
This is awesome!

However, I don't think " mute " should be in the command.

vladimirralev: Well, I thought this at first too, but after testing this is the only reliable way to mute and unmute. If you use level it switches to another mode which messes up things and using start/stop is more complicated with same result. I have this working in my training VM with FS 1.7 but it also works identically in F 1.4 if you want to test it.

I have this metaflow as example for mute and unmute which is the primary use-case. With more testing other configs could work too but may also need fixes in freeswitch.

      "4": {
        "module": "audio_level",
        "data": {
          "action": "start",
          "mode": "read",
          "level": "-4"
        }
      },
      "5": {
        "module": "audio_level",
        "data": {
          "action": "start",
          "mode": "read",
          "level": "0"
        }
      }
    },

It produces for mute ```uuid_audio(Zj3a1bneFKXwhhNnfwTzRypWNz0MAOLS start read mute -4)``` and ```uuid_audio(Zj3a1bneFKXwhhNnfwTzRypWNz0MAOLS start read mute 0)``` for unmute. Do you have a batter idea to do mute/unmute?

vladimirralev:
Would you guys accept a konami module that sends raw unstructured freeswitch instructions? I could make sure the those are only available for uuid-scoped commends to avoid syswide security issues.

For example

```
  "4": {
    "module": "freeswitch_command",
    "data": {
      "action": "uuid_audio",
      "data": "$UUID start read mute 0"
    }
  },
  "5": {
    "module": "freeswitch_command",
    "data": {
      "action": "uuid_break",
      "data": "$UUID"
    }
  }
```

Same for channels API, possibly with result delivered in the response as well. It will satisfy so many needs at once.

applications/ecallmgr/src/ecallmgr_call_command.erl
jamesaimonetti: I'm going to run this by the FreeSWITCH team to see if they have a better way to accomplish this. I don't fully understand the `uuid_audio` functionality, so can't comment yet if there's a better approach. If they don't have suggestions, we'll merge this in today.

jamesaimonetti:
@vladimirralev regarding the unstructured freeswitch commands, I don't think we're ready to allow it. I think it introduces too broad of a surface for unverified activity. In what ways could this functionality be abused by a third party to bypass billing, for instance?

At this point, unless @k-anderson or @lazedo see differently, I don't think we want to open up unvalidated access to FreeSWITCH. I'd rather continue to write abstracted APIs that go through ecallmgr to build the resulting FreeSWITCH commands.

Happy to hear more arguments for including this if you can address security/fraud concerns.

jamesaimonetti:
Nothing definitive yet. Since it works for you, we can merge it for now and improve it once we get a better idea of uuid_audio's functionality.

PR #1716, opened 2016-03-18
KAZOO-4594: set number of pages to zero on crash

icehess:
and catch crashes properly


applications/fax/src/fax_worker.erl
```diff
@@ -830,7 +829,7 @@ prepare_contents(JobId, RespHeaders, RespContent) ->
                                                  ),
             Cmd = io_lib:format(ConvertCmd, [OutputFile, InputFile]),
             lager:debug("attempting to convert pdf: ~s", [Cmd]),
-            case os:cmd(Cmd) of
+            case catch os:cmd(Cmd) of
```
fenollp: How `try "success" = os:cmd(Cmd) of` instead of `case catch`?
Then the `_Else ->` part goes into the `catch` part

icehess: What is the deference here?

PR #1722, opened 2016-03-23
KAZOO-4575: use check_sync instead of registrar_flush when device doc…

link89:
use check_sync instead of registrar_flush when device doc is updated

### DIFF
--- a/applications/crossbar/src/crossbar_util.erl
+++ b/applications/crossbar/src/crossbar_util.erl
@@ -384,6 +384,7 @@ flush_registration(Username, <<_/binary>> = Realm) ->
                 ,{<<"Username">>, Username}
                 | wh_api:default_headers(?APP_NAME, ?APP_VERSION)
                ],
+    whapps_util:amqp_pool_send(FlushCmd, fun wapi_switch:publish_check_sync/1),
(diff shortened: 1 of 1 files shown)


jamesaimonetti:
@link89 Sorry for the delay! I'm not sure this is an appropriate change as check_sync and registrar_flush are two different functions. I imagine you want the phone to re-register when you change the SIP credentials, which is why you have changed which function is called here? I would be okay with doing both the check_sync and the registrar_flush, but I don't think we'll accept swapping them out.

Would modifying this pull request to call both functions work for you? If not, can you describe why?

link89:
Hi @jamesaimonetti , I create this PR because it is reported that when sip credential of a device is changed, that device would remain registered for a while and still able to recevie phone calls until next register cycle. It turns out that crossbar would flush registrar of that device in this case, which will remove the registrar entry of that device from ecallmgr, but the device is still registered on kamailio. We found solution from KAZOO-3514, in which a sip notify packet is send to the device whose configuration is changed, so that the device would send a register to kamailio and turn offline immediately. So I think it would be ok to call both of the functions.

jamesaimonetti:
@link89 Yup! I agree, calling both functions is the way to go. Can you edit the PR to reflect that?

PR #1725, opened 2016-03-23
KAZOO-4633: return all fields for GET req on a number

icehess:
@fenollp @k-anderson I think it’s better to rename `KNM_DB`(`WNM_DB`) fields, so when using `wh_json:public_fields` it would return necessary fields(for example rename `pvt_state` to `state`)


jamesaimonetti:
@icehess LGTM. You and @fenollp will need to figure out how this change makes its way into kazoo_number_manager.

fenollp:
Fixed by https://github.com/2600hz/kazoo/pull/1762

PR #1726, opened 2016-03-24
KAZOO-4615: disable bypass media when it is loopback

icehess:


applications/trunkstore/src/ts_from_offnet.erl
```diff
@@ -48,13 +48,13 @@ proceed_with_endpoint(State, Endpoint, JObj) ->
     Q = ts_callflow:get_my_queue(State),
     'true' = wapi_dialplan:bridge_endpoint_v(Endpoint),
 
-    MediaHandling = case wh_json:get_value([<<"Custom-Channel-Vars">>, <<"Offnet-Loopback-Number">>], JObj) of
-                        'undefined' ->
+    MediaHandling = case wh_json:is_true([<<"Custom-Channel-Vars">>, <<"Executing-Extension">>], JObj) of
```
jamesaimonetti: Could this become

```erlang
case is_true(ExecutingExt, JObj) orelse is_false(BypassMedia, Endpoint) of
    true -> process
    false -> bypass
end
```

fenollp:
forward port me maybe

PR #1729, opened 2016-03-25
HELP-21736: attempt to log/catch the ets:counter_update issue

k-anderson:


jamesaimonetti:
@k-anderson happy with this? LGTM

PR #1732, opened 2016-03-28
KAZOO-4601: Callflow module to branch on the callee's account/user/device document property value

siplabs:


applications/callflow/doc/cf_branch_property.md
```diff
@@ -0,0 +1,42 @@
...
+
+The `branch_property` callflow enables you to branch on value of some field inside callee's document (device, user or account).
+
+### Mandatory fields
+
+**property** - name of the property we are looking for
+
+### Optional fields
+
+**action** - "branch" - the only action supported now
```
lazedo: i don't see this being used anywhere

applications/callflow/src/module/cf_branch_property.erl
```diff
@@ -0,0 +1,92 @@
+%%%-------------------------------------------------------------------
+%%% @copyright (C) 2013-2016, 2600Hz INC
+%%% @doc
+%%%
+%%% "data":{
+%%%   "action":"branch", % the only one action supported currently
```
lazedo: i don't see this being used anywhere

lazedo:
why the use of `targeted_lookup / automatic_lookup` ? why don't you use the the `cf_endpoint` ?

applications/callflow/src/module/cf_branch_property.erl
```diff
@@ -0,0 +1,92 @@
...
+
+-include("../callflow.hrl").
+
+%% API
+-export([handle/2]).
+
+-spec handle(wh_json:object(), whapps_call:call()) -> 'ok'.
+handle(Data, Call) ->
+    Scope = wh_json:get_binary_value(<<"scope">>, Data),
+    PointedField = wh_json:get_binary_value(<<"property">>, Data),
```
lazedo: you are only supporting simple fields at the root level.

applications/callflow/src/module/cf_branch_property.erl
```diff
@@ -0,0 +1,92 @@
...
+                        'undefined'
+                end,
+    case is_binary(ChildName) of
+        'true' -> cf_exe:continue(ChildName, Call);
+        'false' -> cf_exe:continue(Call)
+    end.
+
+-spec lookup_child_name(api_binary(), whapps_call:call(), ne_binary()) -> api_binary().
+lookup_child_name(Scope, Call, Field) ->
+    case Scope of
```
jamesaimonetti: This should be pattern-matched in the function clauses instead.

applications/callflow/src/module/cf_branch_property.erl
```diff
@@ -0,0 +1,92 @@
...
+-include("../callflow.hrl").
+
+%% API
+-export([handle/2]).
+
+-spec handle(wh_json:object(), whapps_call:call()) -> 'ok'.
+handle(Data, Call) ->
+    Scope = wh_json:get_binary_value(<<"scope">>, Data),
+    PointedField = wh_json:get_binary_value(<<"property">>, Data),
+    ChildName = case is_binary(PointedField) of
```
jamesaimonetti: I would prefer you call a new function with `PointedField` and pattern match, versus all of these `is_*` case clauses.

```erlang
ChildName = find_child_name(PointedField),
...


find_child_name(Data, Call, 'undefined') -> 'undefined';
find_child_name(Data, Call, PointedField) ->
    Scope = wh_json:get_binary_value(<<"scope">>, Data),
    lookup_child_name(Scope, Call, PointedField).
```

Then call another well-named function to pattern match the `ChildName` returned.

applications/callflow/src/module/cf_branch_property.erl
```diff
@@ -0,0 +1,92 @@
...
+        _Else ->
+            lager:debug("can not lookup endpoint for caller"),
+            'undefined'
+    end.
+
+-spec targeted_lookup(ne_binary(), whapps_call:call(), ne_binary()) -> api_binary().
+targeted_lookup(Scope, Call, Field) ->
+    AuthId = whapps_call:authorizing_id(Call),
+    DocId = case {Scope, whapps_call:authorizing_type(Call)} of
+                      {<<"account">>, _} -> whapps_call:account_id(Call);
```
jamesaimonetti: Why not pattern-match `Scope = <<"account">>` in the function clause and return `whapps_call:account_id(Call)` directly?

applications/callflow/src/module/cf_branch_property.erl
```diff
@@ -0,0 +1,92 @@
...
+                      {Same, Same} -> AuthId;
```
jamesaimonetti: If you pattern match as above, this could be a second function pattern match:

```erlang
targeted_lookup(Scope, Call, Field) ->
    ChildDocId = targeted_lookup(Scope, Call, Field, whapps_call:authorizing_id(Call)),
    lookup_doc(Scope, Call, Field, ChildDocId).

targeted_lookup(Scope, Call, Field, Scope) -> whapps_call:authorizing_id(Call);
targeted_lookup(_Scope, Call, _Field, <<"device">>) -> lookup_device_owner(Call, whapps_call:authorizing_id(Call));
targeted_lookup(_Scope, _Call, _Field, <<"user">>) -> 'undefined';
targeted_lookup(_Scope, _Call, _Field, _AuthType) ->
    lager:debug("unsupported authorizing type: ~s", [_AuthType]),
    'undefined'.

lookup_doc(_Scope, _Call, _Field, 'undefined') ->
    lager:debug("cannot find document for current scope: ~s", [_Scope]),
    'undefined';
lookup_doc(_Scope, Call, Field, DocId) ->
    lookup_childname(Call, Field, DocId).
```

Also, I prefer it when function arguments aren't re-arranged for no reason. `lookup_childname/3` in your code should follow the order of arguments from the calling function; I think it reads better and the compiler may be able to optimize the callstack a little(?) but mostly I find it more consistent.

applications/callflow/src/module/cf_branch_property.erl
siplabs: Yes, it just marks that it's reserved for future use.
Future possibilities - branch_and_update or stuff like that.

lazedo: future use of what ? if you're not using it in the code just remove it. please elaborate on `branch_and_update or stuff like that`

applications/callflow/src/module/cf_branch_property.erl
siplabs: For now - yes. This could be expanded in future versions (we need to invent some syntax for that)

lazedo: @siplabs you can add something like `binary:split( wh_json:get_binary_value(<<"property">>, Data), <<".">>)`

applications/callflow/doc/cf_branch_property.md
amgunn: @lazedo, the customer is stating there's some more detail in the ticket comments regarding branch_property: https://2600hz.atlassian.net/browse/KAZOO-4601 (look at extreme rotary's comments)

lazedo: @amgunn there's no specific questions, we're looking at the implementation of the ticket. i still don't get why not use the merged properties from `cf_endpoint` (what's the use case not addressed by `cf_endpoint`). as to this particular comment, there's nowhere in the code a reference to this property, so it should be removed.

k-anderson:
@siplabs I am late to the game on this one so please forgive me but I would really like to entertain the idea of combining this functionality into cf_branch_variable.  To me they are conceptionally the same.  They look at a parameter (from the db or call) and execute a branch matching the value.

I feel that having lots of callflow modules that have minor differences is confusing for developers and the community as a whole.

I propose that you add all this functionality to the existing cf_branch_variable, which would make it the sole callflow module responsible for branching based on the value of a variable.  Be it sourced from the CCVs, user document, device document, or the merge.

I propose that to make this change you simply extend "scope" to include (and default to if not present) "custom_channel_vars".

This would make the existing functionality supported as:

```
   "data":{  
      "variable":"Call-Forwarding"
   },
   "module": "branch_variable",
   "children":{  
      "true": {},
      "false": {}
   }
```
or
```
   "data":{
      "scope":"custom_channel_vars",
      "variable":"Call-Forwarding"
   },
   "module": "branch_variable",
   "children":{  
      "true": {},
      "false": {}
   }
```

and the new functionality would be achieved as you it was written but changing "property" to "variable" to maintain compatibility:

```
   "data":{  
      "scope":"user",
      "variable":"user_record_outbound"
   },
   "module": "branch_variable",
   "children":{  
      "true":{},
      "false":{},
      "_":{}
   }
```

I am open to discussion on this and can be available for a phone call if you prefer.  I am also willing to leave it as is if you have strong feelings around it but I am really trying to make sure that our callflow modules don't become confusing about which do what...

applications/callflow/doc/cf_branch_property.md
siplabs: ok, it will be removed. But sometimes people leave stuff to be relevant with the idea and support consistency with future versions.

applications/callflow/src/module/cf_branch_property.erl
siplabs: Will remove

siplabs:
@k-anderson Well, I'd argue that... And lots, lots more of callflow modules will appear as time goes... Merging this one with cf_branch_variable will break compatibility... but I understand your intention...
 
My feeling is having many different callflow modules is not bad if we teach our UI tools to replace elements inside the tree, not just remove/re-add entire subtrees. I think different modules here are better because users never forget to set the scope correctly, etc.... 

Let's exhange some emails about that.

k-anderson:
@siplabs Howdy,
I hope that lots and lots of callflow actions are created! :)  My concern ending up with 6 callflow modules that all do the same thing but slightly differently because of a developers specific use-case.  I am all about alternatives but if the module is too specific to a use-case then I feel it makes it confusing.  Again, I do NOT strongly believe this is the case here.

If we made the change we would be able to tell a developer that if you want to branch a callflow based on a parameters value you use cf_branch_variable and set the scope to control where the parameter is found.  As submitted the explanation would be, if you want to branch on CCV parameters use cf_branch_variable but if you want to branch based on database parameters use cf_branch_property (again, I am NOT opposed to this I am just proposing an alternative to have this discussion).

I don't believe that merging would break functionality with cf_branch_variable because the change I am proposing would be to use the scope variable to conditionally change the behavior.  The existing behavior would exist if the scope parameter was set to "custom_channel_vars" and if the scope parameter is missing (as it would be today since cf_branch_variable did not have a scope parameter) it should just default to this value.  I do understand that if cf_branch_property was already being used it would clearly not be compatible because the callflows would already exist.

If you would like to continue this via email then we can but I would encourage us to keep the conversation public in the event that a community member would like to weigh in.  I will also bring your comments up with the team today and see if they are swayed :)

k-anderson:
@siplabs We discussed this further and the team does feel like this would be best as one module.  The justification that we used in the discussion was choosing between the modules was basically just picking the scope of the variable to use.  Except that one module then had additional scope options.  As such we could collapse the scope selection into one parameter.

I look forward to your thoughts on these points!

k-anderson:
@siplabs Howdy, any thought regarding the above comments?

k-anderson:
@siplabs Howdy,

We are trying to prepare for 4.0 so I really need to know if you guys agree with the above and will create the one module or if you feel strong enough to request that we merge "as-is".  If you would like I could create the module as described to clarify... let us know :)

k-anderson:
@siplabs Howdy, any updates?

icehess:
closing this for issue a new PR.

PR #1735, opened 2016-03-29
KAZOO-4595_master: cherry pick PR 1693 from 3.22 to master

fenollp:
"Cherry pick" of #1693 

**Potential problem**: `reconcile` will actually create/move number to set its state as IN_SERVICE instead of  using `create` with this state parameter (because that is not how `create` is able to be used as).
However, during `reconcile` there will be requests made to phonebook and I am unsure that's the intent of reconcile. Maybe @jamesaimonetti or @k-anderson can tune in?


fenollp:
@lazedo please merge

PR #1736, opened 2016-03-30
KAZOO-4571: additional props for user to handle rare channel creates

k-anderson:


jamesaimonetti:
@k-anderson forward port plz

PR #1737, opened 2016-03-30
CB-METRICS: add basic metrics per request

jamesaimonetti:
Tracks memory usage and binary allocations for each request, in addition to time spent.

### DIFF
--- a/applications/crossbar/src/api_resource.erl
+++ b/applications/crossbar/src/api_resource.erl
@@ -94,6 +94,7 @@ rest_init(Req0, Opts) ->
               ,{fun cb_context:set_profile_id/2, ProfileId}
               ,{fun cb_context:set_api_version/2, Version}
               ,{fun cb_context:set_magic_pathed/2, props:is_defined('magic_path', Opts)}
+              ,{fun cb_context:store/3, 'metrics', metrics()}
               ],
 
     Context0 = cb_context:setters(cb_context:new(), Setters),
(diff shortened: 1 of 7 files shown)


core/whistle/Makefile
```diff
@@ -17,6 +17,9 @@ all: compile
 
 include $(ROOT)/make/kz.mk
 
+proper: ERLC_OPTS += -DPROPER
```
fenollp: You know that if you add this to kz.mk it will be added everywhere?

applications/crossbar/src/api_resource.erl
```diff
@@ -111,6 +112,10 @@ rest_init(Req0, Opts) ->
             {'ok', cowboy_req:set_resp_header(<<"x-request-id">>, ReqId, Req9), Context3}
     end.
 
+-spec metrics() -> {non_neg_integer(), non_neg_integer()}.
+metrics() ->
+    {wh_util:bin_usage(), wh_util:mem_usage()}.
```
fenollp: Do we want to activate this for every request? are these calls without impact on perf?

make/kz.mk
```diff
@@ -12,7 +12,7 @@ ERLC_OPTS += -Werror +warn_export_all +warn_unused_import +warn_unused_vars
 ELIBS = $(ERL_LIBS):$(ROOT)/deps:$(ROOT)/core
 EBINS += $(ROOT)/core/whistle/ebin \
 	     $(ROOT)/deps/lager/ebin
-TEST_EBINS += $(EBINS)
+TEST_EBINS += $(EBINS) $(ROOT)/deps/proper/ebin
```
fenollp: If you're doing this then update the travis file accordingly please

core/whistle/Makefile
jamesaimonetti: Then why didn't you do it in the first place? :)

lazedo:
@fenollp  update the travis file accordingly please

PR #1740, opened 2016-04-01
del-auth-tok: fix documentation of DELETE /v1/token_auth

fenollp:


fenollp:
@jamesaimonetti I see that all the code braces with curl calls inside use "\`\`\`curl … ```" blocks. They should read `shell` instead of `curl` though. Also JSON I/Os should maybe be separated and colorized as `json` blocks too.

Was the intent to use GitHub's coloring code blocks?

jamesaimonetti:
@fenollp The intent initially was to be rendered by Slate. And you could tag those code blocks to have the multi-tab right column.

With moving to mkdocs as a target instead, this is probably less relevant and we can do something that makes both mkdocs and github rendering happy.

applications/crossbar/doc/token_auth.md
```diff
@@ -65,4 +76,14 @@ URL segment: `/token_auth`
 
 If you'd like to invalidate an authentication token programmatically (versus letting the system expire the token), you can issue a `DELETE`:
 
-    curl -v -X DELETE -H "X-Auth-Token: {AUTH_TOKEN}" http://server.com:8000/v1/token_auth
+```shell
+curl -v -X DELETE -H "X-Auth-Token: {AUTH_TOKEN}" http://server.com:8000/v1/token_auth -d '{"data": {}}'
+```
+
+```json
```
jamesaimonetti: I'm on the fence about this separation. I guess it shouldn't matter anymore since Slate is out so...let's see how it goes.

fenollp: Want me to do a `sed` PR?

PR #1742, opened 2016-04-04
KAZOO-4609: force ignore early media if require key confirm

icehess:


applications/stepswitch/src/stepswitch_bridge.erl
```diff
@@ -332,6 +332,11 @@ build_bridge(#state{endpoints=Endpoints
 
     FmtEndpoints = stepswitch_util:format_endpoints(Endpoints, Name, Number, OffnetReq),
 
+    IgnoreEarlyMedia = case wh_json:is_true(<<"Require-Ignore-Early-Media">>, CCVs, 'false') of
```
fenollp: How about `andalso`?

PR #1747, opened 2016-04-07
Update cb_phone_numbers to work with KNM

skorobkov:
User with pvt_wnm_allow_additions can add numbers and activate them.
Records `#knm_number{}` & `#knm_phone_number{}` was moved in separte HRL file, for use it in `include_lib` if you need this records.
In this PR `#knm_number` required in `set_response` functions. Now its can handle `knm_number_return()`.


skorobkov:
TODO thing: `knm_number:buy/3` must be updated.
Any advise how to do it in proper way?

skorobkov:
Need advise on `cb_phone_numbers_v2:summary/1`.
It use `crossbar_doc:load(?KNM_PHONE_NUMBERS_DOC, Context)`, it must be changed to `crossbar_doc:load_view`?  But `crossbar_doc:load_view` have issues when used with filters in QS.
If defined QS filters, its [impliccit add](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/crossbar_doc.erl#L355) option `include_docs` and apply filter [to original docments](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/crossbar_doc.erl#L951) (in <<"doc">> object), not view. And `filter_state=in_service` (which use Monster-UI) didn't work, must be used `filter_pvt_state=in_service`.

Waht i have to do? Use

applications/crossbar/src/modules_v1/cb_phone_numbers_v1.erl
```diff
@@ -461,6 +462,8 @@ validate_delete(Context) ->
                           cb_context:context().
 set_response({'ok', {'ok', Doc}}, _, Context) ->
     crossbar_util:response(Doc, Context);
+set_response({'ok', #knm_number{}=Number}, _, Context) ->
```
fenollp: Instead of matching here and making the records public, can you just create `knm_number:is_number/1` + do a `case`?

The rationale is to keep record inside its module in a kind of type-handling manner.

skorobkov: At first, I created this function, but then changed my mind and decided to use record.
Ok. I will redo changes  and add function.

skorobkov:
"Mess" with the knm_carriers & knm_providers:
- function knm_phone_number:delete/1 receive `#knm_phone_number{}` on input, but knm_providers:delete/1 want `#knm_number{}`
```Erlang
delete(#knm_phone_number{dry_run='false'}=Number) ->
    Routines = [fun knm_providers:delete/1
                ,fun delete_number_doc/1
                ,fun maybe_remove_number_from_account/1
               ],
    {'ok', NewPhoneNumber} = setters(Number, Routines),
    NewPhoneNumber.
```
- function knm_number:buy/1 have the similar issue, knm_carriers:acquire/1 want `#knm_number{}`, not `#knm_phone_number{}`
```Erlang
buy(Num, Account, Options) ->
    Updates = [{fun knm_phone_number:set_assigned_to/2, wh_util:format_account_id(Account, 'raw')}
               ,{fun knm_phone_number:set_state/2, ?NUMBER_STATE_IN_SERVICE}
               ,fun knm_carriers:acquire/1
              ],
    update(Num, Updates, Options).
```

How this can be fixed? Modify knm_carriers & knm_providers (and check all modules, which use them)?

fenollp:
Thanks for the patch @skorobkov! It's amazing.
Looking into the API disparity…

fenollp:
For buy see #1751
For delete I think it is safe enough (although dirty) to put the PhoneNumber into an empty Number shell and pass it to delete.

I'm very open to your thoughts & ideas

sergey-safarov:
A has tested master 9a3333d20addd50a37ea69259ee82b95d8028baa and found that numbers cannot be delete. Details in KAZOO-4630

skorobkov:
@sergey-safarov Yes, I didn't found any call to "update" function on delete. Its did a lot of things, but in the end it didn save it anywhere...

fenollp:
I'm working on GET /phone_numbers right now: making it work & be backwards-compatible. Will get to this at some point

PR #1753, opened 2016-04-07
Fix similar to #1747

skorobkov:
Fix similar to #1747
I beleive here `assigned_to` must be changed to `assign_to`.
Without it request can't be authorized.


skorobkov:
And maybe similar issues:
https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/modules/cb_callflows.erl#L387
https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/modules/cb_jobs_listener.erl#L164
https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/modules_v2/cb_devices_v2.erl#L819

???

fenollp:
@skorobkov closing this in favour of https://github.com/2600hz/kazoo/pull/1752

PR #1758, opened 2016-04-08
KAZOO-4657: braintree_request/kz_http binary to string vars change

onnet:
Attempt #2

http:maybe_basic_auth/2 still waits for binaries: 

AuthString = "Basic " ++ base64:encode_to_string(<<Username/binary, ":", Password/binary>>),


fenollp:
Thanks! That's due to new defaults by `kz_http`. Will edit the other places this happens.

PR #1762, opened 2016-04-08
knm-GET-phone_numbers: replace old way using a view

fenollp:
Making `master` work so that it returns [this](https://github.com/2600hz/kazoo/blob/7eabc5ecc9130353f9d2b731add92527e5dfa3a9/applications/crossbar/doc/phone_numbers.md#response-4), as was possible in `3.22`.

Filters [don't work yet somehow](https://github.com/2600hz/kazoo/pull/1747#issuecomment-206884613) please @jamesaimonetti @k-anderson @lazedo help me?

Some code will introduce weird things in the number doc, notably cb_port_requests:1448 and knm_phone_number:254-255 to name a few. (needs work)

mergeable


applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
```diff
@@ -940,14 +867,23 @@ identify(Context, Number) ->
...
-    Result = case knm_number:get(Number, Options) of
-                 {'ok', KNum} -> {'ok', knm_number:to_public_json(KNum)};
-                 {'error', _R}=Error -> Error
-             end,
-    set_response(Result, Number, Context).
+    ViewOptions = [ {'keys', [knm_converters:normalize(Number)]}
+                  ],
+    case
+        cb_context:resp_data(
+          crossbar_doc:load_view(?CB_LIST, ViewOptions, Context, fun normalize_view_results/2)
```
fenollp: @jamesaimonetti this won't `reconcile` number (as `knm_number:get/2` would), is this ok you think?

PR #1763, opened 2016-04-08
DOC-51: IPs API docs

jamesaimonetti:


jamesaimonetti:
@lazedo @fenollp I can't figure out how to get `kz_mime.erl` to compile. I tried adding it explicitly to the [`SOURCES = ](https://github.com/2600hz/kazoo/blob/master/core/whistle/Makefile#L4)` in `core/whistle/Makefile` but that still isn't building the module.

Halp?

jamesaimonetti:
Let's see if this works...

fenollp:
It should work.
The thing is kz_mime.erl doesn't exist yet when `$?` is replaced. So the stuff in COMPILE_MOAR should actually be compiled first.
Try `all: compile | src/kz_mime.erl` (`|` is to ask for sequentiality) and just `all: src/kz_mime.erl compile` if it doesnt do it (but `make -j kazoo` won't work).
COMPILE_MOAR can be removed.

PR #1764, opened 2016-04-09
HELP-21582 use System-Log-ID instead of Call-ID

lazedo:
when call-id is an optional header for a wapi message, we are still adding it as part of the payload because of logging in `gen_listener:handle_event/4`.

### DIFF
--- a/core/whistle/src/wh_util.erl
+++ b/core/whistle/src/wh_util.erl
@@ -135,6 +135,7 @@
 -include_lib("whistle/include/wh_types.hrl").
 -include_lib("whistle/include/wh_log.hrl").
 -include_lib("whistle/include/wh_databases.hrl").
+-include_lib("whistle/include/wh_api.hrl").
 
 -define(WHISTLE_VERSION_CACHE_KEY, {?MODULE, 'whistle_version'}).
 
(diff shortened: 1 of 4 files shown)


core/whistle/src/wh_util.erl
```diff
@@ -632,9 +633,9 @@ put_callid(JObj) -> erlang:put('callid', callid(JObj)).
 get_callid() -> erlang:get('callid').
 
 callid(Prop) when is_list(Prop) ->
-    props:get_first_defined([<<"Call-ID">>, <<"Msg-ID">>], Prop, ?LOG_SYSTEM_ID);
+    props:get_first_defined([?KEY_LOG_ID, <<"Call-ID">>, ?KEY_MSG_ID], Prop, ?LOG_SYSTEM_ID);
```
jamesaimonetti: Call-ID should come first, always. I'd do Call-ID, Msg-ID, then System-Log-ID

lazedo: @jamesaimonetti that defeats the purpose as you will always get Msg-ID in `jonny5` sync. `System-Log-ID` shyould come first because it is only used in `wh_amqp_worker` and fetched from the payload by  `gen_listener`

PR #1769, opened 2016-04-10
HELP-22049 handle fax_settings

lazedo:


core/kazoo_documents/src/kz_account.erl
```diff
@@ -167,16 +168,30 @@ set_language(JObj, Language) ->
...
+-spec timezone(ne_binary() | doc(), Default) -> ne_binary() | Default.
+timezone(AccountId)
+  when is_binary(AccountId) ->
+    {'ok', JObj} = fetch(AccountId),
+    timezone(JObj);
 timezone(JObj) ->
     timezone(JObj, 'undefined').
+timezone(AccountId, Default)
+  when is_binary(AccountId) ->
+    {'ok', JObj} = fetch(AccountId),
```
fenollp: Can we not introduce this sort of thing? until now kz document modules mostly work on `doc()`s and I think we should keep it that way. Hiding DB calls under accessor modules functions is bad

lazedo: @fenollp i tend to agree but in cases where the value may actually be a merge down from hierarchy it hides the complexity of inheritance from whapps.

PR #1772, opened 2016-04-11
Monster-UI requre that field 'used_by' must be in result

skorobkov:
Fix after #1762 
If no 'used_by' field in result, number wont show in list of spare numbers.
For example - https://github.com/2600hz/monster-ui-callflows/blob/master/app.js#L636

### DIFF
--- a/applications/crossbar/priv/couchdb/account/phone_numbers.json
+++ b/applications/crossbar/priv/couchdb/account/phone_numbers.json
@@ -3,7 +3,7 @@
     "language": "javascript",
     "views": {
         "crossbar_listing": {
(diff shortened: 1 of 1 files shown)


lazedo:
@skorobkov i think this should be fixed in `monster-ui` not here

fenollp:
I agree with @lazedo. Moreover I would remove the `features` field from results if its value is `{}`. Thoughts?

fenollp:
However, `3.22` returns `used_by` even for `""`. Afraid of breaking things here.

jamesaimonetti:
I agree that this should be fixed in Monster. Crossbar should not be tied to Monster and I don't like any coupling there.

fenollp:
@JRMaitre can you take a look at this in Monster?

fenollp:
Opened ticket UI-2129

JRMaitre:
Thanks, will fix it it Monster!

PR #1773, opened 2016-04-11
ignore_data_on_empty: ignore if req body is empty

jamesaimonetti:


jamesaimonetti:
@fenollp see if this floats your boat

fenollp:
I rebased it and fixed a spec (ah!!). It allows one to PUT to phone_numbers/DID without data.

2 buts:

1. in one sense you removed the check you introduced (as I understand it, this will now fail iff the data field in not an object, which is not all that useful)
1. what about the crossbar modules you edited to remove the is_data_empty checks?

also... you broke it you fix it, right? :)

jamesaimonetti:
1. The check should only be bypassed if there is no request body at all. What I originally was fixing was when the client submitted a request body with the data at the top-level and not under the "data" key.
2. I'm not sure which modules you are talking about here?

fenollp:
@jamesaimonetti was thinking [about these](https://github.com/2600hz/kazoo/pull/1713/files)

Makes sense. Merging it in

PR #1774, opened 2016-04-11
NM REST

fenollp:
Merge this before it gets bigger

@jamesamonetti 

1. Should I also list user fields on listing all of an account numbers? There is a view made for this. 
1. It seems users are unable to update features themselves. Is this normal? There is a schema for them.


applications/crossbar/doc/phone_numbers.md
```diff
@@ -310,6 +319,85 @@ Possible reasons for failure:
 }
 ```
 
+#### Add a number to the database
```
jamesaimonetti: This isn't following the ref doc's formatting.

fenollp: What is this formatting?

jamesaimonetti: https://github.com/2600hz/kazoo/blob/knm-REST/applications/crossbar/doc/ref/phone_numbers.md#create

Also, can you update any uses of `_` in `allowed_methods/X` to be `_PhoneNumber` or `_WhateverThisIs` so the ref doc generates more meaningful URLs?

PR #1778, opened 2016-04-12
KNM-REST: DELETE

fenollp:


applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
```diff
@@ -433,10 +429,19 @@ delete(Context, ?COLLECTION) ->
...
     Options = [{'auth_by', cb_context:auth_account_id(Context)}
-               ,{'dry_run', not cb_context:accepting_charges(Context)}
               ],
-    Result = knm_number:delete(Number, Options),
-    set_response(Result, Context).
+    case knm_number:release(Number, Options) of
+        {'error', Data}=Error ->
+            case wh_json:is_json_object(Data) andalso
+                knm_errors:error(Data) == <<"invalid_state_transition">> andalso
+                knm_errors:cause(Data) == <<"from available to released">>
```
k-anderson: @fenollp  Not a fan of using the arbitrary error cause as a conditional.  Is there no other way to make this determination?

fenollp: Not really: the KNM lib returns the right thing. It's just that `3.22` works that way.

If you merge this one I have another one coming plz

PR #1786, opened 2016-04-13
KAZOO-4670: answer the call and switch media off callflow module

onnet:


jamesaimonetti:
As @lazedo mentioned on the ML, this really needs to go through ecallmgr to the FreeSWITCH node. I know its convenient as written but it completely breaks the abstraction. In fact, I am not sure we need the switch nodename in the whapps_call record anymore. I believe it was used in the case of transfers?

So, perhaps add a command to wapi_dialplan. Since you're basically putting the leg into bypass media mode with the `uuid_media off {CALL_ID}`, you could name it something like `update_rtp_stream` with a `bypass: true | false` field. Or some better name.

Is that reasonable?

jamesaimonetti:
Also, [`uuid_media`](https://wiki.freeswitch.org/wiki/Mod_commands#uuid_media) is for toggling bypass media on/off, so its strange to use it in this way. I wonder if @vladimirralev's [`audio_levels`](https://github.com/2600hz/kazoo/blob/master/core/whistle/src/api/wapi_dialplan.hrl#L411) command could be used instead to mute the leg?

onnet:
@jamesaimonetti @lazedo this is not the first time you guys remind me it is actually a rule of thumb to talk over amqp messages between apps. Since I am using an easy cluster structure with all the apps running together I don't feel the pain and don't care about these caveats. So, I need to beg your pardon again. Sorry. :)

And yes, since mute command exists now, I certainly will try to use it. Thanks @vladimirralev!

lazedo:
@onnet Thanks! (again)

PR #1788, opened 2016-04-13
knm-REST-POST: fix POST phone_number/DID

fenollp:
@icehess What is that `_read_only` field in `3.22` for? @jamesaimonetti ?

### DIFF
--- a/applications/crossbar/doc/phone_numbers.md
+++ b/applications/crossbar/doc/phone_numbers.md
@@ -131,13 +131,20 @@ curl -v -X DELETE \
 {
     "auth_token": "{AUTH_TOKEN}",
     "data": {
-        "assigned_to": "{ACCOUNT_ID}",
-        "created": 63627691442,
-        "id": "{PHONENUMBER}",
-        "state": "available",
-        "updated": 63627691442
+        "_read_only": {
+            "_id": "{PHONENUMBER}",
+            "pvt_assigned_to": "{ACCOUNT_ID}",
+            "pvt_created": 63627848588,
+            "pvt_db_name": "numbers%2F%2B1424",
+            "pvt_modified": 63627848588,
+            "pvt_module_name": "knm_local",
+            "pvt_previously_assigned_to": "{ACCOUNT_ID}",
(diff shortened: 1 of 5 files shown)


applications/crossbar/doc/phone_numbers.md
```diff
@@ -220,12 +220,30 @@ Possible reasons for failure:
...
+```shell
 curl -v -X POST \
     -H "X-Auth-Token: {AUTH_TOKEN}" \
+    -d '{"data": {"my_field": "some other value"}}' \
     http://{SERVER}:8000/v2/accounts/{ACCOUNT_ID}/phone_numbers/{PHONENUMBER}
 ```
 
+```json
+{
+    "auth_token": "694c7e1938aa213955fe600442c433f1",
```
jamesaimonetti: use the placeholders `{AUTH_TOKEN}`, `{ACCOUNT_ID}`, etc. Sadly, there are people that will complain if their response doesn't match this one unless we use the placeholders.

jamesaimonetti:
@fenollp I believe the `_read_only` object was meant to be where `pvt_*` fields were "leaked" to, instead of setting them on the document response. So if `pvt_state` is leaked, it would be under `{"_read_only":{"state":"foo"},...}` instead of `{"state":"foo",...}`.

What happens is people get the response object, modify it, and save it and the `"state":"foo"` was getting saved too. Granted, it would be overwritten on the next fetch by whatever `pvt_state` was, but it seemed silly to store a value for no purpose.

The other issue was that now "state" couldn't be used to store your own custom value, since it would be overwritten by pvt_state. But "state" isn't in the schema and as such, should be a valid key for the client to use for their own purposes.

The `_read_only` object then became the place to put these leaked pvt fields. They are automatically stripped from the request data if the client includes it so we aren't storing it unnecessarily.

Make sense?

jamesaimonetti:
I should probably add a doc about that, or at least update the basics.md with that info.

fenollp:
Fixed the docs issue. Will update the _read_only stuff in another PR.

PR #1790, opened 2016-04-13
KAZOO-4671: Hotfix to allow bind to IPv6 socket

sergey-safarov:


fenollp:
Does this force IPv6 or does it fallback to IPv4 or does it do both?

sergey-safarov:
Pierre this pull request allo crossbar to accept connection via IPv6 ("::") and IPv4 ("0.0.0.0") socket. Used dual binding.
It tested on CentOS7 host. But this may generate error for distribs that not support IPv6.

But I think at present time all distribs support IPv6

jamesaimonetti:
@sergey-safarov I'll test this today using Debian Jessie. I think I have a CentOS 6 install somewhere too.

jamesaimonetti:
Works for me on Debian Jessie:

`curl 192.168.1.5:8000` and `curl [2601:1c0:5803:3660:1a03:73ff:fef8:3c0f]:8000` worked fine.

PR #1793, opened 2016-04-14
ci-couch: Travis plays with bunny, now on the sofa

fenollp:


jamesaimonetti:
@fenollp Did you see the warning about calling sudo (~line 145)?
Missing beam file at ~1372?
ci_chunk_tests are quite verbose. Can those be trimmed?

fenollp:
Ah didn't notice that sudo call. It's a false positive though. Travis-CI should know better…

```
===> Missing beam file nksip_pbx <<"/home/travis/build/2600hz/kazoo/deps/nksip/samples/nksip_tutorial/ebin/nksip_pbx.beam">>
```
yeah well we don't care about sample code do we?

Tests are tests. You will want as verbose an output as possible when they fail.

lazedo:
weird, we removed `nksip`.

fenollp:
@lazedo latest commit fixed that issue

PR #1796, opened 2016-04-14
handle undefined

lazedo:


jamesaimonetti:
@lazedo why would this be undefined? Can we fix the call sites to do this check and only pretty-print when the number exists and isn't undefined?

lazedo:
@jamesaimonetti teletype templates are blindly calling pretty_print

lazedo:
@jamesaimonetti i would also prefer to do it in the call sites but this prevents one those silent crashes.

PR #1798, opened 2016-04-14
HELP-22165: added park document and consitant keys for presence_type

k-anderson:


applications/callflow/src/module/cf_park.erl
```diff
@@ -643,10 +643,14 @@ unanswered_action(SlotNumber, Slot, Data, Call) ->
 
 -spec presence_type(ne_binary(), wh_json:object(), whapps_call:call()) -> ne_binary().
 presence_type(SlotNumber, Data, Call) ->
-    case wh_json:get_value(?PRESENCE_TYPE_KEY, slot_configuration(Data, SlotNumber)) of
-        'undefined' -> ?ACCOUNT_PARKED_TYPE(whapps_call:account_id(Call));
-        Type -> Type
-    end.
+    JObj = slot_configuration(Data, SlotNumber),
+    DefaultPresenceType = 
+        wh_json:get_integer_value(
```
lazedo: @k-anderson should be `wh_json:get_ne_binary`

applications/callflow/doc/park.md
```diff
@@ -0,0 +1,84 @@
...
+Parking will place a call in a numbered "slot" where it will remain until it is retrieved or the caller hangs up.
+
+### Schema for callflow action
+
+#### Mandatory fields
+**action** - action (see below)
+
+#### Optional fields
+
+* **{ACTION}**: default is `park`
```
jamesaimonetti: This should all be in a table that looks like [this](https://github.com/2600hz/kazoo/blob/master/doc/eng/things.md#schema).

k-anderson: @jamesaimonetti its not an api.... and this is the format you gave me earlier...

jamesaimonetti: Its still options. We do the same table for API schema and query string parameters. See [device quickcall](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/doc/devices.md#execute-a-quick-call).

PR #1802, opened 2016-04-15
Knm leak fix

fenollp:
Fixes for the docs are coming. Please merge other PRs before! @lazedo

### DIFF
--- a/core/kazoo_number_manager/src/knm_phone_number.erl
+++ b/core/kazoo_number_manager/src/knm_phone_number.erl
@@ -259,7 +259,7 @@ to_public_json(Number) ->
         wh_json:from_list(
           props:filter_empty(
             [ {<<"created">>, wh_doc:created(JObj)}
-            , {<<"updated">>, wh_doc:modified(JObj)}
+            , {<<"modified">>, wh_doc:modified(JObj)}
             , State
             , UsedBy
             , {<<"module">>, module_name(Number)}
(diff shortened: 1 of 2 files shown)


core/kazoo_number_manager/src/knm_phone_number.erl
```diff
@@ -253,14 +253,27 @@ authorized_release(PhoneNumber) ->
...
-                )
-              ),
-    wh_json:set_value(<<"_read_only">>, ToLeak, wh_json:public_fields(JObj)).
+    State = {<<"state">>, state(Number)},
+    UsedBy = {<<"used_by">>, used_by(Number)},
+    ReadOnly =
+        wh_json:from_list(
+          props:filter_empty(
+            [ {<<"created">>, wh_doc:created(JObj)}
+            , {<<"updated">>, wh_doc:modified(JObj)}
```
k-anderson: @fenollp where is modified leaked as updated?  In the PR for 3.22 I accepted it but changed this to modified.  Specially because this is a new property... we have not had _read_only on the result before so why do you state it used to be updated?

fenollp: ah right

fenollp:
the updated <> modified thing was that I noticed there was this swap on `GET /v2/accounts/ACCOUNT/phone_numbers` and my mind said "the swap also happens on `/DID` requests!", but it doesn't

PR #1808, opened 2016-04-16
Knm buy

fenollp:
3 possible issues

1. `is_authorized` is modified so that it returns true when number was never assigned. This should be fine though, as the only way to add numbers unassigned seem to be by searching for numbers.
1. `buy` does not return charges (thus the crossbar module needed patching). Charging may not be happening.
1. carrier name & errors are shown to the user. This may break some abstraction.

### DIFF
--- a/applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
+++ b/applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
@@ -378,16 +378,22 @@ put(Context, Number) ->
 
 put(Context, ?COLLECTION, ?ACTIVATE) ->
     Results = collection_process(Context, ?ACTIVATE),
-    CB = fun() -> ?MODULE:put(cb_context:set_accepting_charges(Context), ?COLLECTION, ?ACTIVATE) end,
+    CB = fun() -> put(cb_context:set_accepting_charges(Context), ?COLLECTION, ?ACTIVATE) end,
     set_response(Results, Context, CB);
 put(Context, Number, ?ACTIVATE) ->
     Options = [{'auth_by', cb_context:auth_account_id(Context)}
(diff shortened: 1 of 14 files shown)


applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
```diff
@@ -401,24 +401,30 @@ put(Context, Number) ->
...
-    CB = fun() -> ?MODULE:put(cb_context:set_accepting_charges(Context), Number, ?ACTIVATE) end,
-    set_response(Result, Context, CB);
+    AccountId = cb_context:account_id(Context),
+    case knm_number:buy(Number, AccountId, Options) of
+        {'error', _}=Error ->
+            set_response(Error, Context);
+        {'ok', _} ->
+            NewOptions = [{'dry_run', cb_context:accepting_charges(Context)} | Options],
+            Result = knm_number:buy(Number, AccountId, NewOptions),
+            set_response(Result, Context)
```
lazedo: @fenollp this doesn't seem right to me, what about `charges` ?

fenollp: I agree & that's what the OC says. Forgot to ping @jamesaimonetti about this

fenollp:
Please @jamesaimonetti take a look at this (and top comment)

fenollp:
@jamesaimonetti look at meeee

fenollp:
Had an epiphany. Only point **3** is relevant now.

lazedo:
@fenollp you're almost there, let me know when (3) is solved

fenollp:
@lazedo **3** is actually a question. Should the carrier error get all the way back to the user or just be logged somewhere? What do you think @jamesaimonetti ?

PR #1810, opened 2016-04-18
Crossbar schema validation

lazedo:
optionally tries to convert to numbers

### DIFF
--- a/applications/crossbar/src/cb_context.erl
+++ b/applications/crossbar/src/cb_context.erl
@@ -1251,8 +1251,8 @@ maybe_convert_numbers(Context, SchemaJObj, Errors) ->
     end.
 
 -spec maybe_convert_number(jesse_error:error_reason(), wh_json:object()) -> wh_json:object().
-maybe_convert_number({data_invalid, {Props}, 'wrong_type', Value, Key}, JObj) ->
-    case props:get_value(<<"type">>, Props) of
+maybe_convert_number({'data_invalid', SchemaJObj, 'wrong_type', Value, Key}, JObj) ->
+    case wh_json:get_value(<<"type">>, SchemaJObj) of
         <<"integer">> -> 
             try wh_util:to_integer(Value) of
                 V -> wh_json:set_value(Key, V, JObj)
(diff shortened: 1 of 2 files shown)


applications/crossbar/src/cb_context.erl
```diff
@@ -1233,3 +1236,32 @@ build_error_message(_Version, Message) when is_binary(Message) ->
...
+                                                                         ,Errors
+                                                                        ]),
+            failed(Context, Errors);
+        NewJObj ->
+            validate_request_data(SchemaJObj, set_req_data(Context, NewJObj) )
+    end.
+
+maybe_convert_number({data_invalid, {Props}, 'wrong_type', Value, Key}, JObj) ->
+    case props:get_value(<<"type">>, Props) of
+        <<"integer">> -> case filter_numbers(Value) of
```
jamesaimonetti: Why the filter? If `wh_util:to_integer(Value)` doesn't happen, move on, no?

applications/crossbar/src/cb_context.erl
```diff
@@ -1233,3 +1236,32 @@ build_error_message(_Version, Message) when is_binary(Message) ->
     wh_json:from_list([{<<"message">>, Message}]);
 build_error_message(_Version, JObj) ->
     JObj.
+
+maybe_convert_numbers(Context, SchemaJObj, Errors) ->
```
jamesaimonetti: Specs needed

applications/crossbar/src/cb_context.erl
jamesaimonetti: And what about negative numbers?

applications/crossbar/src/cb_context.erl
```diff
@@ -1233,3 +1236,30 @@ build_error_message(_Version, Message) when is_binary(Message) ->
...
+            lager:debug("request data did not validate against ~s: ~p", [wh_doc:id(SchemaJObj)
+                                                                         ,Errors
+                                                                        ]),
+            failed(Context, Errors);
+        NewJObj ->
+            validate_request_data(SchemaJObj, set_req_data(Context, NewJObj))
+    end.
+
+-spec maybe_convert_number(jesse_error:error_reason(), wh_json:object()) -> wh_json:object().
+maybe_convert_number({data_invalid, {Props}, 'wrong_type', Value, Key}, JObj) ->
```
jamesaimonetti: Why pattern match the tuple? This is a JObj, no? You can use `wh_json` below to get the type.

jamesaimonetti: Tickie the data_invalid atom too

jamesaimonetti:
@lazedo just pushed a little rename/reorg that I think captures the spirit better. Are you on board with it? I can revert if not...

PR #1811, opened 2016-04-19
HELP-22038: use formatted caller id for comparison

jamesaimonetti:


jamesaimonetti:
@k-anderson : does this get the heart of the matter? AFAICT, blacklists are the raw_numbers object merged with formatted numbers from the blacklist doc.

fogeytron:
This will meet our use case. As long as the caller id number is formatted and compared against a formatted number, we should not have any issues. Also that should alleviate the need for the raw_numbers, since that was put in place to solve the same issue.

In the code I saw that you check if the called id is undefined, then always return false, indicating not to block. It would be nice if that could be an option, to block people who do not have a called id defined.

Also, this needs to work with a caller id of all 0's - (0000000000)

jamesaimonetti:
@fogeytron thanks for the feedback. I'll discuss with @k-anderson after our standup about blocking no caller-id, as well as testing the 0s

jamesaimonetti:
@fogeytron after discussing with @k-anderson, we realized caller ID should never be empty, as that would violate the SIP spec.

With anonymous caller IDs, as long as you have them in your blacklist, they should be normalized using the same function that will normalize the caller ID from the call.

jamesaimonetti:
I should add, in master (Kazoo 4.0), we'll remove the `raw_numbers` hack, as well as give the account the ability to toggle whether to blacklist the anonymous caller ID instead of having to manually add it to the blacklist.

PR #1814, opened 2016-04-19
HELP-21360 set call-id for logging in wh_amqp_worker cast

lazedo:


fenollp:
@jamesaimonetti @lazedo note that [`log_stacktrace` uses info, not debug](https://github.com/2600hz/kazoo/blob/2caeec8cad84d3d455e293f9ac5f39f7acb0466d/core/whistle/src/wh_util.erl#L155). Maybe that should change?

lazedo:
@fenollp that's the intent

PR #1816, opened 2016-04-20
log_stacktrace: use error instead of info

fenollp:


core/kazoo_bindings/src/kazoo_bindings.erl
```diff
@@ -647,27 +645,28 @@ apply_responder(#kz_responder{module=M
 
 -spec log_undefined(atom(), atom(), non_neg_integer(), list()) -> 'ok'.
 log_undefined(M, F, Length, [{M, F, _Args,_}|_]) ->
-    lager:debug("undefined function ~s:~s/~b", [M, F, Length]);
+    lager:error("undefined function ~s:~s/~b", [M, F, Length]);
```
jamesaimonetti: This isn't necessarily worth increasing the log level. You don't have to define all the callbacks, so this could log but is benign.

fenollp: I agree for `log_undefined`, however `log_function_clause` only has true positives right?

applications/crossbar/src/modules/provisioner_v5.erl
```diff
@@ -636,7 +636,7 @@ get_schema() ->
     case wh_json_schema:load(?SCHEMA) of
         {'ok', SchemaJObj} -> SchemaJObj;
         {'error', _E} ->
-            lager:debug("failed to find schema ~s: ~p", [?SCHEMA, _E]),
+            lager:error("failed to find schema ~s: ~p", [?SCHEMA, _E]),
```
fenollp: This whole thing should use crossbar utils to load schema anyway

core/kazoo_documents/src/kzd_schema.erl
```diff
@@ -23,7 +23,7 @@ find_schema(<<_/binary>> = Schema) ->
     case wh_json_schema:load(Schema) of
         {'ok', SchemaJObj} -> SchemaJObj;
         {'error', _E} ->
-            lager:debug("failed to find schema ~s: ~p", [Schema, _E]),
+            lager:error("failed to find schema ~s: ~p", [Schema, _E]),
```
fenollp: Note: this code is used nowhere

PR #1822, opened 2016-04-21
knm-quickfix: correct id for phone_numbers schema

fenollp:


applications/crossbar/priv/couchdb/schemas/phone_numbers.json
```diff
@@ -1,6 +1,6 @@
 {
     "$schema": "http://json-schema.org/draft-03/schema#",
-    "_id": "numbers",
+    "_id": "phone_numbers",
```
fenollp: @lazedo things using this are cb_phone_numbers v1 and v2, and actually they have been using `phone_numbers.json` (and that hasn't been useful). I think removing it all together would be better, as I don't yet know what really depends on it (at least in v2). Maybe @jamesaimonetti has an idea quickly?

That can be fixed in a later PR. Let's get this merged so that the real work #1825 gets in line

PR #1823, opened 2016-04-21
Crossbar json

lazedo:


applications/crossbar/src/api_util.erl
```diff
@@ -44,7 +44,7 @@
 -define(DATA_SCHEMA
         ,wh_json:from_list([{<<"type">>, <<"object">>}
                             ,{<<"description">>, <<"The request data to be processed">>}
-                            ,{<<"required">>, 'true'}
+                            ,{<<"required">>, 'false'}
```
fenollp: Can't we still keep this?

core/whistle/src/wh_json.erl
```diff
@@ -117,53 +117,38 @@ decode(JSON, <<"application/json">>) ->
...
             Part = binary:part(Text, Loc, Size-Loc),
-            lager:debug("~s near char # ~b of ~b ('~s'): ~s", [_Err, Loc, Size, Part, Text]),
-            log_big_binary(iolist_to_binary(JSON)),
-            try_converting(JSON);
-        'throw':_E ->
-            lager:debug("thrown decoder error: ~p", [_E]),
-            new();
-        _Error:_Reason ->
-            lager:debug("decoder exited with ~p:~p", [_Error, _Reason]),
-            log_big_binary(iolist_to_binary(JSON)),
```
fenollp: Please keep the logging on

applications/crossbar/src/api_util.erl
lazedo: why ? if the json payload has a `data` element it will validate and context_req_data has its value, otherwise `context:json_request` will have the submitted payload. cb modules handle * req_data`, others like incoming webhooks don't send the `data` object because they have their own schema.

core/whistle/src/wh_json.erl
lazedo: why ? the calling module will get the exception and can log if desired.

fenollp:
So this basically fixes:

If one sends invalid json, master returns 400 `.data field is required but missing`.
Whereas this PR returns `“invalid_trailing_data near char# 84"` which is still a 400.

Not sure reverting all this history is worth it

lazedo:
the commit that introduced the wh_json:new() in the catch from decode is IMHO in error and this reverts it.
again, `400`is the http code, what's your point here ?

core/whistle/src/wh_json.erl
fenollp: The intent of https://github.com/2600hz/kazoo/pull/1668 was to log the impeding JSON that made some kazoo app crash without explanation. By logging here we make sure that next time it happens, we have logs to inspect

jamesaimonetti:
So, we should have the following cases handled with this:

1. PUT/POST/PATCH/DELETE with no request body: req_json = new(), req_data = new()
2. PUT/POST/PATCH/DELETE with request body but no `"data":{}`: 400 error
3. PUT/POST/PATCH/DELETE with invalid JSON: 400 error
4. PUT/POST/PATCH/DELETE with request body and "data":{}: req_json=request JSON, req_data = "data" object

Agreed?

fenollp:
SGTM

PR #1829, opened 2016-04-22
wnm-hide: exclude carrier module name from crossbar responses

fenollp:


jamesaimonetti:
Will this be available in the `_read_only` object?

jamesaimonetti:
I don't know if anyone relies on it, but it should probably be somewhere...maybe

fenollp:
@jamesaimonetti this is to remove it from the `_read_only` json.

No one relies on it: it's too new for that. I thought it would be useful to have in the response payload, but it probably breaks the abstraction too much that it's actually harmful.

That's what we agreed on with @lazedo 
This can be re-added later on but it has already been removed in `master`

PR #1830, opened 2016-04-22
knm-reserve: fix phone_numbers/DID/reserve

fenollp:


core/kazoo_number_manager/src/knm_number.erl
```diff
@@ -195,21 +195,18 @@ create_phone_number(Number) ->
...
-                'false' ->
-                    {'error', {'invalid_state_transition', State}}
-            end;
-        _Other -> _Other
+            attempt(fun do_reserve/1, [Number]);
+        {'error', _R}=E -> E
     end.
 
+-spec do_reserve(knm_number()) -> knm_number_return().
+do_reserve(Number) ->
```
jamesaimonetti: Why remove the state check?

fenollp: Because it is done is `knm_number_states`. It's pretty much the same impl as `knm_number:move`

PR #1831, opened 2016-04-22
prevent looping "Execute-Extension-Original-"

lazedo:


applications/ecallmgr/src/ecallmgr_call_command.erl
```diff
@@ -1192,9 +1192,12 @@ execute_exten_handle_ccvs(DP, _Node, UUID, JObj) ->
         'false' ->
             ChannelVars = wh_json:to_proplist(CCVs),
             [{"application", <<"set ", (ecallmgr_util:get_fs_kv(K, V, UUID))/binary>>}
-             || {K, V} <- ChannelVars] ++ DP
+             || {K, V} <- ChannelVars, not execute_exten_is_original_ccv(K)] ++ DP
     end.
 
+execute_exten_is_original_ccv(<<?CHANNEL_VARS_EXT, _/binary>>) -> 'true';
```
jamesaimonetti: Just needs a spec

applications/ecallmgr/src/ecallmgr_call_command.erl
```diff
@@ -1227,7 +1230,15 @@ create_dialplan_move_ccvs(Node, UUID, DP) ->
 -spec create_dialplan_move_ccvs(wh_proplist(), wh_proplist()) -> wh_proplist().
 create_dialplan_move_ccvs(DP, Props) ->
     lists:foldr(
-      fun({<<"variable_", ?CHANNEL_VAR_PREFIX, Key/binary>>, Val}, Acc) ->
+      fun({<<"variable_", ?CHANNEL_VAR_PREFIX, ?CHANNEL_VARS_EXT, _Key/binary>>, _Val}, Acc) ->
```
jamesaimonetti: Would this be better in a named function now? It doesn't seem like we're importing variables from the outer scope so it should be fine.

applications/ecallmgr/src/ecallmgr_call_command.erl
lazedo: @jamesaimonetti i didn't add it because all other `execute_exte` don't have specs and this is just to prevent the loop.  all these `execute_exte` should actually be deleted in 4.0

applications/ecallmgr/src/ecallmgr_call_command.erl
lazedo: @jamesaimonetti  this is just to prevent looping for now, this code is going away.

PR #1833, opened 2016-04-26
Knm bandwidth2

fenollp:
why is Dialyzer complaining here?


lazedo:
@fenollp please see this after the `great rename`

PR #1841, opened 2016-04-28
HELP-22210: kz_cache backport to 3.22

jamesaimonetti:
Tested thusly:

`kz_tracers:gen_load(N).` will spawn N workers who will create a new database and 1000 documents inside that database, and will perform edit/delete operations against those documents until all are deleted, at which point the database is deleted.

The parent process will track the `whistle_couch_cache` process' largest `message_queue_len` and `current_function`, and will output those values (along with runtime):

```
kz_tracers:gen_load(5).
...processing...
finished test after 531140ms: {23, {lager_format, build2, 3}}
```

In this case, the largest mailbox length found was 23 messages and the whistle_couch_cache process was in `lager_format:build2/3` when that was read.

For perspective, the previous `wh_cache` topped out around 23K messages in the queue, so this PR marks a significant improvement.

### DIFF
--- a/core/whistle-1.0.0/Makefile
+++ b/core/whistle-1.0.0/Makefile
@@ -3,6 +3,8 @@ ROOT = ../..
 
 EBINS = $(shell find $(ROOT)/deps/lager-* -maxdepth 2 -name ebin -print) \
 	$(shell find $(ROOT)/core/whistle-* -maxdepth 2 -name ebin -print) \
+	$(shell find $(ROOT)/core/whistle_couch-* -maxdepth 2 -name ebin -print) \
+	$(shell find $(ROOT)/deps/couchbeam-* -maxdepth 2 -name ebin -print) \
 	$(shell find $(ROOT)/deps/rabbitmq_client-* -maxdepth 2 -name ebin -print)
 PA = $(foreach EBIN,$(EBINS),-pa $(EBIN))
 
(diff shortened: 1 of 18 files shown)


jamesaimonetti:
The other important thing to note is that I was using gen_server:call/2 for store/erase ops. This was in part for the PropEr testing to make it easier to not fail a test while data was in-flight to ETS, but also shows just how far along this version is in performance. We can revert those to cast/2 if desired.

jamesaimonetti:
@k-anderson and @lazedo: please review, test yourselves, and let's chat as soon as @k-anderson is around in the morning.

lazedo:
@jamesaimonetti looks good to me, didn't test it in 3.22 yet.

i would like to see the ops terminate before all docs are empty (random between 40-50 missing docs?) so that we also check that the delete database flushes all cached docs for that database, and would also test a large delete. thoughts ?

jamesaimonetti:
@lazedo good thought. I will test it when I get into the office.

jamesaimonetti:
`kz_tracers:gen_load(Workers, Docs)` to play with concurrent workers and the number of docs to build at the start. There's also a random chance the account db is deleted after all its docs are saved, to test flushing the cache.

lazedo:
@jamesaimonetti check any changes into master, specially the goodies in `kz_tracers`

PR #1844, opened 2016-04-28
Kazoo 4091 part un

fenollp:


fenollp:
I'm sorry I don't think this should be merged. Luis had comments

PR #1845, opened 2016-04-28
qfs-1: fix format errors due to kz_http return type change

fenollp:


core/kazoo_oauth/src/kazoo_oauth_util.erl
```diff
@@ -203,7 +203,7 @@ verify_token(ProviderId, AccessToken) when is_binary(ProviderId) ->
         {'error', _}=Error -> Error
     end;
 verify_token(#oauth_provider{tokeninfo_url=TokenInfoUrl}, AccessToken) ->
-    URL = <<TokenInfoUrl/binary,AccessToken/binary>>,
+    URL = <<TokenInfoUrl/binary, AccessToken/binary>>,
```
lazedo: don't touch unnecessary files

fenollp: Then don't let your fingers eat spaces :)

lazedo: stop running `sed` :)

fenollp: If I was it would be run by Travis and this PR wouldn't exist.

This PR actually fixes things that are broken, in the does-not-work-at-all sense. Can we merge it soon?

PR #1846, opened 2016-04-28
qfs-2: correctly log stack trace

fenollp:


core/kazoo_number_manager/src/carriers/knm_voip_innovations.erl
```diff
@@ -387,8 +387,9 @@ handle_response({'ok', Code, _Headers, "<?xml"++_=Response}) ->
         verify_response(Xml)
     catch
         _:R ->
+            ST = erlang:get_stacktrace(),
             lager:debug("failed to decode xml: ~p", [R]),
-            lager:debug("st: ~p", [erlang:get_stacktrace()]),
+            wh_util:log_stacktrace(ST),
```
lazedo: why are we logging the stack here ?

PR #1851, opened 2016-04-29
Fix cb_sup

skorobkov:
validate/{2,3} calls module_name/1 several times and each added "_maintenance" to module name

### DIFF
--- a/applications/crossbar/src/modules/cb_sup.erl
+++ b/applications/crossbar/src/modules/cb_sup.erl
@@ -215,7 +215,10 @@ module_name(ModuleBin) ->
     %% NOTE: the unsafe convertion to an atom is not an issue
     %%   in this module, despite coming from a user, because
     %%   only the system admin has access...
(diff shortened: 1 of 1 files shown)


lazedo:
adds "_maintenance" to module name to keep the system a bit more safe

skorobkov:
It stil adds. This PR prevent double/triple "_maintenance" adding.

lazedo:
@skorobkov i don't see where it gets duplicated. are you sure that it isn't the user that is adding _maintenance ? i've been playing with this in the last last days without problem. but maybe i'm bypassing some use case.

skorobkov:
Yes, you are right.
There should be fixed only lager:debug output.

When i testing, there was interefernce from other function, which throw `function_clause` and I was deceived with message "failed to run whistle_services_maintenance_maintenance:debit/3: error: function_clause".

lazedo:
@skorobkov thanks

PR #1852, opened 2016-04-29
HELP-22409: if the carrier module is not enabled allow the disconnect

k-anderson:


jamesaimonetti:
@fenollp take note!

fenollp:
I did take note already :)

PR #1858, opened 2016-05-02
fwdport_1852: allow disconnecting numbers from unknown carriers

fenollp:
forward port of https://github.com/2600hz/kazoo/pull/1852

### DIFF
--- a/core/kazoo_number_manager/src/carriers/knm_carriers.erl
+++ b/core/kazoo_number_manager/src/carriers/knm_carriers.erl
@@ -318,19 +318,31 @@ acquire(Number, ?NE_BINARY = Mod, 'false') ->
 -spec disconnect(knm_number:knm_number()) ->
                         knm_number:knm_number().
 disconnect(Number) ->
-    Module = carrier_module(Number),
-    Module:disconnect_number(Number).
+    case knm_phone_number:module_name(knm_number:phone_number(Number)) of
+        ?NE_BINARY=Mod ->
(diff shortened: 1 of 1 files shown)


core/kazoo_number_manager/src/carriers/knm_carriers.erl
```diff
@@ -318,20 +318,32 @@ acquire(Number, ?NE_BINARY = Mod, 'false') ->
...
-    Module = carrier_module(Number),
-    Module:disconnect_number(Number).
+    case knm_phone_number:module_name(knm_number:phone_number(Number)) of
+        ?NE_BINARY=Mod ->
+            Module = wh_util:to_atom(Mod, 'true'),
+            lager:debug("contacting carrier ~s", [Module]),
+            Module:disconnect_number(Number);
+        _Mod ->
+            lager:debug("non-existant carrier module ~p, allowing disconnect", [_Mod]),
+            Number
```
lazedo: shouldn't we release the number ?
from knm_managed ...
```
 update_doc(Number, [{?PVT_STATE, ?NUMBER_STATE_RELEASED}
                        ,{<<"pvt_assigned_to">>,<<>>}
                       ]).
```

fenollp: knm_carriers:disconnect/1 is only called by knm_number:release which calls into number_states which does the state changing stuff

PR #1861, opened 2016-05-02
ci-docs: check how many API endpoints are documented (120 / 475)

fenollp:


jamesaimonetti:
Interesting, I like the idea. There may need to be some handling as not all ref docs map to doc/ versions. For instance, I think `ref/user_auth.md` is contained in `doc/user_authentication.md`

fenollp:
I realised there is an issue: some false positives: `> endpoint` says found when it should look for `> endpoint\n`…

There is no direct mapping. ie: 

1. it finds api endpoints (things starting with `>` and ending with `\n`) under `ref/`
1. then it looks for said enpoints in any file matching `doc/*.md`

jamesaimonetti:
Indeed. I am open to suggestions on how to tie ref-generated docs to
their doc/ counterparts without overly complex string searching.

Pierre Fenoll writes:

> I realised there is an issue: some false positives: > endpoint says
> found when it should look for > endpoint\n…
> There is no direct mapping. ie:
>    1. it finds api endpoints (things starting with > and ending with \n)
>       under ref/
>    2. then it looks for said enpoints in any file matching doc/*.md
> —
> You are receiving this because you commented.
> Reply to this email directly or view_it_on_GitHub


-- 
James Aimonetti

Lead Systems Architect
"If Dialyzer don't care, I don't care"
2600HzPDX | http://2600hz.com
sip:james@2600hz.com
tel:415.886.7905
irc:mc_ @ freenode

fenollp:
False positives fixed.

@jamesaimonetti string searching is close to the simplest here. Also there is no need to try to maintain some filename-API_endpoint associativity here: all endpoints are in the same /v2/… scope and are thus already unique. I don't think they'll ever be documented twice (that would break this algo).

PR #1862, opened 2016-05-02
The great rename final

mark2600:


ANNOUNCEMENTS.md
```diff
@@ -49,9 +53,9 @@ The Blackhole application providing websocket support currently utilizes the soc
 
 The media prompts that we used to keep in the code repository for historical reasons has been moved the kazoo-prompts repository.
 
-#### Consistent naming (removal of references to whistle)
+#### Consistent naming (removal of references to kazoo)
 
-All instances of whistle or variations such as wh_ have been replaced with kazoo or kz_.  This will include all maintenance functions as well as init.d/systemd scripts.  A script is provided to convert any custom code but will require in-depth testing.
+All instances of kazoo or variations such as kz_ have been replaced with kazoo or kz_.  This will include all maintenance functions as well as init.d/systemd scripts.  A script is provided to convert any custom code but will require in-depth testing.
```
swysor: Can we exclude this file from the script? Its supposed to say whistle here because it is explaining that it is migrating from whistle.

lazedo: @swysor will do after merge, thanks for noticing

PR #1864, opened 2016-05-03
fix an issue occured when rebasing

icehess:


lazedo:
@icehess i think the problem may be caused by https://github.com/2600hz/kazoo/commit/f38bc4229dd3763aed9aeabc40561fab2cdbed8f

there seems to be some confusion here.

icehess:
@lazedo there is no `listing_by_username` view in `users.json` as I can see.

lazedo:
@icehess yep, but there's one here https://github.com/2600hz/kazoo/blob/master/applications/crossbar/priv/couchdb/views/user_auth.json

PR #1866, opened 2016-05-03
sup-cmd: add an installable shell file for sup & some documentation

fenollp:


fenollp:
@k-anderson This should help for installation. I can set a better `default_root` but this one works great for dev. I introduced `KAZOO_ROOT` but maybe you have set a hard path for that? I could have it try `default_root` and then `/opt/kazoo`.

doc/installation.md
```diff
@@ -60,6 +60,28 @@ $ make
...
+Make sure that the path to Kazoo's intallation directory is right (in `/usr/bin/sup`).
+Otherwise you can change it by setting the `KAZOO_ROOT` environment variable (not set by default).
+If one needs `KAZOO_ROOT`, an alias should be created:
+
+```shell
+alias sup='KAZOO_ROOT=/opt/kazoo sup'
+```
+
+It is preferable but not required that one also adds autocompletion to the `sup` command.
+In order to do this, see documentation about `make sup_completion`.
```
jamesaimonetti: Is there an anchor we could add to jump to this section directly?

scripts/setup-dev.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-ln -s /opt/kazoo/utils/sup/sup /usr/bin/sup
+ln -s /opt/kazoo/core/priv/sup /usr/bin/sup
```
mark2600: This symlink will fail on systems that already have /usr/bin/sup set (so it will not work to migrate from an existing install).

This is the same for all of the symlinks set in this file, I think that should probably be changed so that it always re-creates the symlinks.

fenollp:
cherry picked to 4.0

sergey-safarov:
@mark2600 could you please look at KAZOO-4726

PR #1870, opened 2016-05-03
require_realm: update schema, require realm on PUT/POST

jamesaimonetti:


applications/crossbar/priv/couchdb/schemas/connectivity.json
```diff
@@ -43,10 +44,12 @@
                     "type": "integer"
                 }
             },
+            "required": true,
             "type": "object"
         },
         "name": {
             "description": "Human-friendly name of the trunkstore account",
+            "required": true,
```
fenollp: why require name? it says "require realm"

jamesaimonetti: Name is required too

fenollp: Your commit message was a lie!

fenollp:
@jamesaimonetti fwd port this plz

jamesaimonetti:
KAZOO-4687 will be the proper fixes for 4.x. This was a 3.22-only patch

PR #1871, opened 2016-05-04
Proposal to make webseq status configurable

tsromana:
Proposal to load webseq status load from a configurable place as keeping it on everytime crashes konami. i am looking into why there is a crash on some particular call transfers

### DIFF
--- a/applications/konami/src/konami_code_fsm.erl
+++ b/applications/konami/src/konami_code_fsm.erl
@@ -56,7 +56,7 @@
                }).
 -type state() :: #state{}.
 
--define(WSD_ENABLED, whapps_config:get_integer(?APP_NAME, <<"webseq_enabled">>, 'true')).
+-define(WSD_ENABLED, whapps_config:get_integer(?APP_NAME, <<"webseq_enabled">>, 'false')).
 
 -define(WSD_ID, ?WSD_ENABLED andalso {'file', get('callid')}).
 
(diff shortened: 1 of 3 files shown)


applications/konami/src/konami_code_fsm.erl
```diff
@@ -56,7 +56,7 @@
                }).
 -type state() :: #state{}.
 
--define(WSD_ENABLED, 'false').
+-define(WSD_ENABLED, kapps_config:get_integer(?APP_NAME, <<"webseq_enabled">>, 'false')).
```
jamesaimonetti: Should use `get_is_true/3`

tsromana: sorry was half asleep while sending the pr. should i move the declaration to a central place in konami.hrl ? as both konami_code_fsm.erl and konami_transfer.erl requires it or there should be different variables for both of them. as with this variable it will either be true or false for both.

jamesaimonetti: moving it to the .hrl is a good idea

tsromana: and should it be false by default ?

lazedo:
@tsromana thanks

lazedo:
@tsromana i merged this but in the future try to squash your commits. thanks

tsromana:
@lazedo yea sure sorry for this !

PR #1877, opened 2016-05-05
KAZOO-4738 handle possible migration issues at startup

lazedo:


core/kazoo_apps/src/kapps_maintenance.erl
```diff
@@ -1217,3 +1220,58 @@ remove_system_media_refs(HangupCause, Config) ->
                                      kz_json:object().
 remove_system_media_ref(Key, <<"/system_media/", Value/binary>>, Acc) -> kz_json:set_value(Key, Value, Acc);
 remove_system_media_ref(Key, Value, Acc) -> kz_json:set_value(Key, Value, Acc).
+
+-spec last_migrate_version() -> ne_binary().
+last_migrate_version() ->
+    kapps_config:get_ne_binary(?MODULE, <<"migrate">>, <<"3.22">>).
```
jamesaimonetti: I think this should be "migrate_from" or "current_version" even.

core/kazoo_apps/src/kapps_maintenance.erl
```diff
@@ -1217,3 +1220,58 @@ remove_system_media_refs(HangupCause, Config) ->
...
+    migrate_system(last_migrate_version(), kz_util:kazoo_version()).
+
+-spec migrate_system(ne_binary(), ne_binary()) -> 'ok'.
+migrate_system(ThisVersion, ThisVersion) ->
+    lager:notice("system config version is ~s", [ThisVersion]);
+migrate_system(PreviousVersion, ThisVersion) ->
+    Routines = migrate_system_version_routines(PreviousVersion, ThisVersion),
+    Result = lists:foldl(
+               fun(Fun, Acc) when is_function(Fun, 2) ->
+                       [Acc | Fun(PreviousVersion, ThisVersion)];
```
jamesaimonetti: Why build these lists like this? Acc is a list. Fun returns...a boolean in handle_midule_rename's case. So you construct [[] | false] which is an improper list.

lazedo: this was stupid copy/paste on the wrong side!

PR #1878, opened 2016-05-05
Fixed an issue with adding a customer

tsromana:
When someone needed to add a customer without having their credit card braintree send an error that make_default(undefined) clause is not defined in braintree_card.erl. now we can add a new customer in braintree by just having using his info and adding a card is not needed.

### DIFF
--- a/core/braintree/src/braintree_card.erl
+++ b/core/braintree/src/braintree_card.erl
@@ -176,11 +176,10 @@ expiring(Start, End) ->
 %% Accessors for field 'make_default'.
 %% @end
 %%--------------------------------------------------------------------
--spec make_default(bt_card() | 'undefined') -> api_boolean().
+-spec make_default(bt_card()) -> api_boolean().
 -spec make_default(bt_card(), boolean()) -> bt_card().
 
-make_default(#bt_card{make_default = 'true'}) -> 'true';
-make_default(_) -> 'false'.
+make_default(#bt_card{make_default = Value}) -> Value.
 
 make_default(#bt_card{}=Card, Value) ->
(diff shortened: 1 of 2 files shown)


tsromana:
would it be nice to check inside the first make_default function to see if the value is either true or false and have a catch all clause instead of undefined ?

core/braintree/src/braintree_card.erl
```diff
@@ -179,7 +179,8 @@ expiring(Start, End) ->
 -spec make_default(bt_card()) -> api_boolean().
```
jamesaimonetti: This needs updating to `bt_card() | 'undefined'`.

tsromana:
something like this

make_default(#bt_card{make_default = 'true'}) -> 'true';
make_default(_) -> 'false'.

tsromana:
@jamesaimonetti will the above code would be better ?

jamesaimonetti:
@tsromana no, I prefer the explicitness of your original version; you either get the card's made_default boolean value. Honestly though, as I think about it more, I wonder who's calling make_default/1 with 'undefined' instead of a bt_card()? Shouldn't we look to fix the caller instead? braintree_card.erl is meant to operation on bt_card() records.

tsromana:
@jamesaimonetti thats what i thought now that why is it undefined at first it should be card record i will look into it. let me look further and i will update the pr.

jamesaimonetti:
@tsromana thanks.

tsromana:
@jamesaimonetti let me do some more tests to make sure everything is fine will update you when everything is good

tsromana:
@jamesaimonetti everythings good, pr can be merged

PR #1879, opened 2016-05-06
Sysconf should send answer when "Value" is undefined.

skorobkov:
Sysconfig doesnt send any answer when return value will be empty, this leads to unnecessary delays in call procedding. In my situation i had 3 seconds delay on every call, because ecallmgr couldnt get value for "should_detect_inband_dtmf".
Manualy set this parameter in DB solves my problem, but does not eliminate the root of the problem.

### DIFF
--- a/applications/sysconf/src/sysconf_get.erl
+++ b/applications/sysconf/src/sysconf_get.erl
@@ -28,22 +28,19 @@ handle_req(ApiJObj, _Props) ->
 
     lager:debug("received sysconf get for ~s:~s from ~s", [Category, Key, Node]),
 
-    case get_value(Category, Key, Default, Node) of
-        'undefined' -> 'ok';
-        Value ->
-            RespQ = kz_json:get_value(<<"Server-ID">>, ApiJObj),
+    Value = get_value(Category, Key, Default, Node),
+    RespQ = kz_json:get_value(<<"Server-ID">>, ApiJObj),
 
-            lager:debug("sending reply for ~s.~s(~s): ~p"
-                        ,[Category, Key, Node, Value]
-                       ),
-            Resp = [{<<"Category">>, Category}
-                    ,{<<"Key">>, Key}
-                    ,{<<"Value">>, Value}
-                    ,{<<"Msg-ID">>, kz_json:get_value(<<"Msg-ID">>, ApiJObj)}
-                    | kz_api:default_headers(?APP_NAME, ?APP_VERSION)
-                   ],
-            kapi_sysconf:publish_get_resp(RespQ, Resp)
-    end.
+    lager:debug("sending reply for ~s.~s(~s): ~p"
+                ,[Category, Key, Node, Value]
+               ),
+    Resp = [{<<"Category">>, Category}
+            ,{<<"Key">>, Key}
+            ,{<<"Value">>, Value}
+            ,{<<"Msg-ID">>, kz_json:get_value(<<"Msg-ID">>, ApiJObj)}
+            | kz_api:default_headers(?APP_NAME, ?APP_VERSION)
+           ],
+    kapi_sysconf:publish_get_resp(RespQ, Resp).
 
 -spec get_value(ne_binary(), ne_binary(), any(), ne_binary()) -> any().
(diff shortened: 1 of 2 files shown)


skorobkov:
Found another interesting issue.
When FS reloads sofia or ACL (for example after you update global resources), its try to get configs `spandsp.conf` and `timezone.conf` (sequential).
Each of config request return `not found` result with 3 seconds delay (after that FS load config from local files).
Thsi timeout result of call `ecallmgr_config:get(<<"configuration_handlers">>)` in `ecallmgr_fs_config.erl`.
While FS not get this configs (or load from local files) you will not be able to make call and kazoo will got error:
```
May 30 10:35:16 t13-dev 2600hz[22086]: |YWUwZmVkNzIwNzc1NjNlOTI1YmJhOGM5YWRlMWU0Yjc.|ecallmgr_call_control:1105 (<0.17268.0>) invalid command bridge: not_found
May 30 10:35:16 t13-dev 2600hz[22086]: |YWUwZmVkNzIwNzc1NjNlOTI1YmJhOGM5YWRlMWU0Yjc.|kz_util:164 (<0.17268.0>) stacktrace:
May 30 10:35:16 t13-dev 2600hz[22086]: |YWUwZmVkNzIwNzc1NjNlOTI1YmJhOGM5YWRlMWU0Yjc.|kz_util:171 (<0.17268.0>) st: ecallmgr_fs_bridge:call_command/3 at (36)
May 30 10:35:16 t13-dev 2600hz[22086]: |YWUwZmVkNzIwNzc1NjNlOTI1YmJhOGM5YWRlMWU0Yjc.|kz_util:171 (<0.17268.0>) st: ecallmgr_call_command:exec_cmd/5 at (36)
May 30 10:35:16 t13-dev 2600hz[22086]: |YWUwZmVkNzIwNzc1NjNlOTI1YmJhOGM5YWRlMWU0Yjc.|kz_util:171 (<0.17268.0>) st: ecallmgr_call_control:execute_control_request/2 at (1082)
May 30 10:35:16 t13-dev 2600hz[22086]: |YWUwZmVkNzIwNzc1NjNlOTI1YmJhOGM5YWRlMWU0Yjc.|kz_util:171 (<0.17268.0>) st: ecallmgr_call_control:handle_dialplan/2 at (868)
May 30 10:35:16 t13-dev 2600hz[22086]: |YWUwZmVkNzIwNzc1NjNlOTI1YmJhOGM5YWRlMWU0Yjc.|kz_util:171 (<0.17268.0>) st: ecallmgr_call_control:handle_cast/2 at (265)
May 30 10:35:16 t13-dev 2600hz[22086]: |YWUwZmVkNzIwNzc1NjNlOTI1YmJhOGM5YWRlMWU0Yjc.|kz_util:171 (<0.17268.0>) st: gen_listener:handle_module_cast/2 at (974)
May 30 10:35:16 t13-dev 2600hz[22086]: |YWUwZmVkNzIwNzc1NjNlOTI1YmJhOGM5YWRlMWU0Yjc.|kz_util:171 (<0.17268.0>) st: gen_server:try_dispatch/4 at (615)
May 30 10:35:16 t13-dev 2600hz[22086]: |YWUwZmVkNzIwNzc1NjNlOTI1YmJhOGM5YWRlMWU0Yjc.|kz_util:171 (<0.17268.0>) st: gen_server:handle_msg/5 at (681)
```
If you wait for 6 seconds (3 seconds timeout for each config file req) after updating resource, call will originate without problems.

k-anderson:
@skorobkov Howdy,

So there is some danger in replying authoritative from a cached value which is why it defers to other syconf by not sending a response.  However, after talking with the team and throwing some ideas around we decided that it could be authoritative if it bypassed the cache AND the database operation was successful or 404.  Since the other sides are caching there really isn't a performance hit (or need) to use the local config cache.  As such I have opened https://2600hz.atlassian.net/browse/KAZOO-4806.

If we modify sysconfig to use fetch operations via the config library then we feel it would be safe to reply with an empty value.

Let me know your thoughts on this :)

lazedo:
@skorobkov as discussed, closing this in favor of KAZOO-4806

PR #1881, opened 2016-05-06
fix warning of unhandled account ID

fenollp:
@lazedo I get a warning every time a number db is created because revise_views tries to convert NumberDb to an account ID (format_account_id fails & returns NumberDb).

Should revise_views really take an account ID? What do you think should be done here?

### DIFF
--- a/core/kazoo_data/src/kz_datamgr.erl
+++ b/core/kazoo_data/src/kz_datamgr.erl
@@ -1094,8 +1094,8 @@ change_notice() ->
 %% @private
 %% @doc
 %% NOTE: the attempt to correct the dbname is not very erlang like, but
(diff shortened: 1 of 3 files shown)


lazedo:
@fenollp where do you see that conversion happening ? i don't see revise_views doing any conversion

lazedo:
@fenollp i see where the problem is. i'll add a commit to this branch. you should actually see the warning everytime a number is saved because kzs_publish:publish_doc tries to get account_id from DbName.

lazedo:
@fenollp check if this is what you're looking for.

fenollp:
@lazedo yes this works great. Thanks!

PR #1882, opened 2016-05-06
knm-features: fix backwards compat issue : {} v. []

fenollp:


fenollp:
These fixes are very important! Needs backport to 4.0 too

core/kazoo_number_manager/src/knm_phone_number.erl
```diff
@@ -372,6 +368,11 @@ setters(Number, Routines) ->
         Result -> {'ok', Result}
     catch
         'throw':{'stop', Error} -> Error;
+        'error':'function_clause' ->
+            {_M, FName, Args, _Info} = hd(erlang:get_stacktrace()),
```
jamesaimonetti: Why not capture the stacktrace, pop the head, then call `log_stacktrace/1`?

PR #1885, opened 2016-05-06
Fixed braintree address duplication issue

tsromana:
When a post request is sent to braintree/customer the customers cards are deleted and new cards are added leaving behind old addresses which can pile up, this change also deletes all the old addresses associated with those cards, making braintree clean and admin happy

### DIFF
--- a/core/braintree/src/braintree_customer.erl
+++ b/core/braintree/src/braintree_customer.erl
@@ -233,12 +233,17 @@ update_card(Customer, Card) ->
                  )
                ],
 
-    %% Delete previous cards /after/ changing subscriptions' payment token.
-    lists:foreach(fun braintree_card:delete/1, OldCards),
+    %% Delete previous cards and addresses /after/ changing subscriptions' payment token.
+    lists:foreach(fun delete_old_card_and_address/1, OldCards),
 
     UpdatedCustomer#bt_customer{credit_cards = NewCards
                                 ,subscriptions = NewSubscriptions
                                }.
+
+-spec delete_old_card_and_address(bt_card()) -> bt_card().
+delete_old_card_and_address(#bt_card{billing_address=OldAddress}=OldCard) ->
+    braintree_address:delete(OldAddress),
+    braintree_card:delete(OldCard).
 
 -spec do_update(bt_customer()) -> bt_customer().
 do_update(#bt_customer{id=CustomerId}=Customer) ->


tsromana:
@jamesaimonetti so this will fix the above mentioned issue, please check the code and let me know if this can be written some other way which you feel is the right way. This also made me think what if we wanted to delete only a single card and that will also delete the card and will leave the address behind and no card shares the address associated with one card so should we add the clause to delete the address when we delete the card ?

jamesaimonetti:
Thanks @tsromana; pinging @k-anderson to weigh in as he'll know more of the nuance.

tsromana:
so i was wrong with sharing of addresses we can share the address with different cards, for that purpose when adding a new card we can try to check existing addresses and instead of adding a new address we just add the id of the previous address if it matches with new address given for the new card and when deleting a card checking if there are other cards utilizing the same address, if not we can safely delete the address as well. Otherwise if we do not want to share an address we can safely delete address.

jamesaimonetti:
@tsromana should this PR be closed then? Or do you plan to amend this one with the address checks you describe above?

tsromana:
@jamesaimonetti the issue still remains these are just some thoughts, there according to me are two options
OPTION 1: The address should be different for each card and delete the address when the card is deleted. (This pr only deletes address when we bulk add new cards using the braintree/customer api, not throught other apis)
OPTION 2: When adding a new card check to see if there are available addresses that are same as the address being added, if yes associate the card with the old address otherwise add a new address and when deleting a card, we check to see if there some cards who are using that address, if not delete it.

Now the first option is good enough if we do not want to use the addresses api to get addresses and just want to associate a card with just postal code of a user. Than we can add multiple addresses even if they are basically the same. This will be less work but will workout. 

The second option i guess is more professional when we want to associate cards with full addresses at that time we can use the addresses api to show users their addresses in a similar fashion as with amazon. Than when a user is adding a card he has the option to just select a previously added card bypassing our first check to see if there is a similar address for the user in the system.

So basically just let me know what you prefer is the right way to do it and i will try my best to do it that way. if you agree i will implement the first version and after that i could try to add new checks pieces by pieces completing the whole procedure.

fenollp:
The intent is to always only keep one card. In that idea there should only ever be one address stored too.

So I'm leaning towards OPTION 2.

tsromana:
‎so i will update the pr with the scenario that whenever somone deletes the card it will check to see if the address is being used by any other card if not it will delete the address.                                                                                                                                                                                                                 From: Pierre FenollSent: Monday, May 9, 2016 4:57 AMTo: 2600hz/kazooReply To: 2600hz/kazooCc: Romana; MentionSubject: Re: [2600hz/kazoo] Fixed braintree address duplication issue (#1885)The intent is to always only keep one card. In that idea there should only ever be one address stored too.

So I'm leaning towards OPTION 2.

—You are receiving this because you were mentioned.Reply to this email directly or view it on GitHub

fenollp:
Note that the supposedly only other card that could be using an address should be the card that is going to be removed in that same call. So this card's address is not to be taken into account ;)

tsromana:
This pr is being moved here https://github.com/2600hz/kazoo/pull/1897 as the old repo is lost,this will be deleted

PR #1886, opened 2016-05-07
Validate js

fenollp:
Heuristic: looks for JS code in values of JSON files (strings starting with `"function"`).

@lazedo do you think this heuristic misses some JS? Is there JS code in our trunk that doesn't start with `"function"`?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,7 @@ after_success:
   - files="$(git diff --name-only master... -- application/ core/)" || true && echo $files
   - make build-plt
   - ./scripts/check-dialyzer.escript .kazoo.plt $files
+  - ./scripts/validate-js.sh $(find applications/ core/ -name '*.json')
   - make sup_completion
   - make build-ci-release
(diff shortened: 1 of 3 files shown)


lazedo:
@fenollp i like this a lot :) 
can you run this in all the views in the repo to make sure it works before merging ?

lazedo:
@fenollp how about adding this to local builds ?

fenollp:
In this PR Travis went through all JSON files of the repo. (see `.travis.yml` in the diff).

Please answer my question: `Is this checking all JS?` (is my heuristic good)

New question now: `should this run only on .json files in views/ ? Probably not given there are views in applications/crossbar/priv/couchdb/account/`

lazedo:
@fenollp well, the diff was only one file. if you're checking all string starting with function then it should be good, but, if you add it to local builds, then you can run it against all views to make sure.

fenollp:
The diff shows one file because it found one bug.

As I said [Travis ran on all json files](https://github.com/2600hz/kazoo/pull/1886/files#diff-354f30a63fb0907d4ad57269548329e3R43) or I misunderstood what you mean by `local builds`?

PR #1887, opened 2016-05-07
added new agent status query element

tsromana:
Added new query element in acdc queue status which will return full agent status

GET /agents/ID/status?full

### DIFF
--- a/applications/crossbar/src/modules/cb_agents.erl
+++ b/applications/crossbar/src/modules/cb_agents.erl
@@ -1,5 +1,5 @@
 %%%-------------------------------------------------------------------
-%%% @copyright (C) 2011-2014, 2600Hz INC
+%%% @copyright (C) 2011-2016, 2600Hz INC
 %%% @doc
 %%%
 %%% CRUD for call queues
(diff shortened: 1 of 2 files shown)


applications/crossbar/src/modules/cb_agents.erl
```diff
@@ -309,6 +312,35 @@ fetch_all_current_stats(Context, AgentId) ->
             ]),
     fetch_stats_from_amqp(Context, Req).
 
+-spec fetch_current_status(cb_context:context(), api_binary(), api_boolean()) -> cb_context:context().
+fetch_current_status(Context, AgentId, Full) ->
+    case Full of
```
fenollp: Use 2 function clauses instead of 2 case clauses please:

```erlang
fetch_current_status(Context, AgentId, 'false') ->
    …;
fetch_current_status(Context, AgentId, 'true') ->
    ….
```

tsromana:
So this will basically add new api parameter but in reality the api have bugs and it is incomplete, the code also look dated if allowed i would like to update the api with some changes as well update the docs to reflect the changes. The reason i want permission to do so is because of current way of implementation of the api and it will change some of its payload and will allign cb_agents more with payloads from other apis. i would like to discuss the issues so that i could work on it.

lazedo:
@tsromana thanks

PR #1897, opened 2016-05-10
Billing pr

tsromana:
This pr will make sure that every address associated with old cards is deleted properly as well as if the old address is being used by the new card it will not delete it.

### DIFF
--- a/core/braintree/src/braintree_customer.erl
+++ b/core/braintree/src/braintree_customer.erl
@@ -217,7 +217,7 @@ update_card(Customer, Card) ->
                         ,get_cards(UpdatedCustomer)
                        ),
 
-    NewSubscriptions =
+    _NewSubscriptions =
         [braintree_subscription:update(
            braintree_subscription:update_payment_token(Sub, NewPaymentToken)
           )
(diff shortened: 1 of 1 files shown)


tsromana:
It is continue of https://github.com/2600hz/kazoo/pull/1885

core/braintree/src/braintree_address.erl
```diff
@@ -92,6 +92,7 @@ update(#bt_address{id=AddressId
 -spec delete(bt_address()) -> bt_address().
 -spec delete(ne_binary() | nonempty_string(), ne_binary() | nonempty_string()) ->  bt_address().
 
+delete(#bt_address{id='undefined'}) -> #bt_address{};
```
fenollp: Could you filter out these beforehand? I would very much prefer we don't add cases like these & just be more careful.

tsromana: @fenollp will do so i was thinking of adding the check to multiple apis and thought every api will have to look here, but i was wrong and now i think this is wrong :p will do as suggested

core/braintree/src/braintree_customer.erl
```diff
@@ -228,17 +228,31 @@ update_card(Customer, Card) ->
...
+
+    %% Delete previous cards and addresses /after/ changing subscriptions' payment token.
+    delete_old_cards_and_addresses(OldCards, NewCard1),
+
+    %%get all the new user info
+    braintree_customer:find(UpdatedCustomer).
+
+-spec delete_old_cards_and_addresses(bt_cards(), bt_card()) -> ok.
+delete_old_cards_and_addresses(OldCards, #bt_card{billing_address_id=NewAddressId}) ->
+    lists:foreach(fun(#bt_card{billing_address=OldAddress, billing_address_id=OldAddressId}=OldCard) ->
```
jamesaimonetti: I would prefer matching `OldAddressId` in the function head instead of a case here. I believe you'll need to do a guard for the `NewAddressId` clause.

```erlang
lists:foreach(
    fun(#bt_card{billing_address_id='undefined'}=OldCard) ->
            braintree_card:delete(OldCard);
        (#bt_card{billing_address_id=BillingAddressId}=OldCard) when BillingAddressId =:= NewAddressId ->
            braintree_card:delete(OldCard);
        (#bt_card{billing_address=OldAddress}=OldCard) ->
            braintree_card:delete(OldCard),
            braintree_address:delete(OldAddress)
        end
       ,OldCards
)
```

Something like that. Removes a nested chunk and I think captures the intent better.

core/kazoo_services/src/kz_services.erl
```diff
@@ -573,7 +573,7 @@ public_json(#kz_services{jobj=ServicesJObj
                          ,cascade_quantities=CascadeQuantities
                         }) ->
     AccountId = kz_doc:account_id(ServicesJObj),
-    InGoodStanding = try maybe_follow_billling_id(AccountId, ServicesJObj) of
+    InGoodStanding = try maybe_follow_billing_id(AccountId, ServicesJObj) of
```
fenollp: BTW these typos have already been fixed. Changes to this file should be reverted (or PR rebased)

tsromana:
will do as suggested

k-anderson:
@tsromana Howdy, I just wanted to make sure that this ticket didn't deadlock ;)  As per the last comment I think you are still working on this but if you there is anything you need from us (or you believe this is ready for a re-review) then please let us know!

Thanks!

tsromana:
@k-anderson sorry for keeping this pr open, i didn't get the time to update it, i will push changes to the branch when i get home :) in few hours

k-anderson:
@tsromana No worries I wasn't rushing ya!  We were just reviewing the pending PRs and I wanted to make sure you were not waiting for us... we have had tickets in the past were both parties were waiting for eachother :)

tsromana:
@k-anderson no problem :) the pr is updated and can be merged, just a ques will there be changes in this, 
https://github.com/2600hz/kazoo/commit/f9f400565cb7f2195c136cabb2898cc83923633c
as the code does not get executed when a user just wants to update his info and not his cards, or it should be kept the same :) ( i currently work with banking domain applications so was thinking of use cases as what might a customer do)

k-anderson:
@tsromana Howdy, not sure I understood your question re: f9f4005.  At this time we don't have any active tickets in the braintree module if that helps.  Perhaps if you could restate it I can give you a better answer!  

Thanks! :)

PR #1899, opened 2016-05-11
Refactoring stepswitch_resources:get_endpoints/3

skorobkov:
Refactoring filtering resources by flags/rules/cid in `get_endpoints/3`.

I planing continue refactoring and adding features to stepswitch. so I have some questions:

1. Since we have formatters, maybe `rules` should be only for filtering number? And all number modification do only by formatters?
2. In our setup we have to route calls to different carriers by lists of prefixes (very long lists). This lists cant be fited in regexes. For now its solved with custom patch. where stepswitch search prefixes in ets tables, but i want make it more scalable and use couchdb for this. So what i want to do:

- Create system-wide DB "offnet_prefixes" (offnet_routing? offnet_... ?) and per account DB (place this prefixes in account DB or use separate DB?)
- Fill this DB via cb_resources (JSON request or upload CSV file)
- Each document in DB contain prefix, resource id, some "prefix-resource-weight" (if we put here price for this resource, then we have LCR routing)
- In stepswitch_resources add request to filter resources which have prefix in DB
- After filtering, we order endpoint with standart resource-wheight and prefix-carrier-weight

### DIFF
--- a/applications/stepswitch/src/stepswitch_resources.erl
+++ b/applications/stepswitch/src/stepswitch_resources.erl
@@ -76,6 +76,7 @@
            ,bypass_media = 'false' :: boolean()
            ,formatters :: api_objects()
            ,proxies = [] :: kz_proplist()
+           ,outbound_number :: api_binary()
          }).
 
 -type resource() :: #resrc{}.
@@ -152,7 +153,9 @@ endpoints(Number, OffnetJObj) ->
 -spec maybe_get_endpoints(ne_binary(), kapi_offnet_resource:req()) -> kz_json:objects().
 maybe_get_endpoints(Number, OffnetJObj) ->
     case kapi_offnet_resource:hunt_account_id(OffnetJObj) of
-        'undefined' -> get_global_endpoints(Number, OffnetJObj);
+        'undefined' -> 
+            lager:debug("attempting to find global resources"),
+            get_endpoints(get(), Number, OffnetJObj);
         HuntAccount -> maybe_get_local_endpoints(HuntAccount, Number, OffnetJObj)
     end.
 
@@ -167,22 +170,29 @@ maybe_get_local_endpoints(HuntAccount, Number, OffnetJObj) ->
             [];
         'true' ->
             lager:info("account ~s is using the local resources of ~s", [AccountId, HuntAccount]),
-            get_local_endpoints(HuntAccount, Number, OffnetJObj)
+            lager:debug("attempting to find local resources for ~s", [HuntAccount]),
+            get_endpoints(get(HuntAccount), Number, OffnetJObj)
     end.
 
--spec get_local_endpoints(ne_binary(), ne_binary(), kapi_offnet_resource:req()) -> kz_json:objects().
-get_local_endpoints(AccountId, Number, OffnetJObj) ->
-    lager:debug("attempting to find local resources for ~s", [AccountId]),
+-spec get_endpoints(resources(), ne_binary(), kapi_offnet_resource:req()) -> kz_json:objects().
+get_endpoints(Resources, Number, OffnetJObj) ->
+    CallerIdNumber = case ?RULES_HONOR_DIVERSION of
+                         'false' -> kapi_offnet_resource:outbound_caller_id_number(OffnetJObj);
+                         'true' -> check_diversion_fields(OffnetJObj)
+                     end,
     Flags = kapi_offnet_resource:flags(OffnetJObj, []),
-    Resources = filter_resources(Flags, get(AccountId)),
-    resources_to_endpoints(Resources, Number, OffnetJObj).
-
--spec get_global_endpoints(ne_binary(), kapi_offnet_resource:req()) -> kz_json:objects().
-get_global_endpoints(Number, OffnetJObj) ->
-    lager:debug("attempting to find global resources"),
-    Flags = kapi_offnet_resource:flags(OffnetJObj, []),
-    Resources = filter_resources(Flags, get()),
-    resources_to_endpoints(Resources, Number, OffnetJObj).
+    Routines = [fun(R) -> maybe_filter_resources_by_flags(Flags, R) end
+                ,fun(R) -> filter_resources_by_rules(Number, R) end
+                ,fun(R) -> filter_resources_by_cid_rules(CallerIdNumber, R) end
+               ],
+    FilteredResources = lists:foldl(
+                          fun(_F, []) -> [];
+                             (F, R) -> F(R)
+                          end
+                          ,Resources
+                          ,Routines
+                         ),
+    resources_to_endpoints(FilteredResources, Number, OffnetJObj).
 
 -spec sort_endpoints(kz_json:objects()) -> kz_json:objects().
 sort_endpoints(Endpoints) ->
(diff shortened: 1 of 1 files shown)


applications/stepswitch/src/stepswitch_resources.erl
```diff
@@ -152,7 +153,9 @@ endpoints(Number, OffnetJObj) ->
 -spec maybe_get_endpoints(ne_binary(), kapi_offnet_resource:req()) -> kz_json:objects().
 maybe_get_endpoints(Number, OffnetJObj) ->
     case kapi_offnet_resource:hunt_account_id(OffnetJObj) of
-        'undefined' -> get_global_endpoints(Number, OffnetJObj);
+        'undefined' -> 
+            lager:debug("attempting to find global resources"),
+            get_endpoints(get(), Number, OffnetJObj);
```
fenollp: Instead of just `get()`, wouldn't it be better to introduce a special key (e.g. `'global_endpoints'`) that contains a set of keys to endpoints? `get()` at L158 would become `sets:to_list(get('global_endpoints'))` (would need initialisation `put('global_endpoints', sets:new())`).

The intent is to not arbitrarily go through the whole process dictionary.

applications/stepswitch/src/stepswitch_resources.erl
```diff
@@ -152,7 +153,9 @@ endpoints(Number, OffnetJObj) ->
 -spec maybe_get_endpoints(ne_binary(), kapi_offnet_resource:req()) -> kz_json:objects().
 maybe_get_endpoints(Number, OffnetJObj) ->
     case kapi_offnet_resource:hunt_account_id(OffnetJObj) of
-        'undefined' -> get_global_endpoints(Number, OffnetJObj);
+        'undefined' ->
+            lager:debug("attempting to find global resources"),
+            get_endpoints(get(), Number, OffnetJObj);
```
fenollp: Instead of just `get()`, wouldn't it be better to introduce a special key (e.g. `'global_endpoints'`) that contains a set of keys to endpoints? `get()` at L158 would become `sets:to_list(get('global_endpoints'))` (would need initialisation `put('global_endpoints', sets:new())`).

The intent is to not arbitrarily go through the whole process dictionary.

skorobkov: This `get` is `stepswitch_resources:get/0` not `erlang:get/0`.
Should it be ranamed?

skorobkov: Renamed ad remove compile directive
```Erlang
-compile({'no_auto_import', [get/0, get/1]}).
```

fenollp: Oh okay thanks

k-anderson:
@skorobkov 

Howdy,

I have not looked at the code yet but I thought I would attempt to answer you questions first

* I like the idea of separating the concerns between formatting and rules.  However, when stepswitch was originally written there were no formatters and as such currently deployments rely on the rules to get the capture group.  I can not foresee a backward compatible way to decouple this.  There are a few things in every project that would be nice to do differently with hindsight but we can't because it would change existing behavior that the community relies on.

* We have discussed similar approaches for 'advanced resource selections', and would love to work with you on this.  At this time there is no formal design so if you would like to tackle it maybe we could establish some meetings to bounce ideas around (in IRC, or by phone)? In the meantime, to address the specifics brought up here:

 * We would not want to added it to the account db, but rather create a new DB per account, similar to the MODBs (like account/xx/xx/xxxxxxxxx-prefixes).  In 4.0 we have removed the last non-configuration changes to the account db which means almost everything is write once, making the system more preformant and reducing the need for compaction.  As far as the naming (offnet_prefixes, offnet_routing? offnet_... ?) that is the hardest part of any project ;)  I don't dislike prefixes but think that something more descriptive would be better as its likely to extended past simple prefix selections as you mentioned.  After a lot of thought I personally propose: resource_discriminators .... (I have a long standing wish to change the "offnet" db to "resources".  But thats another thing that remains as is for historical reasons.)

 * I like the concept, and agree that the discriminators db should be used to get a list of resource ids (and not redefine gateways as has been talked about in the community before).  As a matter of fact I would recommend NOT changing stepswitch_resources but rather creating a new stepswitch_discriminators module.  Then the inbound and outbound modules could conditionally (based on the db existence, system config, AMQP request, ect) choose to use the existing behavior by building the endpoints list from stepswitch_resources OR get the list form stepswitch_discriminators.  This ensures there is no backward incompatablity issue with the new functionality, and you can make any change you desire because everyone who didn't utilize the new feature would continue using an unmodified stepswitch_resources.  There is some overlap with converting the JSON to the endpoint object but that code could be shared in stepswitch_util (or some other better named module) or even stepswitch_discriminators uses exported functions of stepswitch_resources to create the endpoints.

 * This brings up an interesting idea..... ;) IF stepswitch_discriminators was modular then the documents in this db could be a collection of objects that map to stepswitch discriminator modules...  For example, if the document looked like:

```
{  
   "name":"SF low cost for customer-22",
   "discriminators":{  
      "dialed-number":{  
         "prefix":"415"
      },
      "source-ip":{  
         "ip":"10.26.0.88"
      }
   },
   "weights":{  
      "cost":{  
         "weight":15,
         "priority":5
      },
      "quality":{  
         "weight":6,
         "priority":1
      }
   },
   "resource":{  
      "id":"xxxxxxxxxxxxxxx",
      "override":{  
         "weights":true,
         "rules":true
      }
   },
   "pvt_type":"discriminator"
}
```

This is only partially thought out (hence the desire for maybe more live communication) but the initial idea here is the documents are chosen based on the discriminators object.  By making it an object of objects we can continue to expand the functionality there.  Then those that qualify are weighted.  The weights would work similar to SRV records so you could sort by quality then cost within that.  IE: start with all my lowest cost high quality carriers, then my lower quality low cost, ect ect.  This would be a very flexible system for anybody then ;)  Finally, this results in a list of resources to use sorted by weights/priority considerations and filtered by discriminators.  This is then converted to the endpoints array already generated by stepswitch_resources today and all is well!

 * I added resource.override to optionally keep the existing resource weights, I believe that if we implement what I have above then it is likely some admins will want to ignore the "global" weights and rules.  Instead using this new approach to find resources.  Others may want to extend the existing resource selection this way.  All use-cases are supported by optionally overriding the existing behavior.

** TO BE CLEAR ** This is just my initial thoughts, generally through discussion and other input we get to the final design.   Also I am not requesting you build all of this, for example if you agree with some of these points then we can start with just your prefix discriminator but structure the document to support future expansion and build new discriminators as requested.  ect ect.

Thoughts?

k-anderson:
@skorobkov One last thought to leave you with... If we look at the current capabilities in Kamailio perhaps these selections and manipulations could be done there?  Stepswitch would be responsible for loading the data in the appropriate kamailo server and routing to it.... IE: we add a kamailio role to handle outbound traffic, then using our Kamailio kazoo connection we provide native modules the data they need for this type of routing and simple bridge to the appropriate Kamailio server.

skorobkov:
Ok. I will try to make stepswitch_discriminators.
Objectives:

- separate DB for discriminators
- system-DB name: "resource_discriminators"
- user-DB name "account/xx/xx/xxxxxxxxx-discriminators"
- stepswitch_discriminators used only if `system_config/stepswitch/route_by_discriminators : true` (stepswitch_resources used only for fetching resources and build endpoints), otherwise used old stepswitch_resources behaviour
- stepswitch_discriminators modular, in folder "discriminators" we put modules with name "kz_ssd_{NAME}.erl"
- in config system_config/stepswitch.discriminators" list which modules to use (all by default)
- manage discriminators via cb_resources ( /v2/account/XXXXX/resources/XXXX/dicrimintaros/... )

--
> Stepswitch would be responsible for loading the data in the appropriate kamailo server and routing to it....

What benefits does it give?

skorobkov:
@k-anderson could you merge this PR?
I need modified version `stepswitch_resources:resource_to_endpoints/3` (which only convert, and not doing any filtering).

PR #1902, opened 2016-05-11
Listing only enabled UAC gateways

skorobkov:


k-anderson:
@skorobkov Howdy,

What is the use case here?  I thought that stepswitch_resources would filter based on this, could be wrong.  My concern is that property is not required and could be missing on some deployments, which would now be interpreted as 'false'.  That means this has the potential to surprisingly disable a lot of resources on update!  

I would be more comfortable if it checked for the explicit value of 'false'.  IE: the parameter must be present and set to a value of 'false' to skip it in the view.

Thoughts?

skorobkov:
As I see `resources/listing_uac_gateways` list gateways, which have `register:true`.And this list used by `ecallmgr_fs_config:default_sip_profiles/1`.
And current version check only if recource enabled, and list all gateways (enabled and disabled) in this reource.

lazedo:
@skorobkov ` !doc.gateways[index].enabled ` , if enabled is not present it is considered `false`. that may lead to undesired behaviour due to what @k-anderson wrote about `enabled`being required.
you can use something like ` !(doc.gateways[index].enabled || true) ` which would assume enabled if not present.

skorobkov:
Ok.

skorobkov:
Changed to explicit gatewayy have option `enable: false`

BTW @lazedo Sorry for assking you here. but could you look again to https://2600hz.atlassian.net/browse/KAZOO-4792 ?

applications/crossbar/priv/couchdb/views/resources.json
```diff
@@ -15,7 +15,7 @@
...
+            "map": "function (doc) { if (doc.pvt_type != 'resource' || doc.pvt_deleted || !doc.enabled) return; for (var index in doc.gateways) { if (!doc.gateways[index] || !doc.gateways[index].register || doc.gateways[index].enabled == 'false') { continue; } var gateway = doc.gateways[index]; emit(doc._id, {'username': gateway.username, 'auth_username': gateway.auth_username, 'server': gateway.server, 'password': gateway.password, 'realm': gateway.realm, 'expiration': gateway.expiration, 'proxy': gateway.proxy, 'account_id': doc.pvt_account_id, 'id': doc._id + '-' + index, 'resource_id': doc._id, 'register_extension': gateway.register_extension, 'version': doc._rev})}}"
```
lazedo: @skorobkov tried this and it doesn't work, `doc.gateways[index].enabled == 'false' `
it does work without ' like `doc.gateways[index].enabled == false`

skorobkov:
Out of habit, put quotes around atoms...
=\

lazedo:
@skorobkov thanks, please squash into 1 commit

PR #1907, opened 2016-05-12
Added Hold and Move schema

tsromana:


fenollp:
@tsromana Note that these are generated by `cb_api_endpoints:to_ref_doc/0` (writes to `doc/ref/`) and then copy pasted into the corresponding `doc/*.md` file.

If you want to make `cb_api_endpoints` (helps documentation and testing) here is a wishlist:

* Turn into CI-able escript (ie turn `cb_api_endpoints` into something runnable at compile time, not just run time)
* Find all endpoints (also query strings, filters & required content types) (e.g. missed endpoint in `cb_port_requests`: `/descendants/port_requests`)
* Enhance the camelCase to_underscore conversion
* Checks on copy-pasted & hand-completed docs:
    * Are all introduced variables used?
    * Are some variables misspelled? (similar to previous)
    * …

That would be tremendous!

tsromana:
got it will look into it. Thank you

PR #1913, opened 2016-05-13
Knm ports

fenollp:


applications/crossbar/doc/port_requests.md
```diff
@@ -1,262 +1,1030 @@
...
+```json
+{
+    "auth_token": "{AUTH_TOKEN}",
+    "data": {
+        "created": 63630107130,
+        "id": "{PORTREQUEST_ID}",
+        "name": "Porting 202.555.9000",
         "numbers": {
-            "{{NUMBER}}": {}
+            "+12025559000": {}
```
mark2600: Did you intend to actually have a number here ?

fenollp: Hm my rationale is: there is no way to know what will be here from the request alone, so data should stay intact here.

PR #1918, opened 2016-05-16
check code style

fenollp:


fenollp:
Woops, just noticed this check doesn't run properly yet!

> Symbol's function definition is void: cl-mapc

Need to work with Travis' default emacs which should be emacs23 or 22.

lazedo:
i don't agree with this

fenollp:
Travis will not support checking this soon according to https://github.com/travis-ci/travis-ci/issues/6076
Emacs' version is the culprit here.

This PR has to be rethought with one of these possible solutions (in no special order):
* https://github.com/tsloughter/erl_tidy
* https://github.com/jimenezrick/vimerl/blob/master/indent/erlang_indent.erl
* https://github.com/mbucc/erlfmt
* https://github.com/ebengt/erlang_stdin_formatter

Validation criteria:
* MUST be a compile time script
* MUST `style(style(Codebase)) === style(Codebase)`

An proper ticket should be written. Will get back to this at some point.

k-anderson:
@fenollp rebase me!

fenollp:
@k-anderson so this works but needs a freeze of all other PRs otherwise last commit will always need to be redone :)

PR #1919, opened 2016-05-16
KAZOO-4718: add caller and callee numbers to per_minute_call doc

onnet:
We need it to separate per minute calls costs for different invoices by caller and callee numbers combination. 
This data could be separately taken from cdr but it looks like it will be much better to obtain this information from per_minute_call doc without additional queries.


onnet:
Will be reworked after https://github.com/2600hz/kazoo/pull/1923 merge

jamesaimonetti:
@onnet merged #1923 FYI

PR #1924, opened 2016-05-17
KAZOO-4757: fix formatting to return numbers in e164

mark2600:


jamesaimonetti:
@mark2600 don't forget to issue a PR for master and 4.0 too

fenollp:
Actually 4.0 & master are both fine wrt this particular issue ;)

PR #1930, opened 2016-05-17
Kazoo 4750

Shoowa:
I replaced the raw binaries with a function that converts an iolist to a binary.


jamesaimonetti:
Doing this properly by getting tests to pass locally first.

PR #1932, opened 2016-05-18
KAZOO-4718: additional fields for per_minute_call doc

onnet:
https://github.com/2600hz/kazoo/pull/1919 continued


lazedo:
@onnet thanks!

PR #1945, opened 2016-05-19
KAZOO-4760 refactor callflow validation for 3.22

lazedo:


applications/crossbar/src/modules/cb_callflows.erl
```diff
@@ -176,34 +185,42 @@ load_callflow(DocId, Context) ->
...
+                                      ])
+                                   ,Context
+                                   ).
+
+-spec normalize_numbers(ne_binaries(), cb_context:context()) -> cb_context:context().
+normalize_numbers(Ns, Context) ->
+    normalize_numbers(Ns, cb_context:account_id(Context), Context).
+
+-spec normalize_numbers(ne_binaries(), ne_binary(), cb_context:context()) -> cb_context:context().
+normalize_numbers(Ns, 'undefined', Context) ->
```
fenollp: Same. Can the account be undefined though?

PR #1947, opened 2016-05-19
KAZOO-4171

fenollp:
* No persistence yet (should save to DB (where?); will see with CSV-saving phase)
* No direct subscribing to AMQP: uses `kz_globals`

Next step: 4172


core/kazoo_tasks/src/kz_task_worker.erl
```diff
@@ -0,0 +1,163 @@
...
+          fun () ->
+                  _ = kz_tasks:worker_running(TaskId, self()),
+                  try
+                      R = apply(M, F, A),
+                      _ = kz_tasks:worker_finished(TaskId),
+                      lager:debug("R = ~p. Should store this in MoDB", [R])
+                  catch
+                      _E:_R ->
+                          Stacktrace = erlang:get_stacktrace(),
+                          Msg = io_lib:format("(~p) ~p: ~w\n~p"
```
jamesaimonetti: why `\n` instead of `~n` ?

fenollp: That's what my fingers type automatically, also it's a standard escaped char instead of a special occurence from The Prolog Days™

fenollp:
Note: this PR is implementing a now defunct draft that is taken over by the KAZOO-4172 branch, coming to a PR near you soon :)

PR #1951, opened 2016-05-21
per-minute-voip ledgers doc: an attempt to add Rate-Description field. 

onnet:
Please consider to add rounded billing seconds and rate description to per-minute-voip ledger doc.

### DIFF
--- a/applications/jonny5/src/j5_per_minute.erl
+++ b/applications/jonny5/src/j5_per_minute.erl
@@ -149,6 +149,7 @@ create_ledger_usage(Amount, Request, Limits) ->
 metadata(Request) ->
     RateObj = kz_json:from_list(
                 [{<<"name">>, j5_request:rate_name(Request)}
+                 ,{<<"description">>, j5_request:rate_description(Request)}
                  ,{<<"value">>, j5_request:rate(Request)}
                  ,{<<"increment">>, j5_request:rate_increment(Request)}
                  ,{<<"minimum">>, j5_request:rate_minimum(Request)}
--- a/applications/ecallmgr/src/ecallmgr_fs_authz.erl
+++ b/applications/ecallmgr/src/ecallmgr_fs_authz.erl
@@ -17,7 +17,8 @@
 
 -define(RATE_VARS, [<<"Rate">>, <<"Rate-Increment">>
                     ,<<"Rate-Minimum">>, <<"Surcharge">>
-                    ,<<"Rate-Name">>, <<"Base-Cost">>, <<"Pvt-Cost">>
+                    ,<<"Rate-Name">>, <<"Rate-Description">>
+                    ,<<"Base-Cost">>, <<"Pvt-Cost">>
                     ,<<"Discount-Percentage">>, <<"Rate-NoCharge-Time">>
                    ]).
 
--- a/applications/jonny5/src/j5_channels.erl
+++ b/applications/jonny5/src/j5_channels.erl
@@ -67,6 +67,7 @@
                   ,discount_percentage :: api_binary() | '_'
                   ,surcharge :: api_binary() | '_'
                   ,rate_name :: api_binary() | '_'
+                  ,rate_description :: api_binary() | '_'
                   ,rate_id :: api_binary() | '_'
                   ,base_cost :: api_binary() | '_'
                  }).
@@ -423,6 +424,7 @@ to_props(#channel{call_id=CallId
                   ,discount_percentage=DiscountPercentage
                   ,surcharge=Surcharge
                   ,rate_name=RateName
+                  ,rate_description=RateDescription
                   ,rate_id=RateId
                   ,base_cost=BaseCost
                  }) ->
@@ -444,6 +446,7 @@ to_props(#channel{call_id=CallId
        ,{<<"Discount-Percentage">>, DiscountPercentage}
        ,{<<"Surcharge">>, Surcharge}
        ,{<<"Rate-Name">>, RateName}
+       ,{<<"Rate-Description">>, RateDescription}
        ,{<<"Rate-ID">>, RateId}
        ,{<<"Base-Cost">>, BaseCost}
       ]
@@ -534,6 +537,7 @@ handle_cast({'rate_resp', JObj}, State) ->
                ,{#channel.discount_percentage, kz_json:get_value(<<"Discount-Percentage">>, JObj)}
                ,{#channel.surcharge, kz_json:get_value(<<"Surcharge">>, JObj)}
                ,{#channel.rate_name, kz_json:get_value(<<"Rate-Name">>, JObj)}
+               ,{#channel.rate_description, kz_json:get_value(<<"Rate-Description">>, JObj)}
                ,{#channel.rate_id, kz_json:get_value(<<"Rate-ID">>, JObj)}
                ,{#channel.base_cost, kz_json:get_value(<<"Base-Cost">>, JObj)}
               ]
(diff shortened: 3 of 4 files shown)


lazedo:
@onnet this defeats the purpose of ledgers simple api. please do not create extra values like `charged_seconds`, there's already a quantity for that.
also move `calculate_seconds` from  `wht_util` to `j5_per_minute`.
remove kazoo_ledger changes.
i don't see `Amount` being affected by `charged_seconds`, what am i missing ?

onnet:
@lazedo thanks for your comment. 

I am afraid I don't understand your comment in full, therefore here are some comments to describe my intentions

IMHO:

I believe ledgers doc should contain rounded billing seconds as a usage->quantity.

From my point of view per-minute-voip ledgers could be also good for per minute calls reports, just because it contains only charged calls and charged calls are the only calls accountants care about :) 
And accountants prefer to see correlation between minutes, rate and amount in their reports.

On the other hand, if someone would like to know real call duration, it could be taken from cdr.

To not to introduce backward incompatibility and/or break things I don't understand, I just added "usage->charged_quantity" along with already existing "quantity".

Amount is already calculated based on rounded billing seconds (charged_seconds/quantity/... whatever it named)
No need to affect it.

Actually, since we've already got rate_increment and rate_minimum in rate object, rounded billing seconds could be calculated on client's side (only Rate-NoCharge-Time needed).

Since I'm afraid I didn't understand you correctly, could you please specify whether you want me to replace quantity with rounded value or just abandon the idea of populating ledgers doc with rounded billing seconds value.


There is also another field I wanted populate ledgers doc with - <<"Rate-Description">>.
Could you please also comment whether it goes against the intention of ledgers simple api or not.

Thanks in advance.

lazedo:
@onnet Hi. you should populate `quantity` with `charged_seconds`, no need to add extra stuff to ledgers.
as for the <<"Rate-Description">>, why do you need this ? the concern is that you're going to populate Channel-Vars with more data, isn't Rate-Name enough ?

onnet:
@lazedo

Thanks for doubled efforts to make me copy your comment :)
It seems everything is clear now.

Regarding Rate-Description:

I accept Rate-Name as an exact prefix description (even closer to definition) - for example "Russia-Megafon-79219"
At the same time, Rate-Description could be used as an aggregate name of a bunch of prefixes - for example "Russia, Saint-Petersburg, Megafon carrier"

Another use case: 
- for easier debugging I can use Rate-Name as a description in Latin-1 
- for internationalization purposes I can populate Rate-Description fields in UTF-8

And getting back to mentioned above call reports, I can use Rate-Description as a localized and human readable call direction name in these reports.

Regards,
Kirill

applications/jonny5/src/j5_per_minute.erl
```diff
@@ -161,3 +168,18 @@ metadata(Request) ->
...
+calculate_seconds(NoChargeTime, Increment, Minimum, BillingSecs) ->
+    case BillingSecs =< 0 of
+        'true' -> 0;
+        'false' when BillingSecs =< NoChargeTime ->
+            0;
+        'false' when BillingSecs =< Minimum ->
+            Minimum;
+        'false' ->
+            kz_util:ceiling( BillingSecs / Increment ) * Increment
+    end.
```
lazedo: @onnet hi, this seems wrong to me, or i'm missing something. i think `wht_util:call_cost` may also be wrong.
as an example, if BillingSecs = 23 and NoChargeTime = 10 and Increment = 1, shouldn't the result of `calculate_seconds` be 13 ?

lazedo:
@onnet i still think Rate-Description adds unnecessary overhead to the message bus. you can do the same with Rate-ID and Rate-Name

applications/jonny5/src/j5_per_minute.erl
onnet: Hi @lazedo 

AFAIK NoChargeTime is a kind of a protection for customer against wrong (or maybe accidentally) dialed number or any other problem.
It's a minimal number of seconds needed to understand that something is wrong and hangup because of that. So, if your dialed number was wrong and you was quick enough to hangup you would pay nothing, if it was a normal call you pay in full.
So, from this point of view, calculations are correct.
In your example 23 seconds call should be belled as 23 seconds call, but 9 seconds call should be 0.

Regards,

onnet:
@lazedo 

Regarding <<"Rate-Description">>

I certainly can not argue with you, and will abandon this part of PR if you'll insist, but imho ...

-define(RATE_VARS, [<<"Rate">>, <<"Rate-Increment">>
                    ,<<"Rate-Minimum">>, <<"Surcharge">>
                    ,<<"Rate-Name">>, <<"Base-Cost">>, <<"Pvt-Cost">>
                    ,<<"Discount-Percentage">>, <<"Rate-NoCharge-Time">>
                   ]).

..it looks like Rate-ID is also absent.
This means, going your way, one more field should be also added anyway.
With that we returning back to the same overload except <<"Rate-Description">>'s size could be slightly bigger.
But I also believe no one will put novels into Rate-Description. 
Rate-Description is usually rephrased, but almost the same size as Rate-Name.
As maximum it could be twice as big as Rate-Name. 
 
Besides that, if I remember correctly, Rate-ID can't be specified on bulk rate upload. 
It is created based on Rate-Name, which make sense at least because it makes work with ratedeck in Futon convenient.

applications/jonny5/src/j5_per_minute.erl
lazedo: @onnet thanks for the explanation

lazedo:
@onnet makes sense. can you squash your commits ?
i created [KAZOO-4787](https://2600hz.atlassian.net/browse/KAZOO-4787) to refactor the use of these vars.

onnet:
@lazedo will do and thanks for help and spending time for this

lazedo:
@onnet i'm ok to accept this when you squash the commits

onnet:
@lazedo Hi! I already squashed this PR (Certainly, if I correctly understood what you want)

https://github.com/2600hz/kazoo/pull/1968 - this one is rounding billing seconds. Please accept it firstly.

This (current) one consists of what left for <<"Rate-Description">>. If rebase will be needed, I'll do it after you accept https://github.com/2600hz/kazoo/pull/1968

Thanks in advance,

lazedo:
@onnet squash the 5 commits into 1.
something like..
```
git rebase -i HEAD~5
pick
fixup
fixup
fixup
fixup
:wq
git push -f

```

onnet:
@lazedo done ans thanks for explanation!

lazedo:
@onnet sorry if i didn't explain my self. this pr was ok with only the change to Rate-Description, this was after you separated into 2 different PRs. what i asked was , instead of having 5 commits without relevant changeset in each and with the descriptions like `WIP` or `fixup`, we prefer to squash the commits into one.
1st commit "KAZOO-1234 add some cool new feature"
2nd commit "fix error"
3rd commit  "makeup"
...
git rebase allows you to rework your commits.
the sample i gave you before was to squash the commits in your pr at that time.
i see that you now have put back the other changes that i'm not ok to accept now.
can you please, make this pr to change only Rate-Description ?
thanks for your work

onnet:
@lazedo Yep, no problem, will rerevert it back and check calculation at https://github.com/2600hz/kazoo/pull/1968, but a bit later - it's sauna time, kids are waiting :)). 
Thanksfor your help!

onnet:
@lazedo Hi, this one rereverted and squashed. Regards,

lazedo:
@onnet thanks

PR #1955, opened 2016-05-23
sync support (online re-compile)

jamhed:
Allow to dynamically recompile Erlang source code.

Usage:
#!/bin/sh
export KAZOO_CONFIG=$HOME/kazoo/config.ini
make build-dev-release
RELX_REPLACE_OS_VARS=true KZname='-name kazoo@localhost' _rel/kazoo/bin/kazoo console

(kazoo@localhost)1> sync:go().

### DIFF
--- a/core/kazoo/src/kz_rel.erl
+++ b/core/kazoo/src/kz_rel.erl
@@ -1,0 +1,6 @@
+-module(kz_rel).
+-export([start/0]).
+
+start() ->
+	{ok, Cwd} = file:get_cwd(),
+	code:add_path(Cwd).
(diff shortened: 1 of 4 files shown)


fenollp:
There is reloader already, with `scripts/dev-start-apps.sh` which need a `make build-release` beforehand

jamhed:
This one is different: it recompiles *.erl files on the fly and reloads them. So you just start REPL and it automatically compiles and edit beams as you edit the source in editor.

fenollp:
Reloader does exactly this

jamhed:
@fenollp The doc says: it will monitor every loaded Erlang module's BEAM file for changes, and following example says: 
After recompiling `nuclear_ant_server.erl`:
    Reloading nuclear_ant_server ... ok.

So reloader does a bit different thing: it monitors the *.beam, while sync monitors the *.erl, so you don't need to run compile on file change.

PR #1956, opened 2016-05-23
refactor: knm.hrl from include to include_lib

jamhed:


lazedo:
@jamhed why ?

fenollp:
Yeah knm.hrl is in src/ so that outside apps don't try to include it.

jamhed:
@fenollp okay, I put it back to src, but leave include_lib. Will this work? :)

lazedo:
@jamhed but why are changing this ? what's not working for you to make the change ?

jamhed:
@lazedo, to make sync work, see https://github.com/2600hz/kazoo/pull/1955. It fails to recompile on the fly files with include because it doesn't know where to look for (you need to specify the include path as compiler option), and with include_lib it knows.

fenollp:
No include_lib won't work for things in src. that's the point

jamhed:
@fenollp it will work with src like this: include_lib("kazoo_number_manager/src/knm.hrl").

PR #1960, opened 2016-05-24
KAZOO-4706: treat empty wdays as all wdays

jamesaimonetti:


applications/callflow/src/module/cf_temporal_route.erl
```diff
@@ -13,8 +13,8 @@
 %%%-------------------------------------------------------------------
 -module(cf_temporal_route).
 
--include("callflow.hrl").
--include("cf_temporal_route.hrl").
+-include("../callflow.hrl").
+-include("./cf_temporal_route.hrl").
```
fenollp: Why change this?

applications/callflow/src/module/cf_temporal_route.erl
```diff
@@ -747,6 +747,15 @@ to_dow(<<"friday">>) -> 5;
 to_dow(<<"saturday">>) -> 6;
 to_dow(<<"sunday">>) -> 7.
 
+-spec to_wday(wh_daynum()) -> wday().
```
fenollp: `s/wh_/kz_/`

PR #1962, opened 2016-05-24
refactor: sync-compatible changes

jamhed:
use include_lib instead of include


fenollp:
Seems to me sync is missing include paths. How about setting `ERL_LIBS`?

lazedo:
@fenollp is there anything left to do here ?

jamhed:
@lazedo, please hold it, i have more to check.

PR #1963, opened 2016-05-24
sync support (online re-compile)

jamhed:
To recompile project's *.erl files (and then reload *.beam) automatically on changes.
Release must be built with symlinks (make build-dev-release).

In erlang shell:
> lager:set_loglevel(lager_console_backend, info).
> sync:go().

Then edit and save a file (actually just save will do):
> 18:26:38.254 [info] ../../core/kazoo_apps/src/kz_nodes.erl:0: Recompiled.

### DIFF
--- a/rel/sys.config
+++ b/rel/sys.config
@@ -1,4 +1,10 @@
 [
+ {sync, [
+        {src_dirs, {replace, [
+				{"../../applications", []},
+				{"../../core", []}
+			]}}
+ ]},
  {lager, [
           {handlers, [
                       {lager_console_backend, info}
--- a/make/deps.mk
+++ b/make/deps.mk
@@ -1,7 +1,7 @@
 DEPS = lager eiconv gen_smtp amqp_client cowboy jesse jiffy certifi couchbeam wsock zucchini \
        erlsom erlydtl exml escalus folsom detergent erlang_localtime \
        nklib gproc poolboy syslog lager_syslog eflame hep ecsv reloader \
-       proper recon getopt
+       proper recon getopt sync eunit
 
 BUILD_DEPS = parse_trans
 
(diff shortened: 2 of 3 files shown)


core/kazoo/src/kz_rel.erl
```diff
@@ -0,0 +1,6 @@
+-module(kz_rel).
+-export([start/0]).
+
+start() ->
+	{ok, Cwd} = file:get_cwd(),
+	code:add_path(Cwd).
```
fenollp: Could you instead update dev-start-apps.sh? This feels like something that could be done with just L8 from said file

jamhed: Sure.

fenollp:
Sync sounds much more interesting than reloader. How about replacing reloader with sync?

rel/sys.config
```diff
@@ -1,4 +1,10 @@
 [
+ {sync, [
+        {src_dirs, {replace, [
+				{"../../applications", []},
```
fenollp: Compile options can be found in kz.mk

jamhed: It's not a compile option, it's a run time sync configuration.

fenollp: I was refering to the empty lists

fenollp: Also I'm not sure enabling sync on prod is a good idea for now?

jamhed: It is not enabled by default until you manually start it with sync:start() or sync:go(). What you see here is a mere app config, that could or could not be used.

jamhed: What empty lists are you taling about?

fenollp: Good then :) 
The empty lists at `{"../../applications", []}` and the core one. Reading the `sync` doc I understood that's where compile options should be added. Maybe `sync` gets the compilation options from the BEAM chunks already? (That would be great)

jamhed: @fenollp, nothing needs to be there at the moment.

PR #1964, opened 2016-05-24
KAZOO-4777: quickcheck model for kz_globals

jamesaimonetti:


core/kazoo_globals/src/kz_global_proxy.erl
```diff
@@ -0,0 +1,84 @@
...
+terminate(_Reason, _Global) ->
+    lager:debug("global proxy ~p down: ~p", [_Global, _Reason]).
+
+code_change(_OldVsn, Global, _Extra) ->
+    {'ok', Global}.
+
+-spec amqp_send(kz_global:global(), term()) -> 'ok'.
+amqp_send(Global, Message) ->
+    Payload = [{<<"Name">>, kz_global:name(Global)}
+              ,{<<"Message">>, kapi_globals:encode(Message)}
```
fenollp: Instead of using base64 can we stick with JSON? Isn't amqp_client encoding anyway?

core/kazoo_globals/test/kz_globals_pqc.erl
```diff
@@ -0,0 +1,292 @@
...
+        {'value', ?REG(Name, _P, 'local')} -> 'true';
+        _ -> 'false'
+    end;
+precondition(#{?LOCAL_NODE := Registry}
+             ,{'call', 'kz_globals', 'unregister_name', [Name]}
+            ) ->
+    case lists:keysearch(Name, 1, Registry) of
+        {'value', ?REG(Name, _Pid, 'local')} -> 'true';
+        _ -> 'false'
+    end;
```
fenollp: Why only run should-be-successful functions?
Also why not use `name_exists/2`?

core/kazoo_globals/test/kz_globals_pqc.erl
```diff
@@ -0,0 +1,292 @@
...
+    compare_names(Registry, Names);
+compare_names([?REG(_N, _, _)|_Reg], _Names) ->
+    lager:debug("failed to find ~p in ~p (~p)", [_N, _Names, _Reg]),
+    'false'.
+
+remote_register(Remote, Name, Pid) ->
+    Payload = [{<<"Name">>, Name}
+              ,{<<"State">>, 'pending'}
+              ,{<<"Node">>, remote_node_name(Remote)}
+               | kz_api:default_headers(<<"testing">>, <<"4.0.0">>)
```
fenollp: You could use ?APP_VERSION here and 2 other times below

core/kazoo_globals/src/kz_global_proxy.erl
lazedo: @fenollp we need to encode the msg that is going to other node, term_to_binary may contain invalid chars for json.

fenollp: I don't get it. Why use term_to_binary at all? JSON is already accepted.
Why transform json to binary then to base64 if it's just to put it into a json payload?

lazedo: @fenollp how would you pass something like `{'$gen_cast', {'what', [123, 456], <<"qwerty">>}}` as a message to the other node ?

fenollp: Ah! Right! :D

core/kazoo_globals/src/kz_globals.erl
```diff
@@ -412,70 +425,149 @@ maybe_add_zone(Zone, #state{zones=Zones}) ->
...
+    T2 = kz_global:timestamp(Remote),
+    N1 = kz_global:node(Local),
+    N2 = kz_global:node(Remote),
+    case T1 of
+        T1 when N1 =:= N2 -> 'no';
+        T1 when T1 < T2 -> 'no';
+        T1 when T1 > T2 -> 'yes';
+        T1 when T1 =:= T2 andalso N1 < N2 -> 'no';
+        T1 when T1 =:= T2 andalso N1 > N2 -> 'yes'
+    end.
```
fenollp: Is there a way to rewrite that?

lazedo: try it

jamesaimonetti: @lazedo i was planning on addressing this in [KAZOO-4783](https://2600hz.atlassian.net/browse/KAZOO-4783) since it was out of scope of this PR. Can we revert the commit here and address it in the new ticket?

lazedo: @jamesaimonetti accepted as a temp fix ? it solved my issues starting fax at same time in 3 VMs.

core/kazoo_globals/test/kz_globals_pqc.erl
fenollp: So how about removing some of these preconditions?

jamesaimonetti: The nodes won't unregsiter_name if they haven't registered the name already. These conditions check the node's registry to see if calling unregister_name makes sense.

core/kazoo_globals/src/kz_globals.erl
jamesaimonetti: Fair enough

PR #1970, opened 2016-05-25
Fix typo

Shoowa:
Change 'lib' to 'local'

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -28,7 +28,7 @@ $ chmod a+x kerl
 $ mv kerl /usr/bin
 $ kerl list releases
 $ kerl build 18.2 r18.2 # this takes a while
-$ kerl install r18.2 /usr/lib/erlang
+$ kerl install r18.2 /usr/local/erlang
 $ . /usr/lib/erlang/activate
 ```
 


doc/installation.md
```diff
@@ -28,7 +28,7 @@ $ chmod a+x kerl
 $ mv kerl /usr/bin
 $ kerl list releases
 $ kerl build 18.2 r18.2 # this takes a while
-$ kerl install r18.2 /usr/lib/erlang
+$ kerl install r18.2 /usr/local/erlang
```
lazedo: @Shoowa why is this a typo ?

icehess: @lazedo it would conflict with erlang installed by os package manager. I would prefer to install kerl erlang to /opt/erlang/{VERSION} though.

jamhed: I wouldn't recommend to install erlang to /usr using kerl, it will definitely confuse the system admin. It's better to keep it at home.

jamesaimonetti: I'll offer a dissenting opinion - I think OS packages and kerl should be mutually exclusive. I never install OS packages so putting kerl-based builds in the "normal" location makes sense to me. The installation instructions are for setting up a development machine with which to hack on Kazoo. I would again encourage using kerl to install latest and greatest Erlang version.

Remember, this guide is for getting up and running somewhat quickly, not for covering "advanced" scenarios where you need kerl + OS packaged Erlang or multiple versions of kerl-installed Erlang versions.

jamhed: In this case a system administrator must know not to install OS packaged Erlang, and not to install any OS package having Erlang in dependencies. This actions are highly unprobable though, but eventually could happen. However if we use our own Erlang this couldn't happen even in theory, and both ways are equally quick. Just my 2 cents.

lazedo: i'm with @jamesaimonetti on this, its a dev thing. the released package already runs without the need for erlang, it doesn't use it even if installed.

k-anderson:
The /usr/local hierarchy is for use by the system administrator when installing software locally. It needs to be safe from being overwritten when the system software is updated. It may be used for programs and data that are shareable among a group of hosts, but not found in /usr.

IE: the erlang package installs into /usr but we should be compiling and installing (via kerl) to /usr/local/

PR #1971, opened 2016-05-25
KAZOO-4735: api for move voicemail messages to another vmbox

icehess:


lazedo:
@icehess haven't test yet but, shouldn't update_mwi be called for both the old and the new vmbox ?

icehess:
I passed both old and new owner_id to update_mwi and handling that there,

`update_mwi(C, [OldOwnerId, NewOwnerId], 'undefined')`

lazedo:
@icehess yesw, i see that, thanks. what about the subscriptions to vmbox number not necessarily by owner_id ?

icehess:
@lazedo What do you mean by not necessarily by owner_id?

lazedo:
@icehess you're updating mwi based on the owner_id, ie, updates mwi on devices owned by owner_id with mwi totals for owner_id independent of vmbox user has. there may also exist subscriptions to the vmbox number by devices that do not belong to the owner_id.  a use case would be a web dashboard javascript using some sip library, using SUBSCRIBE/NOTIFY to display vmboxes with counters and getting display updated as vmboxes are updated.

icehess:
@lazedo so that means I also should send an api message to ecallmgr_fs_notify for sending mwi update based in vmbox number too?

We didn't to that before, we just was using `cb_modules_util:update_mwi` which calls `cf_util:unsolicited_owner_mwi_update`

lazedo:
@icehess cf_voicemail does this, we should do the same in crossbar

icehess:
Sure, I’ll do it

icehess:
@lazedo I updated mwi stuff, let me know what you think about it.

applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -1793,4 +1793,4 @@ send_mwi_update(New, Saved, BoxNumber, Call) ->
                | kz_api:default_headers(?APP_NAME, ?APP_VERSION)
               ],
     lager:debug("updating MWI for vmbox ~s@~s (~b/~b)", [BoxNumber, Realm, New, Saved]),
-    kz_amqp_worker:cast(Command, fun kapi_presence:publish_mwi_update/1).
+    kapps_util:amqp_pool_send(Command, fun kapi_presence:publish_mwi_update/1).
```
lazedo: why this change ?

icehess: cf_util uses that, just wanted to make it consistence.

lazedo: i would prefer changing in `cf_util`. `kz_amqp_worker` has more options then `kapps_util:amqp..` .
don't get me wrong, `cast` is ok but `call` and `call_collect` are much more flexible in `kz_amqp_worker` and the kapps_util:amqp_...` just calls kz_amqp_worker so we better use it directly.

PR #1972, opened 2016-05-25
rename all remaining cf_attributes to kz_ prefix

icehess:


applications/crossbar/src/modules/cb_channels.erl
```diff
@@ -254,8 +254,8 @@ user_endpoints(Context, UserId) ->
     Options = [{'key', [UserId, <<"device">>]}
                ,'include_docs'
               ],
-    %% TODO: Using the cf_attributes from crossbar isn't exactly kosher
-    Context1 = crossbar_doc:load_view(<<"cf_attributes/owned">>, Options, Context),
+    %% TODO: Using the kz_attributes from crossbar isn't exactly kosher
```
jamesaimonetti: Can probably remove this now!

applications/crossbar/src/modules_v1/cb_users_v1.erl
```diff
@@ -245,8 +245,8 @@ validate(Context, UserId, ?CHANNELS) ->
     Options = [{'key', [UserId, <<"device">>]}
                ,'include_docs'
               ],
-    %% TODO: Using the cf_attributes from crossbar isn't exactly kosher
-    Context1 = crossbar_doc:load_view(<<"cf_attributes/owned">>, Options, Context),
+    %% TODO: Using the kz_attributes from crossbar isn't exactly kosher
```
jamesaimonetti: Remove

PR #1973, opened 2016-05-25
KAZOO-4790: make toll-free determination more restrictive

mark2600:


core/whistle_number_manager-1.0.0/src/carriers/wnm_bandwidth2.erl
```diff
@@ -44,6 +44,26 @@
...
+    Prefix == <<"881">> orelse
+    Prefix == <<"882">> orelse
+    Prefix == <<"883">> orelse
+    Prefix == <<"884">> orelse
+    Prefix == <<"885">> orelse
+    Prefix == <<"886">> orelse
+    Prefix == <<"887">> orelse
+    Prefix == <<"888">> orelse
+    Prefix == <<"889">>
+).
```
fenollp: Is that from 3.22? I had done some research to figures these out there.
Are you also making a PR for master? I'd like to add tests for @k-anderson 's gist too.

mark2600: Yes this is for 3.22 as the defect was reported against it.  I will open a second PR for master shortly.

PR #1975, opened 2016-05-26
KAZOO-4097 fax outbound refactoring

lazedo:


applications/fax/src/fax_jobs.erl
```diff
@@ -85,9 +124,36 @@ handle_call(_Request, _From, State) ->
...
+handle_cast({'job_status',{JobId, <<"end">>, ServerId}}, #state{jobs=#{pending := Pending
+                                                            ,running := Running
+                                                            ,numbers := Numbers
+                                                            }=Jobs
+                                                       }=State) ->
+    #{JobId := #{queue := ServerId
+                ,number := Number
+                }
+     } = Running,
+    {'noreply', State#state{jobs=Jobs#{pending => maps:remove(JobId, Pending)
```
fenollp: Why have a map of jobs instead of a record key per job type?

applications/fax/src/fax_sup.erl
```diff
@@ -33,12 +33,12 @@
 
 -define(CHILDREN, [?WORKER('fax_init')
                    ,?CACHE_ARGS(?CACHE_NAME, ?CACHE_PROPS)
-                   ,?SUPER('fax_worker_pool_sup')
                    ,?SUPER('fax_requests_sup')
                    ,?SUPER('fax_xmpp_sup')
-                   ,?WORKER('fax_jobs')
+                   ,?SUPER('fax_jobs_sup')
+                   ,?SUPER('fax_worker_sup')
```
fenollp: Shouldn't the worker sup be started by the jobs sup?

applications/fax/src/fax_jobs.erl
lazedo: job type ? the maps is to avoid ets table and for quick lookups

applications/fax/src/fax_sup.erl
lazedo: no

PR #1986, opened 2016-05-27
removed whapps reference

tsromana:
There are still whapp variable left inside the kazoo_apps and jonny5 should we remove those too ???
the wh_to_kz.sh script does not contain the case when the letters are WhApp


tsromana:
the wh_to_kz script does not contain the case when the words are WhApp, there are still variable reference inside kazoo_apps and and jonny5

fenollp:
Yes. And there are more to catch!
`git grep -InE Wh'[A-Z]'` should help.
wh_to_kz.sh does not need to be updated however.

lazedo:
@tsromana thanks.

PR #1993, opened 2016-05-31
KAZOO-4794 federate notifications

jamhed:
federate notifications bindings for notify and teletype if
federate option is specified for notify application in system_config document.

### DIFF
--- a/core/kazoo/include/kz_config.hrl
+++ b/core/kazoo/include/kz_config.hrl
@@ -1,0 +1,6 @@
+-ifndef(KZ_CONFIG_HRL).
+
+-define(FEDERATE_BINDING(App), case kapps_config:get_is_true(App, <<"use_federated_listener">>, 'false') of 'true' -> ['federate']; _ -> [] end).
+
+-define(KZ_CONFIG_HRL, true).
+-endif.
--- a/core/kazoo_config/src/kapps_config.erl
+++ b/core/kazoo_config/src/kapps_config.erl
@@ -6,6 +6,7 @@
 %%%   Karl Anderson
 %%%   James Aimonetti
 %%%   Pierre Fenoll
+%%%   Roman Galeev
 %%%-------------------------------------------------------------------
 -module(kapps_config).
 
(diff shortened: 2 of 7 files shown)


core/kazoo/include/kz_config.hrl
```diff
@@ -0,0 +1,6 @@
+-ifndef(KZ_CONFIG_HRL).
+
+-define(FEDERATE_BINDING, case kz_config:get_boolean('amqp', 'use_federation') of 'true' -> ['federate']; _ -> [] end).
```
k-anderson: @jamhed howdy, two notes.  

1) I would expect this parameter to live in the system_config database but (unless I am mistaken) I think you are pulling it from the ini file.  
2) This parameter appears to enable/disable federation globally by its naming but it should be specific to the notifications.  In system_config we have a document (notify) for the notify app and another notifications for the teletype app.  I would be ok with either using one parameter for both (in which case it should be on the notification document) or on parameter per app.  In this case since the context would be on the system_config/notification document then 'use_federation' would be an acceptable parameter name.

If you want we can discuss this further in the standup tomorrow!

core/kazoo_config/src/kapps_config.erl
```diff
@@ -15,6 +16,7 @@
 -export([get_string/2, get_string/3, get_string/4]).
 -export([get_binary/2, get_binary/3, get_binary/4]).
 -export([get_atom/2, get_atom/3, get_atom/4]).
+-export([get_boolean/2, get_boolean/3, get_boolean/4]).
```
lazedo: get_is_true / get_is_false already exist. i think this is not needed.

core/kazoo/include/kz_config.hrl
```diff
@@ -0,0 +1,6 @@
+-ifndef(KZ_CONFIG_HRL).
+
+-define(FEDERATE_BINDING(App), case kapps_config:get_boolean(App, <<"use_federated_listener">>) of 'true' -> ['federate']; _ -> [] end).
```
lazedo: use kapps_config:get_is_true/2

lazedo:
@jamhed please create PR for 4.0 and 3.22

PR #1998, opened 2016-05-31
Kazoo 4701 4.0

jamesaimonetti:


core/kazoo_apps/src/kapps_call.erl
```diff
@@ -665,10 +685,17 @@ callee_id_number(#kapps_call{callee_id_number=CIDNumber}) -> CIDNumber.
...
+                     ,request_user=to_e164(RequestUser)
                      ,request_realm=RequestRealm
                     }.
 
+-ifdef(TEST).
+to_e164(Number) -> Number.
+-else.
+to_e164(Number) ->
+    knm_converters:normalize(Number).
+-endif.
```
fenollp: Why is this necessary?

jamesaimonetti: at least with the wnm_util version, it queried the db to get the regex. We aren't testing to_e164 so skip that check

PR #2010, opened 2016-06-02
KAZOO-4506_4.0: update number to cf_capture_group in cf_dynamic_cid for 4.0

icehess:
also update document as well


icehess:
merge me please

PR #2013, opened 2016-06-02
KAZOO-4506_master: update number to cf_capture_group in cf_dynamic_cid for master

icehess:
also update document as well


icehess:
merge me please

PR #2020, opened 2016-06-03
better dev start/sup scripts

jamhed:


scripts/dev/kazoo.sh
```diff
@@ -0,0 +1,28 @@
...
+
+CMD=$1
+if [ "$CMD" = "" ]
+then
+	CMD=console
+else
+	shift
+fi
+
+RELX_REPLACE_OS_VARS=true KZname="-sname $KAZOO_NODE" _rel/kazoo/bin/kazoo $CMD "$*"
```
lazedo: don't use short names

scripts/dev/kazoo.sh
```diff
@@ -0,0 +1,28 @@
...
+fi
+
+if [ ! -f $KAZOO_CONFIG ]; then
+	echo "Specified kazoo config doesn't exists, please provide readable kazoo config file"
+	exit
+fi
+
+if [ "$KAZOO_NODE" = "" ]
+then
+	KAZOO_NODE=kazoo@localhost
```
lazedo: remove @localhost

scripts/dev/sup.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+if [ "$KAZOO_CONFIG" = "" ]
+then
+	export KAZOO_CONFIG=$PWD/config.ini
+fi
+
+if [ "$KAZOO_NODE" = "" ]
+then
+	KAZOO_NODE=kazoo@localhost
```
lazedo: remove @localhost

scripts/dev/sup.sh
```diff
@@ -0,0 +1,23 @@
...
+fi
+
+IFS='@' read -r -a NH <<< "$KAZOO_NODE"
+NAME="${NH[0]}"
+HOST="${NH[1]}"
```
lazedo: why this ?

jamhed: To supply sup with NAME and HOST deduced from KAZOO_NODE, to make it talk to this node.

lazedo: and why the need for that ?

scripts/dev/kazoo.sh
```diff
@@ -0,0 +1,28 @@
...
+fi
+
+if [ ! -f $KAZOO_CONFIG ]; then
+	echo "Specified kazoo config doesn't exists, please provide readable kazoo config file"
+	exit
+fi
+
+if [ "$KAZOO_NODE" = "" ]
+then
+	KAZOO_NODE=kazoo@$(hostname)
```
lazedo: Default mode should be kazoo_apps

jamesaimonetti:
@lazedo happy with the changes @jamhed has made now? Want to merge?

scripts/dev/sup.sh
```diff
@@ -0,0 +1,16 @@
...
+: ${KAZOO_NODE=kazoo_apps@$(hostname)}
+echo Connecting to node: $KAZOO_NODE
+
+IFS='@' read -r -a NH <<< "$KAZOO_NODE"
+NAME="${NH[0]}"
+HOST="${NH[1]}"
+
+# request can execute longer than 5 seconds, so timeout 10 is required
+
+RELX_REPLACE_OS_VARS=true KZname="-name $KAZOO_NODE" \
```
fenollp: `exec` needs to be prepended here so as to propagate the return code

scripts/dev/sup.sh
```diff
@@ -0,0 +1,16 @@
...
+	_rel/kazoo/bin/kazoo escript lib/sup-*/priv/sup.escript \
+	-n $NAME \
+	-h $HOST \
+	-c change_me \
+	-t 10 "$*"
```
fenollp: Since it is set here (I assume it cannot be re-set in `$*`) setting the timeout to something like 300 helps with kazoo_tasks_maintenance for example.

jamhed: It's better to set it from command-line. See update.

PR #2021, opened 2016-06-03
KAZOO-4816 allow empty list of apps from env

lazedo:


fenollp:
```
if we set the os env variable KAZOO_APPS to empty string the vm takes the defaults. 
i would expect the vm not to start any app. 
```

I disagree with your expectations. This is a breaking change.
Also this should be documented.

Instead I would set `KAZOO_APPS` to `"none"` to obtain the behaviour you desire, thus not breaking old behaviour.

lazedo:
@fenollp where's the breaking change ? if you unset the os variable it loads the default apps

PR #2022, opened 2016-06-03
KAZOO-4817 filter default headers with message headers

lazedo:
Server-ID is optional but if a specific message wants to turn it
mandatory it won't validate

add Queue-ID to default optional list.
when doing a amqp call or call_collect we may still want to send our
queue in the request


fenollp:
backport?

PR #2028, opened 2016-06-03
KAZOO-4750: build iolist of options

jamesaimonetti:


fenollp:
Could you move URI things from kz_util to kazoo_web as part of this PR?
In kz_util (which may be duplicates of kz_http_util functions):
```
uri_encode/1
uri_decode/1
resolve_uri/2
safe_urlencode/1
normalize_amqp_uri/1  %% This one should probably stay in kz_util
uri/2
```

There is also a test there that should be moved too.

jamesaimonetti:
@fenollp sounds like a new ticket would be more appropriate

PR #2032, opened 2016-06-07
KAZOO-4820 default for rpc call should be infinity

lazedo:


core/sup/src/sup.erl
```diff
@@ -196,7 +196,7 @@ option_spec_list() ->
     ,{'host', $h, "host", {'string', localhost()}, "System hostname"}
     ,{'node', $n, "node", {'string', "kazoo_apps"}, "Node name"}
     ,{'cookie', $c, "cookie", {'string', "monster"}, "Erlang cookie"}
-    ,{'timeout', $t, "timeout", {'integer', 5}, "Command timeout"}
+    ,{'timeout', $t, "timeout", {'integer', 0}, "Command timeout"}
```
fenollp: why not set the default to infinity here?

lazedo: because its an integer ?

PR #2041, opened 2016-06-08
KAZOO-4805: Implemented voicemail forwarding.

siplabs:
Implemented option (when in voicemail voice menu) to copy message to some different extension and add voice comment.


lazedo:
@siplabs Hi, i'm ok for accepting this but i'm concerned that this is for 3.33. can you create a pull request for this feature against master ?

thanks

siplabs:
Hello,

Yes, we are working on that but our cusomter agreed to finance it only after it's accepted in 3.22 

Please accept it and we'll provide version for master and v4 asap (several days).

lazedo:
@siplabs thanks for this. please note that you don't need to write specific for 4.0 but only for master.
for master version, i would like to see the sox part (concatenation of mp3 files) moved into kazoo_media.
also, not sure if this will work with wav / mp4 files so you may also add a check to verify the format it is saved/saving and enable/disable the forward based on that.

PR #2050, opened 2016-06-09
KAZOO-4759 ensure account has credit for charges

lazedo:


applications/crossbar/src/crossbar_services.erl
```diff
@@ -106,18 +95,12 @@ accepting_charges(Context, JObj, Services) ->
 %%--------------------------------------------------------------------
 -spec extract_items(kz_json:object()) -> kz_json:objects().
 extract_items(JObj) ->
-    kz_json:foldl(fun extract_items_from_category/3
-                  ,[]
-                  ,JObj
-                 ).
+    kz_json:foldl(fun extract_items_from_category/3, [], JObj).
```
fenollp: Wouldn't it be more explicit to get_values of get_values(JObj)?

lazedo: this is a formatting change. changes that affect how we extract the data will be part of KAZOO-4826

core/kazoo_services/src/kz_services.erl
```diff
@@ -1168,8 +1168,14 @@ get_service_modules() ->
 
 -spec get_filesystem_service_modules() -> atoms().
 get_filesystem_service_modules() ->
+    get_filesystem_service_modules(filelib:wildcard([code:lib_dir('kazoo_services'), "/src/services/*.erl"])).
```
k-anderson: @lazedo We are going to need a new approach here... When the release is packaged it does not contain the src/ directory.

lazedo: @k-anderson yes i know. this is the first approach. if module list  is not available from db, it takes from src/*.erl and if not available take them from default coded below.

PR #2051, opened 2016-06-09
Use correct macros for kz_device:sip_invite_format and kz_device:set_sip_invite_format

skorobkov:
If device configured with `invite_format = e164` (or any other not `username`) we shouldn't add X-KAZOO-AOR header.
If X-KAZOO-AOR header exist, kamailio replace INVITE RURI with stored Contact (from REGISTRATION request).
BTW: kz_device:sip_route/2 and kz_device:set_sip_route/2 must be fixed? Is macro `?IP` should be replaced to something other?

### DIFF
--- a/core/kazoo_documents/src/kz_device.erl
+++ b/core/kazoo_documents/src/kz_device.erl
@@ -50,6 +50,7 @@
 -define(REALM, [?SIP, <<"realm">>]).
 -define(IP, [?SIP, <<"ip">>]).
 -define(INVITE_FORMAT, [?SIP, <<"invite_format">>]).
+-define(ROUTE, [?SIP, <<"route">>]).
 -define(CUSTOM_SIP_HEADERS, <<"custom_sip_headers">>).
 -define(CUSTOM_SIP_HEADERS_KV_ONLY, [?SIP, ?CUSTOM_SIP_HEADERS]).
 -define(CUSTOM_SIP_HEADERS_IN, [?SIP, ?CUSTOM_SIP_HEADERS, <<"in">>]).
@@ -128,7 +129,7 @@ sip_route(DeviceJObj) ->
     sip_route(DeviceJObj, 'undefined').
 
 sip_route(DeviceJObj, Default) ->
-    kz_json:get_value(?IP, DeviceJObj, Default).
+    kz_json:get_value(?ROUTE, DeviceJObj, Default).
 
 -spec custom_sip_headers_inbound(doc()) -> api_object().
 -spec custom_sip_headers_inbound(doc(), Default) -> kz_json:object() | Default.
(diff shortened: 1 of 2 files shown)


lazedo:
@skorobkov this is a non issue, invite_format is independent of AOR. please provide kamailio logs that show this behaviour because we do have it working as expected.

skorobkov:
It is issue on upgrading from 3 to 4
Kamalio in 4.0 call `lookup()` function (which rewrite R-URI) if there "X-KAZOO-AOR" header.
https://github.com/2600hz/kazoo-configs/blob/master/kamailio/default.cfg#L586
3.22 dont
https://github.com/2600hz/kazoo-configs/blob/3.22/kamailio/default.cfg#L566

IN 4.0 set `invite_fromat` to `e164` changes only To field.to DID@SERVER_IP
In 3.22 set `invite_fromat` to `e164` also change INVITE R-URI from USERNAME@SERVER_IP to DID@SERVER_IP, and clients can easily route call by DID on ther sip-server (and in same time use callflow-features from kazoo).

lazedo:
@skorobkov like i said, works as expected. routes to registered contact by device. you can however change kamailio scripts to optionally preserve RURI by implementing a role, this change is not acceptable here. in 4.0, ecallmgr registrar will stop trying to create a full uri for the registered devices and let kamailio do its job. it brings many benefits in tls and wss.

skorobkov:
Ok. We will think what to do with this clients...
But what about changes in kz_device:sip_invite_format/2 and kz_device:set_sip_invite_format/2 ?

lazedo:
@skorobkov sorry, yes, that seems a valid change, looks to me that its a bug fix.

skorobkov:
Reverted changes in kz_endpoint.
But you should look into kzLdevice:sip_route/2 and kz_device:set_sip_route/2
This functions not used anywhere else, but in future can cause the problem.

lazedo:
@skorobkov not sure i understand the problem with `kz_device:set_sip_route/2`
i see it being used by `kzt_twiml_dial`

skorobkov:
set_sip_invite_format and set_sip_route  was copy of set_sip_ip and use the same macro `?IP`
set_sip_invite_format now fixed and used `?INVITE_FORMAT` macro.
But set_sip_route still use `?IP` macro, because there not defined any `?ROUTE` macro and I dont know what key it should use...

Original code:
```Erlang
-spec set_sip_ip(doc(), ne_binary()) -> doc().
set_sip_ip(DeviceJObj, Ip) ->
    kz_json:set_value(?IP, Ip, DeviceJObj).

-spec set_sip_invite_format(doc(), ne_binary()) -> doc().
set_sip_invite_format(DeviceJObj, Ip) ->
    kz_json:set_value(?IP, Ip, DeviceJObj).

-spec set_sip_route(doc(), ne_binary()) -> doc().
set_sip_route(DeviceJObj, Ip) ->
    kz_json:set_value(?IP, Ip, DeviceJObj).
```

lazedo:
@skorobkov got it, thanks. let me try to figure this out.

lazedo:
@skorobkov so it seems that we should create
-define(ROUTE, [?SIP, <<"route">>]).
and use it in `sip_route` and `set_sip_route`
can you do it in this pull request ?
Thanks

skorobkov:
Done.

lazedo:
@skorobkov thanks again, backported to 4.0 and 3.22

PR #2053, opened 2016-06-09
KAZOO-4462 participant events

jamhed:
* generate amqp events from fs in form of conference.event.$conf_id.$call_id
* process these events in conf_participant to reflect the state of participant


applications/conference/src/conf_participant.erl
```diff
@@ -45,12 +47,12 @@
 
 -define(SERVER, ?MODULE).
 
--define(RESPONDERS, [{{?MODULE, 'relay_amqp'}
+-define(RESPONDERS(CallId), [{{?MODULE, 'relay_amqp'}
```
fenollp: CallId is unused?

jamhed: Right,  remnants.

PR #2057, opened 2016-06-09
KAZOO-4734: bulk download voicemails audio in zip format

icehess:


lazedo:
@icehess lets talk over this

PR #2058, opened 2016-06-09
KAZOO-4837 add name if missing

lazedo:


onnet:
Hi @lazedo. Seems to be an imperfection: https://2600hz.atlassian.net/browse/KAZOO-4894

lazedo:
@onnet fixed https://github.com/2600hz/kazoo/pull/2259. thanks for reporting

PR #2062, opened 2016-06-10
escriptize sup

jamhed:


lazedo:
as part of build-release should we copy the generated sup to _rel/kazoo/bin ?
also, bring back the add-alias.sh that existed in 3.22 ?

jamhed:
@lazedo, it won't work in _rel/kazoo/bin by itself (it will use enscript in host search PATH, and it will be not the one from release), so i believe we can put it in project root (and provide a wrapper how to run it with escript from release).

lazedo:
@jamhed not sure what you're saying, i copied the generated sup to _rel/kazoo/bin and rut it from there. what's the catch ? if its a matter of not having erlang and only the released one, then can't we add that withing sup.erl ? ie, walk ../lib ../erts and add to path ?

fenollp:
Why this PR?

lazedo:
@fenollp why not ?

fenollp:
I don't see the need to escriptize sup. One won't be able to run the escript properly as no ERTS is installed on the destination machine. Maybe working around that is possible using the release's ERTS, but then the portability of an escript is moot.

Secondly, sup works (the bash script from priv dir). If it's too complicated a command to type, then let's alias this.

Again, I don't see the need for this PR and would be interested in the rationale behind it.

lazedo:
` ERTS is installed on the destination machine` its not true, its just in a different place.
`If it's too complicated a command to type, then let's alias this`, yes, it is. it just doesn't work.
create a clean environment, build and see for yourself.

fenollp:
What doesn't work?
This works for me: `./_rel/kazoo/lib/sup-4.0.0/priv/sup -n whistle_apps -h qwd tasks_maintenance task_output 7a334faeabca331e4e25180680f9e5` but uses my kerl-installed ERTS.

lazedo:
@fenollp i believe it works for you but not for the rest. start from scratch

fenollp:
I'm working on something related to this and I still don't understand. @lazedo any idea of what the actual error was?

PR #2075, opened 2016-06-14
KAZOO-4824 KAZOO-3862: search local available numbers

fenollp:


core/kazoo_number_manager/src/carriers/knm_local.erl
```diff
@@ -20,56 +20,90 @@
...
-    find_numbers(<<"+1", Number/binary>>, Quanity, Opts);
-find_numbers(_Number, _Quanity, _Opts) ->
-    %% TODO: given the requestor's account, discover knm_local numbers
-    %%        that are available but managed by accendants of the account.
-    {'error', 'not_available'}.
+                          {'ok', knm_number:knm_numbers()} |
+                          {'error', any()}.
+find_numbers(Number, Quantity, Options) when size(Number) < 5 ->
+    find_numbers(<<"+1", Number/binary>>, Quantity, Options);
+find_numbers(Number, Quantity, Options) ->
```
lazedo: why add +1 ? what about searching other countries ? i think that it should got here already with +1 or +44 or +49 ...
depending on what crossbar receives , maybe event adding a country prefix in the api payload for search.

fenollp:
@lazedo wrt to adding +1 or another i18n prefix, do you think it could be fetched from the account document? I know we store country-specific number formatters there.

I only do this to find the corresponding DB & numbers.
I corrected the code so that it behaves ok with queries already starting with `+`.
But it is possible to store numbers not starting with `+`. I am unsure if that is intended though.

What I could do is remove the `+` checking/adding, thus letting users search for numbers that they have stored and that don't start with `+`.
What do you guys think @lazedo @jamesaimonetti ?

core/kazoo_number_manager/src/knm_phone_number.erl
```diff
@@ -483,10 +485,12 @@ set_prev_assigned_to(N, PrevAssignedTo=?MATCH_ACCOUNT_RAW(_)) ->
...
 %% @end
 %%--------------------------------------------------------------------
--spec used_by(knm_phone_number()) -> api_binary().
+-spec used_by(knm_phone_number()) -> api_ne_binary().
 used_by(#knm_phone_number{used_by=UsedBy}) -> UsedBy.
 
--spec set_used_by(knm_phone_number(), ne_binary()) -> knm_phone_number().
+-spec set_used_by(knm_phone_number(), api_ne_binary()) -> knm_phone_number().
+set_used_by(N, UsedBy='undefined') ->
+    N#knm_phone_number{used_by=UsedBy};
```
lazedo: why do you add a function clause that doesn't do anything different ?

core/kazoo_number_manager/src/knm_phone_number.erl
```diff
@@ -612,10 +618,12 @@ update_carrier_data(N=#knm_phone_number{carrier_data = Data}, JObj=?JSON_WRAPPER
...
 %% @end
 %%--------------------------------------------------------------------
--spec region(knm_phone_number()) -> ne_binary().
+-spec region(knm_phone_number()) -> api_ne_binary().
 region(#knm_phone_number{region=Region}) -> Region.
 
--spec set_region(knm_phone_number(), ne_binary()) -> knm_phone_number().
+-spec set_region(knm_phone_number(), api_ne_binary()) -> knm_phone_number().
+set_region(N, Region='undefined') ->
+    N#knm_phone_number{region=Region};
```
lazedo: same here

core/kazoo_number_manager/src/knm_phone_number.erl
```diff
@@ -624,10 +632,12 @@ set_region(N, Region=?NE_BINARY) ->
...
 %% @end
 %%--------------------------------------------------------------------
--spec auth_by(knm_phone_number()) -> api_binary().
+-spec auth_by(knm_phone_number()) -> api_ne_binary().
 auth_by(#knm_phone_number{auth_by=AuthBy}) -> AuthBy.
 
--spec set_auth_by(knm_phone_number(), ne_binary()) -> knm_phone_number().
+-spec set_auth_by(knm_phone_number(), api_ne_binary()) -> knm_phone_number().
+set_auth_by(N, AuthBy='undefined') ->
+    N#knm_phone_number{auth_by=AuthBy};
```
lazedo: and here

core/kazoo_number_manager/src/knm_phone_number.erl
fenollp: Because `when UsedBy == 'undefined'; UsedBy = ?NE_BINARY ->` is an illegal guard.

lazedo: just remove the guards no?

fenollp: No they are the whole point. Look at `setters/2`.

icehess: `cb_callflows` needs update then cuz it sets `used_by` to `<<>>`.

core/kazoo_number_manager/src/carriers/knm_carriers.erl
```diff
@@ -53,145 +53,149 @@ find(Num, Quantity) ->
...
+             );
+        _Carrier ->
+            AssignTo = knm_phone_number:assign_to(PhoneNumber),
+            Add = props:filter_undefined(
+                    [{<<"number">>, DID}
+                    ,{<<"activation_charge">>, activation_charge(DID, AssignTo)}
+                    ,{<<"state">>, knm_phone_number:state(PhoneNumber)}
+                    ]),
+            kz_json:set_values(Add, CarrierData)
+    end.
```
lazedo: why are you treating `?CARRIER_MANAGED` differently ?
no rates for other providers ? or should we remove rate from `?CARRIER_MANAGED` ?

core/kazoo_number_manager/src/carriers/knm_carriers.erl
```diff
@@ -53,145 +53,149 @@ find(Num, Quantity) ->
...
-process_number_result(Number, Acc, Carrier) ->
-    DID = knm_phone_number:number(knm_number:phone_number(Number)),
-    check_for_existing_did(Number, Acc, Carrier, knm_phone_number:fetch(DID)).
+    case knm_phone_number:module_name(PhoneNumber) of
+        ?CARRIER_OTHER ->
+            [found_number_to_jobj(Number) | Acc];
+        Carrier ->
+            DID = knm_phone_number:number(PhoneNumber),
+            check_for_existing_did(Number, Acc, Carrier, knm_phone_number:fetch(DID))
+    end.
```
lazedo: why are you treating `?CARRIER_OTHER` differently ?
no carrier module should have an exception treatment, maybe you could extend the carrier behaviour for not having to deal with exceptions here.

core/kazoo_number_manager/src/carriers/knm_carriers.erl
```diff
@@ -307,6 +323,43 @@ disconnect(Number) ->
...
+%% @doc
+%% Create a number in a discovery state.
+%% @end
+%%--------------------------------------------------------------------
+-spec create_discovery(ne_binary(), module(), ne_binary(), kz_json:object()) ->
+                              knm_number_return().
+create_discovery(DID=?NE_BINARY, Carrier, ?MATCH_ACCOUNT_RAW(AuthBy), Data=?JSON_WRAPPER(_))
+  when is_atom(Carrier) ->
+    case knm_number:get(DID) of
+        {'ok', _Number}=Ok -> Ok;
```
lazedo: ensure existing number is not assigned already ?

core/kazoo_number_manager/src/carriers/knm_carriers.erl
fenollp: I'm not sure why OTHER and MANAGED are treated differently. That's @jamesaimonetti's code from a year ago.

core/kazoo_number_manager/src/carriers/knm_carriers.erl
fenollp: Yes. This is important. We may have a number in db that is actually available to buy (client released it without going through us). We are never the most up to date: the actual carrier is.

So, if we already have the number, we return it here. Somewhere else in knm_carriers the number will be updated (see transition_existing_to_discovery).

core/kazoo_number_manager/src/knm_phone_number.erl
fenollp: @icehess where exactly?

core/kazoo_number_manager/src/knm_phone_number.erl
```diff
@@ -95,7 +95,8 @@ new(DID, Options) ->
                 ,{fun set_number_db/2, knm_converters:to_db(NormalizedNum)}
                 ,{fun set_assign_to/2, knm_number_options:assign_to(Options)}
                 ,{fun set_state/2, ?NUMBER_STATE_DISCOVERY}
-                ,{fun set_module_name/2, knm_number_options:module_name(Options)}
+                ,{fun set_module_name/2, ?CARRIER_OTHER}
```
lazedo: why this revert ? and why `?CARRIER_OTHER` ? the system can be configured not to use this carrier, right ?
if not available in options, shouldn't we get it from config (maybe `available_module_name`) ?

fenollp: Because I don't want this PR to conflict with #2126 which has the fix :)

core/kazoo_number_manager/src/carriers/knm_carriers.erl
jamesaimonetti: This was carried over from [wnm_managed](https://github.com/2600hz/kazoo/blob/3.22/core/whistle_number_manager-1.0.0/src/carriers/wnm_managed.erl#L70)

core/kazoo_number_manager/src/carriers/knm_carriers.erl
```diff
@@ -359,6 +360,21 @@ create_discovery(DID=?NE_BINARY, Carrier, ?MATCH_ACCOUNT_RAW(AuthBy), Data=?JSON
...
+%% @end
+%%--------------------------------------------------------------------
+-spec is_local(ne_binary()) -> boolean().
+is_local(?CARRIER_LOCAL) -> 'true';
+is_local(?CARRIER_RESERVED) -> 'true';
+is_local(?CARRIER_RESERVED_RESELLER) -> 'true';
+is_local(?CARRIER_MANAGED) -> 'true';
+is_local(?CARRIER_INUM) -> 'true';
+is_local(_ForeignCarrier) -> 'false'.
+
```
lazedo: how about implementing carrier_module:is_local() ?

fenollp: I like this!

PR #2085, opened 2016-06-15
generate_numbers: exposed as knm_maintenance

jamesaimonetti:


fenollp:
Could you also update https://github.com/2600hz/kazoo/blob/master/core/kazoo_number_manager/src/carriers/knm_managed.erl#L127 so that `AccountId` matches `?MATCH_ACCOUNT_RAW(_)`?

fenollp:
4.0 plz

PR #2090, opened 2016-06-15
Fix incorrect movement of descendants

kalda341:
There were two bugs in update_descendants_tree.
- The recursive call passed ToTree instead of Tree
- Descendants were assumed to be children, but could be any level of descendant


k-anderson:
@lazedo Hey, did you get a chance to take a look at this?  lgtm

lazedo:
@k-anderson also lgtm, will try to test it over the weekend

lazedo:
@kalda341 great finding and fix. Thanks.

PR #1991, opened 2016-05-30
KAZOO-4804: stepswitch resource selectors

skorobkov:
Stepswitch Resource Selectors(SRS) is new "routing engine" for stepswitch.
Enable it with command:
`sup kapps_config set stepswitch route_by stepswitch_resource_selectors` (or `kapps_config:set(<<"stepswitch">>, <<"route_by">>, <<"stepswitch_resource_selectors">>).` in erlang console).
Documentation will be added soon.
SRS is modular, each module can add, filter or order (sort) resources.
List of called modules and their behaviour controled with `rules` from system_config/stepswitch.resource_selectors document (or from `resource_selectors` in AccountDB).
Default rules:
``` JSON
"rules": [
           {
               "get_resources": {
               }
           },
           {
               "filter_flags": {
                   "source": "resource"
               }
           },
           {
               "filter_regex_rules": {
                   "source": "resource",
                   "deny_on_empty_rules": true,
                   "update_number": true
               }
           },
           {
               "filter_regex_cid_rules": {
                   "source": "resource",
                   "deny_on_empty_rules": false,
                   "update_cid_number": false
               }
           },
           {
               "order_weight": {
                   "source": "resource",
                   "sort_order": "ascend"
               }
           }
       ]
```
Default rules replicate old stepswitch_resources behaviour, load resources from DB, filter by flags, filter with regex rules in resource docment and order by `weight`.
Existing modules:
kz_srs_filter_cid_prefix.erl
kz_srs_filter_flags.erl
kz_srs_filter_prefix.erl
kz_srs_filter_regex_cid_rules.erl
kz_srs_filter_regex_rules.erl
kz_srs_get_resources.erl
kz_srs_null.erl
kz_srs_order_weight.erl

Modules kz_srs_filter_prefix.erl and kz_srs_filter_cid_prefix.erl get list of prefixes from DB (`resource_selectors` or `{{ACCOUNT_ID}}_selectors`) which match number (or Caller-ID number) and return list of resources.
kz_srs_null.erl - testing module< which retrun empty list.

"Slectors" updates via cb_resources module (WIP).
URL: `/resources/{{RESOURCE_ID}}/selectors/{{SELECTOR_TYPE}}` (or /accounts/{{ACCOUNT_ID}}/esources/{{RESOURCE_ID}}/selectors/{{SELECTOR_TYPE}}`

{{SELECTOR_TYPE}} - can be `prefix`, `cid_prefix`, `flag`, `weight`, `regex`, `regex_cid`.

[...]

PR #2092, opened 2016-06-16
KAZOO-4804: new version of resource selectors

skorobkov:
Reworked #1991


k-anderson:
@skorobkov This is amazing work!  I was remarking to the team how impressed I am with this and how we could use this as a gold standard (example of code done right)!  

Thank you, your effort on this is greatly appreciated!

P.S> Master and 4.0 have deviated to such an extent that we are going to merge master into 4.0 after the final core module renames, so this will be in 4.0.1!

PR #2093, opened 2016-06-16
KAZOO-4850 allow array in data field

lazedo:


fenollp:
I'm not sure it's a good idea to de-enforce that simple rule that API always listens for `{data: {...}}`.

What collections is the ticket about?
In phone numbers a collection [is sent like this](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/doc/phone_numbers.md#get-locality-information-for-a-collection-of-numbers): `{data: {numbers: [...]}}`

lazedo:
@fenollp resources

PR #2097, opened 2016-06-17
fix some vm issues found during testing

lazedo:
type checking for older messages
maybe_load_vmboxes recursion

### DIFF
--- a/core/kazoo_voicemail/src/kz_vm_message.erl
+++ b/core/kazoo_voicemail/src/kz_vm_message.erl
@@ -43,6 +43,7 @@
 -define(MODB_COUNT_VIEW, <<"mailbox_messages/count_per_folder">>).
 -define(COUNT_BY_VMBOX, <<"mailbox_messages/count_by_vmbox">>).
 -define(BOX_MESSAGES_CB_LIST, <<"vmboxes/crossbar_listing">>).
+-define(PVT_LEGACY_TYPE, <<"private_media">>).
 
 -define(RETENTION_DURATION
         ,kapps_config:get_integer(?CF_CONFIG_CAT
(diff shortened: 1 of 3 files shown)


core/kazoo_voicemail/src/kz_vm_message.erl
```diff
@@ -43,6 +43,7 @@
 -define(MODB_COUNT_VIEW, <<"mailbox_messages/count_per_folder">>).
 -define(COUNT_BY_VMBOX, <<"mailbox_messages/count_by_vmbox">>).
 -define(BOX_MESSAGES_CB_LIST, <<"vmboxes/crossbar_listing">>).
+-define(PVT_LEGACY_TYPE, <<"private_media">>).
```
fenollp: Is string missing "pvt_" ?

lazedo: what ?

fenollp: macro is called `PVT_LEGACY_TYPE` but content reads `private_media`. I was wondering then why call it PVT_ ?

lazedo: really ?
[check here](https://github.com/2600hz/kazoo/blob/master/core/kazoo_documents/src/kzd_callflow.erl#L22)
[or here](https://github.com/2600hz/kazoo/blob/master/core/kazoo_documents/src/kz_notification.erl#L48)
or any other in kazoo_documents

lazedo:
@fenollp pull request was assigned to @icehess

fenollp:
ah sorry I missed that. Want be to restore?

lazedo:
@fenollp no worries, @icehess can look at the commit and correct anything that he may find.

PR #2103, opened 2016-06-18
handle a single vmbox id as well

icehess:


applications/crossbar/src/modules_v1/cb_vmboxes_v1.erl
```diff
@@ -474,6 +474,7 @@ load_vmbox_summary(Context) ->
 %%--------------------------------------------------------------------
 -spec maybe_load_vmboxes(ne_binaries(), cb_context:context()) -> cb_context:context().
```
fenollp: Then you should update the first arg to `ne_binaries() | ne_binary()`.

PR #2122, opened 2016-06-21
KAZOO-4858 use correct xpath for order

lazedo:
also changes the way that we return number for search and after processing


fenollp:
Shouldn't we check if the number size `byte_size(Num) == 10` before adding `+1`?

lazedo:
bandwidth numbers are always 10 digit, they only do US

PR #2123, opened 2016-06-21
KAZOO-4858 use correct xpath for order

lazedo:
also handles the way we return the number from search and activate


core/kazoo_number_manager/src/carriers/knm_bandwidth2.erl
```diff
@@ -137,7 +142,7 @@ find_numbers(Search, Quantity, Options) ->
 
 -spec process_search_response(xml_el(), kz_proplist()) -> knm_number:knm_numbers().
 process_search_response(Result, Options) ->
-    AccountId = props:get_value(<<"account_id">>, Options),
+    AccountId = props:get_first_defined([<<"Account-ID">>, <<"account_id">>], Options),
```
fenollp: Please revert this as it is fixed in #2075

PR #2124, opened 2016-06-21
Fast re-arm when binding digit is detected

vladimirralev:
I think you may find this useful. Impatient people tend to get stuck in armed state redialing expecting to start as quickly as possible but the binding digit just gets added towards the collected action. Especially useful if you set interdigit timeout high to accomodate slower users.


applications/konami/src/konami_code_fsm.erl
```diff
@@ -486,26 +486,32 @@ arm_bleg(#state{digit_timeout=Timeout}=State) ->
                 ,b_leg_armed='true'
                }.
 
+-spec maybe_fast_rearm(ne_binary(), ne_binary(), ne_binary())
```
fenollp: badspec

jamesaimonetti:
I'm debating in my head how likely it would be for people to use the arming DTMF in a pattern as well. For instance, would I ever think someone would arm with '*' and use a pattern that included using '*'?

If this PR is accepted, it will require updating the docs to explicitly mention the restriction of using the binding digit in any metaflow patterns. For now, can you include an update to the [docs](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/doc/metaflows.md) so we can at least have it ready? I'll discuss with the team on our next standup whether to go down this route.

vladimirralev:
Alright. I was thinking about the same and the only case that I came up with is some operators like to have a shortcut such as "**" (double start) which is the fastest metaflow to type that could exist. That shortcut is actually used in campaign dialing sometimes to move on quickly to next target number. It is also used when people want to reduce the chance of interfering with voicemail/IVR systems where the binding digit will have another purpose but runs the risk of konami taking over the call in unexpected ways. There you can have * binding digit, and all number that are configure will also start with a * so basically the arming digit itself will be double-star so to speak. These cases will suffer a bit, but I can code an exception for double-binding-key if you want, I think it will address almost all concerns.

jamesaimonetti:
@vladimirralev thanks for the update. I will discuss this today with the team to make sure there aren't any objections.

jamesaimonetti:
@vladimirralev: @lazedo had a good point in our meeting. We're looking at leveraging FreeSWITCH's [bind_meta_app](https://wiki.freeswitch.org/wiki/Misc._Dialplan_Tools_bind_meta_app) or [bind_digit_action](https://wiki.freeswitch.org/wiki/Misc._Dialplan_Tools_bind_digit_action) in newer versions of Konami/metaflows (4.x).

As such, we think it would be advisable to put this feature behind a config flag. So the function could become something like:

```erlang
-spec maybe_fast_rearm(ne_binary(), ne_binary(), binary()) -> binary().
-spec maybe_fast_rearm(ne_binary(), ne_binary(), binary(), boolean()) -> binary().

-define(USE_FAST_REARM, kapps_config:get_boolean(?APP_NAME, <<"use_fast_rearm">>, 'false')).

maybe_fast_rearm(DTMF, BindingDigit, Collected) ->
    maybe_fast_rearm(DTMF, BindingDigit, Collected, ?USE_FAST_REARM).

maybe_fast_rearm(_DTMF, _BindingDigit, Collected, 'false') -> <<Collected/binary, DTMF/binary>>;
maybe_fast_rearm(DoubleBindingDigit, DoubleBindingDigit, <<>>, 'true') -> DoubleBindingDigit;
maybe_fast_rearm(DTMFisBindingDigit, DTMFisBindingDigit, _Collected, 'true') -> <<>>;
maybe_fast_rearm(DTMF, _BindingDigit, Collected, 'true') -> <<Collected/binary, DTMF/binary>>.
```

Untested, but something along those lines would be acceptable.

applications/konami/src/konami_code_fsm.erl
```diff
@@ -486,26 +486,40 @@ arm_bleg(#state{digit_timeout=Timeout}=State) ->
                 ,b_leg_armed='true'
                }.
 
+-spec maybe_fast_rearm(ne_binary(), ne_binary(), binary()) -> binary().
+-spec maybe_fast_rearm(ne_binary(), ne_binary(), binary(), boolean()) -> binary().
+
+maybe_fast_rearm(DTMF, BindingDigit, Collected) ->
+    maybe_fast_rearm(DTMF, BindingDigit, Collected
+        ,kapps_config:get_is_true(<<"metaflows">>, <<"use_fast_rearm">>, 'false')).
```
lazedo: please use `?CONFIG_CAT` instead of metaflows.

vladimirralev: I just finished testing this, but CONFIG_CAT from inside konami points to konami's config section, while most other settings are in the metaflows section.The config is most similar to the binding digit so I thought it should be in the same place https://github.com/2600hz/kazoo/blob/master/applications/konami/src/konami_config.erl#L64

lazedo: @vladimirralev its a good point but, the config is for application specific behaviour.

vladimirralev: OK. I see this is more konami-scoped config. I will need to change the docs too in this case.

jamesaimonetti: @vladimirralev just pinging you to see if you'll get to this today?

jamesaimonetti: @vladimirralev ping again ;)

vladimirralev: Sorry, updated it now.

jamesaimonetti:
@vladimirralev i manually merged this; still haven't figured out why github isn't detecting it properly.

Thanks for all the effort here!

PR #2125, opened 2016-06-21
superfluous-escape: backman says no to backspace

fenollp:


applications/ecallmgr/src/ecallmgr_channel_move.erl
```diff
@@ -77,10 +77,9 @@ resume(UUID, NewNode, Evt) ->
 %% and its corresponding '>' to %3C and %3E as they should be
 -spec fix_metadata(ne_binary()) -> ne_binary().
 fix_metadata(Meta) ->
-    Replacements = [
-                    {<<"\<sip\:">>, <<"%3Csip:">>}
-                    ,{<<"\>\<sip">>, <<"%3E<sip">>}
-                    ,{<<"\>;">>, <<"%3E;">>} % this is especially nice :)
+    Replacements = [{<<"<sip\:">>, <<"%3Csip:">>}
```
jamesaimonetti: @fenollp How did you test that this works as expected? This code has been untouched since 2013 so I am hesitant to accept this commit without proof of testing. Show your work! :)

fenollp:
```erlang
1> <<"\> \<">> == <<"> <">>.
true
```

fenollp:
I rebased because I missed a `\:`.

applications/ecallmgr/src/ecallmgr_channel_move.erl
lazedo: you should test as part of regex

lazedo:
you should test as part of regex, i think the backspaces are for regex expression.

fenollp:
@lazedo 

```erlang
1> <<"\> \<">> == <<"> <">>.
true
```

lazedo:
@fenollp that is not a test with regex. you should test it as part of the re:replace

```
Replacements = [
 -                    {<<"\<sip\:">>, <<"%3Csip:">>}
 -                    ,{<<"\>\<sip">>, <<"%3E<sip">>}
 -                    ,{<<"\>;">>, <<"%3E;">>} % this is especially nice :)
 +    Replacements = [{<<"<sip:">>, <<"%3Csip:">>}
 +                    ,{<<"><sip">>, <<"%3E<sip">>}
 +                    ,{<<">;">>, <<"%3E;">>} % this is especially nice :)
                      %% until such time as FS sets these properly
                      ,{<<"<dialplan></dialplan>">>, <<"<dialplan>XML</dialplan>">>}
                      ,{<<"<context>default</context>">>, <<"<context>context_2</context>">>}
                    ],
     lists:foldl(fun({S, R}, MetaAcc) ->
                         iolist_to_binary(re:replace(MetaAcc, S, R, ['global']))
                 end, Meta, Replacements).
```
if `<` and '>' are not special characters in regex, then i think your change is ok.
if they are, then you may actually need a double `\`

fenollp:
<> are not special regex chars, and if they were it doesn't matter as @jamesaimonetti said, this code has been working as expected, untouched since 2013.

My changes actually don't change any of the semantics of this code.

PR #2126, opened 2016-06-22
qf-knm: log setter name in a human-readable way

fenollp:


jamesaimonetti:
@fenollp not sure I understand the need for this changeset? Are functions crashing and you can't figure out which one did?

fenollp:
No, you;re right. Moreover when a setter is wrong the reason gets logged. I'll close this and do some cleaning tomorrow

PR #2127, opened 2016-06-22
qf-tasks: fix race condition where handle_info would happen before ha…

fenollp:
…ndle_cast

### DIFF
--- a/core/kazoo_tasks/include/kazoo_tasks.hrl
+++ b/core/kazoo_tasks/include/kazoo_tasks.hrl
@@ -13,7 +13,7 @@
 -define(KZ_TASKS_INPUT_ERROR_MIME, <<"expected_content">>).
 
 -define(KZ_TASKS_WAIT_AFTER_ROW,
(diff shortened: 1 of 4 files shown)


core/kazoo_tasks/include/kazoo_tasks.hrl
```diff
@@ -13,7 +13,7 @@
 -define(KZ_TASKS_INPUT_ERROR_MIME, <<"expected_content">>).
 
 -define(KZ_TASKS_WAIT_AFTER_ROW,
-        ?MILLISECONDS_IN_SECOND * kapps_config:get_integer(?KZ_TASKS_DOC_TYPE, <<"wait_seconds_after_row">>, 2)).
+        kapps_config:get_integer(?KZ_TASKS_DOC_TYPE, <<"wait_milliseconds_after_row">>, 500)).
```
jamesaimonetti: These values have traditionally been the name with unit as a suffix. I would use "wait_after_row_ms" in this case.

PR #2132, opened 2016-06-23
allow to mark voicemail as deleted without soft-deleting

icehess:


applications/crossbar/src/modules_v2/cb_vmboxes_v2.erl
```diff
@@ -338,7 +344,7 @@ validate_messages(Context, DocId, ?HTTP_DELETE) ->
         ,ToDelete
     ).
 
--spec get_folder_filter(cb_context:context(), ne_binary()) -> ne_binary().
+-spec get_folder_filter(cb_context:context(), ne_binary()) -> ne_binary() | {ne_binary(), boolean()}.
```
fenollp: How about using vm_folder/0?

PR #2134, opened 2016-06-23
fix permissions for phone number activate

lazedo:


applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
```diff
@@ -863,6 +863,7 @@ collection_process(Context, Action, Numbers) ->
                             knm_numbers:numbers_return().
 numbers_action(Context, ?ACTIVATE, Numbers) ->
     Options = [{'auth_by', cb_context:auth_account_id(Context)}
+               ,{'assign_to', cb_context:account_id(Context)}
```
fenollp: Actually, why do you set assign_to here? Looking at knm_number:move/3 it sets assign_to from the 2nd argument.

PR #2135, opened 2016-06-23
add a country to search numbers parameters

lazedo:
adds a country to search numbers parameters so we can infer the country
prefix to pass to number search.

used knm_managed to generate +1555 numbers (Hollywood)
simple test in monster-ui 

before patch
search area code 424. it correctly returns 424 US numbers (bandwidth2)
search area code 555. no knm_managed numbers returned
search area code 155. knm_managed numbers returned

after patch
search area code 424. it correctly returns 424 US numbers (bandwidth2)
search area code 555. knm_managed numbers returned

### DIFF
--- a/applications/crossbar/priv/couchdb/schemas/find_numbers.json
+++ b/applications/crossbar/priv/couchdb/schemas/find_numbers.json
@@ -2,6 +2,13 @@
     "$schema": "http://json-schema.org/draft-03/schema#",
     "_id": "find_numbers",
     "properties": {
+        "country": {
+            "default": "US",
+            "maxLength": 2,
+            "minLength": 2,
+            "required": false,
+            "type": "string"
+        },
         "prefix": {
             "maxLength": 10,
             "minLength": 3,
--- a/core/kazoo_number_manager/src/knm_util.erl
+++ b/core/kazoo_number_manager/src/knm_util.erl
@@ -13,6 +13,7 @@
 
 -export([pretty_print/1, pretty_print/2
          ,fixture/1
+         ,prefix_for_country/1
         ]).
 
 -include("knm.hrl").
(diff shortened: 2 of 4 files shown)


applications/crossbar/priv/couchdb/schemas/find_numbers.json
```diff
@@ -2,6 +2,13 @@
     "$schema": "http://json-schema.org/draft-03/schema#",
     "_id": "find_numbers",
     "properties": {
+        "country": {
+            "default": "US",
```
fenollp: Doesn't this pre-fill Context?

core/kazoo_number_manager/src/knm_util.erl
```diff
@@ -126,3 +127,9 @@ read_fixture({'ok', Contents}, _F) ->
     kz_util:to_list(Contents);
 read_fixture({'error', 'enoent'}, F) ->
     throw({'error', 'missing_fixture', F}).
+
+%% TODO
+%% this should be replaced with a call to elibphonenumber
+%% when/if we integrate that lib or do it ourselves
+-spec prefix_for_country(ne_binary()) -> ne_binary().
+prefix_for_country(<<"US">>) -> <<"+1">>.
```
fenollp: I think this should go in converters:normalize. This is where we can swap default (regexp) with libphonenumber normalization. When we do have libpn we'll use it as default I guess?

applications/crossbar/priv/couchdb/schemas/find_numbers.json
lazedo: yes, if not provided, that's the intent.

core/kazoo_number_manager/src/knm_util.erl
lazedo: no conversion involved here. we are not normalizing numbers.
even if/when we use libpn, regex will always be there for the cases where libpn doesn't work like dialplans

PR #2138, opened 2016-06-23
Api endpoints update

jamesaimonetti:


applications/crossbar/doc/ref/tasks.md
```diff
@@ -4,9 +4,6 @@
 
 #### Schema
 
-Key | Description | Type | Default | Required
---- | ----------- | ---- | ------- | --------
-`records` | List the rows of input data | `array(object)` |   | `true`
```
fenollp: Uh you sure?

jamesaimonetti: Uh, is there a schema file?

fenollp: Yes but it's called tasks.records. Should I rename it to tasks?

jamesaimonetti: If you have a custom schema that doesn't follow the convention, it should go direct into the `doc/tasks.md` version and not the auto-generated version.

PR #2140, opened 2016-06-23
use send_file instead of to_binary

lazedo:
also uses new send_file


applications/crossbar/src/modules/cb_presence.erl
```diff
@@ -142,7 +142,7 @@ validate(Context) ->
     validate_thing(Context, cb_context:req_verb(Context)).
 
 validate(Context, ?MATCH_REPORT_PREFIX(Report)) ->
-    load_report(Context, Report);
+    load_report(Context, Report, cb_context:req_verb(Context));
```
jamesaimonetti: Is this necessary if we're keeping the "report-" pattern in `allowed_methods`?

PR #2147, opened 2016-06-24
added otp utils

tsromana:
this pr contains helper methods to generate secret for otp as well generating token from that secret, sorry i didn't get time to create module to use it for the auth, will create that but in mean time wanted to get suggestions on it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ core/kazoo/mime.types
 
 applications/*/priv/archive_data/*
 
+.idea/
--- a/applications/crossbar/priv/couchdb/account/users.json
+++ b/applications/crossbar/priv/couchdb/account/users.json
@@ -6,10 +6,10 @@
     "language": "javascript",
     "views": {
         "creds_by_md5": {
-            "map": "function(doc) { if(!doc.pvt_md5_auth || (doc.pvt_type != 'user' || doc.pvt_deleted || !doc.enabled)) return; emit( doc.pvt_md5_auth, {'owner_id': doc._id, 'account_id': doc.pvt_account_id} ); }"
+            "map": "function(doc) { if(!doc.pvt_md5_auth || (doc.pvt_type != 'user' || doc.pvt_deleted || !doc.enabled)) return; emit( doc.pvt_md5_auth, {'owner_id': doc._id, 'account_id': doc.pvt_account_id, 'otp_secret': 'doc.pvt_otp_secret} ); }"
         },
         "creds_by_sha": {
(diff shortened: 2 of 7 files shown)


lazedo:
@tsromana hey, thanks for this. some initial thoughts.

`base32` i would prefer to add to deps something like [base32](https://github.com/aetrion/base32_erlang) or other or add base32 to `core/kazoo`

`create_qr_code` there is already code in `notify_util` and  `cb_port_requests` that use the google api to create qr codes. we should move this to a common place (maybe  `core/kazoo`)

fenollp:
How does https://github.com/komone/qrcode (or another) compares to Google's? We could save us an HTTP call with this!

lazedo:
@fenollp nice finding.

tsromana:
@lazedo yes i will do, i thought so but its better to take an expert's view, as far as the qr code the api we are using i guess is being depricated by google so i was thinking to have an inbuilt lib to generate qr code but couldnt find a lib which can do so Directly as i found projects with qr codes built in so was thinking of converting that as @fenollp mentioned i was thinking of using the above lib but at this time it generates a png so didnt knew to just get qrcode

jamesaimonetti:
@tsromana please include a doc in `applications/crossbar/doc` that explains what cb_otp is for, how to call the endpoints, etc. See most of the other docs for formatting. You can also run, from the Erlang shell, `cb_api_endpoints:to_ref_doc().` and it will build a stub version in `applications/crossbar/doc/ref` that you can copy into `doc/` and modify from there.

jamesaimonetti:
@tsromana how's this coming along? Need help generating the docs? Also, make sure all your functions (esp `cb_otp_util.erl`) have proper specs.

tsromana:
@jamesaimonetti sorry for being late again. i created a login crossbar module which works same as the normal, user login but also check a pvt otp secret field which if exists return user reply to send out the login details with the otp password otherwise will login the user after matching the credentials hash. What i was thinking is to have 'cb_otp_util' as a helper method and integrate the otp login logic inside the 'cb_user_auth' making it backward compatible so it doesn't complain if the there is no otp field inside user's doc or should i create a new login module ?? and i will move base32 to kazoo core. than the only thing left is to create module to generate and save the otp code should i do it inside the cb_user or create a different module for that as well ? and for the qr code i guess we should just send out the secret and let the ui generate the qr code. Thank you

tsromana:
i have updated the code to match the otp token recieved from user and added necessary errors will update the code to for generation of secret for user, please let me know if there is any issue, i will do some testing tomorrow on the code.

tsromana:
i will create another pr with the updated code as i am having trouble rebasing the code

PR #2152, opened 2016-06-24
Iso3166a2 to itu

fenollp:


applications/crossbar/src/modules_v1/cb_phone_numbers_v1.erl
```diff
@@ -418,8 +418,8 @@ get_find_numbers_req(Context) ->
 get_numbers(QueryString) ->
     PrefixQuery = kz_json:get_ne_value(?PREFIX, QueryString),
     Country = kz_json:get_ne_value(?COUNTRY, QueryString, ?DEFAULT_COUNTRY),
-    CountryPrefix = knm_util:prefix_for_country(Country),
-    Prefix = <<CountryPrefix/binary, PrefixQuery/binary>>,     
+    CountryPrefix = knm_iso3166a2_itu:to_itu(Country),
+    Prefix = <<CountryPrefix/binary, PrefixQuery/binary>>,
```
lazedo: i would prefer to leave `knm_util:prefix_for_country` and change its implementation to call ` knm_iso3166a2_itu:to_itu`
if/when we change to libpn we only have to change `knm_util:prefix_for_country`

core/kazoo_number_manager/src/knm_util.erl
```diff
@@ -127,9 +126,3 @@ read_fixture({'ok', Contents}, _F) ->
     kz_util:to_list(Contents);
 read_fixture({'error', 'enoent'}, F) ->
     throw({'error', 'missing_fixture', F}).
-
-%% TODO
-%% this should be replaced with a call to elibphonenumber
-%% when/if we integrate that lib or do it ourselves
--spec prefix_for_country(ne_binary()) -> ne_binary().
-prefix_for_country(<<"US">>) -> <<"+1">>.
```
lazedo: as commented above, leave this and change it to call  `knm_iso3166a2_itu:to_itu`

PR #2153, opened 2016-06-24
Apply tidier --boolean to core dir(s)

kostis:


fenollp:
Can you put the orelse/andalso on new lines?

```erlang
    lists:member(self(), Pids)
        orelse hook_it_up(Hook).
```

kostis:
Why?  They fit nicely in one line.  Are you guys getting paid by the lines your code base contains? :)
Do you also write the following?
<pre>
   SomeVariable
       + SomeOtherVar</pre>
(`+` and `orelse` are both binary operators after all...)

fenollp:
It's easier to read as the lines get longer. Some of these are already a bit long.

We don't, but I think @jamesaimonetti is paid by the number of -specs he writes :)

PR #2154, opened 2016-06-24
Various tidier simplifications and other cleanups

kostis:


core/amqp_cron/src/amqp_cron.erl
```diff
@@ -362,20 +363,20 @@ send_tasks(Tasks, Election) ->
             ok
     end.
 
--spec stop_tasks(State :: #state{}) -> #state{}.
+-spec stop_tasks(State :: state()) -> state(}.
```
fenollp: `state(}`

core/kazoo_documents/src/kz_attachment.erl
```diff
@@ -44,9 +44,9 @@ get_content_type(MediaType, _) ->
     end.
 
 -spec corrected_base64_decode(ne_binary()) -> ne_binary().
-corrected_base64_decode(Base64) when byte_size(Base64) rem 4 == 3 ->
+corrected_base64_decode(Base64) when byte_size(Base64) rem 4 =:= 3 ->
     base64:decode(<<Base64/binary, "=">>);
-corrected_base64_decode(Base64) when byte_size(Base64) rem 4 == 2 ->
+corrected_base64_decode(Base64) when byte_size(Base64) rem 4 =:= 2 ->
```
fenollp: `rem` works on floats? What does this change change?

core/kazoo_tasks/src/kz_task_noinput_worker.erl
```diff
@@ -56,7 +56,7 @@ start(TaskId, Module, Function, ExtraArgs, OrderedFields) ->
                                                                                   {'error', any()}.
 init(TaskId, Module, Function, ExtraArgs, _) ->
     case
-        kz_util:try_load_module(Module) == Module andalso
+        kz_util:try_load_module(Module) =:= Module andalso
```
fenollp: try_load_module returns `atom()`. Why this change?

kostis:
@fenollp Almost all uses of `==/2` in Erlang should be avoided.  It's more kosher to use `=:=/2` instead, unless you really want to compare integers and floats, in which case you actually signify to the reader that you really want that behavior (as opposed to being just an accident of using `==` all over the place).

Moreover, `=:=/2` does not have to check for the coersion situation and therefore it's in principle a leaner/faster BIF (and its implementation is definitely faster in native code).

fenollp:
Nice to know! Well, there are a lot to catch in this repo!

Have you seen a larger & open source Erlang project than Kazoo?

kostis:
Well, I've been working for 15 years now on Erlang/OTP...

I do not want to brag about sizes, but the HiPE compiler is 100 KLOC and dialyzer about 30 KLOC.

kostis:
Oh, and my latest and therefore dearest project, [CutEr](https://github.com/aggelgian/cuter), contains this [monster Erlang file](https://github.com/aggelgian/cuter/blob/master/src/cuter_binlib.erl).

PR #2160, opened 2016-06-24
Apply tidier --structs to modules of the core dirs

kostis:
This factors out common structure creations, which typically both
shortens the text of the source and makes the code faster by avoiding
unnecessary term creations.

DO NOT merge yet -- more to come here...

### DIFF
--- a/core/apns/src/apns_connection.erl
+++ b/core/apns/src/apns_connection.erl
@@ -119,8 +119,8 @@ open_out(Connection) ->
     ssl_opts(Connection),
     Connection#apns_connection.timeout
   ) of
-    {ok, OutSocket} -> {ok, OutSocket};
-    {error, Reason} -> {error, Reason}
+    {ok, _OutSocket} = OK -> OK;
+    {error, _Reason} = Er -> Er
   end.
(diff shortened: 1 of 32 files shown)


core/apns/src/apns_connection.erl
```diff
@@ -119,8 +119,8 @@ open_out(Connection) ->
     ssl_opts(Connection),
     Connection#apns_connection.timeout
   ) of
-    {ok, OutSocket} -> {ok, OutSocket};
-    {error, Reason} -> {error, Reason}
+    {ok, _OutSocket} = OK -> OK;
+    {error, _Reason} = Er -> Er
```
kostis: This was done by hand, in order to have the `->` aligned.  Tidier chooses variable names based on the tag of the structure (`Ok` and `Error` in this case), possibly with an appended number (e.g. `Error2`) if the variable is already used in the clause.

jamesaimonetti: Thanks!

kostis:
You can merge this now if you like it.  That's it from tidier!

PR #2161, opened 2016-06-24
convert include to include_lib

jamhed:


fenollp:
Okay so this is to get `sync` to understand where include files are located. I have always assumed that include has a "include anything" v. include_lib which has a "include only purposely public things".

What makes it not possible for `sync` to work with the code (as it is outside this PR)? (I know you told me already but I can't remember what is was).
I fear that other tools / projects wouldn't be able to work with Kazoo apps if this change is merged.

jamhed:
@fenollp both include and include_lib allows to include anything, the only difference is how they search for file to include: include_lib uses application path from code:lib_dir/1. Later allows the file to be [re]compiled independently of current working directory. There is a convention though to use include for "internal" includes, and include_lib for "publicly available", but it's mere a convention (e.g. nothing prevets to alter include paths with -I options).

fenollp:
How is not handling cwd not a bug in @rustyio's sync then?
I understand it fetches include paths from the file's BEAM chunks. Are they not stored as absolute? Are some missing?

jamhed:
> Are they not stored as absolute?

They stored as absolute. Actually it turns out he handles include options completely wrong: he uses only one of them and ignores rest. And I think this is the case, so I close this PR.

PR #2159, opened 2016-06-24
check name against _id of JSON files

fenollp:
Catches errors like file callflows.voicemail.json using `_id` "callflows.language" and ensuing name clashes

PR #2163, opened 2016-06-24
fix callflows.voicemail schema id

icehess:


fenollp:
This is also fixed by #2159

PR #2167, opened 2016-06-24
Tidier applications subdirs

kostis:


applications/crossbar/src/modules/cb_signup.erl
```diff
@@ -134,16 +134,12 @@ resource_exists(_) -> 'true'.
...
-    end.
+    (?HTTP_PUT) = cb_context:req_verb(Context),
+    validate_new_signup(cb_context:set_account_db(Context, ?SIGNUP_DB)).
 
 validate(Context, ActivationKey) ->
-    case cb_context:req_verb(Context) of
-        ?HTTP_POST ->
-            check_activation_key(ActivationKey, cb_context:set_account_db(Context, ?SIGNUP_DB))
-    end.
+    (?HTTP_POST) = cb_context:req_verb(Context),
```
jamesaimonetti: Interested in why Tidier wraps the return in the parentheses?

kostis: Tidier uses [erl_prettypr](https://github.com/erlang/otp/blob/maint/lib/syntax_tools/src/erl_prettypr.erl) for printing and it's this program that does this.  Note that this happens only for macros.  Do not know why there are cases where it is needed -- it's definitely not needed for `?HTTP_POST`.

jamesaimonetti:
@kostis thanks. I merged this [manually](https://github.com/2600hz/kazoo/commit/a447dcfd81833cf84736e5ae643622207674d941) but GitHub seems to not recognize it and close this automatically.

PR #2170, opened 2016-06-25
Apply tidier --all to application/src/*.erl files

kostis:


applications/crossbar/src/api_resource.erl
```diff
@@ -152,10 +152,8 @@ is_proxied(Peer) ->
 
 is_proxied(_Peer, []) -> 'false';
 is_proxied(Peer, [Proxy|Rest]) ->
-    case kz_network_utils:verify_cidr(Peer, kz_network_utils:to_cidr(Proxy)) of
-        'true' -> 'true';
-        'false' -> is_proxied(Peer, Rest)
-    end.
+    kz_network_utils:verify_cidr(Peer, kz_network_utils:to_cidr(Proxy))
+	orelse is_proxied(Peer, Rest).
```
fenollp: indent?

kostis: By the way the indentation was correct, but I used emacs whose default configuration is with tabs instead of spaces... Now it fixed.

PR #2176, opened 2016-06-27
fix is_matched_account_mod

lazedo:
we are checking encoded/encoded and unencoded/unencoded
this adds encoded/unencoded unencoded/encoded encoded/raw unencoded/raw


fenollp:
So it's #2176 xor #2174 ?

lazedo:
i think this works best for `is_matched_account_mod`. merge this and remove `is_matched_account_mod` from #2174

PR #2179, opened 2016-06-28
Revert "fix is_matched_account_mod"

fenollp:
Reverts 2600hz/kazoo#2176

kapps_util:get_account_mods(AccountId / AccountDb) only gives back [AccountDb * X] for me
where X = number of modbs for said account


fenollp:
I say let's merge https://github.com/2600hz/kazoo/pull/2174 instead

PR #2184, opened 2016-06-28
KAZOO-4118

fenollp:
Notes:

when either BoM or EoM is not present (but the other is) it will put the value that exists and an empty value:

```
a4349ir30r934ir0349sub_account_idwew2323,201606,number_services,local,,130
```
(note the 2 following commas `,,`).

However, if both documents are not present, no row is generated for this month.

Waiting on https://github.com/2600hz/kazoo/pull/2183 to fix CI issues.

### DIFF
--- a/core/kazoo_services/src/kz_services_tasks.erl
+++ b/core/kazoo_services/src/kz_services_tasks.erl
@@ -86,11 +86,6 @@ descendant_quantities(_, [SubAccountMoDB | DescendantsMoDBs]) ->
            ,Category
            ,Item
            ,integer_to_binary(kz_json:get_integer_value(Item, BoMItem, 0))
-           ]
-          ,[AccountId
-           ,YYYYMM
-           ,Category
-           ,Item
            ,integer_to_binary(kz_json:get_integer_value(Item, EoMItem, 0))
            ]
           ]
--- a/core/kazoo_services/src/kz_services_tasks.erl
+++ b/core/kazoo_services/src/kz_services_tasks.erl
@@ -73,8 +73,8 @@ descendant_quantities(_, [SubAccountMoDB | DescendantsMoDBs]) ->
     ?MATCH_MODB_SUFFIX_ENCODED(A, B, Rest, Year, Month) = SubAccountMoDB,
     AccountId = ?MATCH_ACCOUNT_RAW(A, B, Rest),
     YYYYMM = <<Year/binary, Month/binary>>,
-    BoM = modb_service_quantities(SubAccountMoDB, 'true'),
-    EoM = modb_service_quantities(SubAccountMoDB, 'false'),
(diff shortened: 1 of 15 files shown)


fenollp:
It's ready!

core/kazoo_services/src/kz_services_tasks.erl
```diff
@@ -0,0 +1,152 @@
...
+descendant_quantities(_, []) ->
+    'stop';
+
+descendant_quantities(_, [SubAccountMoDB | DescendantsMoDBs]) ->
+    ?MATCH_MODB_SUFFIX_ENCODED(A, B, Rest, Year, Month) = SubAccountMoDB,
+    AccountId = ?MATCH_ACCOUNT_RAW(A, B, Rest),
+    YYYYMM = <<Year/binary, Month/binary>>,
+    BoM = modb_service_quantities(SubAccountMoDB, 'true'),
+    EoM = modb_service_quantities(SubAccountMoDB, 'false'),
+    Data =
```
jamesaimonetti: This comprehension of a comprehension, imo, is too dense and hard to parse (thus brittle and hard to edit later on). I'd like to see some intermediary steps, function calls, etc; something to break up the work into smaller, more discrete chunks that can be individually reasoned about.

jamesaimonetti: I'd also like to see the 'work' portion broken out into its own function and a unit test made to show what is being built here. This will aid in refactoring, I think.

core/kazoo_services/src/kz_services_tasks.erl
```diff
@@ -0,0 +1,152 @@
...
+                   (Id = kz_doc:id(JObj)) =/= AccountId
+            ];
+        {'error', _R} ->
+            lager:debug("unable to get descendants of ~s: ~p", [AccountId, _R]),
+            []
+    end.
+
+-spec modb_service_quantities(ne_binary(), boolean()) -> kz_json:object().
+modb_service_quantities(MoDB, IsBoM) ->
+    Id = case IsBoM of
```
jamesaimonetti: This should be its own function. Maybe `snapshot_id(IsBom :: boolean()) -> ne_binary().`?

jamesaimonetti: Or two macros, `?BOM_ID` and `?EOM_ID` instead of passing in the booleans?

jamesaimonetti:
I'd also like to ask for a doc that describes the engineering of `kazoo_tasks`, how to use them, etc. Similar to the [amqp workers](https://github.com/2600hz/kazoo/blob/master/core/kazoo_amqp/doc/amqp_workers.md) doc.

jamesaimonetti:
A few [dialyzer complaints](https://gist.github.com/jamesaimonetti/19c87b9038427276a298b6c4b65fb2bc) that you may want to address (just for kazoo tasks and the like). Ignore the ones not related.

PR #2190, opened 2016-06-29
KAZOO-4859: conferences REST API

jamhed:


applications/crossbar/src/modules/cb_conferences.erl
```diff
@@ -89,8 +96,17 @@ allowed_methods(_, <<"mute">>) ->
...
+resource_exists(_, <<"lock">>) -> 'true';
+resource_exists(_, <<"unlock">>) -> 'true';
+resource_exists(_, <<"hangup">>) -> 'true';
 resource_exists(_, <<"unmute">>) -> 'true';
 resource_exists(_, <<"mute">>) -> 'true'.
+resource_exists(_, <<"unmute">>, _) -> 'true';
+resource_exists(_, <<"mute">>, _) -> 'true';
+resource_exists(_, <<"undeaf">>, _) -> 'true';
+resource_exists(_, <<"deaf">>, _) -> 'true';
+resource_exists(_, <<"hangup">>, _) -> 'true'.
```
fenollp: When binaries are used multiple times we prefer to MACRO-ize them

applications/crossbar/src/modules/cb_conferences.erl
```diff
@@ -233,18 +271,47 @@ create_conference(Context) ->
...
+    UnMuted = [
+        perform_conference_action(Conference, <<"unmute">>, kz_json:get_value(<<"Participant-ID">>, P))
+            || P <- Participants, kz_json:is_false(<<"Is-Moderator">>, P), kz_json:is_false(<<"Speak">>, P)
+    ],
+    kz_json:set_value(<<"resp">>, UnMuted, kz_json:new()).
+
+-spec lock_conference(ne_binary()) -> kz_json:object().
+lock_conference(ConfId) ->
+    kapps_conference_command:lock(ConfId),
+    kz_json:set_value(<<"resp">>, <<"ok">>, kz_json:new()).
```
k-anderson: @jamhed 
Not a fan of these return payloads. 
{"data": {
  "resp": "ok"
 }
}
seems redundant considering the HTTP request was successful.

Also in keeping with our new convension I would argue that all of these 'resp' keys (lock, unlock, mute, ect) should be '_read_only'.

Maybe the rest of the team could weigh in?

applications/crossbar/src/modules/cb_conferences.erl
```diff
@@ -64,14 +64,25 @@ init() ->
...
-    [?HTTP_POST].
+-spec allowed_methods(path_token(), path_token(), path_token()) -> http_methods().
+
+allowed_methods() -> [?HTTP_GET, ?HTTP_PUT].
+
+allowed_methods(_) -> [?HTTP_GET, ?HTTP_POST, ?HTTP_PATCH, ?HTTP_DELETE].
+
+allowed_methods(_, <<"details">>) -> [?HTTP_GET];
+
+allowed_methods(_, <<"lock">>) -> [?HTTP_POST];
```
k-anderson: @jamhed Howdy, I understand that there was precedent but having actions in the REST api is not good practice and something we avoid...  We should discuss alternatives on the bridge tomorrow (or here)...

For example I would propose something such as:

Lock/Unlock Conference:
PUT v2/accounts/{account_id}/conferences/{conference_id}/lock 
DELETE v2/accounts/{account_id}/conferences/{conference_id}/lock

Kick:
DELETE v2/accounts/{account_id}/conferences/{conference_id}/participants/{participant_id

Kick All:
DELETE v2/accounts/{account_id}/conferences/{conference_id}/participants

Mute/Unmute
PUT v2/accounts/{account_id}/conferences/{conference_id}/participants/{participant_id}/mute
DELETE v2/accounts/{account_id}/conferences/{conference_id}/participants/{participant_id}/mute

Deaf/Undeaf
PUT v2/accounts/{account_id}/conferences/{conference_id}/participants/{participant_id}/deaf
DELETE v2/accounts/{account_id}/conferences/{conference_id}/participants/{participant_id}/deaf

jamhed: @k-anderson I think it's counterintuitive to denote unmute action as "delete mute", and here is why: common pattern for REST is: GET -- get object, POST -- update object, PUT/DELETE -- create/delete object. In face we are talking here about participant state "object" (which is created and persisted by server automatically, therefore no DELETE/PUT verbs should be used). We can introduce "mute" object of boolean type, and POST true/false values to it though.

applications/crossbar/src/modules/cb_conferences.erl
jamhed: Agree on ok payload, will replace it. Could you please elaborate on '_read_only' construction? What is it for, what should it tell the user?

jamesaimonetti: `_read_only` is an object for values the server generates. For instance, we used to "leak" pvt_ fields from documents; however, this caused issues when clients tried to set the field themselves. If an endpoint leaked `pvt_state` as `state` and a client was setting `state` themselves (as we don't validate `state` on the doc).

applications/crossbar/src/modules/cb_conferences.erl
jamesaimonetti: Also, in `allowed_methods`, please don't use `_` for variables. `cb_api_endpoints` uses those names to generate the ref docs. `_` will become `{ID}` which isn't too instructive. `_ConferenceId` would become `{CONFERENCE_ID}` which is much nicer. Also, if you match explicit tokens, please do so explicitly.

fenollp: How about 
PATCH /mute
PATCH /unmute
and so on?

jamhed: @fenollp What is the difference with POST? Also, I can understand mute and deaf states, but what to do with kick? It's not a state, it's an action (from any point of view). To put it broader: how should we denote actions and function calls in terms of HTTP API, once and for all?

applications/crossbar/src/modules/cb_conferences.erl
```diff
@@ -164,6 +169,40 @@ handle_participant_uri(?HTTP_POST, Context, ConfId, _ParticipantId) ->
...
+    JObj0 = crossbar_doc:update_pvt_parameters(JObj, Context),
+    case kz_datamgr:save_doc(cb_context:account_db(Context), JObj0) of
+        {'error', Error} ->
+            crossbar_doc:handle_datamgr_errors(Error, kz_doc:id(JObj0), Context);
+        {'ok', JObj1} ->
+            crossbar_doc:handle_datamgr_success(JObj1, Context)
+    end.
+
+create_conference(JObj) ->
+    SchemaJObj = find_schema(<<"conferences">>),
```
fenollp: Can't you use `Ctx = cb_context:validate_request_data(<<"conferences">>, Context)` ?

jamhed: Not directly, because request is not a conference JSON. Anyway I don't think put everything into Context is a good idea. E.g. to create/validate a json object of type "conference" you need to pass Context. Why? It's not used at all at this operations.

PR #2205, opened 2016-06-30
KAZOO-4860: make sure logging doesn't crash

jamesaimonetti:


k-anderson:
@jamesaimonetti looks like we need to update this with the same changes that went to master.

k-anderson:
@jamesaimonetti we plan to merge master into 4.0 shortly so lets actually not worry about this.

PR #2206, opened 2016-06-30
KAZOO-4860-master: make sure logging doesn't crash

jamesaimonetti:


applications/teletype/src/teletype_renderer.erl
```diff
@@ -218,7 +218,7 @@ log_warnings(Ws, Template) ->
 -spec log_infos(string(), string(), [info()], binary()) -> ['ok'].
```
fenollp: Return type is now a lie. Also to help the VM discard unused memory early, make this function return `ok`?

PR #2207, opened 2016-07-01
wait for globals ets

lazedo:


core/kazoo_globals/src/kazoo_globals_init.erl
```diff
@@ -0,0 +1,22 @@
...
+-module(kazoo_globals_init).
+
+-export([start_link/0]).
+
+-include("kazoo_globals.hrl").
+
+start_link() ->
+    wait_for_globals('false').
+
+wait_for_globals('true') ->
```
mark2600: Isn't it a bit aggressive to call this repeatedly without a sleep or wait ?

lazedo: could be, but this is a different process and we are not expecting other processing from same process. added 1 second sleep

PR #2214, opened 2016-07-03
Modernize calls to random-generating functions

kostis:
In Erlang/OTP 19 the 'random' module is marked deprecated and the 'rand'
module is expected to be used instead.  This module also exists in 18.x.

Also, the compiler issues warnings for crypto:rand_bytes/1 and prompts for
the use of crypto:strong_rand_bytes/1 instead.

This set of changes allows the kazoo code base to be compiled without errors
in 19.0 (because the current compiler options consider warnings as errors)
but most likely disables support for 17.x (because the 'rand' module is not
present there).

It should be merged to 'master' *only* if support for 17.x is not desirable.
However, if support for 17.x is still needed, the code should take out the
compiler option that turns warnings into errors, so that the code can also be
compiled in 19.0.

### DIFF
--- a/applications/acdc/src/acdc_init.erl
+++ b/applications/acdc/src/acdc_init.erl
@@ -117,7 +117,7 @@ init_agents(AccountId, {'error', _E}) ->
 init_agents(AccountId, {'ok', As}) ->
     [acdc_agents_sup:new(AccountId, kz_doc:id(A)) || A <- As].
 
-wait_a_bit() -> timer:sleep(1000 + random:uniform(500)).
+wait_a_bit() -> timer:sleep(1000 + rand:uniform(500)).
 
 try_queues_again(AccountId) ->
     try_again(AccountId, <<"queues/crossbar_listing">>, fun init_queues/2).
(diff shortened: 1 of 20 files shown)


kostis:
Now it's ready to be considered for inclusion.

fenollp:
LGTM

k-anderson:
@kostis Howdy, thanks for this work!  Unfortunately we can't merge it due to conflicts, please rebase to master and we will merge it in!

kostis:
I have no clue why some (validate json?) checks are failing and no time to deal with such issues...  sorry.  If you want this pull request please fix it and merge it, otherwise drop it.

fenollp:
@kostis From the looks of it removing your latest commit `2bda3cd` should do the trick :)

kostis:
Why should I remove the last commit?  This is the proper way to fix an unmatched return dialyzer warning involving a list comprehension that is performed only for its side-effects.  There is absolutely no reason to create a list of results only to ignore it.

fenollp:
It was just to make sure that wouldn't conflict with a pending PR, but I see now it was merged long ago.
Let me see what I can do

kostis:
I still do not understand what the CircleCI failure is about, by the way.  I suspect I would not be the only one.  If you want others to contribute with pull requests, it's not very friendly to confront them with failures they have no clue what they are about.

fenollp:
I understand & CI is pretty much always enhancing. These issues are discussed here though & that has never been an issue. 

The issue you saw is very peculiar & was due to Circle wrongly finding the file names changed in your PR.
Every CI has its own way of cloning the repo & maybe this issue appears with only PRs from forks. I'll investigate :)

fenollp:
Thanks!

PR #2220, opened 2016-07-05
readable-bins: change representation of some strings to readable UTF-8

fenollp:


k-anderson:
@fenollp nice improvement!

PR #2239, opened 2016-07-06
KAZOO-4842: migrate wh_ to kz_ while updating system_configs/modb rout…

icehess:
…ines

PR #2240, opened 2016-07-06
KAZOO-4842: migrate wh_ to kz_ while updating system_configs/modb routines for 4.0

icehess:
…ines


k-anderson:
@icehess can you issue this PR against master?

icehess:
@k-anderson PR for master is already merged #2239

PR #2241, opened 2016-07-06
qf-recovery: properly fetch user doc

fenollp:


k-anderson:
@fenollp does this return the user doc as a result of the token auth?

fenollp:
@k-anderson It returns account_id, account_name, apps, … & the public fields of the user doc.
Tested with latest Monster: totally works :)

PR #2242, opened 2016-07-06
KAZOO-4741: send transaction audit log notice after changes to service for 4.0

icehess:


k-anderson:
@icehess can you issue this PR against master?

icehess:
@k-anderson: PR for master is already merged #2223

PR #2248, opened 2016-07-07
KAZOO-4892: access call recording by media_name

onnet:


fenollp:
Hi @onnet 
CI caught a formatting issue in recordings.json. We force a certain formatting so that files are easier to diff.

Can you `make` again & commit the changes please?

lazedo:
@onnet hey, this seems to be a misuse of recordings. Can you please add  your use case to jira ticket?
Thanks

onnet:
@lazedo Done. https://2600hz.atlassian.net/browse/KAZOO-4892

jamesaimonetti:
@onnet I worry about using `/attachment` in the path. I think we made a mistake (and an inconsistent one - there is `attachement` on some and `raw` other endpoints) putting that path token in for downloading attachments.

Instead, I would love to see it use the Accept header from the client (or tunneled in the query string with `?accept=audio/mp3"). What are your thoughts on going that route over the `/attachment` path?

Sorry to be a pain but I don't want to propagate our missteps onto future developers.

lazedo:
@onnet hey, just to let you know this is not forgotten. hope to have an update next week.

jamesaimonetti:
@onnet Hi, just coming back to this to see how it was going? Do you need help with processing GET requests to `/recordings/{CDR_ID}/{MEDIA_NAME}` using the "Accept" header (or `?accept=` querystring var) instead of `/recordings/{CDR_ID}/{MEDIA_NAME}/attachment` ?

onnet:
@jamesaimonetti Hi!  Believe it or not I did it right after your comment, but it is absent somehow now everywhere :)
Maybe forgot to commit or squashed poorly. So, no help needed, will redo it again. Thanks!

jamesaimonetti:
@onnet Yeah, I hear ya on messing up when squashing commits. Ping me when you have it ready and we'll get it merged.

onnet:
@jamesaimonetti ping :)

applications/crossbar/src/modules/cb_recordings.erl
```diff
@@ -68,8 +64,8 @@ init() ->
 -spec allowed_methods(path_token(), path_token()) -> http_methods().
 
 allowed_methods() -> [?HTTP_GET].
-allowed_methods(_RecordingId ) -> [?HTTP_GET].
-allowed_methods(_RecordingId, ?ATTACHMENT) -> [?HTTP_GET].
+allowed_methods(_) -> [?HTTP_GET].
+allowed_methods(_, _) -> [?HTTP_GET].
```
lazedo: by removing `_RecordingId` the generated api doc is ID instead of `RECORDING_ID`. any reason for trying to do that ?

lazedo: if `?ATTACHMENT` is not being used, we should remove `allowed_methods/2` , same for other methods

onnet: > by removing _RecordingId the generated api doc is ID instead of RECORDING_ID. any reason for trying to do that ?
@lazedo Hi Luis, No reasons. Just didn't know why bind to _VariableName. Thanks!

PR #2253, opened 2016-07-07
KAZOO-4859: update conferences API

jamhed:
* PUT is used to perform an action on conference/participant


applications/crossbar/src/modules/cb_conferences.erl
```diff
@@ -240,15 +233,8 @@ maybe_create_conference(Context) ->
...
-                    cb_context:add_validation_error(
-                      [<<"numbers">>]
-                                                   ,<<"unique">>
-                                                   ,kz_json:from_list([
-                                                                       {<<"message">>, <<"Number already in use">>}
-                                                                      ,{<<"cause">>, Number}
-                                                                      ])
-                                                   ,Context
-                     )
+                    Error = kz_json:from_list([{<<"message">>, <<"Number already in use">>},{<<"cause">>, Number}]),
```
fenollp: Could you breakup this on at least 2 lines? to separate the two pairs

PR #2257, opened 2016-07-08
qf-style: fix indentation

fenollp:


jamesaimonetti:
@fenollp well, I think I figured out the squash/merge that Github does locally but its still not being detected here in the PR.

PR #2258, opened 2016-07-08
ecallmgr route refactor

lazedo:


applications/ecallmgr/src/ecallmgr_fs_node_sup.erl
```diff
@@ -56,8 +56,11 @@ node_srv(Supervisor) ->
 authn_srv(Supervisor) ->
     srv(supervisor:which_children(Supervisor), "nhtua_").
 
+route_sup_srv(Supervisor) ->
```
jamesaimonetti: spec!

jamesaimonetti:
@lazedo the good news is this is merged and I think I've figured out how to squash/merge these like github is doing. The bad news is I can't remove the squashed commit as I pushed it before I had really figured out what was happening. Since master is a protected branch, I can't force push the proper commits up either. So it goes...

PR #2260, opened 2016-07-08
ci-rel: generic RPC-ing to a running release

fenollp:
original idea of this PR was to incorporate cb_api_endpoints calls to CI but doc/ is not present in release...


jamesaimonetti:
@fenollp hmm, maybe I need the -no-ff in the merge. Will try on next PR

PR #2266, opened 2016-07-11
Qf tasks

fenollp:
Still race issues!

Anyone can enlighten me on how to avoid db conflicts when saving attachments & non-attachments??
These are always the source of kazoo_tasks db conflicts.

So on the server node I may be saving new doc fields occasionally.
But when the task ends it both tries to
* update some of these doc fields
* and add an output.csv attachment.

How can I explicitly say to couch to merge these changes in case of conflicts?
Like, "oh you are putting an attachment & saving some json fields? Nice, here, merge this attachment stuff on its own & these other fields with the other data" Boom.

@k-anderson

### DIFF
--- a/core/kazoo_number_manager/src/knm_tasks.erl
+++ b/core/kazoo_number_manager/src/knm_tasks.erl
@@ -30,7 +30,6 @@
         ,delete/3
         ,reserve/4
         ,add/5
-                                                %,update_features/5
         ]).
 
 -include("knm.hrl").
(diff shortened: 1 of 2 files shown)


k-anderson:
@fenollp there are no direct APIs like that but we do have an 'ensure_saved' that will handle the conflict and resubmit the save with the new _rev.  It would be best to make your doc changes first then add the attachment as the attachment takes longer to propagate into the cluster and can lead to conflicts...

PR #2267, opened 2016-07-11
caller_id_name and caller_id_number fields

jamhed:
Each participant in participants list in conference http api
have caller_id_name and caller_id_number fields as passed
by correspondent freeswtich event.


rel/relx.config.src
```diff
@@ -9,6 +9,7 @@ Filter = ['rabbitmq_codegen'
           ,'.settings'
           ,'skel'
           ,'parse_trans'
+          ,'sup'
```
fenollp: Why? This removes SUP from the generated release

jamhed: Because:
$ make && make build-release
===> Failed to solve release:
 Dependency sup is specified as a dependency but is not reachable by the system.

lazedo: @jamhed so instead of fixing it you just remove it ? whatif another app fails to build for release ? would you remove it ?

jamhed: @lazedo, this was unintended, this commit just slipped in pull request from my working branch.

lazedo: @jamhed no worries but, you could have make that comment on @fenollp comment

jamhed: @lazedo, to make a notice that make build-release is broken (starting from now)

jamhed: ... or maybe not.

PR #2270, opened 2016-07-13
KAZOO-4857: move account_crawler to ananke and use amqp_cron for scheduling

icehess:


icehess:
@fenollp Is there a way to turn off circleci for 4.0 branch?

fenollp:
I thought we were targeting master, then @k-anderson was going to make master 4.0 so we won't have to do multiple PRs.

But I'll push a circle.yml for 4.0.

icehess:
I don't about making master 4.0, but this circleci could be and issue for 3.22 as well.

fenollp:
Merge #2277 ;)

PR #2272, opened 2016-07-13
KAZOO-4857(master): move account_crawler to ananke

icehess:
and use amqp_cron for scheduling


fenollp:
You'll have to rebase to fix circle: fetch master then rebase on top of origin/master.

icehess:
weird, i did rebase. doing it again.

applications/notify/src/notify_low_balance.erl
```diff
@@ -77,7 +77,7 @@ send(CurrentBalance, Account) ->
 -spec create_template_props(integer(), kz_json:object()) -> kz_proplist().
 create_template_props(CurrentBalance, Account) ->
     AccountDb = kz_util:format_account_id(kz_doc:id(Account), 'encoded'),
-    Threshold = notify_account_crawler:low_balance_threshold(AccountDb),
+    Threshold = ananke_account_crawler:low_balance_threshold(AccountDb),
```
k-anderson: @icehess you should not use functions from another application ;)  Perhaps this should be in core or in kazoo_documents...

applications/ananke/src/ananke.app.src
```diff
@@ -1,12 +1,21 @@
 {application, ananke,
  [
-  {description, "Ananke was the personification of destiny, necessity and fate, depicted as holding a spindle in ancient Greek religion. This application to schedule calls on events."},
```
k-anderson: @icehess why did you remove the description?  I liked it, maybe put it back?

applications/ananke/src/ananke_callback_sup.erl
```diff
@@ -1,88 +0,0 @@
-%%%-------------------------------------------------------------------
```
k-anderson: @icehess is this the task sup below now?  the callback feature still works correct?

icehess: @k-anderson yes this is moved to `ananke_tasks_sup`. vm_callblack should be working as before but it needs refactoring and code clean up anyway in another PR maybe. I fixed some broken code there though.

k-anderson:
@icehess Looks like we need to resolve:
Dependency ananke is specified as a dependency but is not reachable by the system

PR #2274, opened 2016-07-13
qf-tasks: save some time when converting JSON to CSV

fenollp:


k-anderson:
@fenollp why not just handle JSON in the task?

fenollp:
Doable but more trouble than I can do right now.
It would mean doing the conversion when fetching input.csv is done... some assumptions change too

PR #2276, opened 2016-07-13
KAZOO-4854: try to connect default fs node if none configured

jamhed:


applications/ecallmgr/src/ecallmgr_fs_nodes.erl
```diff
@@ -760,25 +760,47 @@ start_node_stats(#node{}) ->
...
+                     'success';
+                 'undefined' ->
+                     lager:debug("failed to receive a response for node configs");
+                 _E ->
+                     lager:debug("received a non-list for fs_nodes: ~p", [_E])
+             end,
+    case Result of
+        'success' -> 'ok';
+        _ ->
+            _ = maybe_add_default_fs(),
```
k-anderson: @jamhed if you are able to connect to the default then don't enter the polling stage

applications/ecallmgr/src/ecallmgr_fs_nodes.erl
```diff
@@ -760,25 +760,47 @@ start_node_stats(#node{}) ->
...
+-spec maybe_add_default_fs() -> 'skip' | 'ok' | {'error', 'no_connection'}.
+maybe_add_default_fs() ->
+    Node = default_fs_node(),
+    lager:info("attempting to connect default freeswitch node ~p", [Node]),
+    case net_adm:ping(Node) of
+        'pong' -> add(Node);
+        _ -> 'skip'
+    end.
+
 start_node_from_config(MaybeJObj) ->
```
k-anderson: @jamhed I know that its not your function, but if you seeing something without a spec please correct it ;)

applications/ecallmgr/src/ecallmgr_fs_nodes.erl
jamhed: @k-anderson what do you mean? the logic here is that first we try a configuration parameter and if it fails for any reason then we try to connect to default node.

applications/ecallmgr/src/ecallmgr_fs_nodes.erl
jamhed: @k-anderson ok

applications/ecallmgr/src/ecallmgr_fs_nodes.erl
k-anderson: @jamhed Howdy, I meant something like:

case maybe_add_default_fs() of
    'ok' -> 'ok';
    _Else ->
             timer:sleep(5 * ?MILLISECONDS_IN_SECOND),
              start_preconfigured_servers()
end.

PR #2279, opened 2016-07-14
Fix crossbar versioned endpoint token parsing

kalda341:
In some cases a versioned crossbar endpoint (for example cb_lists_v2) will not work as the non versioned atom (in this case cb_lists) does not exist. This fixes this bug.

### DIFF
--- a/applications/crossbar/src/api_util.erl
+++ b/applications/crossbar/src/api_util.erl
@@ -605,8 +605,8 @@ is_cb_module(Context, Elem) ->
     try (kz_util:to_atom(<<"cb_", Elem/binary>>)):module_info('exports') of
         _ -> 'true'
     catch
-        % try module_version, regardless of whether it was the atom or the
-        % module which did not exist
+        %% try module_version, regardless of whether it was the atom or the
+        %% module which did not exist
         _E:_R -> is_cb_module_version(Context, Elem)
     end.
 
--- a/applications/crossbar/src/api_util.erl
+++ b/applications/crossbar/src/api_util.erl
@@ -605,7 +605,8 @@ is_cb_module(Context, Elem) ->
     try (kz_util:to_atom(<<"cb_", Elem/binary>>)):module_info('exports') of
         _ -> 'true'
     catch
-        'error':'badarg' -> 'false'; %% atom didn't exist already
+        % try module_version, regardless of whether it was the atom or the
+        % module which did not exist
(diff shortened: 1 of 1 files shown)


k-anderson:
@kalda341 Howdy, thanks for this.  I recently saw this in another environment but the root issue was they had misconfigured the system.  Just out of curiosity in you crossbar modules list in system_config did you have the root module name (cb_lists) or do you have each version listed (cb_lists_v1, cb_lists_v2)?

The system is designed to have the root module name (just cb_lists) in the start up and will discover the version specific modules then load the atom table with cb_lists.  

If you load the two version modules directly then this doesn't happen, and for the previous reporter, caused the same experience you are reporting. 

I think we could expand the system to support both since this appears to have happened to independent operators ;) Just need to have someone double check this change in behavior has no unexpected side-effects!

jamesaimonetti:
I think you need to use '%%' for comments so the formatting doesn't get messed up. '%' is for end-of-line comments.

kalda341:
You are correct in that cb_lists is not present in the autoload_modules. We have cb_lists_v1 and cb_lists_v2 instead.

kalda341:
Thanks for the comment about comments - I had wondered when to use "%" and "%%".

kalda341:
I believe cb_lists_v1 and cb_lists_v2 appeared in our autoload_modules as a result of running sup "crossbar_maintenance start_module" with the module name containing _v1 and _v2 at the end.
If you decide not to support both then perhaps it makes sense to disallow running the commands in this way?

k-anderson:
@kalda341 indeed, so that would do it.  The suggestion to disallow makes sense but give me some time to check with the team to see how we want to proceed. :)

lazedo:
I believe the root cause is the registration process for cb modules that
does handle versioned modules, that's why people use the version when
registering, there is already a ticket to fix this.

lazedo:
@kalda341 hey, check if #2285 fixes all your issues

k-anderson:
@kalda341 Howdy, with the recent PR from @lazedo (#2285) I think this issue has been addressed.  Can you test it out and see if everything is working as expected for ya now?  If so please close this PR otherwise we can continue the discussion here!
Thanks :)

kalda341:
Yes it does :)

lazedo:
@kalda341 thanks for testing.

PR #2280, opened 2016-07-14
KAZOO-4807: skip temporal rule

jamhed:
Skip temporal rule when when rule.enable == undefined.

### DIFF
--- a/applications/callflow/src/module/cf_temporal_route.erl
+++ b/applications/callflow/src/module/cf_temporal_route.erl
@@ -76,10 +76,10 @@ handle(Data, Call) ->
 %%--------------------------------------------------------------------
 -spec process_rules(temporal(), rules(), kapps_call:call()) ->
                            'default' | binary().
-process_rules(Temporal, [#rule{enabled='false'
+process_rules(Temporal, [#rule{enabled=Enabled
                               ,id=Id
                               ,name=Name
-                              }|Rs], Call) ->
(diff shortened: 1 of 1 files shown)


applications/callflow/src/module/cf_temporal_route.erl
```diff
@@ -76,10 +76,10 @@ handle(Data, Call) ->
 %%--------------------------------------------------------------------
 -spec process_rules(temporal(), rules(), kapps_call:call()) ->
                            'default' | binary().
-process_rules(Temporal, [#rule{enabled='false'
+process_rules(Temporal, [#rule{enabled=Enabled
```
k-anderson: @jamhed SO enabled in this context is a little confusing in that it SHOULD be a 'tri-state' parameter.

TRUE: ignore the time-of-day settings and assume the rule matches ALL TIMES
FALSE: ignore the time-of-day settings and assume the rule can NEVER match
'undefined': process the time-of-day-settings

As a simplistic example, if an account has a time-of-day for their office hours then they could set this to false during a holiday to keep the office closed.  If, instead, they were in the office when they would otherwise be closed and want to receive calls (despite the time-of-day not matching) they would set it to true.  If they want to use the configured time of day to determine if the office is open or closed then enabled parameter is removed and the rule operates normally.

Think of it more like "force enabled"....the name was fairly poor but we need to maintain for compatibility.

PR #2284, opened 2016-07-18
KAZOO-4905 handle transfer events in local extension

lazedo:


applications/stepswitch/src/stepswitch_local_extension.erl
```diff
@@ -177,19 +190,46 @@ handle_info(_Info, State) ->
...
+        {<<"call_event">>, <<"CHANNEL_DESTROY">>, CallId} ->
+            lager:debug("channel was destroyed while waiting for bridge"),
+            Result = case kz_json:get_value(<<"Disposition">>, JObj)
+                         =:= <<"SUCCESS">>
+                     of
+                         'true' -> local_extension_success(Request);
+                         'false' -> local_extension_failure(JObj, Request)
+                     end,
+            gen_listener:cast(RequestHandler, {'local_extension_result', Result});
+        {<<"call_event">>, <<"CHANNEL_EXECUTE_COMPLETE">>, CallId} ->
```
fenollp: I don't get it. Why turn a 2-tuple comparison into a 3-tuple comparison when you don't even guard against the 3rd value?

This line and the clause just before indeed use the 3rd value you introduced, but these clauses could have done the same by using `CallId = kz_call_event:call_id(JObj)` themselves.
Also, the 3rd value isn't even read by anything else. Should revert to 2-tuple IMO.

lazedo: @fenollp you couldn't be more wrong

fenollp: Ah right I just realised I missed CallId being bound before the `case`.
Thanks for the explaination

applications/stepswitch/src/stepswitch_local_extension.erl
```diff
@@ -177,19 +190,46 @@ handle_info(_Info, State) ->
...
                        ,[kz_util:to_binary(kz_json:encode(JObj))]),
             gen_listener:cast(RequestHandler, {'local_extension_result', local_extension_error(JObj, Request)});
-        {<<"call_event">>, <<"CHANNEL_DESTROY">>} ->
-            gen_listener:cast(RequestHandler, {'local_extension_result', local_extension_success(Request)});
-        {<<"call_event">>, <<"CHANNEL_EXECUTE_COMPLETE">>} ->
+        {<<"call_event">>, <<"CHANNEL_TRANSFEREE">>, _} ->
+            Transferee = kz_call_event:other_leg_call_id(JObj),
+            gen_listener:cast(RequestHandler, {'replaced', Transferee}),
+            gen_listener:add_binding(RequestHandler, ?CALL_BINDING(Transferee));
+        {<<"call_event">>, <<"CHANNEL_TRANSFEROR">>, _} ->
```
k-anderson: @lazedo Tracking replaced makes sense but why treat the other leg of the transferee/transferor as the stepswitch_local_extension.

For example, if we a local extension A -> B was created,  then transferred such that A -> C we should track A as the local extension.  However, if we receive the transferee event (for A) and take the other leg (C) then the local extension would then be tracking C, no?

For the transferor situation, if A -> B and A preforms the transfer such that B -> C then the original local extension (again A) has been destroyed.  Why would local extension track B in this case?

(Please note these are from  my head I will need to deploy this branch to confirm... just surprised that the replaced event was not sufficient)

applications/stepswitch/src/stepswitch_local_extension.erl
```diff
@@ -30,6 +30,8 @@
                ,response_queue :: api_binary()
                ,queue :: api_binary()
                ,timeout :: reference()
+               ,call_id :: api_binary()
+               ,call_ids :: api_binaries()
```
k-anderson: @lazedo Why track the call_ids, is this for future use?  I looked for where you use this and didn't initially see anything other than collection.

applications/stepswitch/src/stepswitch_local_extension.erl
lazedo: @k-anderson code was `copied` from `stespswitch_bridge` as attended xfers from A were not working

applications/stepswitch/src/stepswitch_local_extension.erl
lazedo: @k-anderson same as above

applications/stepswitch/src/stepswitch_local_extension.erl
k-anderson: @lazedo then the question expands to that code as well ;) 

This does not feel right, am I incorrect that we would be tracking the call id of channel that was not generated by the local_extension (and by extension stepswitch_bridge) in these situations?

PR #2285, opened 2016-07-18
KAZOO-4841 handle versioned modules

lazedo:


applications/crossbar/src/crossbar_config.erl
```diff
@@ -25,7 +25,24 @@ autoload_modules() ->
...
+remove_module_version(Module) ->
+    maybe_remove_module_version(lists:reverse(binary_to_list(Module))).
+
+-spec maybe_remove_module_version(list()) -> binary().
+maybe_remove_module_version("1v_" ++ Module) ->
+    list_to_binary(lists:reverse(Module));
+maybe_remove_module_version("2v_" ++ Module) ->
+    list_to_binary(lists:reverse(Module));
+maybe_remove_module_version(Module) ->
+    list_to_binary(lists:reverse(Module)).
```
fenollp: These 3 calls to `list_to_binary(lists:reverse(` should be moved to `remove_module_version/1`

applications/crossbar/src/crossbar_init.erl
```diff
@@ -63,10 +63,54 @@ start_link() ->
...
+is_versioned_module(Module) ->
+    Mod = lists:reverse(binary_to_list(Module)),
+    case Mod of
+        "1v_" ++ _ -> 'true';
+        "2v_" ++ _ -> 'true';
+        _ -> false
+    end.
+
+-spec start_mod(atom() | string() | binary()) -> 'ok' | {'error', any()}.
+start_mod(CBMod) when not is_atom(CBMod) ->
```
fenollp: Given `is_versioned_module/1` I take `when not is_atom(` should actually be `when is_binary(`

applications/crossbar/src/crossbar_config.erl
lazedo: @fenollp why ?

applications/crossbar/src/crossbar_init.erl
lazedo: @fenollp say again ?

fenollp: Since is_versioned_module expects binaries (or it will crash), then the `not is_atom` guard should actually be `is_binary`

applications/crossbar/src/crossbar_config.erl
fenollp: This way binary_to_list and list reversing stay in the same function. It makes it easier to understand.
Also it saves reading 3x `list_to_binary(lists:reverse(`

lazedo: @fenollp and ? and no it doesn't make it easier to understand. but thanks for the thought.

k-anderson:
@lazedo howdy! I see that in this change you have added a few new notice log lines.  There is a standing ask to only log overall failure/success to start the crossbar module at notice because the intermediate "failures" are confusing.

As such I would like to request that you downgrade the logs related to seeking alternative modules to debug and only have the final status logged at notice.

lazedo:
@k-anderson updated

PR #2286, opened 2016-07-18
bugfix-attributes-view: follow-up of #2283

fenollp:


k-anderson:
@fenollp howdy, so this is the second change moving to the view name to attributes.  Without checking but knowing that @jamesaimonetti recently moved this to the core library I would have expected the view name kz_attributes to be the more recent (expected) view name.  As such are we patching code because the intended view is not being added to the system and as such a bug in the new core library?

PR #2287, opened 2016-07-18
qf-tasks: great speedup when creating tasks by relaxing requirements

fenollp:
This gives 5x and maybe more (theoretically a quadratic improvement)


k-anderson:
@fenollp this is interesting, I can see how the check for unique-ness could be expensive on such large jobs.  Please file a ticket with your findings so that we can investigate alternatives.  However, since the jobs themselves will fail non-unique entries I do agree that this is a redundant check at this time.

PR #2288, opened 2016-07-18
qf-tasks-speed: lighting fast task input validation

fenollp:
Even "better" than https://github.com/2600hz/kazoo/pull/2287 This is a XOR pull request.


k-anderson:
@fenollp removing validation is not a valid approach... :)

fenollp:
I agree :)

PR #2289, opened 2016-07-18
KAZOO-4826 handle activation charges

lazedo:


fenollp:
The fold for `create_transactions` L80 could be done as a LC (or lists:filter | lists:filtermap) now that it is only copying `Acc` and not reading it anymore.

PR #2295, opened 2016-07-19
propagate call parking events via websockets

jamhed:


jamesaimonetti:
Accepted, still can't figure out how to get GH to auto-recognize that fact

PR #2296, opened 2016-07-19
kapps: better time complexity for account_has_descendants/1

fenollp:


core/kazoo_apps/src/kapps_util.erl
```diff
@@ -207,12 +207,16 @@ account_depth(Account) ->
...
     ViewOptions = [{'startkey', [AccountId]}
                   ,{'endkey', [AccountId, kz_json:new()]}
                   ],
-    {'ok', JObjs} = kz_datamgr:get_results(?KZ_ACCOUNTS_DB, <<"accounts/listing_by_descendants">>, ViewOptions),
-    length([JObj || JObj <- JObjs, kz_account:id(JObj) =/= AccountId]) > 0.
+    {'ok', JObjs} = kz_datamgr:get_results(?KZ_ACCOUNTS_DB, View, ViewOptions),
+    [] =/=
+        [JObj || JObj <- JObjs,
+                 kz_account:id(JObj) =/= AccountId
+        ].
```
jamesaimonetti: What size view result did you test against? What were the time comparisons?

jamesaimonetti: Also, why not write a recursive function looking for the first instance of JObj that isn't AccountId? Then recursion can terminate early if a non-AccountId account was found.

fenollp: It's just something I came about when fixing something else. No benchmarking involved.
But yeah I should rewrite this with recursion and return ASAP.

PR #2300, opened 2016-07-19
KAZOO-4886: add PATCH support to devices

jamesaimonetti:
refactor fold to own named function
fix function call, specs
formatting


core/kazoo_endpoint/src/kz_endpoint.erl
```diff
@@ -568,21 +569,26 @@ maybe_rewrite_caller_id(Endpoint, Call) ->
 -spec should_create_endpoint(kz_json:object(), kz_json:object(), kapps_call:call()) ->
                                     'ok' | {'error', any()}.
 should_create_endpoint(Endpoint, Properties, Call) ->
+    case evaluate_rules_for_creation(Endpoint, Properties, Call) of
+        {Endpoint, Properties, Call} -> 'ok';
```
fenollp: Is the matching to already bound `Endpoint` `Properties` & `Call` intensional?

Other than that PR LGTM

PR #2306, opened 2016-07-20
qf-views: update view names in KNM

fenollp:


fenollp:
Since last commit, I'm not sure it will pick up views like `applications/crossbar/priv/couchdb/account/numbers.json` as I am not sure where / what revises it.

jamesaimonetti:
Merged.

PR #2312, opened 2016-07-20
cccp update for 4.0

onnet:
Unfortunately I wasn't able to divide these changes into several steps, 'cause it evolved in time and interconnected. Tested on my single node 4.0 (master).

### DIFF
--- a/applications/cccp/doc/index.md
+++ b/applications/cccp/doc/index.md
@@ -4,14 +4,32 @@ Title: Calling Card Callback Platform
 Language: en-US
 */
 
-cccp
+cccp - Calling Card Callback Platform
 ====
-Calling Card Callback Platform
 
 Could be good for road warriors:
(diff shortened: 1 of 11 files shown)


jamesaimonetti:
Hi @onnet it looks like there's some trailing whitespace in cccp_util.erl and a couple dialyzer complaints (see CircleCI's output for details).

onnet:
Hi @jamesaimonetti  Yep, oh those specs.. :)

jamesaimonetti:
@onnet this latest failure is a bit of an annoyance, but `%` should be for comments at the end of a line while `%%` should be whole-line comments.

Even better, though, is to remove dead code/specs!

onnet:
@jamesaimonetti 
So, do you have any idea of how to get rid of this 'annoyance'?

PR #2320, opened 2016-07-21
KAZOO-4933: adding fields beta/alpha to apps store view

JRMaitre:


lazedo:
@JRMaitre why do you add `alpha: doc.alpha, beta: doc.beta` and not something like `state` or `another_field_that_reflects_stage_of_app` ?  can an application be alpha and beta at the same time ?

PR #2328, opened 2016-07-21
Help 23396

jamesaimonetti:


core/kazoo_endpoint/Makefile
```diff
@@ -3,4 +3,9 @@ PROJECT = kazoo_endpoint
 
 all: compile
 
+test: kze-test
+
+kze-test:
+	$(MAKE) compile-test -C $(ROOT)/core/kazoo_number_manager/
```
fenollp: I am not sure I like this. How to run the test suite is [documented here](https://github.com/2600hz/kazoo/blob/master/doc/installation.md#longer-version) and [demonstrated here](https://github.com/2600hz/kazoo/blob/master/.travis.yml#L31-L32).

It shows a test dependency and that I like. However KNM itself requires the whole Kazoo to be `make compile-test`-ed for its own tests to run…

Yeah maybe that can work until we have a real test setup…
otherwise PR LGTM

jamesaimonetti: The explicit test dep is what I was going for here. I don't need all of kazoo test-compiled when running kazoo_endpoint tests, so this was the shortest path to having kazoo_endpoint tests run successfully.

PR #2330, opened 2016-07-22
More uses of kzd_app

fenollp:
also formatting, some refactor into LCs & less funs.

### DIFF
--- a/applications/crossbar/src/cb_apps_util.erl
+++ b/applications/crossbar/src/cb_apps_util.erl
@@ -44,8 +44,10 @@ allowed_apps(AccountId) ->
 %%--------------------------------------------------------------------
 -spec allowed_app(ne_binary(), ne_binary()) -> api_object().
 allowed_app(AccountId, AppId) ->
-    FindApp = fun(App) -> AppId == kz_doc:id(App) end,
-    case lists:filter(FindApp, allowed_apps(AccountId)) of
+    case [App || App <- allowed_app(AccountId),
+                 AppId =:= kzd_app:id(App)
+         ]
+    of
         [App] -> App;
         [] -> 'undefined'
     end.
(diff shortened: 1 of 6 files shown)


applications/crossbar/src/cb_apps_util.erl
```diff
@@ -298,33 +281,20 @@ get_plan_apps(ServicePlan) ->
...
-            lager:debug("excluding app ~s(~s)", [AppName, AppId]),
-            Acc;
-        'undefined' ->
-            lager:debug("excluding app ~s(~s)", [AppName, AppId]),
-            Acc;
-        AppJObj ->
-            lager:debug("including app ~s(~s) from plan", [AppName, AppId]),
-            [kz_json:delete_key(<<"published">>, AppJObj)|Acc]
-    end.
+    [kz_json:delete_key(<<"published">>, MaybeAppJObj)
```
k-anderson: @fenollp Howdy, this PR looks good except for this... I would not describe this as more readable.  Why the change to list comprehensions?

fenollp: Ah yes I agree I have made a bad choice here. Will revert

fenollp:
@k-anderson done

core/kazoo_documents/src/kzd_app.erl
```diff
@@ -225,18 +225,26 @@ account_id(JObj) ->
 %% @doc
 %% @end
 %%--------------------------------------------------------------------
--spec users(kz_json:object()) -> api_binaries().
+-spec users(kz_json:object()) -> api_objects().
```
lazedo: @fenollp isn't users a list of Ids ? `is_authorized` in cb_apps_util does a `lists:member(UserId, UserIds);`

PR #2334, opened 2016-07-25
KAZOO-4935 add transfer to dialplan

lazedo:


applications/ecallmgr/src/ecallmgr_call_command.erl
```diff
@@ -544,6 +545,12 @@ get_fs_app(_Node, _UUID, JObj, <<"fax_detection">>) ->
             end
     end;
 
+get_fs_app(Node, UUID, JObj, <<"transfer">>) ->
```
fenollp: spec

lazedo: @fenollp look at top of get_fs_app/4

applications/ecallmgr/src/ecallmgr_call_command.erl
```diff
@@ -1529,3 +1536,48 @@ tone_duration_on(Tone) ->
...
+           ],
+    Props = [KV || {K,_V} = KV <- CCVs,
+                   lists:member(K, CCVList)
+            ]  ++ Vars,
+    [Export | Exports] = ecallmgr_util:process_fs_kv(Node, UUID, Props, 'set'),
+    Arg = list_to_binary([Export
+                         ,[[",", Exported] || Exported <- Exports]
+                         ]
+                        ),
+    {<<"att_xfer">>, list_to_binary(["{", Arg, "}loopback/", TransferTo, <<"/">>, transfer_context(JObj)])};
```
fenollp: Looks like Arg does not need to be turned into binary as it will be done here. Or maybe with iolist_to_binary

applications/ecallmgr/src/ecallmgr_call_command.erl
fenollp: ah I missed that, right

PR #2336, opened 2016-07-25
KAZOO-4912: filter out private commentns from port_request API GET's

mark2600:


lazedo:
@mark2600 how about returning these private comments if the auth-id  is super_duper ?

k-anderson:
@mark2600 as discussed we should filter patch as well.

PR #2341, opened 2016-07-25
KAZOO-4936: use set_values/2

jamesaimonetti:
from_list/1 wasn't using the key as a path but as an array-as-key.
merge error into errors in cb_context
handle wrong_length validation error


lazedo:
@fenollp did you test this before merging ?

PR #2343, opened 2016-07-25
KAZOO-4914: activate numbers after transition ports to complete

icehess:


core/kazoo_number_manager/src/knm_numbers.erl
```diff
@@ -30,17 +30,19 @@
 -include("knm.hrl").
 
 -type number_return() :: {ne_binary(), knm_number:knm_number_return()}.
--type numbers_return() :: [number_return()].
+-type knm_numbers_return() :: [number_return()].
```
fenollp: Why this? Also, it breaks

```
applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl:843:                            knm_numbers:numbers_return().
core/kazoo_number_manager/src/knm_port_request.erl:336:-spec was_number_operation_successul(knm_numbers:number_return()) -> boolean().
```

core/kazoo_number_manager/src/knm_port_request.erl
```diff
@@ -332,6 +332,19 @@ transition_numbers(PortReq) ->
...
     end.
 
+-spec create_and_activate(ne_binaries(), knm_number_options:options()) -> knm_numbers:knm_numbers_return().
+create_and_activate(Numbers, Options) ->
+    CreationResults = knm_numbers:create(Numbers, Options),
+    {_OK, Errored} = lists:partition(fun was_number_operation_successul/1, CreationResults),
+    case Errored of
+        [] ->
+            lager:debug("all numbers created, activating numbers"),
+            knm_numbers:move(Numbers, knm_number_options:assign_to(Options), Options);
```
fenollp: This is something that should be achieved by knm_numbers:create's assign_to already. The bug is elsewhere *X-Files theme*.

Since ticket was created multiple versions of this file ago, I'm curious of what crash caused this.

icehess: @fenollp There was a problem with transition (pending/schedule)->complete ports which I fixed by setting `state` to `available` but it causes the numbers to be in `reserved` state, so here I used `knm_numbers:move` to set the state to `in_service`, just like what `activate` is doing in api.

icehess:
@fenollp with changes in #2345 I think there is no need for this pr because we can set `state` to `in_service` when we are creating the numbers, am I correct?

fenollp:
@icehess That is true! So I'll open a new PR reverting this & setting state to in_service.

fenollp:
Actually, once #2345 is merged.

PR #2351, opened 2016-07-27
KAZOO-4914: transition ported in numbers to in_service

fenollp:
Completion of #2340 and #2343.


lazedo:
@icehess merged this to get it in sandbox. please, still verify the changes

PR #2368, opened 2016-07-28
KAZOO-4908: check for all querystring flag and return entire config d…

mark2600:
…oc if set to true


k-anderson:
@mark2600 sean would really like to have all the default and I would be ok with that.  Thoughts?  I know that it would change the API so if we are worried about backward compatability perhaps have it the default for a v2 API variant?

fenollp:
How about add a new API: system_configs/defaults/all or system_configs/all_defaults or system_configs/defaults ?

mark2600:
Lets discuss in today's standup and come to an agreement.

PR #2369, opened 2016-07-28
swagger: drop API endpoints that are no longer provided

fenollp:


fenollp:
Please try `make apis` locally before merging. I'm afraid something is weird

lazedo:
@fenollp needs rebase

core/kazoo_ast/src/cb_api_endpoints.erl
```diff
@@ -465,7 +451,6 @@ path_name(<<_/binary>>=Module) ->
                ,[{'capture', 'all_but_first', 'binary'}]
                )
     of
-        {'match', [<<"about">>]} -> <<?CURRENT_VERSION/binary, "/about">>;
```
jamesaimonetti: If you look at cb_about, you will see there is nothing related to accounts going on in there. If `/v2/about` isn't working then `cb_about` should be fixed to return properly.

applications/crossbar/doc/ref/about.md
```diff
@@ -8,11 +8,11 @@
 
 #### Fetch
 
-> GET /v2/about
+> GET /v2/accounts/{ACCOUNT_ID}/about
```
lazedo: '/v2/about' is the valid one.

applications/crossbar/priv/api/swagger.json
```diff
@@ -12702,37 +12702,9 @@
         "version": "2.0"
     },
     "paths": {
-        "/v2/about": {
-            "get": {}
-        },
```
lazedo: is `make apis` failing to include `"/v2/about"` ? do we know the reason ? are there more endpoints in the same condition ?
thanks

fenollp: there are more. Apparently all paths without masquerading part are not discovered but hard coded

jamesaimonetti: It was easiest to code by assuming the API endpoint operated under an account, and then manually update cb_api_endpoints to include logic to override that assumption. There's nothing in the endpoint modules themselves that would easily identify those APIs that operate in an account, without an account, or with either. [This](https://github.com/2600hz/kazoo/pull/2369/files/057c90b0d82dc492918e6da1684301b1da34e136#diff-78a153b2268aba9e7897d49d708b1f26L468) change is why the endpoint above is being removed.

jamesaimonetti: And to be clear, cb_api_endpoints was only meant to augment/hint at what was found in the code; it was never meant to be the authority on it.

lazedo: @jamesaimonetti yep, we know that.
how about using module attributes to mark these ? something like 
```
-api(root).
```
and then we could use Module:module_info(attributes) and look for `api` ?

fenollp: I like the attributes idea. 
But what if a module provides both a root and masqueradable API?

What is the mechanism that enables or disables prepending accounts in path?
Ah I remember: it's s matter of whether account_id is defined in Context. 

Maybe we can add an explicit way of creating APIs to root?
Something like allowed_root_methods(). 
That would be parsable but requires updating all cb_*

lazedo: @fenollp the attribute(s) can what ever you want, it was just an example.
i would go first with marking the root api's.

jamesaimonetti: I wanted to avoid magic ways of finding relevant information. I don't want developers to have to remember to add an attribute to tag the module as root, masqueraded, both, etc. I think encoding the "root" modules in cb_api_endpoints makes the most sense. When a new "root" endpoint is developed, the ref doc will have `/v2/accounts/{ACCOUNT_ID}/new_endpoints` and such. The dev creating the new endpoint will see that and have to adjust cb_api_endpoints to account for the new endpoint. I don't see this being a frequent-enough occurrence to warrant compiler attributes.

We can discuss more after the standup.

PR #2370, opened 2016-07-28
some more input checking for KNM

fenollp:
(disregard branch name)


core/kazoo_number_manager/src/knm_phone_number.erl
```diff
@@ -424,7 +424,7 @@ setters_fold_apply(Fun, Args) ->
 number(#knm_phone_number{number=Num}) -> Num.
 
 -spec set_number(knm_phone_number(), ne_binary()) -> knm_phone_number().
-set_number(N, NormalizedNum=?NE_BINARY) ->
+set_number(N, <<"+",_:8,_/binary>>=NormalizedNum) ->
```
lazedo: @fenollp why the 8 ?

PR #2376, opened 2016-07-29
cb_channels: Call to target device and intercept given UUID to it, after answer.

skorobkov:
This combination of `quickcall` and `call pickup` for `cb_channels`.
For now its only call to one device, `user` target will be added later.

### DIFF
--- a/applications/konami/src/module/konami_intercept.erl
+++ b/applications/konami/src/module/konami_intercept.erl
@@ -1,0 +1,197 @@
+%%%-------------------------------------------------------------------
+%%% @copyright (C) 2016, 2600Hz
+%%% @doc
+%%% data:
+%%%   target_type: "user" or "device"
+%%%   target_id: Device-ID or User-ID where Call-ID will be bridged (intercepted)
+%%% @end
+%%% @contributors
+%%%   Peter Defebvre
+%%%   Sergey Korobkov
+%%%-------------------------------------------------------------------
+-module(konami_intercept).
+
+-export([handle/2]).
+
+-include("konami.hrl").
+
+-spec handle(kz_json:object(), kapps_call:call()) ->
+                    {'continue', kapps_call:call()} |
+                    {'stop', kapps_call:call()}.
+handle(Data, Call) ->
+    case send_originate_req(get_originate_req(Data, Call), Call) of
+        {'ok', Result} ->
+            maybe_update_metaflow(Data, Call, Result);
+        {'error', _E} ->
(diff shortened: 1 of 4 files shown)


lazedo:
@skorobkov hey, this is cool! some thoughts.
originate+intercept is more like  `move` and there's currently an implementation available in konami/ konami_move action.

i know its not available in cb_channels but, the other implementations (`callflow`, `hangup`, `transfer`) follow the pattern of collecting the information and publish a metaflow request which will be listened by konami and execute the action.

this way, a new action will be available for both channels api and dtmf bindings. 

about the formatting, you may want to run "make fmt"

Best

skorobkov:
Ok. I will try rewrite to metaflow/konami.

lazedo:
@skorobkov great. let us know if there's anything we can help you with.

skorobkov:
Problem with `konami_move` it's bridge **opposite leg** of given Call-ID. And it wont work in scenario where is no other leg (ex. parked call).
What i should do? Modify konami_move (add some special paramaters in `data`)? Or create konami_intercept (based on konami_move)?

skorobkov:
Trying use konami_move.
Adding paramter `intercept_logic:`true` to "data" will swap channels in chosing which one is should be moved.

lazedo:
@skorobkov Hi,  the `konami_intercept` module seems to be missing ?

skorobkov:
:sweat:

skorobkov:
Rebased to master & squashed.

applications/konami/src/module/konami_intercept.erl
```diff
@@ -0,0 +1,197 @@
...
+
+-spec build_originate(kz_json:objects(), ne_binary(), kapps_call:call()) -> kz_proplist().
+build_originate([], _CallId, _Call) -> [];
+build_originate(Endpoints, CallId, Call) ->
+    lager:debug("targeting ~s for intercept", [CallId]),
+    props:filter_undefined(
+      [{<<"Application-Name">>, <<"bridge">>}
+      ,{<<"Endpoints">>, Endpoints}
+      ,{<<"Existing-Call-ID">>, CallId}
+      ,{<<"Intercept-Unbridged-Only">>, 'false'}
```
lazedo: @skorobkov this should be `true` for an `intercept` action or at least fetched from Data (cb_channels should get from params to build it)

lazedo:
@skorobkov thanks for this. please note that we may refactor the `build_endpoints` and `build_originate` in order to de-duplicate code existing in `konami_move` and `konami_intercept`

PR #2383, opened 2016-07-29
KAZOO-4916 (also with KAZOO-4201)

jamhed:
- keep binding in ws context and insert it into every handled message
- adapt bh_call to ws protocol changes: low-case keys, subscribe/unsubscribe, binding field

### DIFF
--- a/applications/blackhole/src/modules/bh_call.erl
+++ b/applications/blackhole/src/modules/bh_call.erl
@@ -22,13 +22,14 @@
                    ]).
 
 -spec handle_event(bh_context:context(), kz_json:object()) -> 'ok'.
-handle_event(Context, EventJObj) ->
+handle_event(#bh_context{binding=Binding} = Context, EventJObj) ->
     kz_util:put_callid(EventJObj),
     lager:debug("handle_event fired for ~s ~s", [bh_context:account_id(Context), bh_context:websocket_session_id(Context)]),
     'true' = kapi_call:event_v(EventJObj)
         andalso is_account_event(Context, EventJObj),
     lager:debug("valid event and emitting to ~p: ~s", [bh_context:websocket_pid(Context), event_name(EventJObj)]),
-    blackhole_data_emitter:emit(bh_context:websocket_pid(Context), event_name(EventJObj), EventJObj).
+    NormJObj = kz_json:normalize_jobj(kz_json:set_value(<<"Binding">>, Binding, EventJObj)),
+    blackhole_data_emitter:emit(bh_context:websocket_pid(Context), event_name(EventJObj), NormJObj).
 
 is_account_event(Context, EventJObj) ->
     kz_json:get_first_defined([<<"Account-ID">>
@@ -44,13 +45,13 @@ event_name(JObj) ->
 -spec subscribe(bh_context:context(), ne_binary()) -> 'ok'.
 subscribe(Context, <<"call.*.*">>) ->
     AccountId = bh_context:account_id(Context),
-    add_call_binding(AccountId, ?LISTEN_TO),
+    add_call_binding(AccountId, Context, ?LISTEN_TO),
     {'ok', Context};
 subscribe(Context, <<"call.", Binding/binary>>) ->
     case binary:split(Binding, <<".">>, ['global']) of
         [Event, <<"*">>] ->
             AccountId = bh_context:account_id(Context),
-            add_call_binding(AccountId, [Event]);
+            add_call_binding(AccountId, Context, [Event]);
         _ ->
             blackhole_util:send_error_message(Context, <<"unmatched binding">>, Binding)
     end,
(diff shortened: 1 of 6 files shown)


applications/blackhole/src/modules/bh_call.erl
```diff
@@ -22,63 +22,64 @@
...
+    blackhole_util:send_error_message(Context, <<"unmatched binding">>, Binding),
+    {'ok', Context}.
 
--spec add_call_binding(ne_binary(), [ne_binary()]) -> ok.
-add_call_binding(_AccountId, []) -> ok;
-add_call_binding(AccountId, [Event | Events]) ->
+-spec add_call_binding(ne_binary(), bh_context:context(), [ne_binary()]) -> ok.
+add_call_binding(_AccountId, _Context, []) -> ok;
+add_call_binding(AccountId, Context, [Event | Events]) ->
+    blackhole_bindings:bind(<<"call.", Event/binary, ".*">>, ?MODULE, 'handle_event', Context),
```
lazedo: shouldn't we include the accountid in the binding key ? this way, this subscription will get notified for every account.
maybe that's why there was a check for the accountid.

applications/blackhole/src/modules/bh_call.erl
```diff
@@ -22,63 +22,64 @@
                    ]).
 
 -spec handle_event(bh_context:context(), kz_json:object()) -> 'ok'.
-handle_event(Context, EventJObj) ->
+handle_event(#bh_context{binding=Binding} = Context, EventJObj) ->
     kz_util:put_callid(EventJObj),
     lager:debug("handle_event fired for ~s ~s", [bh_context:account_id(Context), bh_context:websocket_session_id(Context)]),
-    'true' = kapi_call:event_v(EventJObj)
-        andalso is_account_event(Context, EventJObj),
+    'true' = kapi_call:event_v(EventJObj),
```
lazedo: see comment about binding

applications/blackhole/src/modules/bh_call.erl
```diff
@@ -22,63 +22,64 @@
...
-    add_call_binding(AccountId, Events).
+    add_call_binding(AccountId, Context, Events).
 
--spec rm_call_binding(ne_binary(), [ne_binary()]) -> ok.
-rm_call_binding(_AccountId, []) -> ok;
-rm_call_binding(AccountId, [Event | Events]) ->
+-spec rm_call_binding(ne_binary(), bh_context:context(), [ne_binary()]) -> ok.
+rm_call_binding(_AccountId, _Context, []) -> ok;
+rm_call_binding(AccountId, Context, [Event | Events]) ->
+    blackhole_bindings:unbind(<<"call.", Event/binary, ".*">>, ?MODULE, 'handle_event', Context),
```
lazedo: same as in `blackhole_bindings:bind` above

applications/blackhole/src/modules/bh_conference.erl
```diff
@@ -17,11 +18,11 @@
 -include("blackhole.hrl").
 
 -spec handle_event(bh_context:context(), kz_json:object()) -> 'ok'.
-handle_event(Context, EventJObj) ->
+handle_event(#bh_context{binding=Binding}=Context, EventJObj) ->
     lager:debug("handling conference event ~s", [get_response_key(EventJObj)]),
     blackhole_data_emitter:emit(bh_context:websocket_pid(Context)
                                ,get_response_key(EventJObj)
-                               ,kz_json:normalize_jobj(EventJObj)
+                               ,kz_json:normalize_jobj(kz_json:set_value(<<"Binding">>, Binding, EventJObj))
```
lazedo: can we change `Binding` to `consumer_key` ?

applications/blackhole/src/modules/bh_call.erl
```diff
@@ -22,63 +22,64 @@
...
     lager:debug("valid event and emitting to ~p: ~s", [bh_context:websocket_pid(Context), event_name(EventJObj)]),
-    blackhole_data_emitter:emit(bh_context:websocket_pid(Context), event_name(EventJObj), EventJObj).
-
-is_account_event(Context, EventJObj) ->
-    kz_json:get_first_defined([<<"Account-ID">>
-                              ,[<<"Custom-Channel-Vars">>, <<"Account-ID">>]
-                              ], EventJObj
-                             ) =:=
-        bh_context:account_id(Context).
+    NormJObj = kz_json:normalize_jobj(kz_json:set_value(<<"Binding">>, Binding, EventJObj)),
```
lazedo: can we change `Binding` to `consumer_key` ?

jamhed: @lazedo how consumer_key is better than binding (what it actually is)?

lazedo: is this not related to `KAZOO-4916` ? the client binds to a key and gets a consumer_key. the events generated on that subscription should include the consumer_key

lazedo: i also don't see the consumer_key being generated and sent back in result of subscribe.

jamhed: > is this not related to KAZOO-4916
It is related. Binding value is unique in connection and therefore is an "unique key". There is no need to generate unique key on unique key. And what @jr actually wants is to have binding value in each event.

applications/blackhole/src/modules/bh_call.erl
jamhed: > this subscription will get notified for every account

No, it will not -- see next line, where kz_hook is used with account_id.

PR #2386, opened 2016-08-01
KAZOO-3866

fenollp:
Note: delete is only available through Tasks UI & knm_number:delete.

Should I:

1. Add a maintenance function?
1. Add a DELETE API (though phone_numbers' already calls knm_number:release)

### DIFF
--- a/applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
+++ b/applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
@@ -826,19 +826,19 @@ collection_process(Context, Action, Numbers) ->
     lists:foldl(
       fun
           ({Number, {'ok', KNMNumber}}, {ServicesAcc, JObjAcc}) ->
-                       JObj = knm_number:to_public_json(KNMNumber),
-                       {ServicesAcc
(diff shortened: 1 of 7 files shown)


applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
```diff
@@ -878,13 +878,17 @@ fold_dry_runs(ServicesList) ->
...
--spec pick_release_or_delete(knm_number_options:options()) -> 'release' | 'delete'.
-pick_release_or_delete(Options) ->
+-spec pick_release_or_delete(cb_context:context(), knm_number_options:options()) -> 'release' | 'delete'.
+pick_release_or_delete(Context, Options) ->
     AuthBy = knm_number_options:auth_by(Options),
-    case kz_util:is_system_admin(AuthBy) of
-        'false' -> 'release';
-        'true' -> 'delete'
-    end.
+    Pick = case props:is_true(<<"hard">>, kz_json:to_proplist(cb_context:query_string(Context)))
```
lazedo: @fenollp  how about `kz_util:is_true(cb_context:req_param(<<"hard">>, Context, 'false')) ` ?

PR #2387, opened 2016-08-01
gen-inum: add maintenance cmd to generate inum numbers

fenollp:


lazedo:
@fenollp the `generate_numbers` was added during lisbon training as an example. if we would go to extend it, maybe we can add a generate_numbers/4 instead taking the simplified carrier module (inum, managed) and try to call Module:generate/3. i don't feel a need to maintain generate_numbers/3.
thoughts ?

fenollp:
Agreed

PR #2395, opened 2016-08-02
KAZOO-4961: move core/kazoo_tasks to applications/tasks

fenollp:


fenollp:
@lazedo ready!

core/kazoo_bindings/src/kazoo_bindings.erl
```diff
@@ -710,7 +713,9 @@ map_processor_fold(#kz_binding{binding_parts=BParts
                   ,_Routing
                   ,RoutingParts
                   ) ->
-    case matches(BParts, RoutingParts) of
+    case matches(BParts, RoutingParts)
+        orelse matches(RoutingParts, BParts)
```
jamesaimonetti: @fenollp can you explain this one? What was failing that required this? Where are the unit test additions that show the failing case is now succeeding?

jamesaimonetti: This code hasn't needed updates in a long time, which is why I'm concerned about this having been merged with no updates to the tests.

core/kazoo_bindings/src/kazoo_bindings.erl
```diff
@@ -100,6 +101,10 @@
 map(Routing, Payload) ->
     map_processor(Routing, Payload, get_binding_candidates(Routing)).
 
+-spec map(ne_binary(), payload(), kz_bindings()) -> map_results().
```
jamesaimonetti: Nothing appears to call this in Kazoo. Why was it added?

applications/tasks/src/tasks_listener.erl
```diff
@@ -51,20 +61,83 @@ start_link() ->
...
-                          ,{<<"Msg-ID">>, kz_api:msg_id(JObj)}
-                           | kz_api:default_headers(?APP_NAME, ?APP_VERSION)
-                          ]),
-    kapi_tasks:publish_help_resp(kz_api:server_id(JObj), RespJObj).
+-spec handle_lookup_req(kz_json:object(), kz_proplist()) -> 'ok'.
+handle_lookup_req(JObj, _Props) ->
+    'true' = kapi_tasks:lookup_req_v(JObj),
+    Help =
+        case
+            case {kapi_tasks:category(JObj), kapi_tasks:action(JObj)} of
```
jamesaimonetti: This should have been its own function.

core/kazoo_bindings/src/kazoo_bindings.erl
fenollp: This is due to the fact that kazoo_bindings:matches/2 ignores *s and #s (their special meaning anyway) in the second argument. That mean kazoo_bindings:map/2 & fold/2 's first argument isn't properly matched.
matches/2 only interprets these chars as special in its first argument.

There are no test cases per se but the tasks_bindings:apply/3 code.

jamesaimonetti: The first argument to matches/2 is the binding key; the second argument is the routing key. Binding keys allow wildcards; routing keys are concrete and do not have wildcards. Did you look at the [tests](https://github.com/2600hz/kazoo/blob/master/core/kazoo_bindings/test/kazoo_bindings_test.erl)?

I don't understand what you're trying to accomplish. Give me something concrete here. As-is, your changes have introduced ambiguity and muddied the water for how routing keys are matched to binding keys.

fenollp: You're right I may have misunderstood how bindings work, and I didn't think of looking at the tests.

Here is the code that made me think I could write tasks_bindings:apply/3:

```
applications/crossbar/src/crossbar_maintenance.erl:432:    Context1 = crossbar_bindings:fold(<<"v1_resource.validate.accounts">>, Payload),
applications/crossbar/src/crossbar_maintenance.erl:459:    Context1 = crossbar_bindings:fold(<<"v1_resource.validate.users">>, Payload),
applications/crossbar/src/crossbar_maintenance.erl:478:    Context1 = crossbar_bindings:fold(<<"v1_resource.execute.put.accounts">>, [Context]),
applications/crossbar/src/crossbar_maintenance.erl:503:    Context1 = crossbar_bindings:fold(<<"v1_resource.execute.put.users">>, [Context]),
applications/crossbar/src/modules/cb_bulk.erl:305:    InterimContext = crossbar_bindings:fold(Binding, Payload),
applications/crossbar/src/modules/cb_onboard.erl:160:    Context1 = crossbar_bindings:fold(<<"*.validate.accounts">>, Payload),
applications/crossbar/src/modules/cb_onboard.erl:191:    Context1 = crossbar_bindings:fold(<<"*.validate.phone_numbers">>, Payload),
applications/crossbar/src/modules/cb_onboard.erl:228:            Context1 = crossbar_bindings:fold(<<"*.validate.braintree">>, Payload),
applications/crossbar/src/modules/cb_onboard.erl:281:    Context1 = crossbar_bindings:fold(<<"*.validate.users">>, Payload),
applications/crossbar/src/modules/cb_onboard.erl:340:    Context1 = crossbar_bindings:fold(<<"*.validate.devices">>, Payload),
applications/crossbar/src/modules/cb_onboard.erl:390:    Context1 = crossbar_bindings:fold(<<"*.validate.vmboxes">>, Payload),
applications/crossbar/src/modules/cb_onboard.erl:434:    Context1 = crossbar_bindings:fold(<<"*.validate.callflows">>, Payload),
applications/crossbar/src/modules/cb_onboard.erl:453:    Context1 = crossbar_bindings:fold(<<"*.execute.put.accounts">>, [cb_context:set_resp_status(Context, 'error')]),
applications/crossbar/src/modules/cb_onboard.erl:470:                    catch (crossbar_bindings:fold(<<"*.execute.delete.accounts">>, [Context1, AccountId])),
applications/crossbar/src/modules/cb_onboard.erl:494:    Context1 = crossbar_bindings:fold(<<"*.execute.put.phone_numbers">>, Payload),
applications/crossbar/src/modules/cb_onboard.erl:511:    Context1 = crossbar_bindings:fold(<<"*.execute.post.braintree">>, Payload),
applications/crossbar/src/modules/cb_onboard.erl:524:    Context1 = crossbar_bindings:fold(<<"*.execute.put.", Event/binary>>, Payload),
```

I just need kazoo_bindings:map/2 except that it didn't work with wildcards. Looking at the code I thought it handles wildcards (see ^) but it didn't. So I added some code to fix that.

I understand my additions have muddied waters however I don't know how to better fix that.
Maybe remove `matches(BParts, RoutingParts)` and leave `matches(RoutingParts, BParts)`?

core/kazoo_bindings/src/kazoo_bindings.erl
lazedo: @jamesaimonetti that's part of a possible optimization for csv processing. you call candidates once and call map/3 for each row

core/kazoo_bindings/src/kazoo_bindings.erl
lazedo: kazoo_bindings worked in the premise that the routing key passed for processing is concrete, however some operations may actually want to run all possible bindings based on wildcards. `tasks.help.*` is one of them and `flush` within kazoo_bindings is another (cleaning "tasks.#" on startup). the `orelse` added makes this happen. 

@jamesaimonetti since matches/1 and candidates/1 help us to debug these, you can try
kazoo_bindings:matches(<<"tasks.help.*">>) with and without the `orelse`

jamesaimonetti: The only misuse I see is in `cb_onboard` which as far as I know is a deprecated, unused module. All the other instances of map/fold use concrete keys, built from the version in cb_context. Not to throw Pierre D. under the bus here, but all instances of using a wildcard in a routing key were done by him (and we failed to catch it in review).

kazoo_bindings is an attempt to have local AMQP-like routing - would you publish an AMQP message with a wildcard in the routing key? kazoo_bindings does have an optimization for crossbar bindings wherein kazoo_bindings uses the first two segments of the binding key to create a pared-down result set when comparing against a routing key (otherwise we have to use the whole ETS table of binding keys to see which match).

Modules that bind should be able to clean up after themselves by using the same binding key(s) they used when calling bind/3. flush/1 takes a binding key and flushes all responders bound to that key. flush_mod/1 takes a module and checks all bindings' responders for that module, removing it from the responders (and removing the binding if responders is empty).

If there are binding keys registered that need cleaning up, that's a failing of the application that put them there. So I would ask, why does `tasks.#` need cleaning up at startup? Why is it there in the first place, who put it there, and why didn't they clean it up?

core/kazoo_bindings/src/kazoo_bindings.erl
jamesaimonetti: But where is the call to the function? When I removed the export, `make xref` failed to complain about another module calling map/3.

lazedo: ` possible optimization` not implemeted yet.

core/kazoo_bindings/src/kazoo_bindings.erl
lazedo: i agree on the possible misuse of  `taskd.#` but that was used as an example and the shortcut for cleaninbg can be removed/refactored. so what you're saying is that you can use wildcard as log as it is the first routing part ? that would only service the purposes of crossbar with the version. it's `AMQP-like` not `AMQP-restricted`. i don't see querying `tasks.help.*` as a misuse of kazoo_bindings. how would you query bindings in this use case ?

applications/tasks/src/tasks_listener.erl
fenollp: Addressed in https://github.com/2600hz/kazoo/pull/2409 along with `flushing tasks.#`

core/kazoo_bindings/src/kazoo_bindings.erl
jamesaimonetti: I am saying, never once was kazoo_bindings built to support wildcards in the `routing_key`; why cb_onboard has it is beyond me but it is the **only** place I have seen. The whole purpose of kazoo_bindings is to provide flexible routing of payloads to interested callback modules. So you can bind to `tasks.help.*` and your provided callback will be called when another process calls map/fold with `tasks.help.about` or `tasks.help.foo`.

I think what you're trying to do is query the existing bindings as if it was a database instead of a routing topology. `SELECT * from kazoo_bindings where binding = 'tasks.help.%'`

If that is the case, a new interface must be made to query the existing bindings for matches (where you could do the `matches(RoutingParts, BindParts)` instead.

So I would propose, chers @fenollp and @lazedo, a new set of functions that does binding queries instead of payload routing. I would even consider adding a module to kazoo_bindings, something like `kz_bindings_db` or whatever, to be explicit about using the bindings table as a database vs routing table.

Thoughts?

lazedo: i agree on that proposal and in fact, after a closer look, we misused bindings in tasks. in `tasks`, modules should bind to `tasks.help` and implement proper arity's for `help` and the `tasks.help` usage in `tasks_scheduler` should use `fold` instead of `map` passing [], [Category] or [Category, Action] so proper arity would be called. it should also pass the acumulator in the first param, so help/0 in modules should become help/1, help/2, help/3.

for the bindind queries, we could use matches/1 with the `orelse` implemented only there.

fenollp: I agree with creating a "copy with `orelse` added" even though as I see it that means lots of code duplication. I think it requires kazoo_bindings:map/N+1 (+1 = routes matched). A new module is overkill & baaa code duplication.

I disagree with multiple help arities however: it was first written this way but it makes it impossible to get good error report on bad input: in [Category, Action] if Action is wrong you have no way to reply "action is wrong" because it could also mean "category is wrong" too.

lazedo: if you want that level of error detail, that can be handled in the module arity's with function clauses.

fenollp: I fail to see the benefit of multiple arities here.

lazedo: check here https://github.com/2600hz/kazoo/commit/b166172a8f0dd9f8e5f0dc020d7d0e110fccce60 for a rough implementation

lazedo: but, i'm starting to feel that this should be done with schemas. jobs should have a proper schema.

PR #2398, opened 2016-08-02
KAZOO-4194 fax to ws

jamhed:


applications/blackhole/src/modules/bh_fax.erl
```diff
@@ -5,52 +5,54 @@
...
-add_amqp_binding(<<"fax.status.", FaxId/binary>>, Context) ->
-    blackhole_listener:add_binding('fax', [{'restrict_to', ['status']}
-                                          ,{'account_id', bh_context:account_id(Context)}
-                                          ,{'fax_id', FaxId}
-                                          ,'federate'
-                                          ]);
-add_amqp_binding(_Binding, _Context) ->
-    lager:debug("unmatched binding ~s", [_Binding]).
+-spec subscribe(bh_context:context(), ne_binary()) -> {'ok', bh_context:context()}.
+subscribe(Context, <<"fax.status.*">> = Binding) ->
```
lazedo: @jamhed it's ok to subscribe to all fax jobs

applications/blackhole/Makefile
```diff
@@ -1,5 +1,6 @@
 ROOT = ../..
 PROJECT = blackhole
+ERL_LIBS = ../fax
```
fenollp: How about `ERL_LIBS += $(ROOT)/applications/fax`?

PR #2400, opened 2016-08-03
KAZOO-4956: qs filter for cb_vmboxes

icehess:


fenollp:
Note: formatter properly indents comments starting with 2 %. It dislikes the ones starting with only one %.

PR #2401, opened 2016-08-03
obtain conference participant info by id

onnet:


lazedo:
@onnet hey, here's a good chance to run `make apis` and `make fmt`, want to try it ?

onnet:
@lazedo tried it somewhere in the middle just for fun, but didn't notice any changes and finally forget about it in the end.  Will continue to play with it. Is there a single command to run all checks in one?
Thanks for reminding me!

applications/crossbar/src/modules/cb_conferences.erl
```diff
@@ -324,6 +330,17 @@ perform_participant_action(Conference, ?KICK, ParticipantId) ->
...
 
 %% add real-time call-info to participants
+-spec enrich_participant(ne_binary(), ne_binary(), cb_context:context()) -> cb_context:context().
+enrich_participant(ParticipantId, ConferenceId, Context) ->
+    Participants = extract_participants(
+                     request_conference_details(ConferenceId)
+                    ),
+    [Normalized|_] = [kz_json:normalize_jobj(JObj)
+                      || JObj <- request_call_details(Participants)
+                             ,kz_json:get_binary_value(<<"Participant-ID">>, JObj) == ParticipantId
```
fenollp: So everywhere else ParticipantId is considered an integer(). I'd prefer get_integer_value/2 & integer() in the specs, but that's up for debate…

lazedo:
@onnet Thanks! i asked @fenollp to make something like `make pr` when he finds the time.

PR #2403, opened 2016-08-04
qf-dupes-headers: add only LC headers to ensure no dupes

fenollp:
Cherry pick of https://github.com/2600hz/kazoo/pull/2402
Was supposedly fixed by https://2600hz.atlassian.net/browse/KAZOO-3426


fenollp:
Note: previous fix didn't seem to handle when sending back PDFs (to_pdf).

PR #2404, opened 2016-08-05
Bug fixed

N4COM:
If value of limit is setted to -1 (no limits), the app will read 0. (-1 when -1 < 0 => true, so return 0)
Changing this condition, will fix this bug.


lazedo:
@N4COM this is by design, only private fields allow -1 (unlimited). we have no plans to change that for now.

lazedo:
@N4COM we're closing this, let us know if you have any questions concerning limits and thanks for contributing.

PR #2409, opened 2016-08-06
qf-tasks: split into smaller function

fenollp:


lazedo:
@fenollp how do you feel about creating a module tasks_help that will handle the lookup and also move the `kz_tasks_scheduler:help()` routines there ?

fenollp:
@lazedo done

PR #2410, opened 2016-08-06
cccp autodial direct to conference

onnet:
... makes possible connect participant to the conference over API without creating callflow for this conference


lazedo:
@onnet rocking as usual! thanks.

PR #2412, opened 2016-08-07
swagger: more info & refactor

fenollp:


core/kazoo_ast/src/cb_api_endpoints.erl
```diff
@@ -13,16 +13,15 @@
...
-       ,filename:join([code:lib_dir('crossbar')
-                      ,"doc"
-                      ,"ref"
-                      ,<<Module/binary, ".md">>
-                      ])
-       ).
+-define(REF_PATH(Module),
+        filename:join([code:lib_dir('crossbar'), "doc", "ref", <<Module/binary,".md">>])).
+
+-define(SCHEMAS_PATH(Schema),
```
jamesaimonetti: I think the other kazoo_ast modules have a schemas path builder of sorts too. Let's put this as a function in kz_ast_util and update those modules as well.

fenollp:
rebased

PR #2415, opened 2016-08-08
uniformly handle subscription errors

jamhed:


lazedo:
@jamhed why are you removing debug lines ? these help when looking for problems.

jamhed:
I will re-add them to each module in follow-up commit

PR #2418, opened 2016-08-08
look for missed uses of ?MODULE

fenollp:


core/kazoo/src/gen_listener.erl
```diff
@@ -1090,7 +1090,7 @@ update_existing_listeners_bindings(Listeners, Binding, Props) ->
 -spec update_existing_listener_bindings(federator_listener(), binding_module(), kz_proplist()) -> 'ok'.
 update_existing_listener_bindings({_Broker, Pid}, Binding, Props) ->
     lager:debug("updating listener ~p with ~s", [Pid, Binding]),
-    gen_listener:add_binding(Pid, Binding, Props).
+    ?MODULE:add_binding(Pid, Binding, Props).
```
lazedo: @fenollp in some modules you just removed the module name (ex: ecallmgr_util), and others like this you replace the module name with ?MODULE. what's the criteria ?

lazedo:
@fenollp nice!

core/kazoo/src/gen_listener.erl
fenollp: @lazedo more or less that gen_listener may actually need these `?MODULE:` : running latest version of code.
I was just careful, but since we don't really hotload prod we might as well remove these.

fenollp:
news?

fenollp:
ready

PR #2419, opened 2016-08-08
KAZOO-4924 fix for ring_group schema check

jamhed:


applications/crossbar/src/modules/cb_groups.erl
```diff
@@ -154,7 +154,7 @@ delete(Context, _) ->
 -spec create(cb_context:context()) -> cb_context:context().
 create(Context) ->
     OnSuccess = fun(C) -> on_successful_validation('undefined', C) end,
-    cb_context:validate_request_data(<<"groups">>, Context, OnSuccess).
+    cb_context:validate_request_data(<<"callflows.ring_group">>, Context, OnSuccess).
```
fenollp: Weird. This file with same name & id also exists in 3.22 (`"groups"` schema is also queried there).
Maybe that's not the right schema?

@lazedo @jamesaimonetti @k-anderson

jamhed: Probably true, looks very similar, but json doesn't contain music_on_hold field.

lazedo: yes, it seems that it worked like this from the beginning, which means it doesn't validate anything.
i didn't look yet in `callflows.ring_group`  so confirm that it should be this one. `callflow.*`schemas are supposed to validate what is passed to cf_actions.

applications/crossbar/priv/couchdb/schemas/groups.json
```diff
@@ -0,0 +1,44 @@
...
+        },
+        "name": {
+            "description": "A friendly name for the group",
+            "maxLength": 128,
+            "minLength": 1,
+            "name": "Name",
+            "required": true,
+            "type": "string"
+        },
+        "resources": {
```
lazedo: what's `resources` in this context ?

jamhed: @lazedo, I have no idea. groups.md says it's {}, I was unable to find use of "resources" key in kazoo either.

jamesaimonetti:
merged

PR #2420, opened 2016-08-08
KAZOO-4922: ignore events from nodes that have running webhooks app

icehess:


applications/webhooks/src/webhooks_shared_listener.erl
```diff
@@ -297,7 +297,12 @@ handle_info(_Info, State) ->
 %% @end
 %%--------------------------------------------------------------------
 handle_event(_JObj, _State) ->
-    {'reply', []}.
+    WHZones = lists:sort([kz_util:to_binary(Z)
+                          || Z <- kz_nodes:whapp_zones(?APP_NAME)
+                         ]),
+    Options = [{<<"webhooks_zones">>, WHZones}
+              ] ++ should_handle_federated_events(WHZones),
```
lazedo: @icehess since we're doing all logic to have a 'should_handle_federated', there's no need to include 'zones' in Props.
this could be as simple as [{'should_handle_federated', should_handle_federated_events(WHZones)}] or
{'reply', should_handle_federated_events(WHZones)}.

applications/webhooks/src/webhooks_channel_util.erl
```diff
@@ -11,7 +11,31 @@
...
+handle_event(JObj, Props) ->
+    case kz_api:is_federated_event(JObj) of
+        'false' ->
+            continue_handle_event(JObj, Props);
+        'true' ->
+            EventZone = kz_api:event_zone(JObj),
+            WebHooksZones = props:get_value(<<"webhooks_zones">>, Props),
+            ZoneMatched = lists:member(EventZone, WebHooksZones),
+            ShouldHandle = props:get_is_true(<<"should_handle_federated">>, Props),
+            maybe_handle_federated_event(JObj, Props, ZoneMatched, ShouldHandle)
```
lazedo: in this group you only need to check `should_handle_federated`, the logic was already done in `handle_event`

applications/webhooks/src/webhooks_shared_listener.erl
```diff
@@ -327,3 +332,15 @@ code_change(_OldVsn, State, _Extra) ->
...
 %%%===================================================================
+
+-spec should_handle_federated_events(binaries()) -> kz_proplist().
+should_handle_federated_events([]) -> [];
+should_handle_federated_events([H | T]) ->
+    case kz_util:to_binary(kz_nodes:local_zone()) == H of
+        'true' ->
+            [{<<"should_handle_federated">>, 'true'}];
+        _ when T == [] ->
+            [{<<"should_handle_federated">>, 'true'}];
```
lazedo: if T == [] then we should never get a federated event, no need for this clause

icehess: What if webhooks is only running just in one zone? I'm checking here if this is the only zone that has webhooks running we should process the event.

lazedo: the only way that the case is not true is when T is not empty. Zones will be our and others, others may be empty but then our will be at the top.

PR #2421, opened 2016-08-08
KAZOO-4950: process kazoo modules for raw jobj

jamesaimonetti:


applications/call_inspector/src/parsers/ci_parser_kamailio.erl
```diff
@@ -282,7 +282,7 @@ cleanse_data_and_get_timestamp(Data0) ->
 -spec cleanse_data_fold({ne_binary() | kz_now()} | ne_binary()
                        ,cleanse_acc()
                        ) -> cleanse_acc().
-cleanse_data_fold({RawTimestamp}, {Acc, TS}) ->
+cleanse_data_fold({'timestamp', RawTimestamp}, {Acc, TS}) ->
```
fenollp: Your check is too naive: this is not a JObj. BTW spec needs to be updated with your changes.

applications/ecallmgr/src/ecallmgr_fs_node_sup.erl
```diff
@@ -31,11 +31,16 @@
...
+-define(CHILDREN, [<<"authn">>
+                  ,<<"channel">>
                   ,<<"conference">>
+                  ,<<"config">>
                   ,<<"event_stream_sup">>
                   ,<<"msg">>
+                  ,<<"node">>
+                  ,<<"notify">>
+                  ,<<"resource">>
+                  ,<<"route">>
```
fenollp: Changing the order is OK here?

core/kazoo_number_manager/src/knm_phone_number.erl
```diff
@@ -520,7 +520,8 @@ features_list(N) ->
     sets:to_list(sets:from_list(kz_json:get_keys(features(N)))).
 
 -spec set_features(knm_phone_number(), kz_json:object()) -> knm_phone_number().
-set_features(N, Features=?JSON_WRAPPER(_)) ->
+set_features(N, Features) ->
+    'true' = kz_json:is_json_object(Features),
```
fenollp: Isn't pattern checking a macro value way less expensive? What is wrong with the macro?
Also if you remove the macro please remove the now useless include_lib(kz_json.hrl).

core/kazoo_config/src/kapps_config.erl
```diff
@@ -102,17 +102,24 @@ get_binary(Category, Key, Default, Node) ->
...
-        'undefined' -> 'undefined';
-        ?JSON_WRAPPER(_)=V -> V;
-        _Else -> 'undefined'
+    V = get(Category, Key),
+    as_json_value(V).
+
+-spec as_json_value(any()) -> api_object().
+as_json_value('undefined') -> 'undefined';
+as_json_value(V) ->
+    case kz_json:is_json_object(V) of
```
fenollp: Same here wrt the macro. No without: code is now more bloated

core/kazoo_ast/src/cb_api_endpoints.erl
```diff
@@ -42,7 +42,7 @@ api_to_ref_doc(Module, Paths, ?CURRENT_VERSION) ->
     DocPath = ?REF_PATH(BaseName),
     'ok' = file:write_file(DocPath, Doc);
 api_to_ref_doc(_Module, _Paths, _Version) ->
-    io:format("skipping ~p version ~p~n", [_Module, _Version]).
+    'ok'.
```
fenollp: Why remove such information? On the contrary I believe this is worth mentioning to the user: "no v1 endpoints end up in swagger.json".

Makefile
```diff
@@ -172,6 +172,7 @@ $(FMT):
 
 fmt: TO_FMT ?= $(shell find applications core -iname '*.erl' -or -iname '*.hrl' -or -iname '*.app.src')
 fmt: $(FMT)
+	@ERL_LIBS=deps/:core/:applications/ $(ROOT)/scripts/no_raw_json.escript
```
fenollp: I'd say this should go into a `code_checks` target which you should create.
`no_raw_json` doesn't write to files, does it?

fenollp:
BTW your build fails in part due to trying to run your escript before Kazoo is even compiled

applications/call_inspector/src/parsers/ci_parser_kamailio.erl
jamesaimonetti: Well, yes and no. It would have required a lot more effort to figure out if the variable inside the `{...}` was a list or not. This was an easier fix, and tagged tuples are more idiomatic, I believe. Why wrap a timestamp in a 1-tuple to being with?

applications/ecallmgr/src/ecallmgr_fs_node_sup.erl
jamesaimonetti: AFAIK none of the supervisors rely on the others being up yet. It is a one-for-one supervisor so they should operate independently anyway.

core/kazoo_number_manager/src/knm_phone_number.erl
jamesaimonetti: I believe the macro usage was added to pacify Dialyzer but I could be mis-remembering.

core/kazoo_ast/src/cb_api_endpoints.erl
jamesaimonetti: If you deduplicated all the logging, perhaps. As it was, it was quite chatty and offered little value imo.

fenollp: I agree it is chatty but not that much. Anyway these are unrelated changes, could you move them to a different PR?

PR #2423, opened 2016-08-09
KAZOO-4973bis: fix to also list released numbers

fenollp:


fenollp:
Closed in favour of https://github.com/2600hz/kazoo/pull/2424

PR #2425, opened 2016-08-09
KAZOO-4946 change field requirements

jamhed:
for callflows.conference:
media_id is optional because we don't need it with play=false
play is required because without it command makes no sense


jamesaimonetti:
Merged

PR #2431, opened 2016-08-11
qf-swagger: remove superfluous redefinition of same already-global value

fenollp:


fenollp:
circle fails due to warnings already fixed in master.

PR #2436, opened 2016-08-11
KAZOO-4988: replicate account definition to accounts

jamesaimonetti:


jamesaimonetti:
4.0 version is in #2437

PR #2437, opened 2016-08-11
Kazoo 4929

jamesaimonetti:
Includes teletype updates from all related tickets in the sprint


jamesaimonetti:
@fenollp I am confused as to why `make fmt` is messing with the binary strings. When I run `M-x indent-region` on these regions, no formatting changes are made (but it does format unruly code). So I'm not sure what's up with that. Anyway, if you or @lazedo would review and merge despite that formatting discrepancy, that'd be swell.

lazedo:
LGTM.
about the default templates, maybe we should put them in individual files in priv dir ?

lazedo:
i created [KAZOO-4994](https://2600hz.atlassian.net/browse/KAZOO-4994) to move the templates to priv dir

jamesaimonetti:
@lazedo I assigned that ticket to me; can you put it in the sprint? Should be easy enough to make the switch.

core/kazoo_data/src/kz_data.hrl
```diff
@@ -68,6 +68,7 @@
 
 -type view_option() :: 'conflicts' |
                        'descending' |
+                       'first_when_multiple' |
```
fenollp: That's a lie. `first_when_multiple` is already added to the spec of the only function that uses it. Revert this

core/kazoo_data/src/kz_datamgr.erl
```diff
@@ -1124,7 +1127,7 @@ get_result_keys(JObjs) ->
 %% - otherwise `{error, multiple_results}' is returned.
 %% @end
 %%--------------------------------------------------------------------
--spec get_single_result(ne_binary(), ne_binary(), [view_option()|'first_when_multiple']) ->
+-spec get_single_result(ne_binary(), ne_binary(), view_options()) ->
```
fenollp: Revert this

core/kazoo_transactions/src/wht_util.erl
```diff
@@ -272,13 +274,13 @@ current_account_dollars(Account) ->
...
 %%--------------------------------------------------------------------
--spec get_balance_from_account(ne_binary(), couch_util:view_options()) -> units().
+-spec get_balance_from_account(ne_binary(), kazoo_modb:view_options()) -> units().
 get_balance_from_account(Account, ViewOptions0) ->
     View = <<"transactions/credit_remaining">>,
     AccountId = kz_util:format_account_id(Account),
     AccountDb = kz_util:format_account_db(AccountId),
     ViewOptions = ['first_when_multiple' | ViewOptions0],
-    case kz_datamgr:get_results(AccountDb, View, ViewOptions) of
+    case kz_datamgr:get_results(AccountDb, View, kazoo_modb:strip_modb_options(ViewOptions)) of
```
fenollp: Why say it's `kazoo_modb:view_options()` then?

core/kazoo_voicemail/src/kz_vm_message.erl
```diff
@@ -815,17 +816,21 @@ check_doc_type(_Doc, _ExpectedType, _DocType) ->
     lager:debug("not expected type : ~s , ~s", [_ExpectedType, _DocType]),
     {'error', 'not_found'}.
 
--spec fetch_vmbox_messages(ne_binary(), ne_binary()) -> db_ret().
-fetch_vmbox_messages(AccountId, BoxId) ->
+-spec fetch_vmbox_messages(ne_binary(), ne_binary() | kz_json:object()) -> db_ret().
+fetch_vmbox_messages(AccountId, <<_/binary>> = BoxId) ->
```
fenollp: There's `?NE_BINARY` for these kinds of matchings

fenollp:
Formatting issue looks like a bug in OTP's erlang.el. I'll minify & report.

fenollp:
Formatting bug [reported](https://bugs.erlang.org/browse/ERL-210).

Workaround: split the multiline string into multiple one-line strings, the compiler will concatenate them.

core/kazoo_transactions/src/wht_util.erl
lazedo: this is correct, it gets here after trying to get current_balance from modb. since it goes to accountdb it strips the modb options used before in the code

lazedo: but `first_when_multiple` won't get processed in `kz_datamgr:get_results` so maybe we need to call `kz_datamgr:get_single_result`.

core/kazoo_data/src/kz_data.hrl
lazedo: i agree putting `first_when_multiple` here.

core/kazoo_data/src/kz_datamgr.erl
lazedo: and removing it from  here.

lazedo:
@jamesaimonetti i would prefer not to merge with these formatting issues. if you think KAZOO-4994 is easy enough to make the switch, can you add the required commits to this pr ?

jamesaimonetti:
@lazedo sounds good, will add KAZOO-4994 here to handle the formatting issues.

PR #2446, opened 2016-08-15
Swagger authtoken

fenollp:


core/kazoo_ast/src/cb_api_endpoints.erl
```diff
@@ -297,36 +297,49 @@ to_swagger_paths(Paths, BasePaths) ->
...
+                         ],
+                    Parameter <- [F(Path, Method)],
+                    not kz_util:is_empty(Parameter)
+                ]
+               ).
+
+compare_parameters(Param1, Param2) ->
+    kz_json:get_value(<<"name">>, Param1) =< kz_json:get_value(<<"name">>, Param2).
+
+maybe_add_schema(_Path, <<"put">>=Method, Schema)
```
jamesaimonetti: This seems strange...pattern match `<<"put">>` in the header but have two conditional guards?

core/kazoo_ast/src/cb_api_endpoints.erl
```diff
@@ -339,6 +352,16 @@ swagger_params(PathMeta) ->
                               ])
     end.
 
+auth_token_param(<<"/accounts/{ACCOUNT_ID}/",_/binary>>, _Method) ->
```
jamesaimonetti: I think the clicktocall endpoint doesn't require an auth token when executing the clicktocall; its toggle-able whether the auth token is required. So maybe add a clause for the /connect of a clicktocall and have it be optional.

PR #2457, opened 2016-08-18
Apps of app

jamesaimonetti:
Searches all Kazoo apps for their dependancies and updates the .app.src file accordingly.


make/kz.mk
```diff
@@ -58,8 +58,7 @@ compile: $(COMPILE_MOAR) ebin/$(PROJECT).app json
 ebin/$(PROJECT).app: $(SOURCES)
 	@mkdir -p ebin/
 	ERL_LIBS=$(ELIBS) erlc -v $(ERLC_OPTS) $(PA) -o ebin/ $?
-	@sed "s/{modules, \[\]}/{modules, \[`echo ebin/*.beam | sed 's%\.beam ebin/%, %g;s%ebin/%%;s/\.beam//'`\]}/" src/$(PROJECT).app.src > $@
-
+	@sed "s/{modules,\\s*\[\]}/{modules,\[`echo ebin/*.beam | sed 's%\.beam ebin/%, %g;s%ebin/%%;s/\.beam//'`\]}/" src/$(PROJECT).app.src > $@
```
fenollp: why just do this here? there's the test target too

core/kazoo_ast/src/kast_app_deps.erl
```diff
@@ -0,0 +1,257 @@
...
+-export([process_project/0
+        ,process_app/1
+
+        ,fix_project_deps/0
+        ,fix_app_deps/1
+        ]).
+
+-include_lib("kazoo_ast/include/kz_ast.hrl").
+
+fix_project_deps() ->
```
fenollp: Please -spec at least exported functions

fenollp:
```
===> Errors generating release 
          Circular dependencies among applications: [{kazoo_number_manager,"4.0.0"},
                                           {kazoo_services,"4.0.0"},
                                           {kazoo_transactions,"4.0.0"},
                                           {kazoo_modb,"4.0.0"},
                                           {kazoo_apps,"4.0.0"},
                                           {kazoo_documents,"4.0.0"},
                                           {kazoo_config,"4.0.0"},
                                           {kazoo_data,"4.0.0"},
                                           {kazoo,"4.0.0"},
                                           {kazoo_globals,"4.0.0"}]
```

jamesaimonetti:
Yeah, I figured there would be. I'm hoping to go through and move some things around to alleviate those circular deps.

fenollp:
Why have the list of deps "flat" though? When an app depends on depA you're trying to also list depA's deps right? (that's what you're trying to do right?)

jamesaimonetti:
The code is straightforward:

1. Parse all modules in an application for MFA calls
2. Find the application name for those "remote" modules
3. Create unique list of apps
4. Compare the existing .app contents with the detected list of apps - add what's missing and remove what isn't needed (kernel is implicitly loaded so don't list it)

fenollp:
Note http://erlang.org/doc/design_principles/applications.html#id80593 says

> Notice that all applications have dependencies to at least Kernel and STDLIB.

jamesaimonetti:
@lazedo I plan to continue working on this as I can but will remove the PR for now. I'll issue PRs for reworking app dependencies so we don't have these circular deps, then I'll reopen this PR for the final phase of setting all apps/core .app.src files with their proper application dependencies.

PR #2464, opened 2016-08-18
Small fix to make ACDC usable on 3.22

remcovanvugt:
This is probably not the most sophisticated solution, but hey it's acdc ;).
In 3.22 the agent listener crashes when an agent has multiple devices or is is hotdesking. This results in the call bouncing back into the queue and ghost calling or channel hijack when another agent answers.

I did a quick fix to just ignore the hangup just as when it is undefined. Again, not the best coding but probably acceptable for a component that is being discontinued but still in use in some places.


remcovanvugt:
@jamesaimonetti Got it right after a couple of tries ;) Git in specific or development in general is not my day to day work, so sorry for the multiple PRs.

PR #2465, opened 2016-08-18
kazoo_call: move call-related modules to own core lib

jamesaimonetti:
This is removing circular references as part of #2457 work. When merged,
the work in #2457 will take care of updating the app.src file
appropriately.


fenollp:
@jamesaimonetti needs rebasing

jamesaimonetti:
@fenollp rebased

PR #2469, opened 2016-08-19
KAZOO-4913: add port request numbers to phone_number results

icehess:


fenollp:
Can you confirm GET phone_numbers?filter_state=available properly filters out ports?

applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
```diff
@@ -441,6 +442,61 @@ view_account_phone_numbers(Context) ->
...
+        andalso knm_port_request:account_active_ports(AccountId)
+    of
+        'false' -> kz_json:new();
+        {'error', _} -> kz_json:new();
+        {'ok', Ports} ->
+            PortNumberList = [create_port_number_object(P, Context) || P <- Ports],
+            lists:foldl(fun kz_json:merge_jobjs/2, kz_json:new(), PortNumberList)
+    end.
+
+-spec check_filter_state(api_ne_binary() | ne_binaries()) -> boolean().
```
fenollp: This says api_, shouldn't it handle undefined?

applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
```diff
@@ -441,6 +442,61 @@ view_account_phone_numbers(Context) ->
     cb_context:set_resp_data(Context1, NewRespData).
 
 %% @private
+-spec maybe_add_port_request_numbers(cb_context:context()) -> kz_json:object().
+maybe_add_port_request_numbers(Context) ->
+    AccountId = cb_context:account_id(Context),
+    FilterState = kz_json:get_value(<<"filter_state">>
+                                   ,cb_context:query_string(Context)),
+    case check_filter_state(kz_util:to_binary(FilterState))
```
fenollp: I don't believe this will do. Any field of the response can be used as a filter. Not just state.
Maybe this could go into the function getting the view?

fenollp: Or maybe it is possible to apply the filter function crossbar_doc:load_view uses on a JObj of port numbers.

applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
```diff
@@ -441,6 +442,61 @@ view_account_phone_numbers(Context) ->
...
+    Numbers = kz_json:get_value(<<"numbers">>, Port, kz_json:new()),
+    Fun = fun(N, Meta) -> port_number_object(AccountId, Port, N, Meta) end,
+    kz_json:map(Fun, Numbers).
+
+-spec port_number_object(ne_binary(), kz_json:object(), ne_binary(), kz_json:object()) ->
+                                {ne_binary(), kz_json:object()}.
+port_number_object(AccountId, Port, Number, JObj) ->
+    Meta = kz_json:from_list(
+             props:filter_undefined(
+               [{<<"state">>, <<"in_service">>}
```
fenollp: So every port numbers are "in_service", regardless of pvt_port_state?

applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
icehess: port request state are different from phone number's state and needs especial handling. I can use `crossbar_doc:filtered_doc_by_qs` for the rest of the fields.

applications/crossbar/src/modules/cb_modules_util.erl
```diff
@@ -579,15 +579,40 @@ remove_plaintext_password(Context) ->
...
+-spec maybe_assign_to_port_number(ne_binary(), api_binary()) ->
+                                         assignment_update().
+maybe_assign_to_port_number(DID, Assign) ->
+    Num = knm_converters:normalize(DID),
+    case knm_port_request:get(Num) of
+        {'error', _} ->
+            {DID, knm_number:assign_to_app(DID, Assign)};
+        {'ok', JObj} ->
+            NewPortJObj = update_port_request(Num, Assign, JObj),
+            {DID, kz_datamgr:save_doc(?KZ_PORT_REQUESTS_DB, NewPortJObj)}
```
fenollp: Just noticed: this is `{ne_binary(), {ok, kz_json:object()} | {error, any()}}`, which does not map to `assignment_update()`.
Also this assignment code should probably go to knm_number.erl as there already are ports stuff there, and this has to do with general use of ports.

applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
```diff
@@ -441,6 +442,56 @@ view_account_phone_numbers(Context) ->
...
+    kz_json:foldl(Fun, kz_json:new(), Numbers).
+
+-spec port_number_meta_object(ne_binary(), kz_json:object(), kz_json:object()) ->
+                                     {ne_binary(), kz_json:object()}.
+port_number_meta_object(AccountId, Port, JObj) ->
+    kz_json:from_list(
+      props:filter_undefined(
+        [{<<"state">>, <<"port_in">>}
+        ,{<<"features">>, []}
+        ,{<<"assign_to">>, AccountId}
```
fenollp: Don't you mean assign**ed**_to?

applications/crossbar/src/modules/cb_modules_util.erl
icehess: Do you mean it should go into knm_port_request.erl ?

fenollp: Yes, the code updating the port request should be moved there

applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
```diff
@@ -407,14 +408,68 @@ delete(Context, Number) ->
...
+-spec port_number_summary(kz_json:object(), cb_context:context(), boolean()) -> boolean().
+port_number_summary(PhoneNumber, Context, 'true') ->
+    crossbar_util:response(PhoneNumber, Context);
+port_number_summary(_PhoneNumber, Context, 'false') ->
+    reply_number_not_found(Context).
+
+-spec normalize_port_number(kz_json:object(), ne_binary(), cb_context:context()) ->
+                                   knm_phone_number:knm_phone_number().
+normalize_port_number(JObj, Number, Context) ->
+    Features = lists:foldl(fun (FeatureKey, Acc) ->
```
fenollp: There is knm_phone_number:set_feature/3 (which may need setters/2 to be updated to support 3-tuples).
I would prefer if that translation between list to jobj only happened in KNM

applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
```diff
@@ -460,6 +547,13 @@ normalize_view_results(JObj, Acc) ->
      | Acc
     ].
 
+%% @private
+-spec normalize_port_view_result(kz_json:object()) -> kz_json:object().
+normalize_port_view_result(JObj) ->
+    Number = kz_json:get_value([<<"key">>, 2], JObj),
```
fenollp: 2?

core/kazoo_number_manager/src/knm_port_request.erl
```diff
@@ -554,3 +582,11 @@ fetch_docs(StartKey, Limit) ->
                   ,'include_docs'
                   ],
     kz_datamgr:all_docs(?KZ_PORT_REQUESTS_DB, ViewOptions).
+
+-spec save_doc(kz_json:object()) -> {'ok', kz_json:object()} |
+                                    {'error', any()}.
+save_doc(JObj) ->
+    case kz_datamgr:save_doc(?KZ_PORT_REQUESTS_DB, JObj) of
```
fenollp: Useless use of case

applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
icehess: keys in this view are: [AccountId, Port_Number]

lazedo: @icehess we should try to document these things or make use of macros that show what it means.

PR #2470, opened 2016-08-19
KAZOO-4985 rate limits

jamhed:
- limit number of bh connections per ip, per account on node
- limit message rate per connection on node
- app redesign in message handling and context usage

==

more to come.


applications/blackhole/src/blackhole_socket_callback.erl
```diff
@@ -10,215 +10,94 @@
...
-%% @private
-%% @doc
-%% @end
-%%--------------------------------------------------------------------
--spec maybe_subscribe(bh_context:context(), kz_json:object()) -> cb_return().
--spec maybe_subscribe(bh_context:context(), kz_json:object(), boolean()) -> cb_return().
+-spec process_message(ne_binary(), kz_json:object(), bc_context:context()) -> bc_context:context().
+process_message(<<"subscribe">>, JMsg, Context) -> maybe_subscribe(Context, JMsg);
+process_message(<<"unsubscribe">>, JMsg, Context) -> maybe_unsubscribe(Context, JMsg);
+process_message(<<"authorize">>, _JMsg, Context) ->
```
lazedo: there's a diff between authenticate and authorize. this should be authenticate.  authenticate is a command we handle, authorize is a check we do for the command.

fenollp:
Elvis failed due to

```
# applications/blackhole/src/blackhole_limit.erl [FAIL]
  - no_trailing_whitespace
    - Line 48 has 1 trailing whitespace characters.
# applications/blackhole/src/modules/bh_call.erl [FAIL]
  - no_trailing_whitespace
    - Line 46 has 1 trailing whitespace characters.
```

PR #2471, opened 2016-08-19
KAZOO-5019 add presence probe if needed

lazedo:


applications/omnipresence/src/omnip_util.erl
```diff
@@ -35,3 +37,19 @@ is_valid_uri(Uri) ->
...
+        [Username, Realm | _] -> request_probe(Package, Username, Realm);
+        _Other -> lager:warning("request probe for ~p failed", [_Other])
+    end.
+
+-spec request_probe(binary(), binary(), binary()) -> 'ok'.
+request_probe(Package, Username, Realm) ->
+    API = [{<<"Event-Package">>, Package}
+          ,{<<"Username">>, Username}
+          ,{<<"Realm">>, Realm}
+	   | kz_api:default_headers(?APP_NAME, ?APP_VERSION)
```
fenollp: Great, that slow/buggy elvis script isn't even catching the TABs we ask it to catch..

lazedo: it was formatted with `make fmt`

fenollp: Yup. Just noting we may as well drop Elvis

PR #2478, opened 2016-08-22
Qf elvis

fenollp:
Elvis takes either a long time or times out in CI.
From recent PRs it even seems its TABs check doesn't even work.

Either I'm doing something wrong or elvis isn't alpha enough.
Trying to upgrade it to see what's what. Otherwise will probably revert to a TABs checking script.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,7 +153,7 @@ sup_completion: kazoo
 
 
 $(ELVIS):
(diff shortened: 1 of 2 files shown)


fenollp:
Waiting on https://github.com/inaka/elvis/issues/393

fenollp:
Ready now

PR #2481, opened 2016-08-22
Move knm_sip.erl to kazoo_sip/

fenollp:


lazedo:
@fenollp i think this move should extend kzsip_uri with accessors and replace usage of `#uri ` record with those accessors in system if needed. i don't see a need for kzsip_uris and kzsip_uri

k-anderson:
@fenollp Hey can you update doc/announcements.md with the move?

fenollp:
@lazedo @k-anderson done

PR #2482, opened 2016-08-22
recordings updates

lazedo:


applications/callflow/src/cf_exe.erl
```diff
@@ -706,10 +722,22 @@ cf_module_task(CFModule, Data, Call, AMQPConsumer) ->
...
+-spec amqp_send_message(api_terms(), kz_amqp_worker:publish_fun(), ne_binary()) -> 'ok'.
+amqp_send_message(API, PubFun, Q) ->
     PubFun(add_server_id(API, Q)).
 
+-spec amqp_call_message(api_terms(), kz_amqp_worker:publish_fun(), kz_amqp_worker:validate_fun(), ne_binary()) -> 'ok'.
+amqp_call_message(API, PubFun, VerifyFun, Q) ->
+    Routines = [{fun add_server_id/2, Q}
+               ,fun add_message_id/1
+               ],
+    Request = lists:foldl(fun({Fun, Arg1}, Acc) ->
```
fenollp: Could you turn that fun into a named function? Makes crash reports more readable

PR #2486, opened 2016-08-24
improve logging

jamhed:


fenollp:
I kinda like the separators as they are, plus external tools may depend on this format...
What other changes are there? (not easy to read)

jamhed:
@fenollp:
- don't show module:line if not defined
- use spaces as items separators
- remove various braces

Also, if you'd like to stick with | as separator, I think it's better to use it uniformly, like " | " everywhere.

lazedo:
this has impact on external tools, lets keep what we have now.

PR #2489, opened 2016-08-25
KAZOO-4980: bulk document operations for numbers tasks

fenollp:


fenollp:
After discussion:

1. instead of doing a per-task solution, let's see how changing kz_datamgr to aggregate writes ( and deletes?) with per-node gen_server could handle this for all apps
1. also: make sure `kz_datamgr:save_docs` actually does a bulk write

fenollp:
Ready for merge @lazedo

fenollp:
rebased

PR #2493, opened 2016-08-26
qf-dev: use long name when host is short

fenollp:


fenollp:
Made this due to Ubuntu having troubles setting a FQDN hostname. Set mine in /etc/hostname and that fixed it. 
Closing this for this reason.

PR #2495, opened 2016-08-28
missing ticks

jamhed:


fenollp:
Could you add this in circle.yml around where it does code_checks / make fmt ?

fenollp:
Also CI has found some code check issues:

1. andalso/orelse need to be dropped like `A\nandalso B`
1. ?MODULE should be used instead of the literal module atom

lazedo:
there's no need for these changes.

fenollp:
Reopenning for discussion.

I remember @jamesaimonetti wrote some coding style markdown somewhere (can't seem to find it) within which "ticki-ing atoms" was mentioned.
If this is no longer a requirement does it mean we can skip tickiing atoms in future code additions?

jamhed:
My 2 cents:
http://i.stack.imgur.com/MyQki.jpg

lazedo:
please stop spending your energy in superfluous things, focus on the essential.
if/when the ticking atoms policy will be lifted then we can discuss if we want to put these kind of things in the repo.

jamhed:
@lazedo, this thing is to enforce the existing quoting policy, not to lax it.

jamhed:
If there is policy 'all atoms in certian places should be quoted', then this PR is valid. If there is no such policy, then this PR is unnecessary alongside with any future atom's quotes.

jamesaimonetti:
The [Contributing](https://github.com/2600hz/kazoo/blob/master/CONTRIBUTING.md) doc does ask for tickied atoms in all Kazoo code, so I think this PR is useful. However, I also agree with @lazedo that this probably wasn't the best use of time.

@jamhed If you have a script/make target/etc that is doing this work, great! I don't see it in the PR though, so at this point I think @lazedo is right to call out the appearance of wasted time. If the method of detecting/fixing the untickied atoms can be shared, then this PR has value for everyone going forward.

That's my thought at the moment.

core/kazoo_token_buckets/src/kz_token_bucket.erl
```diff
@@ -213,7 +213,7 @@ handle_call({'consume_until', Req}, _From, #state{tokens=Current}=State) ->
             {'reply', 'true', State#state{tokens=N}};
         _N ->
             lager:debug("not enough tokens (~p) to consume ~p, zeroing out tokens", [Current, Req]),
-            {'reply', 'false', State#state{tokens=0}}
+            {'reply', 'false', State#state{'tokens'=0}}
```
jamesaimonetti: We don't ticky record names, just raw atoms.

core/kazoo_services/src/kz_services.erl
```diff
@@ -448,7 +448,7 @@ set_billing_id(BillingId, #kz_services{jobj=ServicesJObj}=Services) ->
             throw({'invalid_billing_id', <<"Requested billing id is not the parent of this account">>})
     catch
         {'EXIT', _} ->
-            throw({'invalid_billing_id', <<"Unable to determine if billing id is valid">>})
+            'throw'({'invalid_billing_id', <<"Unable to determine if billing id is valid">>})
```
jamesaimonetti: I don't think `throw/1` should be tickied

jamhed:
@jamesaimonetti, https://github.com/jamhed/erl-tools

core/kazoo_token_buckets/src/kz_token_bucket.erl
jamhed: Yep, it's fixed already in tool.

core/kazoo_amqp/src/amqp_util.erl
```diff
@@ -451,31 +451,31 @@ basic_publish(Exchange, RoutingKey, ?NE_BINARY = Payload, ContentType, Props)
...
     MsgProps =
         #'P_basic'{
            content_type = ContentType % MIME content type
-                  ,content_encoding = ?P_GET('content_encoding', Props) % MIME encoding
-                  ,headers = ?P_GET('headers', Props) % message headers
-                  ,delivery_mode = ?P_GET('delivery_mode', Props) % persistent(2) or not(1)
-                  ,priority = ?P_GET('priority', Props) % message priority, 0-9
-                  ,correlation_id = ?P_GET('correlation_id', Props) % correlation identifier
-                  ,reply_to = ?P_GET('reply_to', Props) % address to reply to
+                  ,'content_encoding' = ?P_GET('content_encoding', Props) % MIME encoding
```
jamesaimonetti: Unnecessary tickies on record field names.

jamesaimonetti:
@jamhed ok, I wonder if you and @fenollp can integrate it into the Makefile (like `make fmt`) so Circle can run it as part of building, and we can run it manually before we push to Github.

fenollp:
Eh, I'd say use this script to drop tickies. @jamhed said that is possible.
Also, wasted time? more like gained time. And it's sunday...

jamhed:
@fenollp, yes. The tool wokrs both ways. Actually, to dequote atoms is much easier task to do (because not all atoms are being quoted: function names, record names, record fields, etc).

lazedo:
one more time..., this pr is not useful at this time as it issued. if there's a tool to integrate to enforce quoting policy then the changes would be only to Makefile, any tools coding, any circleci integration (not wished at all). the massive change that this makes into the existing brings no benefit at all, in fact it makes life harder to use git blame.
so, unless there is something useful to add, do not insist on these pull requests.

lazedo:
any updates here ? any tool to integrate ? otherwise, i'll close this

PR #2496, opened 2016-08-29
KAZOO-4996: save http req and resp debug for webhook attempts

icehess:


applications/crossbar/src/modules/cb_webhooks.erl
```diff
@@ -495,7 +495,7 @@ get_start_key(Context, Default, Formatter) ->
 -spec summary_attempts_fetch(cb_context:context(), crossbar_doc:view_options(), ne_binary()) ->
                                     cb_context:context().
 summary_attempts_fetch(Context, ViewOptions, View) ->
-    Db = kz_util:format_account_mod_id(cb_context:account_id(Context), kz_util:current_tstamp()),
+    Db = get_modb(Context),
```
lazedo: @icehess i think there's a option to format modb range in cb_modules_util. can we use that instead ?

PR #2497, opened 2016-08-30
KAZOO-5027 enrich conferences in view with ui_flags field

jamhed:


applications/ecallmgr/src/ecallmgr_fs_conference.erl
```diff
@@ -196,6 +196,7 @@ handle_info({'event', ['undefined' | Props]}, #state{node=Node}=State) ->
     {'noreply', State};
 handle_info({'event', [CallId | Props]}, #state{node=Node, publish_participant_event=EventsToPublish}=State) ->
     Action = props:get_value(<<"Action">>, Props),
+    lager:error("PROPS:~p", [Props]),
```
fenollp: caught that pokemon

lazedo: @jamhed @fenollp there's actually a `props:to_log(Props, <<"THIS IS MYDEBUG">>)` that i find useful while developing

jamesaimonetti:
There should be a corresponding update to the conferences JSON schema. Otherwise who's to know what should go in this field.

jamhed:
@jamesaimonetti, actually, no one knows what should go in this field. I believe UI team will put there whatever they want.

fenollp:
how about just adding `"additionalProperties": true` to the schema? That's how it's done everywhere else (but maybe we don't want users to set other fields than these in the schema?)

jamesaimonetti:
I think `additionalProperties` is assumed to be `true` if missing.

My concern is that this is a MonsterUI-only field. Why is the backend making a special exception for this UI?

jamhed:
@jamesaimonetti I believe they want to store some presentation-specific data. What is the alternative?

fenollp:
Usually a crossbar module with `"additionalProperties": true` returns the whole doc, not just the schema-ed fields. I think this is what @jamesaimonetti meant with "making a special exception".

Wrt `"additionalProperties"` defaulting to true if undefined sounds about right: https://tools.ietf.org/html/draft-fge-json-schema-validation-00#page-13

jamesaimonetti:
They can store whatever they like on the document; they (and others) already do. I'm wondering why we have to explicitly support their field name in the view. That seems to make a lie of the oft-stated property of Crossbar not being tied to a UI.

jamhed:
@jamesaimonetti, ah. They want their stuff to be presented in a view, not in a document, and I've just added the field to the view.

lazedo:
@jamhed and that's correct. the name by itself `ui-flags` could be better. this is a way to provide a way to api users building other apps to include whatever they want in that object and have it return as part of the list operation, without returning the full doc.

by adding to the schema, we are stating that it does matter for the core to have this object which is not true.

fenollp:
So what is the problem with returning the whole document in the view? Sounds like it solves the issue

jamhed:
@fenollp, I believe to save bandwidth. And this list potentially could be huge.

jamesaimonetti:
@lazedo my concern stems from the fact that this object will live in the view indexes as well as the document, which will increase pressure on the Couch nodes storing the database. This also impacts speed as view results come across at least two network hops: Couch->Kazoo, and Kazoo->Client. So bandwidth usage would increase accordingly.

These may all be bogeyman scare-tactics, of course.

I would think, to properly support this, we should support a generic object for clients to put custom data in that will be returned in summary listings. Similar to how we're using `_read_only:{...}` for leaked/immutable fields, we could include a `summary_data:{...}` or `collection_data:{...}` that would be added to all `crossbar_listing` views.

The schema then is `collection_data:{"type":"object", "required":"false"}` which forces it to be set to an object, doesn't require it (doesn't break backwards capabilities), but doesn't restrict the key/value pairs storable inside the object.

Additionally, we could allow optional restraints, within Crossbar, on the size of the object, to allay fears of abuse/mis-use.

Those are my thoughts at the moment.

lazedo:
@jamesaimonetti can you create a ticket detailing the implementation so we can schedule it in jira ?

meanwhile this is being merged and treated as an exception

jamesaimonetti:
Created in [KAZOO-5065](https://2600hz.atlassian.net/browse/KAZOO-5065)

PR #2498, opened 2016-08-30
qf-knm: notes following #2469

fenollp:


fenollp:
@icehess Could you comment on the leftover note at https://github.com/2600hz/kazoo/pull/2469 ? plz

applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
```diff
@@ -29,6 +29,7 @@
 
 -define(CB_LIST, <<"phone_numbers/crossbar_listing">>).
 -define(PORT_NUM_LISTING, <<"port_requests/phone_numbers_listing">>).
+-define(KEY_PORT_NUMBER, 2).
```
lazedo: @icehess sorry, but can we rename it to `KEY_PORT_NUMBER_IDX_KEY` ? i think its more meaningful.

icehess:
@fenollp how do feel about passing an option like `include_ports` to `knm_number:get` and deal with finding port number in number_manager instead?

fenollp:
@icehess I think I would be ok with that. Some thoughts:

1. Flag would be in extra_options
2. Flag only marks "action", not "state": something returning extra_options() like {outbound, true} should never return this flag. I. E. Use this flag on query, shouldn't appear on reply. 
3. There are other places in knm_number that query port requests. Your code should do the same kind of things

PR #2499, opened 2016-08-30
no_magic_numbers: document magic numbers

jamesaimonetti:


fenollp:
I keep thinking that the longer this document gets, less and less people are going to read it. I might be wrong too..

icehess:
I think it's better to break it down to separate files: Coding Style, Contributing, etc... .

I really like FreeSWITCH documentation: https://freeswitch.org/confluence/display/FREESWITCH/Contributing+Code

jamesaimonetti:
I don't disagree @fenollp; this is a first iteration to get things written down _somewhere_. I agree with @icehess on FreeSWITCH's layout being a good option to emulate. It is always a balance between having a one-stop shop for things versus breaking it down into multiple files. I suspect the CONTRIBUTING file will become an index with links off to the various sub-topics. Anyway, at the moment, better to have the content to rearrange later than to not have it at all.

PR #2501, opened 2016-08-30
KAZOO-5024 propagate moderator setting

jamhed:
- pass moderator field to blackhole
- deduce moderator state from conf_participant process
- sync moderator status with ecallmgr_fs_conferences table


lazedo:
this will be manually merged

PR #2509, opened 2016-09-02
KAZOO-5051: try to compile templates before saving attachments

icehess:


jamesaimonetti:
I would maybe think about extracting [the rendering](https://github.com/2600hz/kazoo/blob/master/applications/teletype/src/teletype_renderer.erl#L118-L168) into a core lib so both teletype and crossbar could use it. Would save a round trip on AMQP but still ensure compilation is consistent between apps.

lazedo:
@icehess lets follow @jamesaimonetti suggestion for master

core/kazoo_templates/src/kztpl_pdf.erl
```diff
@@ -7,7 +7,7 @@
 %%%   Peter Defebvre
 %%%   James Aimonetti
 %%%-------------------------------------------------------------------
--module(kz_pdf).
+-module(kztpl_pdf).
```
lazedo: @icehess my fault for misleading you. lets keep the original name `kz_pdf`.

PR #2510, opened 2016-09-02
KAZOO-5051-3.22: try to compile templates before saving attachments

icehess:


applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -825,20 +825,38 @@ update_notification(Context, Id) ->
...
                              cb_context:context().
 update_template(Context, Id, FileJObj) ->
+    DbId = kz_notification:db_id(Id),
     Contents = wh_json:get_value(<<"contents">>, FileJObj),
     CT = wh_json:get_value([<<"headers">>, <<"content_type">>], FileJObj),
     lager:debug("file content type for ~s: ~s", [Id, CT]),
-    Opts = [{'headers', [{'content_type', wh_util:to_list(CT)}]}],
 
-    AttachmentName = attachment_name_by_content_type(CT),
+    Opts = [{'content_type', kz_util:to_list(CT)}],
```
fenollp: kz_util

PR #2511, opened 2016-09-02
KAZOO-4495 check plans to delete to be assigned

jamhed:
Instead of validating their existance in DB.


lazedo:
@jamhed lets talk about this.

PR #2512, opened 2016-09-02
KAZOO-4359 restore filtered empty value

jamhed:


core/kazoo/src/api/kapi_sysconf.erl
```diff
@@ -129,6 +129,13 @@ set_req(Prop) when is_list(Prop) ->
 set_req(JObj) ->
     set_req(kz_json:to_proplist(JObj)).
 
+-spec fix_req(kz_props:props(), api_terms(), function()) -> {'ok', iolist()} | {'error', string()}.
+fix_req(Api, Props, Fun) ->
+    case {props:get_value(<<"Key">>, Api), props:get_value(<<"Value">>, Api)} of
+        {<<"fs_nodes">>, []} -> Fun([{<<"Value">>, []} | Props]);
```
lazedo: `kapi_sysconf`is not about `fs_nodes`. this is not a valid fix

jamhed: @lazedo, why does it fix the issue then?

lazedo: you are fixing the effect and creating an exception for something outside the scope of this app.

lazedo:
@jamhed same here

PR #2513, opened 2016-09-02
KAZOO-5014 decode to field of cdr legs

jamhed:


applications/crossbar/src/modules/cb_cdrs.erl
```diff
@@ -777,4 +777,9 @@ load_legs(InteractionId, Context) ->
 -spec normalize_leg_view_results(kz_json:object(), kz_json:objects()) ->
                                         kz_json:objects().
 normalize_leg_view_results(JObj, Acc) ->
-    Acc ++ [kz_json:get_value(<<"doc">>, JObj)].
+    Acc ++ [fix_leg_values(kz_json:get_value(<<"doc">>, JObj))].
```
fenollp: Would it be possible to flip the arguments of `++` here? just in order to save copying `Acc`.

lazedo:
@jamhed your are trying to fix the effect not the cause

jamhed:
@lazedo, please explain the cause.

lazedo:
fixed elsewhere

PR #2519, opened 2016-09-05
KAZOO-4515 allow to specify conference pin timeout

jamhed:
- per conference
- per account
- system-wide


lazedo:
@jamhed please do not push changes not related to the work you've done.

PR #2521, opened 2016-09-06
KAZOO-5041 enforce entered conf-number not empty

jamhed:
- check entered conference number not to be an empty string, play error message otherwise


fenollp:
circle complains about `A andalso B`: could you `sed` that into `A\nandalso B`?

fenollp:
Thanks!

PR #2525, opened 2016-09-07
qf-swag: define swagger path parameters

fenollp:
Without this, using the swagger file in the "swagger editor" does not ask for the missing path(URL) parameters ({}-enclosed VARs).

Also, with this we are at 62% of APIs documented (from ~32%, as per scripts/state-of-docs.sh)


fenollp:
Ready now

PR #2532, opened 2016-09-09
crossbar storage api and several fixes

lazedo:
fixes schemas issues between draft3/draf4

when merging, please remove the `wip` and other irrelevant commit messages from the squashed commit


applications/crossbar/doc/storage.md
```diff
@@ -0,0 +1,103 @@
...
+#### Schema
+
+Key | Description | Type | Default | Required
+--- | ----------- | ---- | ------- | --------
+`attachments` |   | `#/definitions/storage.attachments` |   | `false`
+`connections` |   | `#/definitions/storage.connections` |   | `false`
+`plan` |   | `#/definitions/storage.plan` |   | `false`
+
+
+#### Remove
```
fenollp: Please don't push "empty" docs like this: it makes tools & humans believe storage is documented when actually only ref/storage.md was copied

applications/crossbar/priv/api/swagger.json
```diff
@@ -6851,6 +7172,11 @@
                     "description": "crossbar default language",
                     "type": "string"
                 },
+                "ensure_valid_schema": {
+                    "default": true,
```
fenollp: Shouldn't that default to false?

applications/crossbar/src/modules/cb_storage.erl
```diff
@@ -0,0 +1,388 @@
...
+-spec allowed_methods() -> http_methods().
+allowed_methods() ->
+    [?HTTP_GET, ?HTTP_PUT, ?HTTP_POST, ?HTTP_DELETE].
+
+-spec allowed_methods(path_token()) -> http_methods().
+allowed_methods(?PLANS_TOKEN) ->
+    [?HTTP_GET, ?HTTP_PUT].
+
+-spec allowed_methods(path_token(), path_token()) -> http_methods().
+allowed_methods(?PLANS_TOKEN, _PlanId) ->
```
fenollp: _StoragePlanId?

PR #2536, opened 2016-09-12
tasks-help: revert to expected kazoo_bindings & other fixes

fenollp:


jamesaimonetti:
Some are easily addressed, others may require more effort:

https://gist.github.com/jamesaimonetti/4c2aa849066d08ee19e4586402ed100d

core/kazoo_bindings/src/kazoo_bindings.erl
```diff
@@ -819,13 +818,14 @@ bindings(Routing) ->
                              }=Bind, Acc) ->
                       case Binding =:= Routing
                           orelse matches(BParts, RoutingParts)
-                          orelse matches(RoutingParts, BParts)
```
lazedo: @fenollp this one should not be removed. `bindings` should look both ways. `candidates` and `get_candidates` should behave before our changes.

lazedo: better..., change bindings to work like `fold`and `map `. bindings/0 use rt_options, bindings/1 should take Options. `matches`above should be replaced with `kazoo_bindings_rt:matches(Options, BParts, RoutingParts)`

lazedo: andalso to `candidates` please

fenollp: This line was introduced by https://github.com/2600hz/kazoo/pull/2395/files#diff-361a0ac2e37e647a171e2f5518a79ee7 And as per @jamesaimonetti bindings should not "look both ways"...

lazedo: that's why i asked you to add same logic as in `fold`and `map `.
`bindings` was first introduced as `matches/1` mostly for debugging purposes. it had the same logic applied as the `orelse` changes introduced by that time.

PR #2541, opened 2016-09-12
KAZOO-4359 make value optional

jamhed:


lazedo:
don't merge this yet.

lazedo:
@jamhed please rebase and lets talk about this. in sysconf we need to ensure we always read from the database and not from kapps_config.

k-anderson:
@lazedo where we at with this?  comment, close or merge :)

PR #2547, opened 2016-09-13
Fs headers

jamesaimonetti:
build a header file with all known freeswitch event headers used.


jamesaimonetti:
@lazedo this will build the file on each invocation: `sup fs_prop_usage to_header_file` to regen the header file.

applications/ecallmgr/src/ecallmgr_fs_conference.erl
```diff
@@ -211,6 +205,18 @@ handle_info(_Info, State) ->
...
+-spec handle_conf_event(kzd_freeswitch:data(), atom()) -> 'ok'.
+handle_conf_event(Props, Node) ->
+    Action = props:get_value(<<"Action">>, Props),
+    _ = case process_conference_event(Action, Props, Node) of
+            'stop' -> 'ok';
+            'continue' -> send_conference_event(Action, Props);
+            {'continue', CustomProps} ->
+                send_conference_event(Action, Props, CustomProps)
+        end,
+    'ok'.
```
fenollp: No need for that as send_conference_event/2,3 return ok too

applications/ecallmgr/src/ecallmgr_fs_conference.erl
```diff
@@ -617,7 +630,8 @@ get_conf_command(<<"stop_play">>, _Focus, _ConferenceId, JObj) ->
                    end,
             {<<"stop">>, Args}
     end;
-get_conf_command(Say, _Focus, _ConferenceId, JObj) when Say =:= <<"say">> orelse Say =:= <<"tts">> ->
+get_conf_command(Say, _Focus, _ConferenceId, JObj) when Say =:= <<"say">>
+                                                        orelse Say =:= <<"tts">> ->
```
fenollp: How about using `;`?

core/kazoo_ast/src/fs_prop_usage.erl
```diff
@@ -0,0 +1,648 @@
...
+
+-export([process/0, process/1
+        ,to_header_file/0
+        ]).
+
+-include_lib("kazoo_ast/include/kz_ast.hrl").
+-include_lib("kazoo/src/kz_json.hrl").
+
+-define(DEBUG(_Fmt, _Args), 'ok').
+%% -define(DEBUG(Fmt, Args), io:format([$~, $p, $  | Fmt], [?LINE | Args])).
```
fenollp: How about uncommenting and bracing these two lines in `-if(true). ...else...endif.`?

core/kazoo_ast/src/fs_prop_usage.erl
```diff
@@ -0,0 +1,648 @@
...
+
+to_header_file() ->
+    Usage = process(),
+    write_usage_to_header(Usage).
+
+write_usage_to_header(Usage) ->
+    {'ok', IO} = file:open(event_filter_filename()
```
fenollp: No real need for this: use `file:write_file` with `append` option

core/kazoo_ast/src/fs_prop_usage.erl
```diff
@@ -0,0 +1,648 @@
...
+                          ,['write'
+                           ,'delayed_write'
+                           ]),
+    write_usage_to_header(Usage, IO).
+
+write_usage_to_header(Usage, IO) ->
+    EventFilters = lists:foldl(fun write_mod_usage/2, sets:new(), Usage),
+    [First|Sorted] = lists:usort(sets:to_list(EventFilters)),
```
fenollp: lists:sort should be just as good

core/kazoo_ast/src/fs_prop_usage.erl
```diff
@@ -0,0 +1,648 @@
...
+    'ok' = file:close(IO),
+    io:format("wrote ~s~n", [event_filter_filename()]).
+
+write_mod_usage({_Mod, Usages}, AccSet) ->
+    Keys = usage_keys(Usages),
+    sets:union(AccSet, Keys).
+
+usage_keys(Usages) ->
+    sets:from_list(
+      lists:usort(
```
fenollp: useless given `sets:from_list`

core/kazoo_ast/src/fs_prop_usage.erl
```diff
@@ -0,0 +1,648 @@
...
+               ,functions = [] %% AST functions loaded
+               ,visited = [] %% MFAs visited (to stop recursion)
+               }).
+
+-spec process() -> [{module(), list()}].
+process() ->
+    io:format("processing ecallmgr FreeSWITCH Props usage: "),
+    Usages = [{Module, Usages} ||
+                 Module <- kz_ast_util:app_modules('ecallmgr'),
+                 (Usages = process(Module)) =/= 'undefined'
```
fenollp: Note: equivalent / less contrived code:

```erlang
Usages <- [process(Module)],
Usage =/= undefined
```

core/kazoo_ast/src/fs_prop_usage.erl
```diff
@@ -0,0 +1,648 @@
...
+
+
+process_action(_M, 'undefined', _As, Acc) -> Acc;
+process_action(Module, Function, Args, Acc) ->
+    #usage{usages=Us} = process_mfa_call(#usage{current_module=Module}
+                                        ,Module, Function, Args
+                                        ),
+    Us ++ Acc.
+
+%% define entry points for modules
```
fenollp: When ecallmgr modules get updated / created, will these lists catch / warn / error on the changes? Or is filling these lists something that requires constant attention?

core/kazoo_documents/src/kzd_freeswitch.erl
```diff
@@ -52,11 +52,14 @@
 
 -include("kz_documents.hrl").
 
+-type data() :: kz_proplist().
```
fenollp: We're in kazoo_documents & all other modules define a `doc/0` type. How about we continue doing this?

core/kazoo_ast/src/fs_prop_usage.erl
fenollp: Also one of these two could spit stuff out to `standard_error`

core/kazoo_ast/src/fs_prop_usage.erl
```diff
@@ -0,0 +1,648 @@
...
+
+    file:write(IO, io_lib:format("-define(FS_EVENT_FITLERS~n       ,[~p~n", [First])),
+    lists:foreach(fun(Filter) ->
+                          file:write(IO, io_lib:format("        ,~p~n", [Filter]))
+                  end
+                 ,Sorted
+                 ),
+    file:write(IO, "        ])."),
+    'ok' = file:close(IO),
+    io:format("wrote ~s~n", [event_filter_filename()]).
```
fenollp: should use ?DEBUG

core/kazoo_ast/src/fs_prop_usage.erl
```diff
@@ -0,0 +1,648 @@
...
+               ,data_var_name = 'Props' %% Tracks current var name
+               ,data_var_aliases = [] %% typically when kz_json:set_value is used
+               ,current_module %% what module are we currently in
+               ,functions = [] %% AST functions loaded
+               ,visited = [] %% MFAs visited (to stop recursion)
+               }).
+
+-spec process() -> [{module(), list()}].
+process() ->
+    io:format("processing ecallmgr FreeSWITCH Props usage: "),
```
fenollp: should use ?DEBUG

core/kazoo_ast/src/fs_prop_usage.erl
```diff
@@ -0,0 +1,648 @@
...
+    Usages = [{Module, Usages} ||
+                 Module <- kz_ast_util:app_modules('ecallmgr'),
+                 (Usages = process(Module)) =/= 'undefined'
+             ],
+    io:format(" done~n"),
```
fenollp: should use ?DEBUG

core/kazoo_ast/src/fs_prop_usage.erl
```diff
@@ -0,0 +1,648 @@
...
+    process_expressions(process_expression(Acc, Expr)
+                       ,Qualifiers
+                       );
+process_expression(Acc, ?LC_GENERATOR(Pattern, Expr)) ->
+    process_expressions(Acc, [Pattern, Expr]);
+process_expression(Acc, ?LC_BIN_GENERATOR(Pattern, Expr)) ->
+    process_expressions(Acc, [Pattern, Expr]);
+
+process_expression(#usage{current_module=_M}=Acc, _Expression) ->
+    io:format("~nskipping expression in ~p: ~p~n", [_M, _Expression]),
```
fenollp: should use ?DEBUG

fenollp:
How about making this a part of Makefile / circle.yml

applications/ecallmgr/src/ecallmgr_fs_conference.erl
jamesaimonetti: Meh, was just fixing formatting

core/kazoo_ast/src/fs_prop_usage.erl
jamesaimonetti: Meh, just a quick macro for logging when developing.

core/kazoo_ast/src/fs_prop_usage.erl
jamesaimonetti: I wanted control over the formatting of the hrl file, and `~p` doesn't format as desired. Hence the open + write.

core/kazoo_ast/src/fs_prop_usage.erl
jamesaimonetti: meh, just being __really__ sure the list is sorted and only contains the unique items.

core/kazoo_ast/src/fs_prop_usage.erl
jamesaimonetti: At the moment, these are the function(s) in the modules that provided the best way to find keys used to inspect the proplist. Most, if not all, are functions that will likely not change at this point. I would expect the code to crash if the function signature changes.

Another thought I had was improving our naming in ecallmgr modules to denote FreeSWITCH proplists as `FSProps` versus `gen_listener`-provided proplists (`GLProps`?). Then the code could look for any function that has an arg of `FSProps` and go from there, thus removing the need to have module-specific filters.

core/kazoo_documents/src/kzd_freeswitch.erl
jamesaimonetti: I thought about that too;.most of the `kazoo_document` modules manipulate JSON docs from Couch; this module handles FreeSWITCH proplists. We could move this to ecallmgr since its specific to that application (and then we could leverage this .hrl file to maybe auto-generate this module too). Perhaps I'll file a ticket for that as a future improvement.

jamesaimonetti: Filed [here](https://2600hz.atlassian.net/browse/KAZOO-5068)

core/kazoo_ast/src/fs_prop_usage.erl
jamesaimonetti: So the lines not using ?DEBUG are for the escript progress thingies (once this is added to the build).

core/kazoo_ast/src/fs_prop_usage.erl
jamesaimonetti: Agreed here

core/kazoo_ast/src/fs_prop_usage.erl
fenollp: You did not completely answer my question: will this program crash/warn when a new ecallmgr module is added? If not it should, so that we are sure to not miss any FS props

core/kazoo_ast/src/fs_prop_usage.erl
fenollp: ...

core/kazoo_ast/src/fs_prop_usage.erl
fenollp: You are using ~p. Have you tried `~100p`?

core/kazoo_documents/src/kzd_freeswitch.erl
fenollp: Whatever the value of doc/0 we should go with it. That's why we have an alias type.

Agreed on autogen kzd_freeswitch

core/kazoo_ast/src/fs_prop_usage.erl
lazedo: i don't agree. for now is more than enough. `FSProps` is the way to go.

core/kazoo_documents/src/kzd_freeswitch.erl
lazedo: no need for `doc/0` here. this will be moved and its not a doc.

Makefile
```diff
@@ -182,6 +182,7 @@ apis:
 	@$(ROOT)/scripts/format-json.sh applications/crossbar/priv/couchdb/schemas/*.json
 	@ERL_LIBS=deps/:core/:applications/ $(ROOT)/scripts/generate-api-endpoints.escript
 	@$(ROOT)/scripts/format-json.sh applications/crossbar/priv/api/swagger.json
+	@ERL_LIBS=deps/:core/:applications/ $(ROOT)/scripts/generate-fs-headers-hrl.escript
```
fenollp: Since you're generating an HRL for ecallmgr, move that call to ecallmgr's Makefile

lazedo: maybe in the future, not needed for now, its not being used. if it works as expected, then maybe we'll put it into Makefile

fenollp: It's not being used? it should be otherwise this PR is mostly dead code

lazedo: maybe i should have said, `not being used for now`. the work that will take benefit from this pr will start briefly.

core/kazoo_ast/src/fs_prop_usage.erl
jamesaimonetti: @fenollp the code will silently skip modules not explicitly handled.

If someone adds code/modules that utilize a new FS header, that will not be communicated to mod_kazoo at startup (which is the point of this PR - to provide the config to mod_kazoo).

I will add some documentation for how to do that.

Makefile
jamesaimonetti: My thoughts were that this was dealing with the interface between Kazoo and FreeSWITCH, which is close enough to an API. I don't think it needs to be part of the compilation process though.

jamesaimonetti:
@lazedo I think this is ready to go as-is.

Makefile
fenollp: So when fs_event_filters.hrl is actually used (this PR builds it but does not use it!), we should move that line to whatever app that's using this HRL. Otherwise builds will not be checking in the proper version of this HRL file, over time. Reminder: our build system does not and has never supported tracking HRL-ERLs build dependencies.

jamesaimonetti: @fenollp the expectation is that, when someone adds code to ecallmgr to access a new field in the FS proplist, they will add the field to the .hrl. Then, when the build job runs, the generated hrl will continue to match the committed hrl. If they don't match (just like if there was a formatting change or API doc change) then its on the committer to update the appropriate place. Plus I think the code needs to be compiled prior so the AST can be traversed.

PR #2552, opened 2016-09-14
kvm-migrate: migrate voicemails in bulk to current modb

icehess:


core/kazoo_documents/src/kzd_box_message.erl
```diff
@@ -74,21 +74,23 @@
 %%--------------------------------------------------------------------
 -spec new(ne_binary(), kz_proplist()) -> doc().
 new(AccountId, Props) ->
-    {Year, Month, _} = erlang:date(),
+    UtcSeconds = props:get_value(<<"Message-Timestamp">>, Props, kz_util:current_tstamp()),
+    Timestamp  = props:get_value(<<"Document-Timestamp">>, Props, UtcSeconds),
+    {Year, Month, _} = kz_util:to_date(Timestamp),
```
fenollp: Note the existence of https://github.com/2600hz/kazoo/blob/master/core/kazoo/src/kz_util.erl#L243.

core/kazoo_documents/src/kzd_box_message.erl
```diff
@@ -74,21 +74,23 @@
...
-    MediaId = <<(kz_util:to_binary(Year))/binary
-                ,(kz_util:pad_month(Month))/binary
-                ,"-"
-                ,(kz_util:rand_hex_binary(16))/binary
-              >>,
+    MsgId = <<(kz_util:to_binary(Year))/binary
+              ,(kz_util:pad_month(Month))/binary
+              ,"-"
+              ,MediaId/binary
+            >>,
```
fenollp: https://github.com/2600hz/kazoo/blob/master/core/kazoo/include/kz_types.hrl#L326 should be used to create modbs. BTW this one is PREFIX, are you instead looking for SUFFIX?

core/kazoo_voicemail/src/kvm_migrate.erl
```diff
@@ -0,0 +1,372 @@
...
+
+    ViewOpts = props:filter_undefined(
+                 [{'limit', ?MAX_BOX_PROCESS}
+                 ,{'skip', Offset}
+                 ,'include_docs'
+                 ]),
+    case get_account_vmboxes(AccountId, ViewOpts) of
+        {'ok', Results} ->
+            do_migrate(AccountId, Results),
+            {'ok', length(Results) - Offset};
```
fenollp: Offset can be undefined (as per spec), and first clause does not rule it out: `undefined < 0 =:= false`.
This will give badarith

core/kazoo_documents/src/kzd_box_message.erl
icehess: I need Year and Month to create Db and MessageId

core/kazoo_documents/src/kzd_box_message.erl
icehess: This is intended, I just add MODB prefix to previous message_id, so if for any reason migration is failed and we ran the migration again, it's going to write to same doc which is safely returns conflict.

fenollp: ok for prefix, but how about using a macro?
Also, I'd push for code comments for behavior such as you describe

icehess:
@fenollp or anyone have any idea why `make apis` is not picking up `MAX_BOX_PROCESS` from `kvm_migrate` and add it to `system_config.callflow.json` and swagger?

jamesaimonetti:
@icehess I'll take a look

jamesaimonetti:
@icehess thanks for noticing the discrepancy. I've adjusted the code and it has found more options. Thanks.

core/kazoo_voicemail/src/kvm_migrate.erl
fenollp: @icehess Notes on this?

fenollp:
@jamesaimonetti Unfortunate that these side changes are now part of this PR

jamesaimonetti:
@fenollp I think they are benign as far as the PR is concerned. There's no new functionality introduced, just enhanced JSON schemas.

core/kazoo_voicemail/src/kvm_migrate.erl
icehess: Offset should be only integer, I fixed the type.

PR #2554, opened 2016-09-14
trunkstore_honor_diversion by ensure valid cid

onnet:
At the moment, if connected over trunkstore PBX adds diversion field, it is ignored on valid caller ID check. Looks like diversions header also even absent in offnet resource request. Would like to give a try to address this.


lazedo:
thanks @onnet

PR #2557, opened 2016-09-15
add search path for kazoo_documents.hrl

jamhed:


lazedo:
why are you adding this ? circleci and travis don't complain about it?

jamhed:
I'm unable to compile on my laptop clean reopo checkout with issuing make.

lazedo:
check you kz.mk file. travis and ci, always take a clean copy, right ?

fenollp:
Yep, builds fine from fresh clone here..
What is the issue you have? Which OTP do you have installed? Does your $PWD contain spaces?

PR #2558, opened 2016-09-15
KAZOO-4767 return dialplans for account for accounts/xxx uris

jamhed:


applications/crossbar/src/modules/cb_dialplans.erl
```diff
@@ -63,11 +64,16 @@ resource_exists() -> 'true'.
...
-    Doc = maybe_add_name(kapps_config:get_all_kvs(<<"dialplans">>)),
+    Doc = get_dialplans(cb_context:account_id(Context)),
     cb_context:setters(Context, [{fun cb_context:set_resp_data/2, Doc}
                                 ,{fun cb_context:set_resp_status/2, 'success'}
                                 ]).
 
+get_dialplans('undefined') ->
+    maybe_add_name(kapps_config:get_all_kvs(<<"dialplans">>));
+get_dialplans(AccountId) ->
+    kapps_account_config:get_global(AccountId, ?ACCOUNT_CONFIG, <<"dialplans">>).
```
lazedo: this can return the same the 'undefined' clause. is it the same format ? shouldn't it go thru the `maybe_add_name` routine ?

lazedo:
the results of `kapps_config:get_all_kvs` and `kapps_account_config:get_global` are different

jamhed:
> the results of kapps_config:get_all_kvs and kapps_account_config:get_global are different

It's depends what is set as value to kapps_account_config:set. It seems you want the account's dialplan value to be similar to system one.

applications/crossbar/src/modules/cb_dialplans.erl
```diff
@@ -18,6 +18,7 @@
         ]).
 
 -include("crossbar.hrl").
+-define(ACCOUNT_CONFIG, <<"config">>).
```
lazedo: why this definition and its usage below ?

applications/crossbar/src/modules/cb_dialplans.erl
```diff
@@ -63,11 +64,23 @@ resource_exists() -> 'true'.
...
+-spec get_joined_dialplans(api_binary()) -> [kz_json:object()].
+get_joined_dialplans('undefined') -> get_dialplans('undefined');
+get_joined_dialplans(AccountId) ->
+    get_dialplans(AccountId) ++ get_dialplans('undefined').
+
+-spec get_dialplans(api_binary()) -> [kz_json:object()].
+get_dialplans('undefined') ->
+    kapps_config:get_all_kvs(<<"dialplans">>);
+get_dialplans(AccountId) ->
+    JObj = kapps_account_config:get_global(AccountId, ?ACCOUNT_CONFIG, <<"dialplans">>, {[]}),
```
lazedo: system_schemas/CONFIG gets transalated to accountdb/configs_CONFIG document.
please review kapps_account_config and see if all paths are valid in getting/setting the value.
thanks

PR #2560, opened 2016-09-15
Help 24822 4.0

jamesaimonetti:


applications/ecallmgr/src/ecallmgr_config.erl
```diff
@@ -30,7 +30,7 @@
 -include_lib("kazoo/include/kz_databases.hrl").
 
 -spec flush() -> 'ok'.
--spec flush(kz_json:key()) -> 'ok' | {'error', any()}.
+-spec flush(kz_json:keys()) -> 'ok' | {'error', any()}.
```
fenollp: This type name change (`key()` -> `keys()`) in `kz_json` was really unfortunate IMO: now any function taking a `JObj`'s "key path" must accept a plural "`keys()`", which is very misleading. Connotation between singular path & plural keys will lose more than one newcomer.

Also, now functions that do take a list of "key paths" have to use `[ keys() ]`...
Why not have stuck with `key() :: [json_string()] | json_string()`?
We know that a "key path" can be either a json_string or a list of these. The name of the "key path" type does not need to reflect some type in its union definition is a list, does it?

jamesaimonetti: I agree that the types could be improved, naming-wise, to increase clarity. I think the original decision was made because having `-type foo() :: whatever()` made having many `foo` have type `-type foos() :: [foo()]`.

That said, I agree with the proposal of making `key() :: [json_string()] | json_string().`

fenollp: Addressed in https://github.com/2600hz/kazoo/pull/2563

PR #2565, opened 2016-09-19
Conference

lazedo:


applications/ecallmgr/src/ecallmgr_conference_command.erl
```diff
@@ -0,0 +1,258 @@
...
+            {<<"play">>, Arg}
+    end;
+
+%% The following conference commands can optionally specify a participant
+get_conf_command(<<"play">>, _Focus, ConferenceId, JObj) ->
+    case kapi_conference:play_v(JObj) of
+        'false' ->
+            {'error', <<"conference play failed to execute as JObj did not validate.">>};
+        'true' ->
+            UUID = kz_json:get_ne_value(<<"Call-ID">>, JObj, ConferenceId),
```
fenollp: How about local functions to fetch this and Participant-ID too (or any fields fetched in more than one place)?
This way they would be handled with the same code.
Also, what about get_ne_binary_value in both cited examples?

applications/ecallmgr/src/ecallmgr_conference_control.erl
```diff
@@ -0,0 +1,255 @@
...
+    Server = props:get_value('server', Props),
+    ConferenceId = props:get_value('conference_id', Props),
+    Event = kz_json:get_value(<<"Event">>, JObj),
+    handle_conference_event(Node, Server, ConferenceId, Event, JObj).
+
+handle_conference_event(_Node, Server, ConferenceId, <<"conference-destroy">>, _JObj) ->
+    lager:debug("RECEIVED CONFERENCE-DESTROY FOR ~s", [ConferenceId]),
+    gen_server:stop(Server);
+handle_conference_event(_Node, _Server, ConferenceId, Event, JObj) ->
+    lager:debug("EVENNT ~s FOR ~s NOT HANDLED : ~p", [Event, ConferenceId, JObj]).
```
fenollp: debug code?

applications/ecallmgr/src/ecallmgr_conference_control.erl
```diff
@@ -0,0 +1,255 @@
...
+handle_conference_event(JObj, Props) ->
+    'true' = kapi_conference:event_v(JObj),
+    Node = props:get_value('node', Props),
+    Server = props:get_value('server', Props),
+    ConferenceId = props:get_value('conference_id', Props),
+    Event = kz_json:get_value(<<"Event">>, JObj),
+    handle_conference_event(Node, Server, ConferenceId, Event, JObj).
+
+handle_conference_event(_Node, Server, ConferenceId, <<"conference-destroy">>, _JObj) ->
+    lager:debug("RECEIVED CONFERENCE-DESTROY FOR ~s", [ConferenceId]),
```
fenollp: debug code?

applications/ecallmgr/src/ecallmgr_conference_command.erl
lazedo: this code was moved, not refactored

applications/ecallmgr/src/ecallmgr_conference_control.erl
lazedo: yep, needs cleaning

applications/ecallmgr/src/ecallmgr_fs_conference.erl
```diff
@@ -234,11 +248,14 @@ publish_event(Action, Props, Node) ->
...
 
-publish_event(Action, #conference{handling_locally='true'} = Conference, Props, _Node) ->
-    Event = conference_event(Action, Conference, Props),
-    publish_event(Event);
-publish_event(Action, _Conference, _Props, _Node) ->
-    lager:debug("conference control on another node, not publishing event ~s", [Action]).
+publish_event(Action, #conference{handling_locally=IsLocal} = Conference, Props, _Node) ->
+    case props:is_true(<<"Force-Publish-Event-State">>, Props, 'false')
+        orelse props:is_true(<<"Publish-Event-State">>, Props, 'true')
+        andalso IsLocal
```
fenollp: I understand this `andalso`'s left hand side only encloses `props:is_true(<<"Publish-Event-State">>, Props, 'true')`.
The `Force...` expression is thus the top-level expression: `A orelse B andalso C` has this precedence: `A orelse (B andalso C)`.

Is this expected? Should we add braces for clarity?

PR #2568, opened 2016-09-19
KAZOO-5078: move modules dealing with AMQP

jamesaimonetti:
Modules that deal with AMQP should be in kazoo_amqp, mostly to simplify
dependency graphs at this point


core/kazoo_amqp/Makefile
```diff
@@ -1,6 +1,7 @@
 ROOT = ../..
 PROJECT = kazoo_amqp
 
+SOURCES = src/gen_listener.erl src/*.erl $(if $(wildcard src/*/*.erl), src/*/*.erl)
```
fenollp: Why conditionally include it?

core/Makefile
```diff
@@ -26,8 +26,8 @@ test: $(MAKEDIRS)
 
 first:
 	$(MAKE) -C kazoo/ $(ACTION)
+	$(MAKE) -C kazoo_amqp/ $(ACTION)
 	$(MAKE) -C kazoo_data/ $(ACTION)
```
fenollp: Now that you're removing the kazoo transform, is the `first` target really needed?

core/kazoo_amqp/Makefile
jamesaimonetti: Good point; think I just copied it over from kz.mk because without it the api modules weren't getting compiled.

core/Makefile
jamesaimonetti: Will try

jamesaimonetti: Needed for kazoo/src/kz_perf_transform.erl and kazoo_data/src/kzs_perf.erl

fenollp:
@jamesaimonetti Needs rebase

jamesaimonetti:
@fenollp done

PR #2580, opened 2016-09-21
Braintree nonce pr

tsromana:
with this update user can add, payment_method_nonce to their json requests without needing to enter credit card info, The payment_method_nonce is provided by the Braintree JS library which validates the credit card info and return the nonce token which can than be sent to the braintree servers and that will add the credit card to the braintree vault. This will help with the compliance issue as the credit card information will no longer be exposed to the kazoo servers.

The user now can be added without any credit card,
{"data":{
"first_name": "John",
"last_name": "Doe",
"company": "ACME CORP",
"phone": "6000000000",
}}
Without any credit card and contains payment method nonce in their json request
{"data":{
"first_name": "John",
"last_name": "Doe",
"company": "ACME CORP",
"phone": "6000000000",
"payment_method_nonce":"valid-nonce"
}}
Payment method nonce is added to the credit card section
{"data":{
"first_name": "John",
"last_name": "Doe",
"company": "ACME CORP",
"phone": "6000000000",
"credit_card":{ 
"payment_method_nonce":"valid-nonce"
}

}}

Credit cards can now be added with just payment method nonce
{"data":{
"payment_method_nonce":"valid-nonce" 
}}

Generating client tokens-
Get /accounts/{ID}/braintree/client_token
Return - { "data":{
         "client_token": "easasas......."
} }

Things working: 
User can add a his info
User can add credit card with his info
User can add credit card
User can create transactions
Client token generation so that the js client knows for which customer we are trying to send request

whats not working: 
¯\_(ツ)_/¯

Please let me know what are your thoughts on this :)

### DIFF
--- a/applications/crossbar/doc/ref/braintree.md
+++ b/applications/crossbar/doc/ref/braintree.md
@@ -8,6 +8,16 @@
 
 #### Fetch
 
+> GET /v2/accounts/{ACCOUNT_ID}/braintree/client_token
+
+```shell
+curl -v -X GET \
+    -H "X-Auth-Token: {AUTH_TOKEN}" \
+    http://{SERVER}:8000/v2/accounts/{ACCOUNT_ID}/braintree/client_token
+```
+
+#### Fetch
+
 > GET /v2/accounts/{ACCOUNT_ID}/braintree/credits
 
 ```shell
@@ -32,16 +42,6 @@ curl -v -X PUT \
 
 ```shell
 curl -v -X GET \
-    -H "X-Auth-Token: {AUTH_TOKEN}" \
-    http://{SERVER}:8000/v2/accounts/{ACCOUNT_ID}/braintree/transactions
-```
-
-#### Create
-
-> PUT /v2/accounts/{ACCOUNT_ID}/braintree/transactions
-
-```shell
-curl -v -X PUT \
     -H "X-Auth-Token: {AUTH_TOKEN}" \
     http://{SERVER}:8000/v2/accounts/{ACCOUNT_ID}/braintree/transactions
 ```
--- a/applications/crossbar/src/modules/cb_braintree.erl
+++ b/applications/crossbar/src/modules/cb_braintree.erl
@@ -29,6 +29,7 @@
 -define(ADDRESSES_PATH_TOKEN, <<"addresses">>).
 -define(TRANSACTIONS_PATH_TOKEN, <<"transactions">>).
 -define(CREDITS_PATH_TOKEN, <<"credits">>).
+-define(CLIENT_TOKEN_PATH_TOKEN, <<"client_token">>).
 
 -define(MOD_CONFIG_CAT, <<(?CONFIG_CAT)/binary, ".braintree">>).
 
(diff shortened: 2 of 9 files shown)


applications/crossbar/src/modules/cb_braintree.erl
```diff
@@ -223,6 +229,18 @@ validate_credits(Context, ?HTTP_PUT) ->
                               maybe_charge_billing_id(Amount, Context1)
                       end.
 
+validate_token(Context, ?HTTP_GET) ->
```
jamesaimonetti: Missing spec

applications/crossbar/src/modules/cb_braintree.erl
```diff
@@ -223,6 +229,18 @@ validate_credits(Context, ?HTTP_PUT) ->
...
+    try braintree_client_token:get_client_token(cb_context:account_id(Context)) of
+        Token ->
+            JObj = kz_json:set_value(<<"client_token">>, Token, kz_json:new()),
+            crossbar_util:response(JObj, Context)
+    catch
+        'throw':{'api_error', Reason} ->
+            crossbar_util:response('error', <<"braintree api error">>, 400, Reason, Context);
```
jamesaimonetti: Unless you differentiate errors, I would say a 5xx error is more appropriate (as the error happens on the server side). If you can add clauses that separate out errors that are due to client-provided information (like bad account id) as 4xx, and leave all others as 5xx (braintree API server down, for instance), that would be even better.

applications/crossbar/doc/ref/braintree.md
```diff
@@ -8,6 +8,16 @@
 
 #### Fetch
 
+> GET /v2/accounts/{ACCOUNT_ID}/braintree/client_token
```
jamesaimonetti: This should be added to doc/braintree.md (or create that doc if missing) and fleshed out with example queries and responses.

jamesaimonetti:
Also, I would add your PR comment to the braintree.md doc when you copy ref/braintree.md over.

applications/crossbar/src/modules/cb_braintree.erl
tsromana: i will look into specific errors related to it and will update it accordingly :)

applications/crossbar/src/modules/cb_braintree.erl
tsromana: i guess there are other funs without specs should i update them too ? :)

applications/crossbar/doc/ref/braintree.md
tsromana: @fenollp got it, thanks :)

tsromana:
@jamesaimonetti will do as suggested and sorry for the delay will be adding other pr's soon :)

tsromana:
@jamesaimonetti i have updated the code and have removed PUT on transactions as there was no function handling the PUT request and i don't see a particular use case here.

jamesaimonetti:
I manually merged this but apparently have forgotten how to get Github to detect that.

PR #2581, opened 2016-09-22
qf-voip_inno: add a test for tollfree, fix issue wrt undefined ResellerId

fenollp:


applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
```diff
@@ -575,17 +575,21 @@ find_numbers(Context) ->
...
-                    Account -> Account
-                end,
+    {AccountId, ResellerId} =
+        case cb_context:account_id(Context) of
+            'undefined' ->
+                Account = cb_context:auth_account_id(Context),
+                {Account, kz_services:find_reseller_id(Account)};
+            Account ->
+                {Account, cb_context:reseller_id(Context)}
+        end,
```
lazedo: what if the auth_account also doesn't have a reseller_id ?
i think you should props:filter_undefined the result below

fenollp:
Rebased to fix CI

PR #2585, opened 2016-09-22
shape credit/debit transaction doc_id to modb style

onnet:


lazedo:
hey @onnet thanks for this.
i would prefer to add this in `kz_transaction:prepare_transaction` as first clause, ie, generate if undefined.
there are other parts calling `kz_transaction:save`
thoughts ?

lazedo:
@onnet great! thank you.

PR #2594, opened 2016-09-26
qf-la: add capability to disable looking up accounts down to port req…

fenollp:
…uests


k-anderson:
There is duplicate that I issued, #2596

PR #2597, opened 2016-09-26
Tracing docs

jamesaimonetti:
Some introductory docs for using the tracing facilities


core/kazoo_data/src/kazoo_data.erl
```diff
@@ -21,22 +21,37 @@
         ]
        ).
 
--spec trace_file() -> any().
+-type level() :: 'debug' |
```
fenollp: How about using `lager:log_level()`

PR #2598, opened 2016-09-27
Fix cleanup timer in tasks

skorobkov:
Cleanup timer set in seconds, but `erlang:start_timer/3` wants milliseconds


fenollp:
Wow my bad! thank you very much. Merging when CI is happy

PR #2603, opened 2016-09-27
Allow extra VM emails to be sent to extra addrs while excl. user VM email

danielfinke:
Previously, disabling voicemail emails on the user meant extra emails specified on the VM box itself were also not receiving the notification emails. These are not tied to the user so therefore they should not be excluded if the user has the notifications disabled.


jamesaimonetti:
@danielfinke Thanks! Good catch. Merged manually

PR #2610, opened 2016-09-28
When vm id not specified (i.e. *97 feature) fix crash in kvm_messages:count/2

danielfinke:
Doc id can be undefined on L1303. Reusing this undefined instead of the MailboxId acquired by get_mailbox_doc/4 (which may automatically find the vmbox id by caller id or owner id) causes a crash


lazedo:
@danielfinke thanks

PR #2611, opened 2016-09-28
Props, scripts, and specs

jamesaimonetti:
Removes usage of proplists in favor of props, documents scripts, and updates some Dialyzer complaints

### DIFF
--- a/applications/cccp/src/cccp_util.erl
+++ b/applications/cccp/src/cccp_util.erl
@@ -189,7 +189,7 @@ cccp_allowed_callee(Number) ->
             'true'
     end.
 
--spec build_request(_,ne_binary(),_,_,_,ne_binary(),_,ne_binary(),ne_binary(),ne_binary()) -> any().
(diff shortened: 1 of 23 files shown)


applications/cccp/src/cccp_util.erl
```diff
@@ -246,33 +247,35 @@ compose_cid(ToDID, RetainCID, RetainNumber, RetainName, AccountId) ->
...
-            {RetainNumber, RetainName}
+            end
     end.
 
 -spec maybe_cid_name(ne_binary(), ne_binary()) -> ne_binary().
 maybe_cid_name(<<Name/binary>>, _) -> Name;
 maybe_cid_name(_, Number) -> Number.
 
--spec build_presence(ne_binary()|'undefined', ne_binary()) -> ne_binary()|'undefined'.
+-spec build_presence(api_binary(), ne_binary()) -> api_binary().
```
fenollp: api_ne_binary()

scripts/conn-to-apps.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
fenollp: why require this?

PR #2613, opened 2016-09-28
qf-vm: call send_mwi_update only when mailbox_id is found

icehess:
When callling *97, we don't know the mailbox_id at first,
mwi should send updates when we found box_id


applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -116,8 +116,7 @@
 
 -define(KEY_LENGTH, 1).
 
--record(mailbox, {
-          mailbox_id :: api_binary()
+-record(mailbox, {mailbox_id = 'undefined' :: api_binary()
```
jamesaimonetti: Just as a note, record fields that don't have a default binding are bound to 'undefined' implicitly on record creation. So this isn't necessary (though it is explicit).

PR #2620, opened 2016-09-29
Qf hold control

lazedo:


applications/ecallmgr/src/ecallmgr_util.erl
```diff
@@ -129,6 +129,9 @@ send_cmd(Node, UUID, "conference", Args) ->
 send_cmd(Node, _UUID, "transfer", Args) ->
     lager:debug("transfering on ~s: ~s", [Node, Args]),
     freeswitch:api(Node, 'uuid_transfer', kz_util:to_list(Args));
+send_cmd(Node, _UUID, "uuid_" ++ _ = API, Args) ->
+    lager:debug("transfering on ~s: ~s", [Node, Args]),
```
fenollp: Logline looks copied from above: should it really say "transfering"?

PR #2621, opened 2016-09-29
qf-docs: properly document some stuff, fixes

fenollp:
@swysor Once this is merged you can use /v2/system_configs for managing custard!


jamesaimonetti:
@fenollp why delete clicktocall.md?

fenollp:
@jamesaimonetti Because it is no different to its ref/ counterpart.

Things in ref/ are automatically generated and hollow.
Things in doc/ are copies of the ref/ AND documented (meaning: example req & rep, non-placeholder headings, eventual notes).

PR #2623, opened 2016-09-29
KAZOO-5063 ensure to-did not being empty

jamhed:
Messages with is_empty To-DID field (<<"0">> as example) should not pass validation.

### DIFF
--- a/core/kazoo_amqp/src/api/kapi_offnet_resource.erl
+++ b/core/kazoo_amqp/src/api/kapi_offnet_resource.erl
@@ -155,7 +155,7 @@
         ,{?KEY_FLAGS, fun is_list/1}
         ,{?KEY_FORCE_FAX, fun kz_util:is_boolean/1}
         ,{?KEY_FORCE_OUTBOUND, fun kz_util:is_boolean/1}
-        ,{?KEY_TO_DID, fun is_ne_binary/1}
+        ,{?KEY_TO_DID, fun kz_util:is_ne_binary/1}
         ,{?KEY_BYPASS_E164, fun kz_util:is_boolean/1}
         ]).
 
@@ -170,10 +170,6 @@
                                      ]).
 -define(OFFNET_RESOURCE_RESP_TYPES, []).
 
--spec is_ne_binary(binary()) -> boolean().
-is_ne_binary(V) ->
-    is_binary(V)
-        andalso kz_util:is_not_empty(V).
 
 %%--------------------------------------------------------------------
 %% @doc Offnet resource request - see wiki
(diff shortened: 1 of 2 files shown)


core/kazoo_amqp/src/api/kapi_offnet_resource.erl
```diff
@@ -187,7 +187,8 @@ req(JObj) -> req(kz_json:to_proplist(JObj)).
 
 -spec req_v(api_terms()) -> boolean().
 req_v(Prop) when is_list(Prop) ->
-    kz_api:validate(Prop, ?OFFNET_RESOURCE_REQ_HEADERS, ?OFFNET_RESOURCE_REQ_VALUES, ?OFFNET_RESOURCE_REQ_TYPES);
+    kz_api:validate(Prop, ?OFFNET_RESOURCE_REQ_HEADERS, ?OFFNET_RESOURCE_REQ_VALUES, ?OFFNET_RESOURCE_REQ_TYPES)
+        andalso not kz_util:is_empty(props:get_value(?KEY_TO_DID, Prop));
```
lazedo: handle `KEY_TO_DID` values in `OFFNET_RESOURCE_REQ_TYPES`, presently is `,{?KEY_TO_DID, fun is_binary/1}`
resuming, you should try not add these kind of conditions here.

jamhed: <<"0">> is binary and kz_util:is_empty at the same time, so these checks are not enough. There is no other way to ensure value to be non-empty (except changing kz_json:get_ne_binary behaviour, what i think is not what we want here).

lazedo: what part did you not understand ?

fenollp: @jamhed I think @lazedo is saying to add this line over at/around https://github.com/2600hz/kazoo/blob/589a54952a51f35d1f7cf7cd4533c9f3bd13eca7/core/kazoo_amqp/src/api/kapi_offnet_resource.erl#L147

core/kazoo_amqp/src/api/kapi_offnet_resource.erl
```diff
@@ -170,6 +170,11 @@
                                      ]).
 -define(OFFNET_RESOURCE_RESP_TYPES, []).
 
+-spec is_ne_binary(binary()) -> boolean().
+is_ne_binary(V) ->
```
lazedo: move this to `kz_util`

PR #2624, opened 2016-09-29
Fixed a Typo

tsromana:
Everything works fine while saving multiple docs


lazedo:
@tsromana thanks for this. save_doc has a function clause for dealing with multiple docs, that's why it just worked. your commit mentions to keep the pr open but, i'm merging it as is, as it is in fact a typo

tsromana:
@lazedo thanks i was playing with it and wanted to test it before this was
merged thats why added that comment :) its good now btw the braintree pr is
also done please let me know if any issue

PR #2625, opened 2016-09-30
KAZOO-5092_3.22 ask provisioner to delete account before deleting it …

jamhed:
…from db


applications/crossbar/src/modules/cb_accounts.erl
```diff
@@ -1543,6 +1543,7 @@ delete_remove_sip_aggregates(Context) ->
 
 -spec delete_remove_db(cb_context:context()) -> cb_context:context() | boolean().
 delete_remove_db(Context) ->
+    _ = delete_remove_from_accounts(Context),
```
lazedo: why are you changing the order the delete is processed ?
i think you should move `_ = provisioner_util:maybe_delete_account(Context),` here, not to change the order related to accoundb / accountsdb

PR #2626, opened 2016-09-30
KAZOO-5092 ask provisioner to delete account before deleting it from db

jamhed:


applications/crossbar/src/modules/cb_accounts.erl
```diff
@@ -1541,6 +1541,7 @@ delete_remove_sip_aggregates(Context) ->
 
 -spec delete_remove_db(cb_context:context()) -> cb_context:context() | boolean().
 delete_remove_db(Context) ->
+    _ = delete_remove_from_accounts(Context),
```
lazedo: same as in #2625

PR #2628, opened 2016-09-30
KAZOO-5039 modb pagination

jamhed:
- applied to cb_notifications as example

### DIFF
--- a/applications/crossbar/src/crossbar_view.erl
+++ b/applications/crossbar/src/crossbar_view.erl
@@ -9,48 +9,49 @@
 
 -module(crossbar_view).
 -include("crossbar.hrl").
--export([load/2, load/3, descending/2, descending/3, ascending/2, ascending/3]).
+-export([load/2, load/3, load/4]).
 
 -define(PAGINATION_PAGE_SIZE, kapps_config:get_integer(?CONFIG_CAT, <<"pagination_page_size">>, 50)).
 -define(DEFAULT_RANGE, kapps_config:get_integer(?CONFIG_CAT, <<"maximum_range">>, (?SECONDS_IN_DAY * 31 + ?SECONDS_IN_HOUR))).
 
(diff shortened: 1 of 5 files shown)


applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -221,7 +221,7 @@ validate(Context) ->
     validate_notifications(maybe_update_db(Context), ReqVerb).
 
 validate(Context, ?SMTP_LOG) ->
-    load_smtp_log(Context);
+    crossbar_view:load(Context, ?CB_LIST_SMTP_LOG, fun(JObj) -> kz_json:get_value(<<"value">>, JObj) end);
```
lazedo: `fun(JObj) -> kz_json:get_value(<<"value">>, JObj) end`why not use something like `normalize_view_results ` ?

jamhed: you mean to name anonymous function? thought it is too simple to have a name.

applications/crossbar/src/crossbar_view.erl
```diff
@@ -0,0 +1,170 @@
...
+%%% @contributors
+%%%   Roman Galeev
+%%%-------------------------------------------------------------------
+
+-module(crossbar_view).
+-include("crossbar.hrl").
+-export([load/3, load/5]).
+
+-define(PAGINATION_PAGE_SIZE, kapps_config:get_integer(?CONFIG_CAT, <<"pagination_page_size">>, 50)).
+-define(DEFAULT_RANGE, kapps_config:get_integer(?CONFIG_CAT, <<"maximum_range">>, (?SECONDS_IN_DAY * 31 + ?SECONDS_IN_HOUR))).
```
icehess: you can use `?MAX_RANGE` which is defined in `crossbar.hrl`

icehess:
@jamhed @lazedo LGTM

PR #2637, opened 2016-10-04
OTP auth pr

tsromana:
This pr will let users utilise OTP to sigin into their accounts. 

Whats done :
Added OTP generation and checking module
OTP auth using user_auth module

TODO :
Add OTP secret generation method into cb_users
Testing
Documentation
Spec check
code optimisation

### DIFF
--- a/applications/crossbar/priv/api/swagger.json
+++ b/applications/crossbar/priv/api/swagger.json
@@ -7258,6 +7258,11 @@
                     "description": "crossbar maximum range",
                     "type": "integer"
                 },
+                "otp_strict_validation": {
+                    "default": false,
+                    "description": "crossbar otp strict validation",
+                    "type": "boolean"
+                },
                 "pagination_page_size": {
                     "default": 50,
                     "description": "crossbar pagination page size",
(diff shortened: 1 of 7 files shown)


lazedo:
@tsromana hey, took a look at this and here are some thoughts. you should not try to change user_auth but instead implement the authentication in cb_otp (or cb_otp_auth). the views should not be changed but instead create a new view to handle otp.  shouldn't the generated otp have the account_id and owner_id encoded ?

tsromana:
@lazedo no problem i could do that, the reason, i added it inside cb_user_auth is so that it is implemented as an extension to the existing module as if i wanted to build a cb_otp_auth it will contain basically all the same fns as user_auth and i didn't wanted to have repeat what's already done and if i wanted to have it completely working as as an alternative to the user_auth i would need to do that, otherwise if i only build it for auth, than the person using it will also needs user_auth enabled as a module for password recovery and using which an attacker knowing the username and password can easily get access to the account of a user just ignoring otp_auth the whole otp process, the code is backward compatible and won't hinder with normal working of user auth, if you want i can make some changes as you feel necessary and if you that it should be a separate module than no problem, also please let me know as for creating a token and saving it inside user doc should i create a separate module or could i do it inside cb_user :)

PR #2641, opened 2016-10-04
check if zone of event is available

lazedo:


applications/webhooks/src/webhooks_shared_listener.erl
```diff
@@ -295,14 +295,14 @@ handle_info(_Info, State) ->
...
                           || Z <- kz_nodes:whapp_zones(?APP_NAME)
                          ]),
-    Options = [{<<"should_handle_federated">>
-               ,kz_util:to_binary(kz_nodes:local_zone()) == hd(WHZones)
-               }
-              ],
+    ShouldHandle = kz_api:is_federated_event(JObj)
+        andalso not lists:member(kz_api:event_zone(JObj), WHZones)
+        andalso kz_util:to_binary(kz_nodes:local_zone()) == hd(WHZones),
+    Options = [{<<"should_handle_federated">>, ShouldHandle}],
```
lazedo: @k-anderson we're deciding if we should take a federated message and do the job if there is no node in that zone running webhooks. i would very much prefer to actually remove federation from webhooks listener and have it deal only with events within the zone in a shared queue with a limited persistence to allow restarts.

PR #2643, opened 2016-10-04
qf-modb (4.0): fix pattern matching crashes if pass encoded db name

icehess:


core/kazoo_modb/src/kazoo_modb.erl
```diff
@@ -272,20 +272,18 @@ maybe_create_current_modb(<<"account%2F", AccountId/binary>>) ->
     maybe_create_current_modb(binary:replace(AccountId, <<"%2F">>, <<>>, ['global'])).
 
 -spec create(ne_binary()) -> 'ok'.
-create(?MATCH_MODB_SUFFIX_RAW(AccountId, _, _) = AccountMODb) ->
```
lazedo: why is this removed? if the problem is encoded dbname, then add a function clause for that.

PR #2644, opened 2016-10-04
qf-KAZOO-5057 (3.22): fix pattern matching crashes if pass encoded db…

icehess:
… name


core/kazoo_modb-1.0.0/src/kazoo_modb.erl
```diff
@@ -221,10 +221,10 @@ maybe_create_current_modb(<<"account%2F", AccountId/binary>>) ->
     maybe_create_current_modb(binary:replace(AccountId, <<"%2F">>, <<>>, ['global'])).
 
 -spec create(ne_binary()) -> 'ok'.
-create(?MATCH_MODB_SUFFIX_RAW(AccountId, _, _) = AccountMODb) ->
```
lazedo: same here as in #2643

PR #2652, opened 2016-10-05
KAZOO-5106: script to dialyze changed files

jamesaimonetti:


scripts/dialyze-changed.bash
```diff
@@ -0,0 +1,27 @@
...
+}
+
+for ERL in $ERL_FILES; do
+    BEAM=$(erl_to_beam $ERL)
+    BEAM_FILES+=($BEAM)
+done
+
+ARGS=${BEAM_FILES[@]}
+echo "dialyzing changed files:"
+dialyzer --plt .kazoo.plt $MOD_BEAM $ARGS $@
```
fenollp: I'm curious: what do you have against `check-dialyzer.escript` or `TO_DIALYZE=... make dialyze`?

jamesaimonetti: AFAICT, `check-dialyzer.escript` dialyzes the individual files. This script dialyzes all the changed files together, in addition to letting you add additional beam files. So I could do:
```shell
./scripts/dialyze-changed.bash core/kazoo/ebin/kz_json.beam
```

And this will dialyze all changed files and include kz_json in the mix.

PR #2653, opened 2016-10-05
KAZOO-5103: fix issue with saving to older modbs

icehess:


core/kazoo_voicemail/src/kvm_message.erl
```diff
@@ -176,6 +176,7 @@ update(AccountId, BoxId, ?NE_BINARY = MsgId, Funs) ->
 update(AccountId, _BoxId, JObj, Funs) ->
     NewJObj = lists:foldl(fun(F, J) -> F(J) end, JObj, Funs),
     Db = kvm_util:get_db(AccountId, NewJObj),
+    io:format("Db ~p", [Db]),
```
lazedo: dev time log ?

icehess: removed

PR #2656, opened 2016-10-06
KAZOO-4711 account_id is not set in context

jamhed:


applications/crossbar/src/modules/cb_user_auth.erl
```diff
@@ -387,7 +387,7 @@ maybe_load_user_doc_by_username(Account, Context) ->
 %% @private
 -spec save_reset_id_then_send_email(cb_context:context()) -> cb_context:context().
 save_reset_id_then_send_email(Context) ->
-    MoDb = cb_context:account_modb(Context),
+    MoDb = kazoo_modb:get_modb(cb_context:account_db(Context)),
```
fenollp: get_modb returns an modb id, but the code after needs an modb db.

jamhed: well, if you look at account_modb/1, then you'll see that it uses get_modb/1.  so the only change is to provide get_modb/1 a valid input.

PR #2658, opened 2016-10-06
return error for schema errors instead of throw

lazedo:
also changes jesse dep to 1.4.2 which handles better the error list


lazedo:
also adds dev time util for loading schema from file

PR #2665, opened 2016-10-06
Qf system configs

fenollp:
For @swysor


applications/crossbar/src/modules/cb_system_configs.erl
```diff
@@ -185,11 +185,11 @@ delete(Context, Id, Node) ->
          ,kz_json:get_ne_value(Node, Doc)
          }
     of
-        {undefined, _} -> crossbar_util:response_bad_identifier(Id, Context);
-        {_, undefined} -> crossbar_util:response_bad_identifier(Node, Context);
+        {'undefined', _} -> crossbar_util:response_bad_identifier(Id, Context);
```
jamesaimonetti: Just FYI, I don't think this clause is necessary. The validate phase would return an error if `Id` was not valid.

PR #2669, opened 2016-10-07
qf-ports_srv_added: fix some issues with ports_request and services_added template

icehess:


applications/crossbar/src/crossbar_services.erl
```diff
@@ -305,22 +305,18 @@ base_audit_account(Context, Services) ->
...
-                      ,AccountName
-                      ,leak_auth_pvt_fields(AuthJObj)
-                     ).
-
--spec leak_auth_pvt_fields(kz_json:object()) -> kz_json:object().
-leak_auth_pvt_fields(JObj) ->
-    kz_json:set_values([{<<"account_id">>, kz_doc:account_id(JObj)}
-                        ,{<<"created">>, kz_doc:created(JObj)}
-                       ]
-                       ,kz_json:public_fields(JObj)
```
fenollp: No longer returning public fields?

icehess: nope, services_added only interested only in a few parameters not all of the accountjobj.

PR #2670, opened 2016-10-07
let my people list available in knm_managed unauthorized

onnet:


applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
```diff
@@ -593,6 +593,24 @@ get_find_numbers_req(Context) ->
       ,{'reseller_id', ResellerId}
       ]).
 
+-spec account_lookup(ne_binary()|'undefined', ne_binary()|'undefined') -> {ne_binary()|'undefined', ne_binary()}.
+account_lookup('undefined', ReqResellerId) ->
```
lazedo: `ReqResellerId` can be 'undefined', you should handle that first.

applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
```diff
@@ -593,6 +593,24 @@ get_find_numbers_req(Context) ->
...
+    case kz_services:is_reseller(ReqResellerId)
+        orelse cb_context:is_superduper_admin(ReqResellerId)
+    of
+        'false' -> {'undefined', kz_services:find_reseller_id('undefined')};
```
lazedo: `kz_services:find_reseller_id('undefined')` this returns {'ok', master} or {'error', _}.
but if `ReqResellerId` =/= 'undefined' and is not `reseller` the request is in `fault` and should not be processed.

applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
```diff
@@ -593,6 +593,24 @@ get_find_numbers_req(Context) ->
...
+        'true' -> maybe_unauthorized_lookup(ReqResellerId)
+    end;
+account_lookup(Account, _) ->
+    {Account, kz_services:find_reseller_id(Account)}.
+
+-spec maybe_unauthorized_lookup(ne_binary()) -> {ne_binary()|'undefined', ne_binary()}.
+maybe_unauthorized_lookup(ReqResellerId) ->
+    case kapps_account_config:get_global(ReqResellerId, ?KNM_CONFIG_CAT, <<"unauthorized_numbers_lookup">>, 'false') of
+        'true' -> {ReqResellerId, ReqResellerId};
+        'false' -> {'undefined', kz_services:find_reseller_id('undefined')}
```
lazedo: same here. if `ReqResellerId` is on query params and not allowed to do  so, the request is at `fault`

onnet:
@lazedo Do you want me to indicate faulty request in response, or empty data is OK, as it actually is now?

lazedo:
return a 404 so it can be tracked down in the logs

onnet:
@lazedo Finished here

PR #2674, opened 2016-10-07
KAZOO-4124: fix accessing number of account below superduper

fenollp:
If these changes are too radical, I'll move them to `knm_phone_number:is_in_account_hierarchy/2`.


fenollp:
Reason for closing:

> Should a subaccount be able to reserve+activate numbers that the MA has as available?

No, MA is in the hierarchy of subaccount , ie, subaccount tree contains MA.

PR #2676, opened 2016-10-07
keep some modules on dev releases

jamhed:


Makefile
```diff
@@ -79,14 +79,17 @@ clean-release:
 
 build-release: $(RELX) clean-release rel/relx.config rel/vm.args
 	$(RELX) --config rel/relx.config -V 2 release --relname 'kazoo'
-build-dev-release: $(RELX) clean-release rel/relx.config rel/vm.args
+build-dev-release: $(RELX) clean-release rel/relx.config-dev rel/vm.args
```
jamhed: @fenollp, no.

rel/relx.config.src
```diff
@@ -4,15 +4,28 @@ Apps = [list_to_atom(App)  || "applications/"++App = Dir <- filelib:wildcard("ap
...
          ,'.erlang.mk'
          ,'.settings'
          ,'skel'
          ,'parse_trans'
          ,'fs_sync'
          ,'fs_event'
          ,'reloader'
-         ],
+         ];
+		_ ->
```
lazedo: use explicit value here.

jamhed: The idea is that either env variable is set or it is not. Why do you want explicit value here?

PR #2678, opened 2016-10-07
Adding kvs cf_capture_group to call when we hunt for branch in cf_menu.

Mike-Dunton:
The idea here is if the menu has to hunt for a callflow that we then set the cf_capture_group on the call so that the flow we are branching to has access to the captured group(s)

Discussion here. https://groups.google.com/forum/#!topic/2600hz-dev/O1CfULJyqgs

keys should be accesiable with these calls. 

``` erlang
kapps_call:kvs_fetch('cf_capture_group', Call)
kapps_call:kvs_fetch('cf_capture_groups', Call)
```

### DIFF
--- a/applications/callflow/src/module/cf_menu.erl
+++ b/applications/callflow/src/module/cf_menu.erl
@@ -249,7 +249,11 @@ hunt_for_callflow(Digits, Menu, Call) ->
             lager:info("callflow hunt succeeded, branching"),
             _ = kapps_call_command:flush_dtmf(Call),
             _ = play_transferring_prompt(Menu, Call),
-            cf_exe:branch(kz_json:get_value(<<"flow">>, Flow, kz_json:new()), Call),
+            Props = [{'cf_capture_group', kz_json:get_ne_value(<<"capture_group">>, Flow)}
+                    ,{'cf_capture_groups', kz_json:get_value(<<"capture_groups">>, Flow, kz_json:new())}
+                    ],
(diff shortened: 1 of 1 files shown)


Mike-Dunton:
I will probably update this to use the dtmf_collection specific kapps_call functions 
```erlang
-spec set_dtmf_collection(api_binary(), call()) -> call().
-spec set_dtmf_collection(api_binary(), ne_binary(), call()) -> call().
set_dtmf_collection(DTMF, Call) ->
    set_dtmf_collection(DTMF, <<"default">>, Call).
set_dtmf_collection('undefined', Collection, Call) ->
    Collections = kvs_fetch(<<"dtmf_collections">>, kz_json:new(), Call),
    kvs_store(<<"dtmf_collections">>
             ,kz_json:delete_key(Collection, Collections)
             ,Call
             );
set_dtmf_collection(DTMF, Collection, Call) ->
    Collections = kvs_fetch(<<"dtmf_collections">>, kz_json:new(), Call),
    kvs_store(<<"dtmf_collections">>
             ,kz_json:set_value(Collection, DTMF, Collections)
             ,Call
             ).
```

lazedo:
HI @Mike-Dunton , i think this will not work as the updatedcall is only used to determine pid of the running cf_exe. you'll need to call cf_exe:update_call before you branch

lazedo:
@Mike-Dunton Thanks! keep them coming. :)

PR #2683, opened 2016-10-08
Numbers state transition to aging problem

onnet:


core/kazoo_number_manager/src/knm_number_states.erl
```diff
@@ -55,7 +55,7 @@ change_state(Number, _State) ->
 to_aging(Number) ->
     to_aging(Number, number_state(Number)).
 
-to_aging(Number, ?NUMBER_STATE_AGING) ->
+to_aging(Number, ?NUMBER_STATE_AVAILABLE) ->
```
lazedo: i think this is another typo, i think it should be `NUMBER_STATE_RELEASED`, but @fenollp would know better.

fenollp:
Hey @onnet sorry this has taken so long!
If you are you happy with https://github.com/2600hz/kazoo/pull/2810, please close this.

onnet:
@fenollp don't know if I'm happy, but predict I will be, therefore closing this one :)
Thanks!

PR #2686, opened 2016-10-11
Add "exp" claim to JWT

skorobkov:
Using `system_config/crossbar/token_auth_expiry` for calculating expiration date.
token_auth_expiry = 0 means no expiration ("exp" claim removed).

### DIFF
--- a/applications/crossbar/src/crossbar_auth.erl
+++ b/applications/crossbar/src/crossbar_auth.erl
@@ -15,6 +15,7 @@
 
 -include("crossbar.hrl").
 
+-define(TOKEN_AUTH_EXPIRY, kapps_config:get_integer(?APP_NAME, <<"token_auth_expiry">>, ?SECONDS_IN_HOUR)).
 
 -spec create_auth_token(cb_context:context(), atom()) ->
                                cb_context:context().
@@ -35,12 +36,17 @@ create_auth_token(Context, AuthModule, JObj) ->
 
     AccountId = kz_json:get_first_defined([<<"account_id">>, [<<"Claims">>, <<"account_id">>]], JObj),
     OwnerId = kz_json:get_first_defined([<<"owner_id">>, [<<"Claims">>, <<"owner_id">>]], JObj),
+    Expiration = case ?TOKEN_AUTH_EXPIRY of
+                     TokenExp when TokenExp > 0 -> erlang:system_time('seconds') + TokenExp;
+                     _ -> 'undefined'
+                 end,
 
     Claims = props:filter_undefined(
                [{<<"account_id">>, AccountId}
                ,{<<"owner_id">>, OwnerId}
                ,{<<"as">>, kz_json:get_value(<<"as">>, Data)}
                ,{<<"method">>, kz_util:to_binary(AuthModule)}
+               ,{<<"exp">>, Expiration}
                 | kz_json:to_proplist(kz_json:get_value(<<"Claims">>, JObj, kz_json:new()))
                ]),
 
@@ -77,3 +83,4 @@ authorize_auth_token(Token) ->
 
 maybe_db_token(AuthToken) ->
     kz_datamgr:open_cache_doc(?KZ_TOKEN_DB, AuthToken).
+


lazedo:
@skorobkov you are forcing a not intended default value . besides that looks good.

lazedo:
@skorobkov actually, this shouldn't be implemented here. it should come out in `Claims` object set by kazoo_auth

skorobkov:
Do you mean "exp" in "Claims" object should add `cb_user_auth` & `cb_api_auth` modules?

skorobkov:
And maybe each cb_*_auth module should use its own configruation parameter for token expiration (cb_user_auth_token_expiration, cb_pi_auth_token_expiration, ...)?

lazedo:
take a look here https://github.com/2600hz/kazoo/blob/master/core/kazoo_auth/src/kz_auth_token_util.erl#L70
to optionally include the exp claim depending on provider https://github.com/2600hz/kazoo/blob/master/core/kazoo_auth/src/kz_auth_providers.erl#L24

skorobkov:
Now I am confused.
For example I am authorizing via `cb_api_auth` (or `cb_user_auth`) module, it calls `crossbar_auth:create_auth_token(Context, ?MODULE)` and i do not see how it uses any of functions you mentioned (`kz_auth_token_util:create_claims` & `kz_auth_providers:kazoo_auth_provider`).

I beleive this functions used by new `cb_auth` module (which is in non working state now), but not old cb_*_auth modules.

lazedo:
`non working state now` ?

skorobkov:
No `auth.authorize.json` schema file, only `auth.callback.json`

skorobkov:
https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/modules/cb_auth.erl#L161
Call `kz_auth:validate_token(cb_context:doc(Context))` always fail, `cb_context:doc(Context)` retrun `kz_json:obect()` but `kz_auth:validate_token` want binary or map.

skorobkov:
To be clear, I want to explain my point of view.
We have billing server which communicate with kazoo using api_key.
Trying to connect it with kazooo4 i realize that X-Auth-Token now is "JSON Web Tokens", not a problem.
But i found that JWT generated by kazoo does not have "exp" field, so i created this PR.
And after you mentoined that "exp" should depend on provider I tried to understand `kazoo_auth` internals.
As I understand kazoo_auth is mostly OAuth(?) + JWT stuff. Complicated stuff. And now i have some questions:

1. How new kazoo_auth and old cb_*_auth modules should coexists and communicate? Will cb_*_auth modules be deprecated in future? Is there any note/post/tweet/... about this? And maybe we should forget about api_key and start migrating to OAuth?
2. `kz_auth:create_token/1` does not have any idea about "provider" it is just generate JWT from claims, and I do not know in what moment it should check provider and generate "exp" field?
3. If i want generate different "exp" field when authenticated via different cb_*_auth modules (ex: cb_api_auth moule generate 24h token, cb_user_auth - 1h token, etc... ), where is the best place for configuration of this, in `system_config/crossbar` or `system_config/auth` configuration? Or maybe use `sytem_auth` DB?

lazedo:
1) they still be supported
2) it does indirectly, i'll make it explicit. check the sign part.
3) as before it should be done depending on provider, it can be configured in `system_config/auth` or `system_auth` .


but..., jwt tokens don't need an explicit expiry if they have the entity_signed. if there's a problem with the token we just change the pvt_secret used to create the entity_signature

skorobkov:
Yes, i looked into `kz_auth:include_claims/1` it calls `kz_auth_identity:sign/1`. Behavior of `kz_auth_identity:sign/1` is not provider dependant, it just calls `kz_auth_providers:kazoo_auth_provider()`.
And I do not understand why and how adding "exp" in `kz_auth:create_token/1` should be provider dependant. This function called only by `crossbat:auth`, so provder always "kazoo".

More I dig into "kazoo_auth", more questions I have. And I do not know where to get answers. It simply is not going into one big picture.

I give up.
Sorry for your time.

jamesaimonetti:
@skorobkov I'm sorry you're frustrated trying to work with this new addition. Please keep asking questions! The more people that do, the better for everyone. @lazedo just arrived in SF for KazooCon so may be a bit jet-lagged so a response may be delayed a bit.

Thanks, in general, for your participation. We all appreciate it and I want to make sure you know it!

skorobkov:
Everything is ok.
I have no claims against @lazedo.
Perhaps the confusion happened because of the low level of my English.

I really need time to "play" with kazoo_auth to better understand it. I aggre with @lazedo that JWT generation should relay on kazoo_auth internals.
I wll try again later, when i have more knowledge about kazoo_auth.

lazedo:
@skorobkov english is also not my first language, so excuse me for any misunderstandings that this may have caused. we're in the week of kazooCon and we're pretty busy here trying to setup everything up. i will come up to this after kazooCon. as i stated before, `2) it does indirectly, i'll make it explicit.`. i believe that after this, it will be easy to add your request there.

skorobkov:
@lazedo ok lets continue after KazooCon.
At that time I will try to rework PR.

lazedo:
@skorobkov lets merge for now and rework provider base later.

PR #2694, opened 2016-10-13
Default timezone for new accounts to the parent account's timezone

danielfinke:
also deprecate system_config/accounts/timezone and use default_timezone instead

The value in timezone will be copied to default_timezone if the latter is undefined.
The core of this change is to make new accounts default to the timezone of the parent account. Removed the default value from the accounts.json schema as it would be merged in prior to the parent timezone check (note that timezone can still be included in the JSON of a new account). As well, we ensure that the timezones fetched by cf_util match. For accounts whose timezone key is blank, kz_account:timezone/1 will deal with fetching the parent or eventually failing back to the system default or even "Los_Angeles". Hence the removal of kz_account:timezone/2 on applications/callflow/src/cf_util.erl#L808 - kz_account:timezone/2 would use the second argument instead of trying the parent first.

### DIFF
--- a/applications/callflow/src/cf_util.erl
+++ b/applications/callflow/src/cf_util.erl
@@ -804,10 +804,8 @@ get_timezone(JObj, Call) ->
 -spec account_timezone(kapps_call:call()) -> ne_binary().
 account_timezone(Call) ->
     case kz_account:fetch(kapps_call:account_id(Call)) of
-        {'ok', AccountJObj} ->
-            kz_account:timezone(AccountJObj, ?DEFAULT_TIMEZONE);
-        {'error', _E} ->
-            kapps_config:get(<<"accounts">>, <<"timezone">>, ?DEFAULT_TIMEZONE)
+        {'ok', AccountJObj} -> kz_account:timezone(AccountJObj);
+        {'error', _E} -> ?DEFAULT_TIMEZONE
     end.
 
 -spec start_task(fun(), list(), kapps_call:call()) -> 'ok'.
--- a/applications/crossbar/doc/ref/accounts.md
+++ b/applications/crossbar/doc/ref/accounts.md
@@ -25,7 +25,7 @@ Key | Description | Type | Default | Required
 `ringtones` | Ringtone Parameters | `object` | `{}` | `false`
 `ringtones.external` | The alert info SIP header added when the call is from internal sources | `string(0..256)` |   | `false`
 `ringtones.internal` | The alert info SIP header added when the call is from external sources | `string(0..256)` |   | `false`
-`timezone` | The default timezone | `string(5..32)` | `America/Los_Angeles` | `false`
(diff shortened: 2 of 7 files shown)


danielfinke:
Can I get a little guidance on the formatting that CircleCI reports around L358 of kapps_maintenance.erl? I see that it is trying to align the second argument with the open parenthesis of the function call. Are we not doing the 2-char indent for long function call lines anymore (see the first argument to that call)?

jamesaimonetti:
@danielfinke yeah, the new formatter makes dropping the first arg in a function call problematic. We're having to revert to putting first arg on the same line as the function. So:

```erlang
V = do_it(Foo
         ,Bar
         ),
```
instead of
```erlang
V = do_it(
      Foo
      ,Bar
    )
```

jamesaimonetti:
@danielfinke I have a couple changes but it would be hard to put them in a comment here, so I'm creating a branch with my suggestions. I've also added a handful of spec/type changes as I've dialyzed the changes here (not necessarily related to your PR changes, more of an FYI). See [here](https://github.com/2600hz/kazoo/compare/voxter-master-pr-account-default-timezone) for a diff against master. Let me know if you have any questions, test the new code out to make sure it works as you expect (I think I was faithful to the operations), and I'll merge this all in if you are satisfied.

danielfinke:
@jamesaimonetti looks good to me for the most part. I identified an issue in both yours and mine, left a comment on that commit

jamesaimonetti:
@danielfinke ok, thanks.

And just for reference, here's the dialyzer stuff I ran:

1. First, dialyze changed files:
    ```shell
    ./scripts/dialyze-changed.bash
    ```
2. Address warnings found
3. Look at the unknown functions/types and include the beam files related to the changes in the code:
    ```shell
./scripts/dialyze-changed.bash core/kazoo_documents/ebin/kz_account.beam core/kazoo_json/ebin/kz_json.beam core/kazoo_data/ebin/kz_data.beam core/kazoo_data/ebin/kazoo_data.beam core/kazoo_documents/ebin/kzd_voicemail_box.beam applications/crossbar/ebin/crossbar_doc.beam applications/crossbar/ebin/cb_context.beam core/kazoo_call/ebin/kapps_call.beam core/kazoo/ebin/props.beam core/kazoo/ebin/kz_util.beam core/kazoo_config/ebin/kapps_config.beam
    ```
4. Go to 2

That's what I did for this particular PR anyway.

jamesaimonetti:
@danielfinke grr, i merged this but github doesn't believe me

danielfinke:
@jamesaimonetti I was pretty sure github only flags PRs as merged if you explicitly use the Merge button in the PR

jamesaimonetti:
I figured it out at one point but now it seems that process isn't working. Anyway, you have Author attribution on the commit, so I'll close this PR for now and revisit my process for merging so github notices.

PR #2695, opened 2016-10-14
Add call to Module:TaskName/3 when reache end of CSV file.

skorobkov:
When CSV file is over `kz_task_worker:loop/2` will try call `Module:TaskName/3` with args (kz_proplist(), task_iterator(), 'eof'). Return values from this call is ignored, after that `teardown/3` is called.

With this you can process lines in "bulk" mode.
Example: you get rows from CSV file and store them in `task_iterator()` on each 1000th row you insert accumulated rows in DB with one request. When you get call `Module:TaskName/3` with 'eof' you can insert rest of accumulated rows.

Also little rewrite actions on return values. Now you can return 'ok', {'ok', ItertorData} when prcessing input data from CSV (kz_task_worker).
Fixed datd which go into task_iterator() in `kz_task_noinput_worker`

### DIFF
--- a/applications/tasks/src/kz_task_noinput_worker.erl
+++ b/applications/tasks/src/kz_task_noinput_worker.erl
@@ -124,22 +124,29 @@ new_state_after_writing(WrittenSucceeded, WrittenFailed, State) ->
 is_task_successful(TaskId, API, ExtraArgs, IterValue) ->
     case tasks_bindings:apply(API, [ExtraArgs, IterValue]) of
         ['stop'] -> 'stop';
-        [{'ok', _Data}=NewIterValue] ->
-            %% For initialisation steps. Skeeps writing a CSV output row.
+        ['ok'] ->
+            {'true', 0, IterValue};
+        [{'ok', NewIterValue}] ->
             {'true', 0, NewIterValue};
-        [{[_|_]=NewRowOrRows, _Data}=NewIterValue] ->
-            Written = store_return(TaskId, NewRowOrRows),
-            {'true', Written, NewIterValue};
-        [{?NE_BINARY=NewRow, _Data}=NewIterValue] ->
-            Written = store_return(TaskId, NewRow),
-            {'true', Written, NewIterValue};
         [{'EXIT', _}] ->
             kz_util:log_stacktrace(),
             Written = store_return(TaskId, ?WORKER_TASK_FAILED),
             {'false', Written, 'stop'};
-        [{Error, _Data}=NewIterValue] ->
+        [{NewRowOrRows, NewIterValue}] when is_list(NewRowOrRows) ->
+            Written = store_return(TaskId, NewRowOrRows),
+            {'true', Written, NewIterValue};
+        [{NewRow, NewIterValue}] when is_binary(NewRow) ->
+            Written = store_return(TaskId, NewRow),
+            {'true', Written, NewIterValue};
+        [{Error, NewIterValue}] ->
             Written = store_return(TaskId, Error),
-            {'false', Written, NewIterValue}
+            {'false', Written, NewIterValue};
+        [NewRowOrRows] when is_list(NewRowOrRows) ->
+            Written = store_return(TaskId, NewRowOrRows),
+            {'true', Written, IterValue};
+        [Error] when is_binary(Error) ->
+            Written = store_return(TaskId, Error),
+            {'false', Written, IterValue}
     end.
 
 %% @private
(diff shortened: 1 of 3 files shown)


applications/tasks/src/kz_task_worker.erl
```diff
@@ -107,7 +107,9 @@ loop(IterValue, State=#state{task_id = TaskId
                             ,extra_args = ExtraArgs
                             }) ->
     case kz_csv:take_row(get(?IN)) of
-        'eof' -> teardown(State, API, IterValue);
+        'eof' ->
+            _ = is_task_successful(TaskId, API, ExtraArgs, FAssoc, 'eof', IterValue),
```
fenollp: Why call with `eof` as input when the same can already be done in `teardown/3`?

applications/tasks/src/kz_task_noinput_worker.erl
```diff
@@ -124,22 +124,29 @@ new_state_after_writing(WrittenSucceeded, WrittenFailed, State) ->
...
-            {'true', Written, NewIterValue};
         [{'EXIT', _}] ->
             kz_util:log_stacktrace(),
             Written = store_return(TaskId, ?WORKER_TASK_FAILED),
             {'false', Written, 'stop'};
-        [{Error, _Data}=NewIterValue] ->
+        [{NewRowOrRows, NewIterValue}] when is_list(NewRowOrRows) ->
+            Written = store_return(TaskId, NewRowOrRows),
+            {'true', Written, NewIterValue};
+        [{NewRow, NewIterValue}] when is_binary(NewRow) ->
```
fenollp: why relax condition on ?NE_BINARY?

fenollp:
Also, one can already return `{ok, NewIterValue}` or `{[], NewIteValuer}` or `[]` as see in kt_numbers.erl. Why change is_task_successful?

applications/tasks/src/kz_task_noinput_worker.erl
skorobkov: Just try to make kz_task_worker & kz_task_noinput_worker follow the same coding style.
Should i change them both to use ?NE_BINARY?

applications/tasks/src/kz_task_worker.erl
skorobkov: Because `teardown/3`  did not call `Module:TaskName/3`.
Tha main idea is to make a call `Module:TaskName/3`. where 3rd argument will be 'eof'.

skorobkov:
In https://github.com/2600hz/kazoo/blob/master/applications/tasks/doc/index.md sayed that return value can be 'ok' or {'ok', Data}. In `kz_task_noinput_worker` we have matching specification for {'ok', Data}, in `kz_task_worker` there is not such matching specification.
Another reason to change `kz_task_noinput_worker:is_task_successful/4` is wrong data going into `task_iterator()` it should be second argument of tuple not whole tuple.

fenollp:
@skorobkov Hey can you look into the recent changes (#2699) to tasks and see if that fits your needs?

On having `teardown/3` call `Module:TaskName/3`: `teardown/3` calls a cleanup function you can implement, with `IterValue` as argument. I would suggest doing your work this way instead of adding a call with eof.
See kt_numbers:import & kt_numbers:cleanup for a usage example.

skorobkov:
@fenollp #2699 only adds logging of {'EXIT', _} and elapsed time.

Ok, I rewert changes in kz_task_worker & kz_task_noinput_worker and try do it in `teardown`.
But i still have a question about functions `is_task_successful` in `kz_task_worker` and `kz_task_noinput_worker`.
`[{NewRowOrRows, NewIterValue}]` and `[{[_|_]=NewRowOrRows, _Data}=NewIterValue]`  what data should be in NewIterValue ? Only second part of tuple or whole tuple?

skorobkov:
BTW, do you think about using map for pass row data to action fun?
So instead using `import(Props, 'init', _1,_2,_3, _4,_5,_6, _7,_8,_9, _10,_11,_12, _13,_14,_15)` we can have `import(Props, 'init', #{cell1 => C1, cell5=C5}=Cells)`.

skorobkov:
Yes, `cleanup/2` do the job.

fenollp:
> [{NewRowOrRows, NewIterValue}] and [{[_|_]=NewRowOrRows, _Data}=NewIterValue] what data should be in NewIterValue ? Only second part of tuple or whole tuple?

True. If https://github.com/2600hz/kazoo/pull/2708 works for you I guess we can close your PR?

skorobkov:
Yes, #2708 is ok.
I am close this PR.
Thank you.

PR #2696, opened 2016-10-14
api-log: shorter log line for every request

fenollp:


jamesaimonetti:
some days GH is happy, other days...

PR #2698, opened 2016-10-14
bugfix: make milliwatt listen to each call request

jamhed:
* as it was before

### DIFF
--- a/applications/milliwatt/src/milliwatt_listener.erl
+++ b/applications/milliwatt/src/milliwatt_listener.erl
@@ -26,10 +26,7 @@
 -record(state, {}).
 -type state() :: #state{}.
 
--define(BINDINGS, [{'route', [{'types', ?RESOURCE_TYPES_HANDLED}
-                             ,{'restrict_to', ['no_account']}
-                             ]
-                   }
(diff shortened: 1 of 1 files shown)


applications/milliwatt/src/milliwatt_listener.erl
```diff
@@ -26,10 +26,7 @@
 -record(state, {}).
 -type state() :: #state{}.
 
--define(BINDINGS, [{'route', [{'types', ?RESOURCE_TYPES_HANDLED}
-                             ,{'restrict_to', ['no_account']}
```
lazedo: why are you removing this ? afaik milliwatt requests are supposed not to be autheticated

jamhed: 1. it was like that before
2. there is a use case in make-busy for that
3. there is no easy way to make make-busy do calls from devices unknown to kazoo
4. karl says so

lazedo: well, karl said to have milliwatt respond to unauthenticated calls, and that's why this is here

PR #2701, opened 2016-10-15
Part 1/2 of KAZOO-5093: store enough information in number doc

fenollp:
... so that a reduce view can be written (as part of PR 2/2) that would return the cumulated quantities.

I'm far from a services expert so please take a look: is all the data needed to compute quantities now stored on the number document?

Note: classifiers need not be in the number doc. They will be part of the reduce view.

Note: depends on #2700. Will need rebase to please Circle.


fenollp:
@lazedo Please review follow up https://github.com/2600hz/kazoo/pull/2722

PR #2703, opened 2016-10-16
Qf vmmessages

lazedo:


applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -60,6 +60,14 @@
                         )
        ).
 
+-define(ACCOUNT_VM_EXTENSION(AccountId)
+       ,kapps_account_config:get_global(AccountId
+                                       ,?CF_CONFIG_CAT
+                                       ,[?KEY_VOICEMAIL, ?KEY_EXTENSION]
+                                       ,<<"mp3">>
```
icehess: it shouldn't be `? DEFAULT_VM_EXTENSION` ?

icehess: aah nvm...

PR #2704, opened 2016-10-16
KazooCon2016 talks

jamesaimonetti:


fenollp:
Should this be in the code repo though? Kazoo's .git is already ~400MB, maybe this should go in its own binary blob friendly empty repo?

jamesaimonetti:
Its a handful of small images + the talks themselves (the org file = plain text). Are we space deficient?

PR #2705, opened 2016-10-18
KAZOO-5120: to_binary bug fix

tsromana:
When retrieving a recording info using /recordings/Id the api tried to convert the json to binary causing issues.

### DIFF
--- a/applications/crossbar/src/modules/cb_recordings.erl
+++ b/applications/crossbar/src/modules/cb_recordings.erl
@@ -90,7 +90,9 @@ content_types_provided(Context, _) ->
 
 -spec content_types_provided_for_download(cb_context:context(), http_method()) -> cb_context:context().
 content_types_provided_for_download(Context, ?HTTP_GET) ->
-    CTP = [{'to_binary', ?MEDIA_MIME_TYPES}],
+    CTP = [{'to_json', ?JSON_CONTENT_TYPES}
+          ,{'to_binary', ?MEDIA_MIME_TYPES}
+          ],
     cb_context:set_content_types_provided(Context, CTP);
 content_types_provided_for_download(Context, _Verb) ->
     Context.


tsromana:
@lazedo yea it works with the headers but when i dont provide it dont send me back the doc but tries to convert the json returned to binary and i recieved error as error badarg.

lazedo:
so isn't this just a matter of having a default ? why all these changes ?

tsromana:
@lazedo i just removed the to_binary from the api when there is no headers so that it doesn't try to convert the json from the doc to binary, let me try to retry the error and will let you know :)

tsromana:
@lazedo check this http://pastebin.com/Drfc5JQh so what happened was according to me, whenever we try to query for just the doc without the accept-audio/mpeg it will know that i need to read only the doc and it did that but when sending back reply it will crash with error that it was not able to convert the doc to binary, so i just made sure that when the request comes to read the doc it will first check if it is to read the doc or the download the file and than process accordingly, let it use to_binary when file otherwise do not covert to binary and than just fetch and down the doc, otherwise get the attachment and convert it to binary. i added the scope variable so that we don't need to check again if the request is read or download. :)

lazedo:
@tsromana thanks for trying to fix this. can you actually open a jira ticket describing your issue ? you said that if you provide accept-header it works, did you try with `application/json` ?

tsromana:
@lazedo yea sure np, yea i have tried that its the same. I get it what you meant by it should default to json, the problem is it sets the default value here for every request here https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/modules/cb_recordings.erl#L93
but when we request a json it defaults to json but does not change the content type provided and only remains to_binary or we can change the content type provided here https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/modules/cb_recordings.erl#L141 so its just a single line fix

tsromana:
@lazedo  opened an issue https://2600hz.atlassian.net/browse/KAZOO-5120 :)

lazedo:
@tsromana it seems that a lot of changes are happening here to solve a simple issue.
maybe @onnet can chime in, because he made the initial change on this and maybe there was something missing from he's pr.

tsromana:
yea sure :)

onnet:
@tsromana @lazedo Hi guys! It's a bit of a Friday wine already in my head now :)
Will try to take a look asap :)

onnet:
@tsromana Hi.

Yep, to_json forgotten:

I.e add {'to_json', ?JSON_CONTENT_TYPES} to line
https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/modules/cb_recordings.erl#L93

And after you'll get the next construction it should be OK:

    CTP = [{'to_json', ?JSON_CONTENT_TYPES}
          ,{'to_binary', ?MEDIA_MIME_TYPES}
          ],

Sorry for inconvenience. I'm just downloading and listening and never worried about recordings metadata. Therefore didn't checked this option.

Let me know if you want me to add these changes.

Regards,
Kirill

tsromana:
@onnet ohhh thank you, learnt something new :D

lazedo:
thanks @onnet . @tsromana want to try that simple change ?

tsromana:
@lazedo yea i will change the code later this night :)

tsromana:
couldn't test it yet as test server is down for some reason :/

tsromana:
all good tested works good 👍

lazedo:
@tsromana @onnet thanks

PR #2709, opened 2016-10-19
KAZOO-5124: optionally filter rates by resource id

onnet:
Would like to be able to apply rates based on gateway, call sent to.

### DIFF
--- a/applications/hotornot/src/hon_rater.erl
+++ b/applications/hotornot/src/hon_rater.erl
@@ -76,7 +76,11 @@ get_rate_data(JObj, ToDID, FromDID, Rates) ->
     RouteOptions = kz_json:get_value(<<"Options">>, JObj, []),
     RouteFlags   = kz_json:get_value(<<"Outbound-Flags">>, JObj, []),
     Direction    = kz_json:get_value(<<"Direction">>, JObj),
-    Matching     = hon_util:matching_rates(Rates, ToDID, Direction, RouteOptions++RouteFlags),
+    ResourceFlag = case kz_json:get_value(<<"Account-ID">>, JObj) of
+                       'undefined' -> [];
+                       AccountId -> maybe_add_resource_flag(JObj, AccountId)
+                   end,
+    Matching     = hon_util:matching_rates(Rates, ToDID, Direction, RouteOptions++RouteFlags++ResourceFlag),
 
     case hon_util:sort_rates(Matching) of
         [] ->
@@ -93,6 +97,17 @@ get_rate_data(JObj, ToDID, FromDID, Rates) ->
                         ]
                        ),
             {'ok', rate_resp(Rate, JObj)}
+    end.
+
+-spec maybe_add_resource_flag(kz_json:object(), ne_binary()) -> kz_proplist().
+maybe_add_resource_flag(JObj, AccountId) ->
+    case kapps_account_config:get_from_reseller(AccountId, ?APP_NAME, <<"filter_by_resource_id">>, 'false') of
+        'true' ->
+            case kz_json:get_value(<<"Resource-ID">>, JObj) of
+                'undefined' -> [];
+                ResourceId -> [ResourceId]
+            end;
+        'false' -> []
     end.
 
 -spec maybe_get_rate_discount(kz_json:object()) -> api_binary().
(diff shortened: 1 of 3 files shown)


lazedo:
@onnet can you please open a jira ticket explaining what are you trying to achieve here ?
from what i see in this pr, i believe this is not the right approach.

onnet:
@lazedo sure. Here it is: https://2600hz.atlassian.net/browse/KAZOO-5124

lazedo:
@onnet can you try putting 'resource_id' as an option instead of changing all these arities ?

applications/hotornot/src/hon_util.erl
```diff
@@ -68,32 +68,33 @@ build_keys(Number) ->
...
                             kz_json:objects().
--spec matching_rates(kz_json:objects(), ne_binary(), api_binary(), trunking_options()) ->
+-spec matching_rates(kz_json:objects(), ne_binary(), api_binary()|'undefined', ne_binary()|'undefined', trunking_options()) ->
                             kz_json:objects().
 matching_rates(Rates, DID) ->
-    matching_rates(Rates, DID, 'undefined', []).
+    matching_rates(Rates, DID, 'undefined', 'undefined', []).
 
-matching_rates(Rates, DID, Direction, RouteOptions) ->
+matching_rates(Rates, DID, Direction, ResourceId, RouteOptions) ->
```
lazedo: @onnet can you try to include `ResourceId` as an option instead ?

applications/hotornot/src/hon_util.erl
```diff
@@ -68,32 +68,33 @@ build_keys(Number) ->
...
     E164 = knm_converters:normalize(DID),
     [Rate || Rate <- Rates,
-             matching_rate(Rate, E164, Direction, RouteOptions)
+             matching_rate(Rate, E164, Direction, ResourceId, RouteOptions)
```
lazedo: @onnet same here. please also look into other places where you extended the funs to include ResourceId and see if you can work with an `option` in `RouteOptions`

onnet:
@lazedo Hi, changed PR to address your request, but this way "optionless rates" will be able to seep into Matching rates list. Anyway still possible to use as a temporary hack. Thanks!

applications/hotornot/src/hon_rater.erl
```diff
@@ -95,6 +99,17 @@ get_rate_data(JObj, ToDID, FromDID, Rates) ->
             {'ok', rate_resp(Rate, JObj)}
     end.
 
+-spec maybe_add_resource_flag(kz_json:object(), ne_binary()) -> kz_proplist().
+maybe_add_resource_flag(JObj, AccountId) ->
+    case kapps_account_config:get_from_reseller(AccountId, ?APP_NAME, <<"filter_by_resource_id">>, 'false') of
+        'true' ->
+            case kz_json:get_value(<<"Resource-ID">>, JObj) of
+                'undefined' -> [];
+                ResourceId -> [ResourceId]
```
onnet: @lazedo Sorry Luis, I'm afaraid I can't get you correctly. 
I'm just adding ResourceId as one more flag which is handled like all other flags at hon_util:matching_options/2
https://github.com/2600hz/kazoo/blob/master/applications/hotornot/src/hon_util.erl#L147

Looks like this:
RateOptions: [<<"721a97ab1628023536dccb2e1fc3aae7">>]
RouteOptions: [<<"fax">>,<<"iamoption">>,<<"721a97ab1628023536dccb2e1fc3aae7">>]

I believe using tuple like {filter_resource_id, ResourceId} will break the process.

I don't see any more ways to pass resource_id to hon_util:matching_rates/4 without arity raise.

Regards,

PS Looks like previous way was a bit more transparent

lazedo: @onnet i missed the hon_util part that does the filtering. previous was more readable but not being a permanent fix, changing the signature of the funs doesn't seem appropriate.

the previous comment is still valid if you handle the tuple in `hon_util:matching_options` i think

onnet:
@lazedo Hi. Is it you was talking about?

lazedo:
@onnet completely missed this part of your comment
```
Looks like this:
RateOptions: [<<"721a97ab1628023536dccb2e1fc3aae7">>]
RouteOptions: [<<"fax">>,<<"iamoption">>,<<"721a97ab1628023536dccb2e1fc3aae7">>]
``` 
sorry about that. the implementation using resource_id as a flag seems the appropriate one to address this use case.

onnet:
@lazedo Hi! No worries, will revert back :)

onnet:
@lazedo Reverted.

lazedo:
@onnet thanks

PR #2710, opened 2016-10-20
KAZOO-5117: convert body to binary to make sure content_lenght sets correctly

icehess:
and some code cleanup

### DIFF
--- a/core/kazoo_web/src/kz_http.erl
+++ b/core/kazoo_web/src/kz_http.erl
@@ -58,6 +58,8 @@
 -export_type([ret/0]).
 -export_type([req_id/0]).
 
+-define(REQ_URL_INDEX, 1).
+
 %%--------------------------------------------------------------------
 %% @public
 %% @doc Send synchronous request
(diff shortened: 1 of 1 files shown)


core/kazoo_web/src/kz_http.erl
```diff
@@ -279,7 +279,11 @@ build_request(Method, Url, Headers, Body) when (Method == 'post');
                                           ]
                                          ,Headers
                                          ,""),
-    {kz_util:to_list(Url), ensure_string_headers(Headers), kz_util:to_list(ContentType), Body}.
+    {kz_util:to_list(Url)
+    ,ensure_string_headers(Headers)
+    ,kz_util:to_list(ContentType)
+    ,kz_util:to_binary(Body)
```
fenollp: How about using `iolist_to_binary` instead?
Also, how about fixing all the calls to kz_http:req calls that call to_binary(Body)? This should save some time/confusion.

fenollp:
@lazedo @icehess I believe the comment on removing superfluous calls to to_binary/1 in code that uses kz_http is still relevant ;)

lazedo:
@fenollp agree. open a jira ticket and work on it :)

icehess:
@fenollp @lazedo Why? most of the time they call `kz_json:encode` or `unicode:*_to_binary` on Body which is not related to kz_http.  kz_http accepts both string(or iolist) and binary and at the end for consistency convert it to binary.

PR #2711, opened 2016-10-20
Json Pretty Print

tsromana:
added pretty printing option for json output

User is able to get the json returned formatted using Header value 'x-pretty-print: true' or inside url or body of the request, 

v1/url?pretty_print=true

### DIFF
--- a/applications/crossbar/src/api_resource.erl
+++ b/applications/crossbar/src/api_resource.erl
@@ -145,10 +145,11 @@ rest_init(Req0, Opts) ->
             {'ok', Req9, Context2};
         {Context1, Req8} ->
             {Req9, Context2} = api_util:get_auth_token(Req8, Context1),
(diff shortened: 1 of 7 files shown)


fenollp:
Hey @tsromana thanks for the patch. However I fail to see why you would want a web server to return pretty printed JSON when any client should be able to do that for you.

When using `curl` for example I would pipe its output to `| python -m json.tool`

tsromana:
@fenollp totally agree with you but there are some use cases where pretty printing the json from the server is a good idea, e.g. using tools such as firebug it is easier for a dev to look into info straight from the tool and also many windows tools used by people for api requests doesn't provide json formatting, so i thought it could help someone :) as many good rest api e.g. github pretty print their json which is more readable

lazedo:
@tsromana thanks

PR #2714, opened 2016-10-21
s3 allow international hosts

tsromana:
s3_bucket_after_host is changed to true as while working with s3 some bucket names gives ssl errors as s3 uses wildcard cert for buckets, the commit changes the bucket name to be appended to the back of the host name and it saves us from the invalid cert issue. The other way to get arround the issue is to use http but i guess this is a good way as it does not break anything and is secure. 

The change to hostname is added which will allow user to use any other bucket besides amazon's us-east which is being used as a default.

### DIFF
--- a/applications/crossbar/priv/couchdb/schemas/storage.attachment.aws.json
+++ b/applications/crossbar/priv/couchdb/schemas/storage.attachment.aws.json
@@ -16,6 +16,9 @@
                 "bucket": {
                     "type": "string"
                 },
+                "host": {
+                    "type": "string"
+                },
                 "key": {
                     "type": "string"
                 },
--- a/core/kazoo_attachments/src/kz_att_s3.erl
+++ b/core/kazoo_attachments/src/kz_att_s3.erl
@@ -13,6 +13,7 @@
 -export([put_attachment/6]).
 -export([fetch_attachment/4]).
 
+-define(AMAZON_S3_HOST, <<"s3.amazonaws.com">>).
 
 %% ====================================================================
 %% API functions
@@ -20,17 +21,13 @@
 
 -spec put_attachment(kz_data:connection(), ne_binary(), ne_binary(), ne_binary(), ne_binary(), kz_data:options()) -> any().
 put_attachment(Params, DbName, DocId, AName, Contents, _Options) ->
-    #{bucket := Bucket
-     ,key := Key
-     ,secret := Secret
-     ,path := Path
-     } = Params,
-    FilePath = kz_util:to_list(list_to_binary([Path, "/", DbName, "/", DocId, "_", AName])),
-    Config = kz_aws_s3:new(kz_util:to_list(Key), kz_util:to_list(Secret)),
+    {Bucket, Key, Secret, Path, Host} = get_map_values(Params),
+    FilePath = get_file_path(Path, DbName, DocId, AName),
+    Config = kz_aws_s3:new(kz_util:to_list(Key), kz_util:to_list(Secret), kz_util:to_list(Host)),
     case kz_aws_s3:put_object(kz_util:to_list(Bucket), FilePath, Contents, Config) of
         {'ok', Props} ->
             Metadata = [ convert_kv(KV) || KV <- Props, filter_kv(KV)],
-            S3Key = base64:encode(term_to_binary({Key, Secret, Bucket, Path})),
+            S3Key = base64:encode(term_to_binary({Key, Secret, Host, Bucket, Path})),
             {'ok', [{'attachment', [{<<"S3">>, S3Key}
                                    ,{<<"metadata">>, kz_json:from_list(Metadata)}
                                    ]}
@@ -43,9 +40,9 @@ fetch_attachment(HandlerProps, DbName, DocId, AName) ->
     case kz_json:get_value(<<"S3">>, HandlerProps) of
         'undefined' -> {'error', 'invalid_data'};
         S3 ->
-            {Key, Secret, Bucket, Path} = binary_to_term(base64:decode(S3)),
-            FilePath = kz_util:to_list(list_to_binary([Path, "/", DbName, "/", DocId, "_", AName])),
-            Config = kz_aws_s3:new(kz_util:to_list(Key), kz_util:to_list(Secret)),
+            {Key, Secret, Host, Bucket, Path} = get_s3_values(S3),
+            FilePath = get_file_path(Path, DbName, DocId, AName),
+            Config = kz_aws_s3:new(kz_util:to_list(Key), kz_util:to_list(Secret), kz_util:to_list(Host)),
             case kz_aws_s3:get_object(kz_util:to_list(Bucket), FilePath, Config) of
                 {'ok', Props} -> {'ok', props:get_value('content', Props)};
                 _E -> _E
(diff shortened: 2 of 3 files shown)


tsromana:
@lazedo according to me, its a bad idea to have key and secret appended to the attachements as the creds can change anytime, should we create a flag if a user wants to have its data found using new keys and creds in case he is storing his attachements and all of a sudden his creds will be useless and it may cause the attachments to be useless as it will always use the encoded creds inside the handler section of a doc ?

tsromana:
@lazedo so i was thinking that could we make a doc with s3 creds and than attach that doc into the handler section for the attachment just like oauth doc, so that in case a change in creds occurs we just need to change the creds inside the doc and everything else will remain the same ?

lazedo:
i prefer to handle it on s3 handler `maps:get(HandlerProps, Default)`.

lazedo:
about the creation of doc, i don't see a need for that as we already that info in system_data. i agree that we we can probably take the secret/key pair out of S#Props attached to the virtual attachment. the intent was to be able to get the attachment without the need for handler or exra lookups, but i guess that fetch_attachment will have the creds, so i'm ok with that change.

tsromana:
@lazedo yea i know, it was a great idea to just use that but there are some stupid people (e.g. me) who could change their s3 creds and than would mess up all their attachments 🤓, with the fetch_attachement you mean to say remove the encoded creds and use the connection instead ? https://github.com/2600hz/kazoo/blob/master/core/kazoo_data/src/kzs_attachments.erl#L42

lazedo:
@tsromana yes, remove them and use the ones from handlerprops

applications/crossbar/priv/couchdb/schemas/storage.attachment.aws.json
```diff
@@ -29,7 +32,8 @@
             "required": [
                 "bucket",
                 "key",
-                "secret"
+                "secret",
+                "path"
```
lazedo: @tsromana lets keep this optional and handle it in s3 handler

tsromana:
@lazedo so i am just a bit confused, if i remove this code, https://github.com/2600hz/kazoo/blob/master/core/kazoo_data/src/kzs_attachments.erl#L42
and jsut send connection details straight to kz_att_s3 and use the handler props from there, i guess i will need to update the google drive code as well to use the connection details from the handler props and not rely on the metadata inside the fake attachment right ?

or i could just add the connection details to the handler props in https://github.com/2600hz/kazoo/blob/master/core/kazoo_data/src/kzs_attachments.erl#L43
and than let google drive keep on using the connection details from the attachement metadata and move s3 to use connection details appended to the handlerprops, this way the google drive will keep on working as configured now and the s3 will use the new creds.

Please let me know what you think is best approach as i dont have experience with google drive as whether the oauth changes will cause any issue with gdrive's attachments. if so i will update the gdrive code as well 😃

lazedo:
@tsromana no need for changes in google drive, no need for changes in kzs_attachment

tsromana:
@lazedo the HandlerProps does not contain the Connection properties and only the  attachment metdata, so thats why i asked as we will need to provide that data from inside kzs_attachments

lazedo:
@tsromana you're right and this is why it was set in pvt_attachment, preventing and extra lookup. i'll figure out the best way later.
so for now, just add `host` support to s3. i'm still not sure about the change of the default that you want to do to `s3_bucket_after_host='true'`

tsromana:
@lazedo i have actually made changes to kzs_attachements which will keep everything same and does not  require dual lookup, i can send a pull here so you can look at it later :) if you like it, than it gold otherwise more room for improvement :) 

so for the 's3_bucket_after_host='true'' part, the url for the request for amazon is 

with true

https://s3.amazonaws.com/bucket

with false its

https://bucket.s3.amazonaws.com

as the url uses wildcard cert it will cause issue when the url is more than one dot e.g.

it will work with this
https://my-bucket.s3.amazonaws.com

but not with this
https://my.bucket.s3.amazonaws.com

but with true it will work however you name the bucket
https://s3.amazonaws.com/my.new.bucket

the other way to get arround the issue is to use http instead of https

Its just there so that someone doesn't complain about tls handshake failures due to bucket names.

lazedo:
@tsromana i understand the prepending of bucket name. i don't understand why do you want to change that, we've been using as is with no ssl issues (live demo at kazooCon 2016).

if you want to try to send a separate PR for kzs_attachments, i'm ok with that.

tsromana:
@lazedo try this url

https://my.new.bucket.s3.amazonaws.com

https://s3.amazonaws.com/my.new.bucket

it will lead to the same bucket address but when you will access both the first one will give tls handshake error or invalid cert error

lazedo:
@tsromana now i got it. but why do you want to use a bucket name with a dot ?

tsromana:
@lazedo hahaha i thought youd ask :) some crazy people like me do use dots in the bucket names (crazy java people just naming everything com.domain.bucket something) :D i have seen some companies use the format so that why just wanted a heads up so no one faces this problems and start scratching their heads if you want i can remove it from the pr

lazedo:
@tsromana i would prefer to have it removed, thanks

core/kazoo_attachments/src/kz_att_s3.erl
```diff
@@ -20,18 +20,15 @@
...
+    {Bucket, Key, Secret, Path, Host} = get_map_values(Params),
     FilePath = kz_util:to_list(list_to_binary([Path, "/", DbName, "/", DocId, "_", AName])),
-    Config = kz_aws_s3:new(kz_util:to_list(Key), kz_util:to_list(Secret)),
+    Config = kz_aws_s3:new(kz_util:to_list(Key), kz_util:to_list(Secret), kz_util:to_list(Host)),
     case kz_aws_s3:put_object(kz_util:to_list(Bucket), FilePath, Contents, Config) of
         {'ok', Props} ->
             Metadata = [ convert_kv(KV) || KV <- Props, filter_kv(KV)],
             S3Key = base64:encode(term_to_binary({Key, Secret, Bucket, Path})),
             {'ok', [{'attachment', [{<<"S3">>, S3Key}
+                                   ,{<<"Host">>, Host}
```
lazedo: this should go into S3Key

core/kazoo_attachments/src/kz_att_s3.erl
```diff
@@ -40,12 +37,13 @@ put_attachment(Params, DbName, DocId, AName, Contents, _Options) ->
 
 -spec fetch_attachment(kz_data:connection(), ne_binary(), ne_binary(), ne_binary()) -> any().
 fetch_attachment(HandlerProps, DbName, DocId, AName) ->
+    Host = get_host(HandlerProps),
```
lazedo: this should go as part of `binary_to_term`

core/kazoo_attachments/src/kz_att_s3.erl
tsromana: got it :) i just thought maybe someone may already be using it in productions so didn't wanna ruing their fun

lazedo: you should check the result of `binary_to_term`, maybe a fun to handle both cases.

tsromana: yea that would be good :) and if i handle the path as undefined should binary_to_term give out 'undefined' after decode or it doesn't handle undefined values ? any ideas, or i will check and see what it says :D

lazedo: path should be handled as optional on the put_attachment and have it default <<>> if not provided. when decoding you're ok.

tsromana: i am handling the filepath in another fun so that an empty path doesn't cause issue here
https://github.com/tsromana/kazoo/blob/838ffc97de1517cc6dbacbecd6fbcd075ed24ff2/core/kazoo_attachments/src/kz_att_s3.erl#L24

tsromana:
@lazedo ready for review, will pr for kzs_attachements later :)

core/kazoo_attachments/src/kz_att_s3.erl
```diff
@@ -65,3 +61,34 @@ convert_kv({K, V})
...
 convert_kv(KV) -> KV.
+
+-spec get_map_values(kz_data:connection()) -> {ne_binary(), ne_binary(), ne_binary(), api_binary(), ne_binary()}.
+get_map_values(#{bucket := Bucket
+                ,key := Key
+                ,secret := Secret
+                ,path := Path
+                ,host := Host
+                }) ->
+    {Bucket, Key, Secret, Path, Host};
```
lazedo: simplify this matching only bucket, key, secret and use `maps:get(.., Map, Default) for path and host

tsromana: got it :)

tsromana:
@lazedo hopefull good this time :p and sorry for any inconvenience, really love working with erlang specially kazoo ;)

core/kazoo_attachments/src/kz_att_s3.erl
```diff
@@ -65,3 +61,22 @@ convert_kv({K, V})
...
+
+-spec get_file_path(api_binary(), ne_binary(), ne_binary(), ne_binary()) -> list().
+get_file_path('undefined', DbName, DocId, AName) ->
+    kz_util:to_list(list_to_binary([DbName, "/", DocId, "_", AName]));
+get_file_path(Path, DbName, DocId, AName) ->
+    kz_util:to_list(list_to_binary([Path, "/", DbName, "/", DocId, "_", AName])).
+
+-spec get_s3_values(ne_binary()) -> {ne_binary(), ne_binary(), ne_binary(), ne_binary(), api_binary()}.
+get_s3_values(S3) ->
+    binary_to_term(base64:decode(S3)).
```
lazedo: @tsromana you still to handle the case where there is no host encoded.
something like 
```
case binary_to_term(base64:decode(S3)) of
    {_Bucket, _Key, _Secret, _Path} = V1 -> V1;
    {_Bucket, _Key, _Secret, _Path, _Host} = V2 -> V2
end.
```

tsromana:
@lazedo sorry forgot to do that, will issue pr regarding kzs_attachments once this is merged :) thank you

lazedo:
@tsromana looks good. i would like one last change if you don't mind.
we try to put constant definitions in macros at the top of the module or in a .hrl
can you please define ?AMAZON_S3_HOST macro with `<<"s3.amazonaws.com">>` and use the macro in the module ?
thanks

tsromana:
@lazedo done! sorry for the trouble, in future will keep these things in mind

lazedo:
@tsromana thank you.

PR #2719, opened 2016-10-21
KAZOO-5125: separate descriptions from system_config schemas

fenollp:
This also involves:

* adding a new system_config flag requires the committer to add corresponding documentation
* fixing lots of system_config settings that were read from different places with different default values.
* new way of creating system_config schemas does not merge them with previous version
* removed system_config flags thus get properly cleansed
* fixes to phone_numbers schema


applications/crossbar/priv/couchdb/schemas/system_config.jonny5.json
```diff
@@ -11,7 +11,7 @@
         },
         "default_{Key}": {
             "description": "jonny5 default {Key}",
-            "type": "integer"
+            "type": "boolean"
```
jamesaimonetti: This should actually be purged from the system config doc. Until we can infer Key, I think it better to leave these off the doc. Thoughts?

core/kazoo_apps/src/kapps_speech.hrl
```diff
@@ -49,7 +49,7 @@
         ,{<<"male/tr-tr">>, <<"eurturkishmale">>}
         ]
        ).
--define(ISPEECH_TTS_URL, kapps_config:get_string(?MOD_CONFIG_CAT, <<"tts_url">>, <<"http://api.ispeech.org/api/json">>)).
+-define(ISPEECH_TTS_URL, kapps_config:get_string(?MOD_CONFIG_CAT, <<"tts_url_ispeech">>, <<"http://api.ispeech.org/api/json">>)).
```
jamesaimonetti: Are you including a migrate command for this? It is possible that there are other values (I doubt it, but would rather address is than get yelled at that we broke someone's TTS engine).

applications/crossbar/priv/couchdb/schemas/system_config.jonny5.json
fenollp: Agreed!

fenollp: So looking into this I'd say I'll do it in another PR.

core/kazoo_apps/src/kapps_speech.hrl
fenollp: Done

applications/crossbar/priv/couchdb/schemas/system_config.jonny5.json
lazedo: @fenollp can you remove it now ?

PR #2720, opened 2016-10-21
Missing originate_uuid event breaks existing apps (noticed on acdc)

danielfinke:
Prior to this patch the originate_uuid event is not fired for uuids generated as part of an originate_req. However, they are used in existing applications and, in particular, event control queues exist for the missed uuids. Without this patch, commands such as hangup on those calls cannot reach FS.

### DIFF
--- a/applications/ecallmgr/src/ecallmgr_originate.erl
+++ b/applications/ecallmgr/src/ecallmgr_originate.erl
@@ -764,8 +764,8 @@ publish_originate_uuid(ServerId, UUID, JObj, CtrlQueue) ->
 
 -spec maybe_send_originate_uuid(created_uuid(), pid(), state()) -> 'ok'.
 maybe_send_originate_uuid({_, UUID}, Pid, #state{server_id=ServerId
-                                                   ,originate_req=JObj
-                                                   }) ->
+                                                ,originate_req=JObj
+                                                }) ->
     CtlQ = gen_listener:queue_name(Pid),
     publish_originate_uuid(ServerId, UUID, JObj, CtlQ).
 
--- a/applications/ecallmgr/src/ecallmgr_originate.erl
+++ b/applications/ecallmgr/src/ecallmgr_originate.erl
@@ -768,6 +768,8 @@ maybe_send_originate_uuid({'fs', UUID}, Pid, #state{server_id=ServerId
                                                    }) ->
     CtlQ = gen_listener:queue_name(Pid),
     publish_originate_uuid(ServerId, UUID, JObj, CtlQ);
+maybe_send_originate_uuid({'api', UUID}, Pid, State) ->
+    maybe_send_originate_uuid({'fs', UUID}, Pid, State);
 maybe_send_originate_uuid(_, _, _) -> 'ok'.
 
 -spec find_originate_timeout(kz_json:object()) -> pos_integer().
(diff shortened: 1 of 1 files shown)


applications/ecallmgr/src/ecallmgr_originate.erl
```diff
@@ -768,6 +768,8 @@ maybe_send_originate_uuid({'fs', UUID}, Pid, #state{server_id=ServerId
                                                    }) ->
     CtlQ = gen_listener:queue_name(Pid),
     publish_originate_uuid(ServerId, UUID, JObj, CtlQ);
+maybe_send_originate_uuid({'api', UUID}, Pid, State) ->
+    maybe_send_originate_uuid({'fs', UUID}, Pid, State);
```
lazedo: @danielfinke if the UUID is api generated isn't it already known by the caller to originate ?

danielfinke: @lazedo yes, however the control queue for the UUID is not - it is delivered by the originate_uuid event after an ecallmgr_call_control proc is spawned/initialized

lazedo: but still delivered right ? i don't see a reason why a short-circuit is needed here.

danielfinke: /only/ delivered if this patch is applied. That L770 publish_originate_uuid doesn't happen if this function is called with {'api', UUID} in the original code. You simply do not get the control queue.

danielfinke: @lazedo the output of dialyzer in CircleCI made me realize this can be written a bit differently. Functionality-wise it still stands that {'api', UUID} should result in the originate_uuid event being published

lazedo:
@danielfinke so you're stating that this breaks existing apps but ifaik this never existed. did you try to use `originate_ready`  instead ?

danielfinke:
@lazedo this was broken when
https://github.com/2600hz/kazoo/blob/3.22/applications/ecallmgr/src/ecallmgr_originate.erl#L795
was changed to
https://github.com/2600hz/kazoo/blob/master/applications/ecallmgr/src/ecallmgr_originate.erl#L807
between 3.22 and 4.0

lazedo:
@danielfinke this was an intended change. did you try using `originate_ready` ?

danielfinke:
@lazedo I can give `originate_ready` a try, but what was the intention of this change? As mentioned it broke any application that expected to get a control queue after originating a call with a custom call id

lazedo:
split applications that fire and forget from those who want control. what any other apps are you talking about ?

danielfinke:
Using `originate_ready` also forces us to disable Originate-Immediate... I think this means we have to publish an originate_execute?

danielfinke:
acdc is the only app within your code base that _I am aware of_ however we have at least two others that need a refactor as a result

danielfinke:
At cursory glance it looks like this will also have broken konami_transfer as it uses Originate-Immediate, an Outbound-Call-ID, and performs hangups on a kapps_call record (containing a control queue received in the originate_uuid)

lazedo:
if you say i take it as good. `konami` is an app that we don't intend to update atm. so, if you want to make a pull request to use originate ready, i'm ok with that.

lazedo:
yes, you get assured that originate is ready to go.

lazedo:
@danielfinke testing this as-is to see if it breaks anything. stay tuned and thanks for your patience.

lazedo:
@danielfinke this is still in staging but looks good. will update soon. don't see a need for originate_ready change. thanks for your patience on this.

PR #2721, opened 2016-10-21
Track/sychronize call position between queue mgrs

danielfinke:
Prerequisite patch to some other features/bug fixes/improvements in acdc.
Ensures an ordered(-ish) list of calls in a queue is shared amongst queue mgrs.


mark2600:
Looks okay, just fix the formatting so it passes make fmt and this can be accepted.

danielfinke:
Done. Thanks @mark2600

PR #2723, opened 2016-10-21
5120: update docs

fenollp:
I'll leave it to the guys at https://github.com/2600hz/kazoo/pull/2705 to actually fix the bug ;)


applications/crossbar/doc/recordings.md
```diff
@@ -9,35 +9,31 @@ Recordings endpoint provides a way to access call recordings.
...
 ```shell
 curl -v -X GET \
     http://{SERVER}:8000/v2/accounts/{ACCOUNT_ID}/recordings/{RECORDING_ID}
 ```
 
-#### Fetch recording media
+Gets a specific recording document attachment if available.
+Mind the `Accept` header in example below.
+For clients that do not support setting the `Accept` header, a querystring parameter can be included: `?accept=text/html`.
```
tsromana: accept=audio/mpeg ?

PR #2724, opened 2016-10-22
removing atom ticking policy

lazedo:
if you agree, add `plus`
don't merge yet

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -68,8 +68,6 @@ Some of the highlights:
 
 Some 2600Hz specifics:
 
-* Tickie all atoms
-  * Use `'foo'` not `foo`
 * Write accurate specs for all functions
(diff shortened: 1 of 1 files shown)


jamhed:
How can I add +100500?

mark2600:
I guess before voting I should have clarified.

Are we INVERTING this policy, or REMOVING it ?

lazedo:
removing the policy, NOT inverting it.
you can still add the tickies :)

jamesaimonetti:
@lazedo looks like its 6 to 8 in favor of keeping tickies! Time to close and resume yelling at PRs without them!

fenollp:
Can we up the ante a bit?
How about: if we keep the policy that REQUIRES tickies then let's build a tool that adds them for us?

lazedo:
@jamesaimonetti start yelling :)
@fenollp make the tool

PR #2726, opened 2016-10-23
Update to kzs_attachement

tsromana:
added use case where users att creds could be changed and at that point amazon s3 will disregard old creds and will use newer ones instead. 

While sending a request to get doc, it will check if the docs att handler and the systems att handler are same if not it will use the docs att handler creds, 
if they are same it will use the systems creds and if somehow the creds are messed up it will revert back to att's creds.

Todo:
Needs thorough testing

### DIFF
--- a/core/kazoo_attachments/src/kz_att_s3.erl
+++ b/core/kazoo_attachments/src/kz_att_s3.erl
@@ -36,11 +36,12 @@ put_attachment(Params, DbName, DocId, AName, Contents, _Options) ->
     end.
 
 -spec fetch_attachment(kz_data:connection(), ne_binary(), ne_binary(), ne_binary()) -> any().
-fetch_attachment(HandlerProps, DbName, DocId, AName) ->
-    case kz_json:get_value(<<"S3">>, HandlerProps) of
+fetch_attachment(Conn, DbName, DocId, AName) ->
+    HandlerProps = kz_json:get_value(<<"handler_props">>, Conn, 'undefined'),
+    case kz_json:get_value(<<"S3">>, Conn) of
         'undefined' -> {'error', 'invalid_data'};
         S3 ->
-            {Key, Secret, Host, Bucket, Path} = get_s3_values(S3),
+            {Key, Secret, Host, Bucket, Path} = get_s3_values(S3, HandlerProps),
             FilePath = get_file_path(Path, DbName, DocId, AName),
             Config = kz_aws_s3:new(kz_util:to_list(Key), kz_util:to_list(Secret), kz_util:to_list(Host)),
             case kz_aws_s3:get_object(kz_util:to_list(Bucket), FilePath, Config) of
@@ -63,7 +64,8 @@ convert_kv({<<"etag">> = K, V}) ->
     {K, binary:replace(V, <<$">>, <<>>, ['global'])};
 convert_kv(KV) -> KV.
 
--spec get_map_values(kz_data:connection()) -> {ne_binary(), ne_binary(), ne_binary(), api_binary(), ne_binary()}.
+-spec get_map_values(kz_data:connection()) ->
+                            {ne_binary(), ne_binary(), ne_binary(), api_binary(), ne_binary()}.
 get_map_values(#{'bucket' := Bucket
                 ,'key' := Key
                 ,'secret' := Secret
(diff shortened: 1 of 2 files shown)


lazedo:
@tsromana thanks. i'll take a look at this next week

tsromana:
no problem I also didn't get time yet to test this properly :) hopefully
will add some changes if I find some bug in it before your review

lazedo:
@tsromana LGTM, can you just rename `system_props` to `handler_props` ?

tsromana:
@lazedo done!

core/kazoo_attachments/src/kz_att_s3.erl
```diff
@@ -21,7 +21,7 @@
 
 -spec put_attachment(kz_data:connection(), ne_binary(), ne_binary(), ne_binary(), ne_binary(), kz_data:options()) -> any().
 put_attachment(Params, DbName, DocId, AName, Contents, _Options) ->
-    {Bucket, Key, Secret, Path, Host} = get_map_values(Params),
+    {Key, Secret, Host, Bucket, Path} = get_map_values(Params),
```
lazedo: @tsromana this is an unrelated change

core/kazoo_attachments/src/kz_att_s3.erl
```diff
@@ -63,26 +64,36 @@ convert_kv({<<"etag">> = K, V}) ->
...
+-spec get_map_values(kz_data:connection()) ->
+                            {ne_binary(), ne_binary(), ne_binary(), ne_binary(), api_binary()}.
 get_map_values(#{'bucket' := Bucket
                 ,'key' := Key
                 ,'secret' := Secret
                 }=Map) ->
     Path = maps:get('path', Map, 'undefined'),
     Host = maps:get('host', Map,  ?AMAZON_S3_HOST),
-    {Bucket, Key, Secret, Path, Host}.
+    {Key, Secret, Host, Bucket, Path}.
```
lazedo: unrelated change

core/kazoo_attachments/src/kz_att_s3.erl
```diff
@@ -63,26 +64,36 @@ convert_kv({<<"etag">> = K, V}) ->
...
 get_file_path(Path, DbName, DocId, AName) ->
     kz_util:to_list(list_to_binary([Path, "/", DbName, "/", DocId, "_", AName])).
 
--spec get_s3_values(ne_binary()) -> {ne_binary(), ne_binary(), ne_binary(), ne_binary(), api_binary()}.
-get_s3_values(S3) ->
+-spec get_s3_values(ne_binary(), kz_data:connection()) ->
+                           {ne_binary(), ne_binary(), ne_binary(), ne_binary(), api_binary()}.
+get_s3_values(S3, 'undefined') ->
     case binary_to_term(base64:decode(S3)) of
         {Key, Secret, Bucket, Path} ->
```
lazedo: i think this should be fixed  to {Bucket, Key, Secret, Path}

tsromana: @lazedo reverted back to same values, sorry was trying to aim for same output on both funs :)

tsromana: @lazedo these values are the default which we had earlier

core/kazoo_attachments/src/kz_att_s3.erl
```diff
@@ -78,11 +80,21 @@ get_file_path('undefined', DbName, DocId, AName) ->
...
             {Key, Secret, ?AMAZON_S3_HOST, Bucket, Path};
         {_Key, _Secret, _Host, _Bucket, _Path} = V2 ->
             V2
+    end;
+get_s3_values(S3, HandlerProps) ->
+    try get_map_values(HandlerProps) of
+        {Bucket, Key, Secret, Path, Host} ->
+            {Key, Secret, Host, Bucket, Path}
```
lazedo: @tsromana we should preserve initial `Host, Bucket, Path` values

tsromana: @lazedo i have changed the values, let me know if the var naming is correct or should i change it something else :)

tsromana:
@lazedo just to clarify the att handler we have for google is for google drive not for google cloud storage right

lazedo:
@tsromana looks good. circleci is failing. can you try to rebase to latest master ? tks

tsromana:
@lazedo will do :)

lazedo:
@tsromana looks good but, still reflecting on a possible issue.
if we save to Host = A with Secret = Y
when we fetch, we may be using Secret = Z configured for Host = B.
i think we should only override the `secret` if the host is the same.
thoughts ?

lazedo:
@tsromana so after IRC discussion, the new `key`and `secret` should only be used if `Host` and `bucket` from handler are equal to what we have in S3 object.
thanks

tsromana:
@lazedo will do and use the old ones if not same :)

core/kazoo_attachments/src/kz_att_s3.erl
```diff
@@ -78,11 +80,24 @@ get_file_path('undefined', DbName, DocId, AName) ->
...
+    {_Key, _Secret, Host, Bucket, Path} = get_s3_values(S3, 'undefined'),
+    try get_map_values(HandlerProps) of
+        {Bucket, Key, Secret, _Path, Host} ->
+            {Key, Secret, Host, Bucket, Path};
+        {_Bucket, Key, Secret, _Path, _Host} ->
+            {_Key, _Secret, Host, Bucket, Path}
+    catch
+        _E:_R ->
+            lager:info("unable to get s3 cred value using att values (~s): ~p", [_E, _R]),
+            {_Key, _Secret, Host, Bucket, Path}
```
lazedo: @tsromana we usualy use `underscore` for variables that we really don't use besides logging
how about
```
    Original =  {_, _, Host, Bucket, Path} = get_s3_values(S3, 'undefined'),
    try get_map_values(HandlerProps) of
        {Bucket, Key, Secret, _Path, Host} ->
            {Key, Secret, Host, Bucket, Path};
        _ -> Original
    catch
...
```

tsromana: @lazedo yea this looks more readable :)

PR #2727, opened 2016-10-24
Code coverage

fenollp:
What it brings: https://coveralls.io/github/fenollp/kazoo
We have to activate https://coveralls.io/github/2600hz/kazoo though. It is totally free even though they try to make it look like you need to register a credit card.

We're at 5.6% coverage with this test suite :/


make/deps.mk
```diff
@@ -6,6 +6,11 @@ DEPS = lager eiconv gen_smtp amqp_client cowboy jesse jiffy certifi couchbeam ws
 BUILD_DEPS = parse_trans
 IGNORE_DEPS = hamcrest
 
+ifeq ($(USER),travis)
```
fenollp: It is travis-only though. Not sure what an env var would change here

.travis.yml
```diff
@@ -33,3 +33,6 @@ script:
   - make eunit
   - # ERL_LIBS="$HOME/proper" ERLC_OPTS='-DPROPER' make compile-test
   - # ERL_LIBS="$HOME/proper" make eunit
+
+after_success:
+  - make coverage-report
```
skorobkov: It is cause a build fail on travis-ci if organisation dont use this service
https://travis-ci.org/dinkor/kazoo/jobs/170391627

fenollp: Should be fixed in https://github.com/2600hz/kazoo/pull/2735 ;)

skorobkov: Yes. It is ok now. Thank you.

PR #2731, opened 2016-10-25
KAZOO-5127: bring back PUT /v1,2/.../phone_numbers/Num/port

fenollp:


jamesaimonetti:
@fenollp some [dialyzer complaints](https://gist.github.com/jamesaimonetti/35492784c5812bd30cad5f4a8a572a8e) that may warrant investigation

jamesaimonetti:
I rebased to latest master to be sure it was dialyzing only your changes, hence the verified commits.

PR #2734, opened 2016-10-25
KAZOO-4804: More bells and whistles

skorobkov:
Added start/stop dates.
Added pagination in cb_resource_selectors
Add kazoo task module for importing/deleting selectotrs (and remove this from crossbar module)
Add filter by service plan
Doc updated


applications/crossbar/src/modules/cb_resource_selectors.erl
```diff
@@ -180,22 +147,56 @@ content_types_accepted_by_verb(Context, _) ->
...
+validate_rules(Context, ?HTTP_POST) ->
+    OnSuccess = fun on_successful_rules_validation/1,
+    cb_context:validate_request_data(<<"resource_selectors.rules">>, Context, OnSuccess).
+
+-spec validate_doc(cb_context:context(), path_token(), http_method()) -> cb_context:context().
+validate_doc(Context, UUID, ?HTTP_GET) ->
+    crossbar_doc:load(UUID, Context, ?TYPE_CHECK_OPTION(?DOC_PVT_TYPE));
+validate_doc(Context, UUID, ?HTTP_POST) ->
+    OnSuccess = fun(C) -> on_successful_doc_validation(C, UUID) end,
+    cb_context:validate_request_data(<<"resource_selector">>, Context, OnSuccess);
```
skorobkov: There is no significant reasons for it. When I was working on it just forgot to rename.

PR #2739, opened 2016-10-26
qf-proper: reactivate PropEr tests but kz_token_buckekts_pqc

fenollp:
Should I remove that module now that we have kazoo_auth?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,10 +29,11 @@ before_install:
   - cd $TRAVIS_BUILD_DIR
 
 script:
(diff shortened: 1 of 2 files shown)


fenollp:
This makes coverage go from 5.67% all the way to 5.85%!

jamesaimonetti:
@fenollp we use token buckets in callflows as a crude loop detector as well.

fenollp:
Yep and in Blackhole too.
@jamesaimonetti Ready!

PR #2747, opened 2016-10-28
ACDC: Improve state and login/logout management in agent

danielfinke:
Move logout operation to agent
Validate that logout is permitted in the current state
Better queueing support for agent pause/resume/logout
Clean up agent stats by publishing status less often, only after state fold
Better approach to auto-login of agents after acdc restart
Better agent queue binding

### DIFF
--- a/applications/acdc/src/acdc_agent_fsm.erl
+++ b/applications/acdc/src/acdc_agent_fsm.erl
@@ -1741,7 +1741,8 @@ get_endpoints(OrigEPs, AgentListener, Call, AgentId, QueueId) ->
 -spec return_to_state(non_neg_integer(), pos_integer()) -> 'paused' | 'ready'.
 return_to_state(Fails, MaxFails) ->
     lager:debug("fails ~b max ~b going to pause", [Fails, MaxFails]),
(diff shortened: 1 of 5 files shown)


applications/acdc/src/acdc_agent_fsm.erl
```diff
@@ -561,8 +555,8 @@ ready({'member_connect_win', JObj}, #state{agent_listener=AgentListener
             case get_endpoints(OrigEPs, AgentListener, Call, AgentId, QueueId) of
                 {'error', 'no_endpoints'} ->
                     lager:info("agent ~s has no endpoints assigned; logging agent out", [AgentId]),
-                    acdc_agent_listener:logout_agent(AgentListener),
                     acdc_agent_stats:agent_logged_out(AccountId, AgentId),
+                    ?MODULE:agent_logout(self()),
```
danielfinke: I may have been mindlessly replacing the old 'acdc_agent_listener' call as ?MODULE. I am happy to take it out as it is unnecessary

mark2600: That would be preferable :)

danielfinke:
@mark2600 made the change to remove ?MODULE

PR #2748, opened 2016-10-28
KAZOO-5133: implement number services map/reduce view

fenollp:
Note: 6af6747 can be reverted. It's still here so as to spawn some kind of discussion.


lazedo:
@fenollp rebase please

fenollp:
done

fenollp:
rebased again

core/kazoo_number_manager/src/knm_maintenance.erl
```diff
@@ -37,10 +38,36 @@
 %% API
 
 %% @public
+-spec maybe_update_number_services_view(ne_binary()) -> ok.
+maybe_update_number_services_view(?MATCH_ACCOUNT_ENCODED(_)=AccountDb) ->
```
lazedo: rename this to `update_number_services_view` since we log the result.
also add a fun clause for raw accountid

lazedo:
needs to hookup with account creation
needs to hookup with `kapps_maintenance:refresh_account_db`, maybe returned from get_all_account_views

core/kazoo_apps/src/kapps_maintenance.erl
```diff
@@ -402,8 +402,8 @@ refresh_account_db(Database) ->
     AccountId = kz_util:format_account_id(Database, 'raw'),
     _ = remove_depreciated_account_views(AccountDb),
     _ = ensure_account_definition(AccountDb, AccountId),
-    Views = get_all_account_views(),
-    _ = kapps_util:update_views(AccountDb, Views, 'true'),
+    _ = kapps_util:update_views(AccountDb, get_all_account_views(), 'true'),
+    _ = knm_maintenance:update_number_services_view(AccountDb),
```
lazedo: @fenollp i would prefer to split knm_maintenance:update_number_services_view if needed to have a fun that returns the view json, so it can be called from `kapps_maintenance:fetch_all_account_views`

fenollp: There's [a TODO](https://github.com/2600hz/kazoo/pull/2748/files#diff-61f44c46777bb6f0a3bd051aafcff1eeR45) that asks for per-account classifiers.
If we do it this way for now, we will have to move things around again when we implement the TODO.
What do you think?

fenollp:
This is waiting on @lazedo / @k-anderson

fenollp:
Note: updated view so that it doesn't require hardcoded lists of features or carrier modules.

PR #2751, opened 2016-10-28
Konami unable to bind bug

tsromana:
konami_listener expects to get Event-Name as bindings instead of bind


lazedo:
@tsromana thanks for the pr but the change should be applied to konami instead

tsromana:
@lazedo do you want me to open new pr? or guys will fix it :)

PR #2762, opened 2016-11-03
HELP-25802: adding support for authorization header

k-anderson:


core/kazoo_services/src/bookkeepers/kz_bookkeeper_http.erl
```diff
@@ -110,6 +109,23 @@ http_payload(#sync{content_type = <<"application/json">>} = Sync) ->
                              ]),
     kz_json:encode(JObj).
 
+-spec http_headers(sync()) -> proplist().
+http_headers(Sync) ->
+    props:filter_undefined(
```
icehess: `props:filter_empty/1` ?

PR #2767, opened 2016-11-04
KAZOO-5156: POST on numbers works as expected & other fixes

fenollp:


core/kazoo_number_manager/src/knm_services.erl
```diff
@@ -63,8 +63,8 @@ activate_feature(Number, {Feature,FeatureData}, BillingId, Services) ->
 
     case kz_services:check_bookkeeper(BillingId, TotalCharges) of
         'false' ->
-            lager:error("not enough credit to activate feature '~s' for $~p"
-                       ,[Feature, wht_util:units_to_dollars(Units)]),
+            lager:error("not enough credit to activate feature '~s' for $~p (~p)"
+                       ,[Feature, wht_util:units_to_dollars(Units), TotalCharges]),
```
lazedo: @fenollp `TotalCharges` should also be converted to $$

PR #2768, opened 2016-11-04
KAZOO-5136: try search cached bandwidth numbers first to support offset

icehess:


k-anderson:
@icehess I am concerned that this approach, while functional, has two drawbacks that I would like to voice.  
1) The initial search will take a considerable time to fetch 5000 numbers.  We should lower that significantly, say 100, and perhaps make it a system configuration parameter
2) The caching of numbers will cause them to become stale and we would likely offer numbers that are no longer available.  A reduction in the allowed discovery age will help this, I think a 1 day would be a good starting point.  

Ideally, we would determine a means to paginate Bandwidths API.  Perhaps we can proceed with this for now and open a discussion with the BW support staff to determine if there is a better approach.

PR #2772, opened 2016-11-06
KAZOO-5125: fix proxy username password missmatch

tsromana:
binary was being matched with a string which always resulted in username password mistmatch

### DIFF
--- a/core/kazoo_media/src/kz_media_store_proxy.erl
+++ b/core/kazoo_media/src/kz_media_store_proxy.erl
@@ -60,8 +60,8 @@ maybe_basic_authentication(Username, Password) ->
     AuthPassword = kapps_config:get_string(?CONFIG_CAT, <<"proxy_password">>, ""),
     not kz_util:is_empty(AuthUsername)
         andalso not kz_util:is_empty(AuthPassword)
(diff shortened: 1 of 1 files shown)


core/kazoo_media/src/kz_media_store_proxy.erl
```diff
@@ -60,8 +60,8 @@ maybe_basic_authentication(Username, Password) ->
     AuthPassword = kapps_config:get_string(?CONFIG_CAT, <<"proxy_password">>, ""),
     not kz_util:is_empty(AuthUsername)
         andalso not kz_util:is_empty(AuthPassword)
-        andalso Username == AuthUsername
-        andalso Password == AuthPassword.
+        andalso kz_util:to_list(Username) == AuthUsername
+        andalso kz_util:to_list(Password) == AuthPassword.
```
lazedo: @tsromana thanks for noticing this. use `kapps_config:get_binary` above instead.

lazedo:
@tsromana check https://github.com/2600hz/kazoo/pull/2773 and close this one, thanks.

tsromana:
@lazedo looks good thanks 👍

PR #2773, opened 2016-11-07
qf-filter out system generated default values

lazedo:
@fenollp 
https://github.com/2600hz/kazoo/commit/026d2548d2e2718c094efae4c9f6885b11baa6e8 broke some functionality in kazoo_media. we should adapt kapps_config_usage to improve schema/swagger generation. we should not change the core code (kazoo_media in this case) to improve schema/swagger generation.

there's still some work to be done in another PR to address the cases where a `case .. of` is used like `ssl_ca_cert` in crossbar_init

### DIFF
--- a/applications/crossbar/priv/api/swagger.json
+++ b/applications/crossbar/priv/api/swagger.json
@@ -7213,10 +7213,6 @@
                     "type": "string"
                 }
             },
-            "required": [
-                "secret_for_user_signature",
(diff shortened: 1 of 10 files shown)


fenollp:
Yeah the intent of these changes were to fix the different default definitions of the same flags throughout Kazoo, thus fixing some potential races.
Fix looks good

PR #2774, opened 2016-11-07
HELP-25764 change bridge timeout defaults

lazedo:


k-anderson:
@lazedo I don't think this is the correct approach.  Looking at the help ticket I see that the callflow has "new defaults" which is not the first time I have heard that.  Looks like the new callflow.x schemas are setting unexpected defaults, including these values:
https://github.com/2600hz/kazoo/blob/master/applications/crossbar/priv/couchdb/schemas/callflows.device.json#L55

Enforcing a minimum would make sense but having the default to 0 doesn't.  If we correct the schema I don't think we will need this patch.

Also, we must review the other schema and validate that their defaults make sense as this is likely not the only one causing trouble.  Maybe @jamesaimonetti can do a first pass since he created them?

PR #2777, opened 2016-11-08
restart file cache if missing

johnwnowlin:
We experienced a problem with prompts not playing after freeswitch cache timed out, then went to whistle to retrieve the file but the whistle_media_file_cache worker had terminated. It then returned a 404.

This change restarts the whistle_media_file_cache if it is stopped.

### DIFF
--- a/core/whistle_media-1.0.0/src/wh_media_continuous_proxy.erl
+++ b/core/whistle_media-1.0.0/src/wh_media_continuous_proxy.erl
@@ -45,9 +45,15 @@ init_from_doc(Db, Id, Attachment, Req) ->
         {'ok', Pid} ->
             {'ok', Req, wh_media_file_cache:continuous(Pid)};
         {'error', _} ->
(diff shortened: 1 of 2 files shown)


jamesaimonetti:
@johnwnowlin sorry for the delay. I don't forsee us accepting this for 3.19 since 3.22 is the currently supported version in the 3.x series. You could apply this as a hotpatch to your 3.19 install or we can include this patch in 3.22 for when you upgrade. Thoughts?

johnwnowlin:
@jamesaimonetti I had forgot I put it here since this was an inactive branch. We applied this in our site, you can apply the patch to 3.22 if you think it would be helpful.

jamesaimonetti:
@johnwnowlin got it! I'll apply to 3.22, thanks!

jamesaimonetti:
Commit [here](https://github.com/2600hz/kazoo/commit/0bb0fe898b469e6603c6b1f5f4c34aa301c2d9ff)

PR #2778, opened 2016-11-08
restart kz_media_file_cache if working has terminated or is missing

johnwnowlin:
We are running kazoo 3.19, so I made the original change in that branch. Since the logic appeared the same, I was told to make the change in the current active branch.

We are experiencing a problem with prompts not playing after the freeswitch cache times out. Freeswitch then goes to whistle to retrieve the file but the whistle_media_file_cache worker had terminated. The wh_media_single_prozy then returned a 404, which caused the prompt to fail to play.

This change restarts the whistle_media_file_cache worker if it is stopped.

### DIFF
--- a/core/kazoo_media/src/kz_media_continuous_proxy.erl
+++ b/core/kazoo_media/src/kz_media_continuous_proxy.erl
@@ -47,9 +47,15 @@ init_from_doc(Url, Req) ->
         {'ok', Pid} ->
             {'ok', Req, kz_media_file_cache:continuous(Pid)};
         {'error', _} ->
-            lager:debug("missing file server: 404"),
-            {'ok', Req1} = cowboy_req:reply(404, Req),
-            {'shutdown', Req1, 'ok'}
+            lager:debug("starting file server ~s/~s/~s", [Db, Id, Attachment]),
+            case kz_media_cache_sup:start_file_server(Db, Id, Attachment) of
+                {'ok', Pid} ->
+                    {'ok', Req, kz_media_file_cache:continuous(Pid)};
+                {'error', Error} ->
+                    lager:debug("start server failed ~s/~s/~s: ~p", [Db, Id, Attachment, Error]),
+                    {'ok', Req1} = cowboy_req:reply(404, Req),
+                    {'shutdown', Req1, 'ok'}
+            end
     end.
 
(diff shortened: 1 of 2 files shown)


core/kazoo_media/src/kz_media_continuous_proxy.erl
```diff
@@ -64,7 +70,7 @@ handle(Req0, {Meta, Bin}) ->
     lager:debug("media: ~s content-type: ~s size: ~b", [MediaName, ContentType, Size]),
 
     Req2 = case ContentType of
-               CT when CT =:= <<"audio/mpeg">> orelse CT =:= <<"audio/mp3">> ->
+               CT when CT =:= <<"audio/mpeg">>; CT =:= <<"audio/mp3">> ->
```
jamesaimonetti: The build job complained because the formatter wants `orelse` and `andalso` dropped to new lines. Replacing `orelse` with `;` changes the behaviour of the code (both conditions are evaluated vs the `orelse` which short-circuits when a 'true' condition is found). Also, we prefer `andalso` and `orelse` as they're more obvious that `,` and `;`, particularly with guards. So could you revert this change and drop the `orelse` condition down a line?

tsromana:
@johnwnowlin the ci error informs you to move orelse to new line

core/kazoo_media/src/kz_media_single_proxy.erl
```diff
@@ -73,7 +79,8 @@ handle(Req0, {Meta, Bin}) ->
     Url = kz_json:get_value(<<"url">>, Meta, <<>>),
 
     Req2 = case ContentType of
-               CT when CT =:= <<"audio/mpeg">> orelse CT =:= <<"audio/mp3">> ->
+               CT when CT =:= <<"audio/mpeg">> orelse
```
tsromana: @johnwnowlin i guess you need to move the whole orelse clause to new line

` CT when CT =:= <<"audio/mpeg">> 
         orelse CT =:= <<"audio/mp3">> ->`

jamesaimonetti:
@johnwnowlin I'll take care of the formatting issues. This looks good conceptually.

jamesaimonetti:
Merged but Github isn't auto-detecting to close the PR. I'll patch master too.

PR #2780, opened 2016-11-08
Fold properly when conference search result is empty

danielfinke:


lazedo:
@danielfinke good catch, thanks.

PR #2781, opened 2016-11-09
KAZOO-4883: receive hold/unhold event from FS and store them in channel record

icehess:


lazedo:
@icehess two things.

1) for a `freeswitch` event to be published, it must belong to `?CALL_EVENTS` defined in `kz_types.hrl` or have a function clause in `ecallmgr_call_events:should_publish/3`.

2) `blackhole` has a restriction on what `* ` means for subscribing to channel events. add also these 2 new events to LISTEN_TO macro in bh_call.

PR #2785, opened 2016-11-10
KAZOO-5141: move kapi_ modules to kazoo_amqp to fix startup of crossb…

mark2600:
…ar without acdc source


core/kazoo_amqp/src/api/kapi_acdc_queue.erl
```diff
@@ -60,7 +60,7 @@
 
 -export([queue_size/2, shared_queue_name/2]).
 
--include("acdc.hrl").
+-include("../../applications/acdc/src/acdc.hrl").
```
lazedo: is this include necessary ?

mark2600: yes, unfortunately

lazedo: we have to figure out a way to decouple this. we can't have core referring an app

jamesaimonetti: Why can't kapi_acdc_queue stay in the acdc application?

lazedo: the problem is that crossbar acdc modules use kapi_acdc but now we have deployments that have only crossbar, thus kapi_acdc is not installed.

this brings back the discussion on queue/agent api abstraction and having apps handling them, acdc or qubicle or other, which is my preferred solution.

PR #2794, opened 2016-11-11
Qf onhold

icehess:
This adds CHANNEL_HOLD and CHANNEL_UNHOLD to list of available websocket bindings in cb_websockets and also publish those events from `ecallmgr_call_events`.

But the published events doesn't have a SIP_HEADERS and CUSTOME_CHANNEL_VARS!!! thus they're not received by bonded modules(e.g., blackhole doesn't received them)

This needs to be addressed in FreeSWITCH itself due to how it treats these events.


lazedo:
@icehess remove the 2nd commit

PR #2798, opened 2016-11-14
Rating trie

jamesaimonetti:


applications/hotornot/src/hon_trie.erl
```diff
@@ -0,0 +1,154 @@
...
+        {'error', _} -> Acc
+    end.
+
+-spec init([]) -> {'ok', state()}.
+init([]) ->
+    PidRef = spawn_monitor(?MODULE, 'build_trie', [self()]),
+    lager:debug("building trie in ~p", [PidRef]),
+    {'ok', PidRef}.
+
+-spec handle_call(any(), pid_ref(), state()) ->
```
fenollp: Note: `handle_call_ret_state/1` types and similar should be used here.

applications/hotornot/src/hon_trie.erl
```diff
@@ -0,0 +1,154 @@
...
+-spec load_rate(ne_binary(), kz_json:objects()) -> kz_json:objects().
+load_rate(RateId, Acc) ->
+    case kz_datamgr:open_cache_doc(?KZ_RATES_DB, RateId) of
+        {'ok', RateDoc} -> [RateDoc | Acc];
+        {'error', _} -> Acc
+    end.
+
+-spec init([]) -> {'ok', state()}.
+init([]) ->
+    PidRef = spawn_monitor(?MODULE, 'build_trie', [self()]),
```
fenollp: I never understand this idea.
Right after init/1 and before build_trie/1,2,4 cast the Trie back to the gen_server, the gen_server will be unusable or in a useless state.
Why spawn?

Side question: why use monitor over trap_exit?

jamesaimonetti: The plan is to have the trie response {error, not_ready} while the trie is building. I also didn't want all the garbage involved with generating the trie in the main trie server process, hence the spawned worker.

A longer term plan is to try the trie and, on error, try the couch view (if trie is enabled). So even if a rate request comes in the first couple seconds of the app being started, it can still be serviced, albeit by the slower view lookup.

I like monitors over links when the spawning process doesn't care about the spawned process' health. The code can be changed to detect the DOWN when state is still the {pid(), reference()} and restart the trie building, for instance. I feel like monitors are lighter than links but that's entirely unsubstantiated.

PR #2799, opened 2016-11-14
Fix wrong server sending invalid conference config

danielfinke:
When the server not hosting the initial participant's conf_participant proc is the one to receive the config req for the conference profile information, it will return an error "no participant worker found for conference '<conference_id>'".
With this patch, if the server cannot find the participant worker (and therefore the conference data), it will nack the AMQP config_req and let another server handle it.

### DIFF
--- a/applications/conference/src/conf_config_req.erl
+++ b/applications/conference/src/conf_config_req.erl
@@ -7,42 +7,49 @@
 %%%-------------------------------------------------------------------
 -module(conf_config_req).
 
--export([handle_req/2]).
+-export([handle_req/3]).
 
 -include("conference.hrl").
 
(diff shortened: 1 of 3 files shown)


lazedo:
@danielfinke Thanks for the PR. after a quick look, it seems to me that the `get_conference` in `conf_config_req` needs to be fixed.

danielfinke:
@lazedo I agree. I'm already revisiting this right now anyway as I noticed a circumstance where this PR can be problematic. I'll approach a change to get_conference

danielfinke:
Reopening with alternative approach. Publish bootstrap_req and have the initial conf_participant reply with conference data. We can't use search_req to acquire the data from ecallmgr_fs_conferences as the conference hasn't been created yet (when looking for the "Profile" for the conference).

PR #2801, opened 2016-11-15
Fix wrong server sending invalid conference config

danielfinke:
Reopening with alternative approach. Publish bootstrap_req and have the initial conf_participant reply with conference data. We can't use search_req to acquire the data from ecallmgr_fs_conferences as the conference hasn't been created yet (when looking for the "Profile" for the conference).

Also note a quick fix in this PR for bad unbind on L1110-1111 of kapi_conference.erl.

### DIFF
--- a/applications/conference/src/conf_config_req.erl
+++ b/applications/conference/src/conf_config_req.erl
@@ -122,16 +122,22 @@ max_members_sound(Conference) ->
 get_conference(?DEFAULT_PROFILE_NAME) -> kapps_conference:new();
 get_conference(?PAGE_PROFILE_NAME) -> kapps_conference:new();
 get_conference(ConferenceID) ->
-    Participants =
-        [Pid ||
-            {_,Pid,'worker',['conf_participant']} <- supervisor:which_children('conf_participant_sup')],
-    Conferences = [conf_participant:conference(Pid) || Pid <- Participants],
-    case [Conference || {'ok',Conference} <- Conferences,
-                        kapps_conference:id(Conference) =:= ConferenceID]
-    of
-        [Conference|_] -> Conference;
-        _Else ->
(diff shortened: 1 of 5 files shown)


lazedo:
@danielfinke Thanks.  i think there's an easier way using `kapps_conference:from_conference_doc/1`. i'll create a PR and let you know for try/review. meanwhile, lets keep this open

danielfinke:
@lazedo I'm aware we could build that info off of the conference doc. However, notice for example that conf_discovery_req modifies the kapps_conference object (setting the play_entry_tone settings). As long as we have the ability to modify conference properties in this code, we can't just rely on what is in the couch doc.

lazedo:
@danielfinke i'm aware of that. play_entry_tone settings seems to be the only relevant. can you please rebase so that circleci warnings go away as we may merge this as is ?

Thanks

lazedo:
@danielfinke acked. this approach is actually our favorite so it should be merged soon.

lazedo:
Hi Daniel,

we evaluated the pros and cons of having conference configuration handled this way, but in the end its only going to allow the customization of  `play_entry_tone ` at the time of inception. we also considered adding a new fanout listener in conjunction with your code to handle this in a way to avoid the extra amqp call. we opted by preserving the shared listener and give every instance of conference app the ability to respond to conference configuration requests. moving forward, we plan to issue a PR for freeswitch mod_conference so that kazoo can receive all the data required to query the conference configuration and remove the encoding of account_id in the conference profile. we believe this will also allow us to provide different profiles based on callflow / user joining in.

the first part moving into that direction is https://github.com/2600hz/kazoo/pull/2817 which just got merged.

thank you for your work on this as it allowed us to properly evaluate/decide the direction we're moving to.

PR #2805, opened 2016-11-16
KAZOO-5150 check charges on activation

lazedo:


lazedo:
closing in favour of https://github.com/2600hz/kazoo/pull/2807

PR #2813, opened 2016-11-17
HELP-25782: migrate unassigned numbers & convert knm_bw to knm_bw2

icehess:


core/kazoo_number_manager/src/knm_phone_number.erl
```diff
@@ -700,8 +701,7 @@ set_module_name(N0, Name, IsBillable)
     N = set_module_name(N0, Name),
     N#knm_phone_number{is_billable = IsBillable};
 set_module_name(N0, Name, 'undefined') ->
-    N = set_module_name(N0, Name),
-    N#knm_phone_number{is_billable = knm_carriers:is_number_billable(N)}.
```
fenollp: Why?

core/kazoo_number_manager/src/knm_phone_number.erl
```diff
@@ -874,6 +874,19 @@ list_attachments(PhoneNumber, AuthBy) ->
 
 %%--------------------------------------------------------------------
 %% @private
+%% @doc Sanitize phone number docs fields and remove deprecated fields
+%% @end
+%%--------------------------------------------------------------------
+-spec sanitize_public_fields(kz_json:object()) -> kz_json:object().
+sanitize_public_fields(JObj) ->
+    kz_json:delete_keys([<<"id">>
+                        ,<<"used_by">>
```
fenollp: `<<"used_by">> =/= ?PVT_USED_BY` so why do this? My thinking is `<<"used_by">>` will be ignored anyway.

core/kazoo_number_manager/src/knm_maintenance.erl
```diff
@@ -89,14 +91,61 @@ fix_account_numbers(Account = ?NE_BINARY) ->
...
+migrate_unassigned_numbers(<<?KNM_DB_PREFIX_encoded, Suffix/binary>>) ->
+    migrate_unassigned_numbers(<<?KNM_DB_PREFIX_ENCODED, Suffix/binary>>);
+migrate_unassigned_numbers(<<?KNM_DB_PREFIX, Suffix/binary>>) ->
+    migrate_unassigned_numbers(<<?KNM_DB_PREFIX_ENCODED, Suffix/binary>>);
+migrate_unassigned_numbers(Number) ->
+    migrate_unassigned_numbers(knm_converters:to_db(Number)).
+
+migrate_unassigned_numbers(NumberDb, Offset) ->
+    ViewOptions = [{limit, kz_datamgr:max_bulk_insert()}
+                  ,{skip, Offset}
```
fenollp: Shouldn't this be `Offset + 1`?

core/kazoo_number_manager/src/knm_services.erl
```diff
@@ -148,7 +148,8 @@ update_services(Number, 'false', _) ->
     _ = kz_services:reconcile(PrevAssignedTo, <<"phone_numbers">>),
     Services = fetch_services(Number),
     Transactions = knm_number:transactions(Number),
-    _ = kz_services:commit_transactions(Services, Transactions),
+    _ = 'undefined' =/= AssignedTo
```
fenollp: How about checking this earlier? Do we want to compute services on unassigned numbers?

core/kazoo_number_manager/src/knm_phone_number.erl
icehess: Why not? it same as the last `set_module_name/2`,  but setting `is_billable` twice

core/kazoo_number_manager/src/knm_phone_number.erl
icehess: Wasn't `<<"used_by">>` replaced by `?PVT_USED_BY` ? I thought that was replaced, so removing it to prevent future confusion.

core/kazoo_number_manager/src/knm_maintenance.erl
icehess: nope, it is setting in the case clause to `Offset + Length`. For the first iteration, it's okay as well and will start at beginning of view result.

core/kazoo_number_manager/src/knm_services.erl
icehess: I have no idea about this one.

core/kazoo_number_manager/src/knm_maintenance.erl
```diff
@@ -151,6 +200,16 @@ fix_docs({ok, NumDoc}, Doc, AccountDb, NumberDb, DID) ->
             ok
     end.
 
+-spec fix_unassign_doc(ne_binary()) -> 'ok'.
+fix_unassign_doc(DID) ->
+    Routines = [{fun knm_phone_number:set_used_by/2, undefined}
```
icehess: Tickies are so 2015 ;)

But anyway I prefer to be consistence with the current format of the file which is not used tickies at all

core/kazoo_number_manager/src/knm_phone_number.erl
fenollp: yes pvt_used_by is the field. I'm saying there is no reason to remove used_by.

core/kazoo_number_manager/src/knm_services.erl
fenollp: @lazedo

core/kazoo_number_manager/src/knm_phone_number.erl
fenollp: ok

PR #2814, opened 2016-11-18
KAZOO-5186 storage lookup

lazedo:


core/kazoo_media/src/kz_media_recording.erl
```diff
@@ -517,14 +517,24 @@ store_url(#state{doc_db=Db
...
-        'true' -> should_store_recording();
+        'true' -> maybe_storage_plan(AccountId);
         'false' -> {'true', 'other', Url}
     end.
 
+-spec maybe_storage_plan(ne_binary()) -> store_url().
+maybe_storage_plan(AccountId) ->
+    AccountDb = kz_util:format_account_mod_id(AccountId),
+    Plan = kzs_plan:get_dataplan(AccountDb, <<"call_recording">>),
+    case maps:get('tag', Plan, 'local') =/= 'local'
```
fenollp: When `tag := Notlocal` it returns `{true, local}`. Is this a typo?

lazedo: no, it has different meanings. {'true', 'local'} means "yes, go ahead and save". `tag := local` is default connection.
if connection is default and no `att_handler` we should look in system_config as before.

PR #2818, opened 2016-11-18
HELP-25940: kazoo updates for cdr summary API

k-anderson:


k-anderson:
Please do not merge yet, requires an update to couchbeam to work properly.

applications/crossbar/doc/ref/cdrs.md
```diff
@@ -72,6 +72,16 @@ curl -v -X GET \
 
 #### Fetch
 
+> GET /v2/accounts/{ACCOUNT_ID}/cdrs/summary
```
jamesaimonetti: Looks like a new endpoint path that should be added to the [doc](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/doc/cdrs.md), eh?

fenollp:
Where are we on the couchbeam patch?
BTW it looks like we're using an outdated version of couchbeam: https://github.com/lazedo/couchbeam/compare/master...1.3.1.2

lazedo:
proper compare https://github.com/lazedo/couchbeam/compare/1.3.1.2...benoitc:master

PR #2756, opened 2016-11-01
Ensure whole project's exported functions are spec-ed

fenollp:
This adds a compilation flag project-wide: `+warn_missing_spec`.
This means that if a module is missing a `-spec` on an exported function, compilation will now fail.

Help requested on how to fix Dialyzer complaints in CI! but chances are they don't really matter.

Spéciale dédicace à @jamesaimonetti

PR #2819, opened 2016-11-19
KAZOO-5188: ensure all exported functions are spec-ed

fenollp:
Superseeds https://github.com/2600hz/kazoo/pull/2756


fenollp:
@lazedo yay!

applications/stepswitch/src/stepswitch_originate.erl
```diff
@@ -71,6 +71,7 @@ start_link(Endpoints, OffnetReq) ->
 %%                     {stop, Reason}
 %% @end
 %%--------------------------------------------------------------------
+-spec init([kz_json:objects() | kapi_offnet_resource:req()]) -> {'ok', state()}.
```
lazedo: why `| ` and not `,` ?

applications/stepswitch/src/stepswitch_sms.erl
```diff
@@ -80,6 +80,7 @@ start_link(Endpoints, OffnetReq) ->
 %%                     {stop, Reason}
 %% @end
 %%--------------------------------------------------------------------
+-spec init([kz_json:objects() | kapi_offnet_resource:req()]) -> {'ok', state()}.
```
lazedo: why `|` and not `,` ?

applications/teletype/src/teletype_maintenance.erl
```diff
@@ -58,8 +58,8 @@ convert_for_printing([_|_]=Vs) -> kz_util:join_binary(Vs, <<",">>).
 -spec receipt_for_printing(ne_binary()) -> ne_binary().
 receipt_for_printing(Receipt) ->
     case re:run(Receipt
-                ,<<"^2.0.0 Ok: queued as ([[:alnum:]]+).*$">>
-                ,[{'capture', 'all_but_first', 'binary'}]
+               ,<<"^2.0.0 Ok: queued as ([[:alnum:]]+).*\$">>
```
lazedo: why adding "\" ?

applications/stepswitch/src/stepswitch_originate.erl
fenollp: In specs, like types, lists need union (`|`) rather than `,` (and if you try to put `,`s there it wouldn't compile).
I tried as much as I could to make the init lists of gen_* modules "look like" the actual list of arguments, but in the end it's just a trick to help documentating and/or readability.

applications/stepswitch/src/stepswitch_sms.erl
fenollp: same as above

applications/teletype/src/teletype_maintenance.erl
fenollp: For formatting reasons only: https://bugs.erlang.org/browse/ERL-204

lazedo:
@fenollp yay!

fenollp:
@lazedo This is amazing! :D

PR #2821, opened 2016-11-21
KAZOO-5180: fix accumulating services for collection ops on numbers

fenollp:


core/kazoo_number_manager/src/carriers/knm_bandwidth2.erl
```diff
@@ -116,7 +116,7 @@ query_local_bw_numbers(Number, Quantity, Options, AccountId)
     ViewOptions = [{'startkey', [?NUMBER_STATE_DISCOVERY, BW2Module, Number]}
                   ,{'endkey', [?NUMBER_STATE_DISCOVERY, BW2Module, <<"\ufff0">>]}
                   ,{'limit', Quantity}
-                  ,{skip, Offset}
+                  ,{skip, Offset+1}
```
lazedo: @fenollp maybe the problem is with the providers implementing the find_more logic ? or i'm missing something, can't see why knm_managed would loop.

core/kazoo_number_manager/src/carriers/knm_managed.erl
```diff
@@ -77,7 +77,7 @@ do_find_numbers_in_account(Prefix, Quantity, Offset, AccountId) ->
     ViewOptions = [{'startkey', [AccountId, ?NUMBER_STATE_AVAILABLE, Prefix]}
                   ,{'endkey', [AccountId, ?NUMBER_STATE_AVAILABLE, <<Prefix/binary,"\ufff0">>]}
                   ,{'limit', Quantity}
-                  ,{skip, Offset}
+                  ,{skip, Offset+1}
```
lazedo: @fenollp why do you think this is looping here ?

lazedo:
@fenollp lets try the other approach we talked

fenollp:
This is not ready. After discussion with @lazedo I'm rewriting `knm_numbers`.

PR #2823, opened 2016-11-22
KAZOO-5133: DANGEROUSLY fix filter_empty

fenollp:


fenollp:
Extracted from https://github.com/2600hz/kazoo/pull/2748

lazedo:
this needs to be here for a while. we should merge it locally and run it in dev env to see if something comes up.

jamesaimonetti:
I think the intent of filter_empty was to emulate ["falsy"](https://www.sitepoint.com/javascript-truthy-falsy/) values in Javascript. Perhaps we can add a 'filter_falsy' that does what 'filter_empty' did prior to this PR, update all filter_empty to filter_falsy, and then selectively update to the new filter_empty as needed?

fenollp:
Actually, there's all the direct usages of `is_empty/1` too that should probably get converted to `is_falsy/1` (as a first step).

k-anderson:
@fenollp this needs to wait until 4.1+ AND we have a complete make-busy test suite to ensure no unforeseen issues arise from this change ;)

PR #2824, opened 2016-11-22
KAZOO-5133: make sure fields set to false get into number docs

fenollp:


fenollp:
Extracted from https://github.com/2600hz/kazoo/pull/2748

PR #2825, opened 2016-11-22
KAZOO-5142: make sure fetching CNAM takes less than 3s

fenollp:


core/kazoo/src/kz_util.erl
```diff
@@ -786,6 +787,25 @@ find_callid(APITerm, GetFun) ->
           ,?LOG_SYSTEM_ID
           ).
 
+%% @public
+%% @doc
+%% Gives `MaxTime' milliseconds to `Fun' of `Arguments' to apply.
+%% If time is elapsed, the sub-process is killed & function returns `false'.
+%% @end
+-spec runs_in(number(), fun(), list()) -> {true, any()} | false.
```
k-anderson: @fenollp for consistency how about?

-spec runs_in(number(), fun(), list()) -> {'ok', any()} | {'error', 'timeout'}.

fenollp: I'd be okay with `… -> {ok, any()} | timeout.`

k-anderson:
@fenollp  consistency is ok

merging!

PR #2831, opened 2016-11-24
ts_amqp_speedup: use AMQP worker to relay events

jamesaimonetti:


fenollp:
Cherry-pick of https://github.com/2600hz/kazoo/pull/2830
2nd attempt after https://github.com/2600hz/kazoo/pull/2832

PR #2833, opened 2016-11-24
ts_amqp_speedup: use AMQP worker to relay events

jamesaimonetti:


fenollp:
Since https://github.com/2600hz/kazoo/pull/2831 is against master I believe we can close this.

PR #2846, opened 2016-11-29
ts_amqp_cleanup-3.22: cleanup binding

jamesaimonetti:


fenollp:
Didn't know you could use Java's try-catch-after pattern in Erlang. Kinda neat. Thx!

PR #2850, opened 2016-11-30
Stop trunkstore process on bridge

jamesaimonetti:
There doesn't seem to be a need for the trunkstore process to stay
alive once the bridge has started (CHANNEL_BRIDGE received). If
received, the process will cleanup and terminate.


jamesaimonetti:
@lazedo @k-anderson thoughts? This will free workers up faster since we won't wait until CHANNEL_DESTROY once CHANNEL_BRIDGE is received.

PR #2851, opened 2016-11-30
acdc: Deliver oldest queue calls before newer

danielfinke:
Rewrite strategy_state tracking to allow for available/ringing/busy/unavailable per agent
Each agent sends updates of their state to queues they are a member of
Using current_member_calls record var, queue manager ensures that only top x (where x=number of available agents) calls are simultaneously distributed. Then, it ensures that older calls are always highest priority for distribution while still allowing multiple calls to be delivered


make/kz.mk
```diff
@@ -74,7 +74,7 @@ test/$(PROJECT).app: ERLC_OPTS += -DTEST
 test/$(PROJECT).app: $(TEST_SOURCES)
 	@mkdir -p test/
 	@mkdir -p ebin/
-	ERL_LIBS=$(ELIBS) erlc -v +nowarn_missing_spec $(ERLC_OPTS) $(TEST_PA) -o ebin/ $(TEST_SOURCES)
+	ERL_LIBS=$(ELIBS) erlc -v +nowarn_missing_spec +nowarn_export_all $(ERLC_OPTS) $(TEST_PA) -o ebin/ $(TEST_SOURCES)
```
fenollp: How is this needed?

applications/acdc/test/test.config
```diff
@@ -0,0 +1,10 @@
+[
+ {lager, [
+          {handlers, [
+                      {lager_console_backend, '=info'}
```
fenollp: I couldn't get this to work (see https://github.com/2600hz/kazoo/compare/test-config )
Do you think it is due to eunit owning the process group?
Is it possible to redirect to group `user` with `lager`?

danielfinke: @fenollp to be clear, what exactly did you run into? Something with the line in my above test.config? Or in your test-config branch did you happen to have a problem?

danielfinke:
@fenollp it looks like I am going to be iterating on this a few times to see where the disconnect is between your CI setup and my development testing. I may need some help from you if that is ok?

applications/acdc/test/test.config
fenollp: On that branch's Travis CI runs you can see that lager prints no logs... but that may be because lager is not started / tests don't get compiled with the lager transform. I'll investigate later :)

fenollp:
Sure, ask away!
Looking at Travis's logs I would try printing `code:all_paths()`.
Also directly calling `application:ensure_all_started(acdc).` might help.

Be aware though that Travis just runs eunit tests without couch nor rabbit & without starting a "kazoo node"

danielfinke:
@fenollp It's definitely the fact that the Travis setup doesn't establish a Kazoo node nor fully compiles the code base that is the reason my test regime doesn't succeed.
I'm thinking for sake of this PR I will rewrite the tests to remove dependency on other applications. Are there any plans on your end for establishing a test environment where what are essentially scripted integration tests can be performed (kinda like some of the stuff that CircleCI is doing - check-release-startup.sh) by defining them at the application level like I was trying for here?

make/kz.mk
fenollp: We prefer to export specific functions rather than using export_all as we believe its usage leads to bad practice :)

fenollp:
Yep we're working on making a test suite public: the make-busy project should soon openly report on PRs passing tests and such. Also everyone will be able to add tests to that project!

danielfinke:
@fenollp I've removed the tests that depend on a Kazoo node for now. We're going to start looking at make-busy and how we might tie similar tests in for future PRs

lazedo:
@danielfinke thanks!

PR #2852, opened 2016-12-01
KAZOO-5141: move ACDC crossbar modules into ACDC application

mark2600:


lazedo:
@mark2600 please rebase and fix it.

PR #2855, opened 2016-12-02
HELP-26081: set CNAME for inbound call in trunkstore

icehess:


applications/trunkstore/src/ts_from_offnet.erl
```diff
@@ -443,8 +444,8 @@ callee_id([JObj | T]) ->
             end
     end.
 
--spec maybe_anonymize_caller_id(ts_callflow:state(), ne_binary(), ne_binary()) -> kz_proplist().
-maybe_anonymize_caller_id(State, OldCallerId, CidFormat) ->
+-spec maybe_anonymize_caller_id(ts_callflow:state(), ne_binary(), ne_binary(), ne_binary()) -> kz_proplist().
```
fenollp: Shouldn't these be `api_ne_binary()`?

PR #2857, opened 2016-12-02
HELP-26318 preserve already set CID

lazedo:


k-anderson:
@icehess seems like this would impact callflows as well, no?  Trunkstore should keep the caller id while callflows should always use the configured caller id.  I think it will need a conditional...

lazedo:
@k-anderson this will only impact callflows. the problem is that CID value set in cf modules are being overwritten by this when it is called by cf_user/cf_device. unless i'm mistaken, i don't think trunkstore uses this.

lazedo:
lets not merge this yet

PR #2867, opened 2016-12-05
qf-mig: log more info

fenollp:


core/kazoo_number_manager/src/knm_search.erl
```diff
@@ -488,7 +488,7 @@ handle_number(JObj) ->
     case local_discovery(Number, []) of
         {'ok', KNumber} ->
             Payload = [{<<"Msg-ID">>, kz_api:msg_id(JObj)}
-                      ,{<<"Results">>, knm_phone_number:to_json(knm_number:phone_number(KNumber))}
+                      ,{<<"Results">>, knm_phone_number:to_public_json(knm_number:phone_number(KNumber))}
```
lazedo: @fenollp does public_data contain carrier and its possible data ? if not, this cannot be.

PR #2869, opened 2016-12-05
http_urls: require HTTP(S) URLs in schemas

jamesaimonetti:


jamesaimonetti:
```erlang
(pdx@pdx.2600hz.com)3> re:run("http://server.com", "^https?://", [{capture, none}]).
match
(pdx@pdx.2600hz.com)4> re:run("https://server.com", "^https?://", [{capture, none}]).
match
(pdx@pdx.2600hz.com)5> re:run("ftp://server.com", "^https?://", [{capture, none}]).
nomatch
```

applications/crossbar/priv/couchdb/schemas/faxbox.json
```diff
@@ -44,7 +44,7 @@
             "type": "string"
         },
         "media": {
-            "default": {},
+            "default": null,
```
fenollp: remove this field entirely maybe?

jamesaimonetti: Weird, I don't remember adding that...

icehess:
@jamesaimonetti If we're going to enforce HTTPS, isn't better to mention in description that the address must be HTTPS?

jamesaimonetti:
@lazedo doesn't require it, 's?' means optional s after 'http'.

icehess:
@jamesaimonetti Hmm you right, I didn't see the regex just said it based on pr title. But anyway it could be good to mention the URL required to be http(s) in description.

k-anderson:
@jamesaimonetti why wouldn't we also support ftp with the other pending pr?

jamesaimonetti:
@k-anderson this was made before I saw @lazedo's work to support FTP backend. If accepted, I can update this to use `format=uri` instead for call recording URLs. The others, I think, are where Kazoo acts as a client to fetch some data (like pivot) and restricting to http(s) makes sense.

jamesaimonetti:
@k-anderson rebased and updated

PR #2870, opened 2016-12-05
26203: add source_ip to registration query responses

mark2600:


mark2600:
NOTE, there is an associated pull request on kazoo-configs that must be accepted with this

lazedo:
@mark2600 don't you need top add `Source-IP`to `kapi_registration` ?
also, maybe we can reuse the existing  `<<"Network-IP">>, <<"Network-Port">>` ?

k-anderson:
@mark2600 IP and port please :)  are these the received parameter from kamailio?

mark2600:
Yes they are passed straight through from Kamailio

PR #2872, opened 2016-12-05
add action callflow schema

jamesaimonetti:
move action schema to its own schema doc

use callflows.action schema

export types for use

add test clause

update ref doc and swagger

### DIFF
--- a/applications/crossbar/priv/couchdb/schemas/callflows.json
+++ b/applications/crossbar/priv/couchdb/schemas/callflows.json
@@ -23,30 +23,8 @@
             "type": "object"
         },
         "flow": {
-            "description": "A callflow node defines a module to execute, data to provide to that module, and one or more children to branch to",
-            "properties": {
-                "children": {
-                    "default": {},
-                    "description": "Children callflows",
-                    "required": false,
-                    "type": "object"
-                },
-                "data": {
-                    "default": {},
-                    "description": "The data/arguments of the callflow module",
-                    "required": true,
-                    "type": "object"
-                },
-                "module": {
-                    "description": "The name of the callflow module to excute at this node",
-                    "maxLength": 64,
-                    "minLength": 1,
-                    "required": true,
-                    "type": "string"
-                }
-            },
-            "required": true,
-            "type": "object"
+            "$ref": "callflows.action",
+            "description": "A callflow node defines a module to execute, data to provide to that module, and zero or more children to branch to"
         },
         "metaflow": {
             "$ref": "metaflows",
--- a/core/kazoo_documents/src/kzd_callflow.erl
+++ b/core/kazoo_documents/src/kzd_callflow.erl
@@ -14,7 +14,9 @@
         ,patterns/1, set_patterns/2
         ,is_feature_code/1
         ,flow/1, flow/2
-        ,set_flow/2, validate_flow/1
+        ,set_flow/2
+        ,validate/1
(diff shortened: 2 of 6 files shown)


core/kazoo_documents/src/kzd_callflow.erl
```diff
@@ -160,15 +170,28 @@ validate_flow(Doc) ->
...
--spec validate_flow_elements(kz_json:object(), validate_acc(), ne_binaries()) -> validate_acc().
+-ifdef(TEST).
 validate_flow_elements(Flow, Acc, Path) ->
     Module = kz_json:get_binary_value(<<"module">>, Flow),
     Data = kz_json:get_json_value(<<"data">>, Flow),
     Children = kz_json:get_json_value(<<"children">>, Flow, kz_json:new()),
 
     validate_action_schema(Module, Data, Children, Acc, Path).
+-else.
+-spec validate_flow_elements(kz_json:object(), validate_acc(), ne_binaries()) -> validate_acc().
```
fenollp: Why not leave the spec out (before ifdef)?

core/kazoo_documents/src/kzd_callflow.erl
```diff
@@ -145,12 +147,20 @@ flow(Doc, Default) ->
 set_flow(Doc, Flow) ->
     kz_json:set_value(?FLOW, Flow, Doc).
 
+-spec validate(doc()) -> {'ok', doc()} |
+                         {'error', kz_json_schema:validation_errors()}.
+validate(Doc) ->
```
fenollp: I can't find the code using this function. Where is it?

applications/crossbar/priv/couchdb/schemas/callflows.json
```diff
@@ -23,30 +23,8 @@
             "type": "object"
         },
         "flow": {
-            "description": "A callflow node defines a module to execute, data to provide to that module, and one or more children to branch to",
-            "properties": {
-                "children": {
-                    "default": {},
```
fenollp: What happened to `{}` & why did it become `null`?

jamesaimonetti: Turns out the json pretty printer in Emacs incorrectly converts `{"foo":{}}` to `{"foo":null}` when pretty printing. There's a bug filed but it hasn't moved very far afaict. I've changed my pretty-printer to shell out to `python -mjson.tool` instead.

jamesaimonetti:
@lazedo updated to handle children too. I don't like the metaflows version because it requires you to enumerate all possible metaflow children in the "oneOf" array.

lazedo:
@jamesaimonetti true but, it ensures that the `data` object for the action at any level is valid. jessie already has the schemas in-memory. the question is, do we want to validate the `data` object for the corresponding callflow action ?

jamesaimonetti:
@lazedo Historically we haven't had callflow schemas for the data portions *and* others have custom callflow actions not part of core. I wouldn't want their ability to modify those callflows to fail because they don't have a schema for the `data`. The current behaviour is, if the schema isn't found, assume that action (assuming module and data exist) is valid.

jamesaimonetti:
@lazedo @k-anderson anything left to do here? I'm ignoring @fenollp because I find validate/1 useful when debugging.

jamesaimonetti:
Rebased and squashed just for funzies.

PR #2876, opened 2016-12-05
KAZOO-5225: handle number documents with "undefined" carrier name

k-anderson:


core/kazoo_number_manager/src/knm_phone_number.erl
```diff
@@ -733,6 +733,8 @@ set_ported_in(N, Ported) when is_boolean(Ported) ->
 module_name(#knm_phone_number{module_name = Name}) -> Name.
 
 -spec set_module_name(knm_phone_number(), ne_binary()) -> knm_phone_number().
+set_module_name(N, <<"undefined">>) ->
+    set_module_name(N, ?CARRIER_LOCAL);
```
fenollp: I would rather this be the first clause of the `/3`. This way only `from_json` is allowed to pass this value in.

PR #2877, opened 2016-12-06
HELP-26607: fall back to callee_number if capture_group is empty

icehess:
cf_dynamic_cid should fallback to callee_id_number if cf_capture_group(s) is empty
Also recollect digits in cf_disa, if collected digits ar empty (because of previous callflow
is crashed during collecting digits, in this case cf_dynamic_cid, and user pressed pound)

### DIFF
--- a/applications/callflow/src/module/cf_dynamic_cid.erl
+++ b/applications/callflow/src/module/cf_dynamic_cid.erl
@@ -82,8 +82,7 @@ handle(Data, Call) ->
 -spec handle_manual(kz_json:object(), kapps_call:call()) -> 'ok'.
 handle_manual(Data, Call) ->
     CID = collect_cid_number(Data, Call),
-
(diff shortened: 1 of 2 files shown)


applications/callflow/src/module/cf_dynamic_cid.erl
```diff
@@ -284,17 +298,17 @@ find_key_and_dest(ListJObj, Data, Call) ->
     case kz_json:get_value(<<"idx_name">>, Data) of
         'undefined' -> find_key_and_dest(ListJObj, Call);
         Idx ->
-            Groups = kapps_call:kvs_fetch('cf_capture_groups', Call),
+            Groups = get_callee_number('cf_capture_groups', Call),
```
jamesaimonetti: The spec for `get_callee_number` says it returns `ne_binary()` but Groups is being used as a JSON object. Can you update the spec to properly reflect this?

PR #2878, opened 2016-12-06
Fixed broken links for API and User Authentication in crossbar basics.md

cando2k6:


cando2k6:
Same changes as another pull request that was just a little earlier.

PR #2881, opened 2016-12-06
do not export inception on bridge

lazedo:


k-anderson:
Please don't accept this yet, I believe that we need more testing.  The inception is used in several places to determine origination/termination.  I don't believe this is backward compatible.

For example:
kz_attributes.erl line 73, 74, 355
kz_endpoint.erl by maybe_add_diversion and maybe_add_alert_info and line 1438
cf_exe, doodle_exe ... ect

jamesaimonetti:
@lazedo any reason to keep this PR around?

PR #2885, opened 2016-12-06
Follow $ref in docs

jamesaimonetti:
Auto-generate schema tables referenced by the main doc's schema. 

This is a first pass to see if this is a viable thing to do. If acceptable, I'll incorporate the changes into the doc/ versions.


fenollp:
I LOVE this!

Side note: we have to find a way to keep these schema refs in sync with the schema in doc files. Maybe a CI check? hehehehe

PR #2894, opened 2016-12-06
Bandwidth sites results have an xpath that is different from the rest

freepenguins:
It could be that I'm not calling `knm_bandwidth2:sites()` incorrectly, but when I call it with `account_id`, `api_username`, and `api_password` set in `system_config/number_manager.bandwidth2`, I get an error:

`** exception error: no match of right hand side value {error,<<"Number not found">>}`

After adding the lines for verifying the response for the sites request, the method returns successfully.


freepenguins:
Closing this request, as peer id request needs modifications as well.

PR #2912, opened 2016-12-07
HELP-26680: filter empty text elements

jamesaimonetti:
Use xmerl's {space, normalize} to reduce #xmlText{} with just
whitespace to " ". Added function to remove those #xmlText{} elements.

Additionally, there was an unintentional pattern match when matching a
single <Number> element. The attributes of the Dial and Number
elements were being bound with the same name. Differentiated those to
make sure the single Number clause is used.

### DIFF
--- a/core/kazoo_xml/src/kz_xml.erl
+++ b/core/kazoo_xml/src/kz_xml.erl
@@ -11,6 +11,7 @@
 -export([elements/1, elements/2
         ,texts_to_binary/1, texts_to_binary/2
         ,attributes_to_proplist/1
+        ,filter_empty_text/1
         ]).
 
 -include_lib("kazoo/include/kz_types.hrl").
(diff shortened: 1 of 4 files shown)


core/kazoo_xml/src/kz_xml.erl
```diff
@@ -41,3 +42,13 @@ texts_to_binary(Vs, Size) when is_list(Vs), is_integer(Size), Size > 0 ->
 -spec attributes_to_proplist(xml_attribs()) -> kz_proplist().
 attributes_to_proplist(L) ->
     [{K, V} || #xmlAttribute{name=K, value=V} <- L].
+
+-spec filter_empty_text(xml_els() | xml_texts()) -> xml_els() | xml_texts().
```
fenollp: This is exactly https://github.com/erlang/otp/blob/b5545594dd90bb6d1e049757050839a368730ecb/lib/xmerl/src/xmerl_lib.erl#L368-L376 How about using it instead?

jamesaimonetti: Not exactly; that one is recursive, this one is only the provided list of xml records

jamesaimonetti:
@lazedo this is ready

PR #2930, opened 2016-12-07
qf-reconcile: only update module name when asked to

fenollp:


core/kazoo_number_manager/src/knm_number.erl
```diff
@@ -414,27 +414,31 @@ reconcile(DID, Options) ->
...
+         }
+        ,{knm_number_options:public_fields(Options)
+         ,knm_phone_number:doc(PhoneNumber)
+         ,fun knm_phone_number:update_doc/2
+         }
+        ,{?NUMBER_STATE_IN_SERVICE
+         ,knm_phone_number:state(PhoneNumber)
+         ,fun knm_phone_number:set_state/2
+         }
+        ],
```
lazedo: @fenollp is the order correct ? `set_state` comes after `update_doc` ?

fenollp: Order doesn't matter. Btw doc here means public fields

fenollp:
@jamhed the makebusy failure is normal?
Can we have a "restart tests" button?
And a html file instead of txt (so that we can later style the txt by writing HTML/js)?
Plz

PR #2932, opened 2016-12-07
HELP-26661: remove default time_limit

jamesaimonetti:
when adding a "stop" action, time_limit can/should be unset from the
client to avoid validation errors. If we include a default, time_limit
will be set and, if the schema changes, the value could become invalid
through no fault of the user.


lazedo:
@jamesaimonetti needs `mase api` ?

PR #2933, opened 2016-12-07
Ensure that both notify and teletype retrieve the configuration value…

mark2600:
…s in the same way

### DIFF
--- a/core/kazoo_config/src/kapps_config.erl
+++ b/core/kazoo_config/src/kapps_config.erl
@@ -31,8 +31,6 @@
 -export([get_non_empty/2, get_non_empty/3, get_non_empty/4]).
 -export([get_ne_binary/2, get_ne_binary/3, get_ne_binary/4]).
 
--export([get_inherited/4, get_inherited/5]).
-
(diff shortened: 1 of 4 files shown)


core/kazoo_config/src/kapps_config.erl
```diff
@@ -31,6 +31,8 @@
 -export([get_non_empty/2, get_non_empty/3, get_non_empty/4]).
 -export([get_ne_binary/2, get_ne_binary/3, get_ne_binary/4]).
 
+-export([get_inherited/4, get_inherited/5]).
```
lazedo: @mark2600 i don't think this should go here. i think it will fit better in `kz_account` since we have the `fetch` method there and you are looking for values in the account document not in configuration

k-anderson: @mark2600 I agree that kapps_config is not a good place, but I disagree that kz_accounts is better ;)
kapps_util would be my vote.

mark2600:
When accepting please also merge #2942

PR #2937, opened 2016-12-08
ensure return codes in attachment handlers

lazedo:


core/kazoo_attachments/src/kz_att_ftp.erl
```diff
@@ -60,22 +60,26 @@ send_request(Host, Port, UserPass, FullPath, Contents) ->
...
                 Routines = [fun() -> ftp:user(Pid, User, Pass) end
                            ,fun() -> ftp:type(Pid, 'binary') end
                            ,fun() -> ftp:cd(Pid, Dir) end
                            ,fun() -> ftp:send_bin(Pid, Contents, File) end
                            ],
                 Res = ftp_cmds(Routines),
-                ftp:close(Pid),
-                Res;
-            {'error', _}=E -> E
+                catch(ftp:close(Pid)),
```
fenollp: Maybe log error on close?

mark2600: Probably not needed since we are closing the connection anyway.

PR #2941, opened 2016-12-08
Revert "keep the origrinal caller id in disa & add some log lines"

mark2600:
Reverted on request from lazedo


lazedo:
@mark2600 thanks

PR #2955, opened 2016-12-09
qf-refrsh: fix +refresh_numbers_db/1

fenollp:


core/kazoo_number_manager/src/kazoo_number_manager_maintenance.erl
```diff
@@ -72,8 +72,8 @@ refresh_numbers_db(<<?KNM_DB_PREFIX_ENCODED, _/binary>> = NumberDb) ->
 refresh_numbers_db(<<?KNM_DB_PREFIX, Suffix/binary>>) ->
     NumberDb = <<?KNM_DB_PREFIX_ENCODED, Suffix/binary>>,
     refresh_numbers_db(NumberDb);
-refresh_numbers_db(<<"+", Suffix/binary>>) ->
-    refresh_numbers_db(Suffix);
+refresh_numbers_db(<<"+", _/binary>> = Num) ->
```
k-anderson: @fenollp any resond for this change?  It diffidently works as is....

core/kazoo_number_manager/src/kazoo_number_manager_maintenance.erl
```diff
@@ -82,10 +82,11 @@ refresh_numbers_db(Suffix) ->
 -spec fix_accounts_numbers([ne_binary()]) -> 'ok'.
 -spec fix_account_numbers(ne_binary()) -> 'ok'.
 fix_accounts_numbers(Accounts) ->
+    _ = purge_discovery(),
```
k-anderson: @fenollp Why was this added?  It should not be executed that frequently....

fenollp: Just because it reduces the amount of things migration will run through, thus making it run faster.
I don't think clearing that cache is so bad: searches are ran much more often than migrations, no?

fenollp:
So previous changes make migration go through inum & managed dbs.
These indeed have number docs, but are not "normal" number docs: these are fake-carriers' number docs.
I think these should not be included in the migration.
Also, migration fails because these dbs don't have the necessary views (since they shouldn't anyway).

PR #2961, opened 2016-12-09
HELP-26813: sum multiple vmboxes if assigned

k-anderson:


core/kazoo_voicemail/src/kvm_messages.erl
```diff
@@ -86,21 +86,29 @@ count_by_owner(?MATCH_ACCOUNT_ENCODED(_)=AccountDb, OwnerId) ->
...
         {'ok', []} ->
-            lager:info("voicemail box owner is not found"),
+            lager:info("no voicemail boxes belonging to user ~s found", [OwnerId]),
             {0, 0};
-        {'ok', [Owned|_]} ->
-            VMBoxId = kz_json:get_value(<<"value">>, Owned),
-            FoldersCounts = count_per_folder(AccountId, VMBoxId),
-            normalize_count_none_deleted(VMBoxId, FoldersCounts);
+        {'ok', Boxes} ->
+            FolderQuantities = count_per_folder(AccountId),
```
icehess: How about use `count_per_folder/2` and loop over it to sum result instead?

k-anderson: @icehess this way its one db request per MODb, instead of looping the voicemails in each MODb.  imho its more efficient this way.

PR #2975, opened 2016-12-12
qf-view: probably dangerous to break map/red sync..

fenollp:
I'm thinking about closing this: it may lead to unsynced map/reduce code. But maybe we'll prefer to have the static part of the view with the rest of the view...


fenollp:
We're keeping the map and the reduce at te same place.

PR #2977, opened 2016-12-12
PISTON-351: validate did ownership before saving callflows/trunks

danielfinke:
Ensure that number docs can't be modified by cb_callflows or cb_connectivity if the authorizing account does have rights to do so.
We noticed that it was possible to update fields in a numbers document that shouldn't be allowed simply by permitting the update of in_service numbers by setting assign_to to an account that /would/ allow those doc updates. This is handled in the knm_number_states change.
cb_callflows and cb_connectivity have validation functions to prevent the callflow from saving if the additional functions (track_assignment and reconciliation) shouldn't be allowed.

### DIFF
--- a/applications/crossbar/src/modules/cb_callflows.erl
+++ b/applications/crossbar/src/modules/cb_callflows.erl
@@ -381,39 +381,61 @@ on_successful_validation(CallflowId, Context) ->
 %%
 %% @end
 %%--------------------------------------------------------------------
+-type assigned_tuple() :: {ne_binary(), ne_binaries()}.
+-type assigned_tuples() :: [assigned_tuple()].
+
 -spec validate_number_ownership(cb_context:context()) -> cb_context:context().
--spec validate_number_ownership(ne_binaries(), cb_context:context()) ->
-                                       cb_context:context().
--spec validate_number_ownership(knm_number:knm_number()
-                               ,ne_binaries()
-                               ,cb_context:context()) -> cb_context:context().
 validate_number_ownership(Context) ->
     NewNums = kz_json:get_value(<<"numbers">>, cb_context:doc(Context), []),
 
-    Assigned =  [Num
+    Assigned =  [knm_converters:normalize(Num)
                  || Num <- NewNums,
-                    Num =/= <<"undefined">>
+                    Num =/= <<"undefined">>,
+                    knm_converters:is_reconcilable(Num)
                 ],
 
-    validate_number_ownership(Assigned, Context).
+    AssignedTo = all_assigned_to_fold(Assigned, []),
+    validate_parent(AssignedTo, Context).
 
-validate_number_ownership([], Context) -> Context;
-validate_number_ownership([Number|Numbers], Context) ->
-    case knm_number:get(Number) of
-        {'ok', NumberObj} ->
-            validate_number_ownership(NumberObj, Numbers, Context);
-        _ -> validate_number_ownership(Numbers, Context)
+-spec all_assigned_to_fold(ne_binaries(), assigned_tuples()) ->
+                                  assigned_tuples().
+all_assigned_to_fold([], Acc) -> Acc;
+all_assigned_to_fold([Number|_]=Numbers, Acc) ->
+    Db = knm_converters:to_db(Number),
+    {SamePrefix, DiffPrefix} = lists:partition(fun(Number1) ->
+                                                       knm_converters:to_db(Number1) =:= Db
+                                               end
+                                              ,Numbers),
+    case kz_datamgr:get_results(Db, <<"numbers/assigned_to">>) of
+        {'ok', Results} ->
+            Acc1 = assigned_to_fold(SamePrefix, Results, Acc),
+            all_assigned_to_fold(DiffPrefix, Acc1);
+        _ -> Acc
     end.
 
-validate_number_ownership(NumberObj, Numbers, Context) ->
+-spec assigned_to_fold(ne_binaries(), kz_json:objects(), assigned_tuples()) ->
+                              assigned_tuples().
+assigned_to_fold([], _, Acc) -> Acc;
+assigned_to_fold([Number|Numbers], ViewJObjs, Acc) ->
+    case kz_json:find_value(<<"id">>, Number, ViewJObjs) of
+        'undefined' -> assigned_to_fold(Numbers, ViewJObjs, Acc);
+        JObj ->
+            AssignedTo = hd(kz_json:get_value(<<"key">>, JObj)),
+            NumbersForAssignedTo = props:get_value(AssignedTo, Acc, []),
+            NumbersForAssignedTo1 = [Number | NumbersForAssignedTo],
+            Acc1 = props:set_value(AssignedTo, NumbersForAssignedTo1, Acc),
+            assigned_to_fold(Numbers, ViewJObjs, Acc1)
+    end.
+
+-spec validate_parent(assigned_tuples(), cb_context:context()) ->
+                             cb_context:context().
+validate_parent([], Context) -> Context;
+validate_parent([{AssignedTo, Numbers}|AssignedTuples], Context) ->
     AuthBy = cb_context:auth_account_id(Context),
-    PhoneNumber = knm_number:phone_number(NumberObj),
-    AssignedTo = knm_phone_number:assigned_to(PhoneNumber),
     case kz_util:is_in_account_hierarchy(AuthBy, AssignedTo, 'true') of
-        'true' -> validate_number_ownership(Numbers, Context);
+        'true' -> validate_parent(AssignedTuples, Context);
         'false' ->
-            Number = knm_phone_number:number(PhoneNumber),
-            Message = <<"unauthorized to add/modify ", Number/binary>>,
+            Message = <<"unauthorized to add/modify ", (hd(Numbers))/binary>>,
             cb_context:add_system_error(403, 'forbidden', Message, Context)
     end.
 
--- a/applications/crossbar/src/modules/cb_callflows.erl
+++ b/applications/crossbar/src/modules/cb_callflows.erl
@@ -361,12 +361,11 @@ add_pattern_conflict(Context, JObj) ->
 -spec validate_callflow_schema(api_binary(), cb_context:context()) -> cb_context:context().
 validate_callflow_schema(CallflowId, Context) ->
     OnSuccess = fun(C) ->
-                        C1 = validate_callflow_elements(on_successful_validation(CallflowId, C)),
-                        C2 = validate_uniqueness(CallflowId, C1),
+                        C1 = validate_uniqueness(CallflowId, on_successful_validation(CallflowId, C)),
 
-                        Doc = cb_context:doc(C2),
+                        Doc = cb_context:doc(C1),
                         Nums = kz_json:get_list_value(<<"numbers">>, Doc, []),
-                        cb_modules_util:validate_number_ownership(Nums, C2)
+                        cb_modules_util:validate_number_ownership(Nums, C1)
                 end,
     cb_context:validate_request_data(<<"callflows">>, Context, OnSuccess).
 
--- a/applications/crossbar/src/modules/cb_callflows.erl
+++ b/applications/crossbar/src/modules/cb_callflows.erl
@@ -433,7 +433,7 @@ track_assignment('post', Context) ->
                     knm_converters:is_reconcilable(Num, AccountId)
                 ],
 
-    Updates = cb_modules_util:apply_assignment_updates(Unassigned ++ Assigned),
+    Updates = cb_modules_util:apply_assignment_updates(Unassigned ++ Assigned, Context),
     cb_modules_util:log_assignment_updates(Updates);
 track_assignment('put', Context) ->
     NewNums = kz_json:get_value(<<"numbers">>, cb_context:doc(Context), []),
@@ -443,7 +443,7 @@ track_assignment('put', Context) ->
                     knm_converters:is_reconcilable(Num, AccountId)
                 ],
 
-    Updates = cb_modules_util:apply_assignment_updates(Assigned),
+    Updates = cb_modules_util:apply_assignment_updates(Assigned, Context),
     cb_modules_util:log_assignment_updates(Updates);
 track_assignment('delete', Context) ->
     Nums = kz_json:get_value(<<"numbers">>, cb_context:doc(Context), []),
@@ -452,7 +452,7 @@ track_assignment('delete', Context) ->
                    || Num <- Nums,
                       knm_converters:is_reconcilable(Num, AccountId)
                   ],
-    Updates = cb_modules_util:apply_assignment_updates(Unassigned),
+    Updates = cb_modules_util:apply_assignment_updates(Unassigned, Context),
     cb_modules_util:log_assignment_updates(Updates).
 
 -spec filter_callflow_list(api_binary(), kz_json:objects()) -> kz_json:objects().
--- a/applications/crossbar/src/modules/cb_callflows.erl
+++ b/applications/crossbar/src/modules/cb_callflows.erl
@@ -361,7 +361,9 @@ add_pattern_conflict(Context, JObj) ->
 -spec validate_callflow_schema(api_binary(), cb_context:context()) -> cb_context:context().
 validate_callflow_schema(CallflowId, Context) ->
     OnSuccess = fun(C) ->
-                        validate_uniqueness(CallflowId, on_successful_validation(CallflowId, C))
+                        C1 = validate_callflow_elements(on_successful_validation(CallflowId, C)),
+                        C2 = validate_uniqueness(CallflowId, C1),
+                        validate_number_ownership(C2)
                 end,
     cb_context:validate_request_data(<<"callflows">>, Context, OnSuccess).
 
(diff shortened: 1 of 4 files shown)


jamesaimonetti:
@danielfinke thanks! I'll review this tomorrow and get you some feedback (or merge!)

k-anderson:
@jamesaimonetti let me know what you find before merging please :)

jamesaimonetti:
@danielfinke sorry, got waylaid. Tomorrow!

danielfinke:
@jamesaimonetti have you some time to look at this PR? (Also why is MakeBusy still going?)

applications/crossbar/src/modules/cb_callflows.erl
```diff
@@ -387,6 +385,48 @@ validate_callflow_elements(Context) ->
...
+        {'ok', NumberObj} ->
+            validate_number_ownership(NumberObj, Numbers, Context);
+        _ -> validate_number_ownership(Numbers, Context)
+    end.
+
+validate_number_ownership(NumberObj, Numbers, Context) ->
+    AuthBy = cb_context:auth_account_id(Context),
+    PhoneNumber = knm_number:phone_number(NumberObj),
+    AssignedTo = knm_phone_number:assigned_to(PhoneNumber),
+    case kz_util:is_in_account_hierarchy(AuthBy, AssignedTo, 'true') of
```
fenollp: Hey since this seems to only need the assigned_to field of a number, can you use the view `numbers/assigned_to` instead of multiple knm_number:get calls? That way you can also group numbers by AssignedTo and do this AuthBy check once per group of numbers.

applications/crossbar/src/modules/cb_connectivity.erl
```diff
@@ -249,6 +253,48 @@ on_successful_validation(Id, Context) ->
...
+                                       cb_context:context().
+-spec validate_number_ownership(knm_number:knm_number()
+                               ,ne_binaries()
+                               ,cb_context:context()) -> cb_context:context().
+validate_number_ownership(Context) ->
+    NewNums = get_numbers(cb_context:doc(Context)),
+
+    Assigned =  [Num
+                 || Num <- NewNums,
+                    Num =/= <<"undefined">>
```
fenollp: What does your request look like?

danielfinke:
@fenollp made changes based on your comments. I'm thinking the core functions in this PR should be moved to cb_modules_util, agreed? Any other thoughts about this PR?

lazedo:
@danielfinke i don't understand what you're trying to do here. the initial description of the PR doesn't seem to be what was coded. this needs careful review

core/kazoo_number_manager/src/knm_number_states.erl
```diff
@@ -234,11 +234,17 @@ in_service_from_reserved_authorize(Number) ->
 in_service_from_in_service_authorize(Number) ->
     PhoneNumber = knm_number:phone_number(Number),
     AssignTo = knm_phone_number:assign_to(PhoneNumber),
+    AssignedTo = knm_phone_number:assigned_to(PhoneNumber),
     AuthBy = knm_phone_number:auth_by(PhoneNumber),
     Sudo = ?KNM_DEFAULT_AUTH_BY =:= AuthBy,
     case Sudo
-        orelse ?ACCT_HIERARCHY(AssignTo, AuthBy, 'true')
```
fenollp: Could you try and see if just using knm_phone_number:is_authorized/1 solves your problem here?
There's a bunch of different ways auth is verified in this app and I'm trying to reduce them all in just one place.

fenollp:
@danielfinke Yes I think adding these functions to cb_modules_util is a great idea.
However their implementation duplicates logic in KNM and misses its optimisations. Could you use knm_numbers:get instead?
Note: knm_numbers has a different API in 4.0 (doesn't return the same things), but then maybe you don't need this on 4.0?

danielfinke:
@fenollp I've addressed the things in your last set of comments. You were right, knm_phone_number:is_authorized/1 works. So I've applied that to knm_number_states.
I've moved back to using knm_number:get for the authorization at the crossbar level, through a shared set of code in cb_modules_util.
The last necessary part was to validate that "track_assignment" validates the account id that is being operated on before changing the used_by on a number. This crops up when a number was assigned to a callflow in one account and moved to another account without being unassigned first. If the first account later removed the number from the callflow, it was still able to clear the used_by on the number. The change to apply_assignment_updates fixes that.

applications/crossbar/src/modules/cb_modules_util.erl
```diff
@@ -563,29 +564,91 @@ remove_plaintext_password(Context) ->
...
+%% operated on is the one that the number is assigned to
+%%
+%% @end
+%%--------------------------------------------------------------------
+-spec maybe_assign_to_app(ne_binary(), api_binary(), ne_binary(), api_binary()) ->
+                                 assignment_update().
+maybe_assign_to_app(DID, Assign, AccountId, AccountId) ->
+    {DID, knm_number:assign_to_app(DID, Assign)};
+maybe_assign_to_app(DID, _, _, _) ->
+    {DID, {'error', 'unauthorized'}}.
```
fenollp: Not sure why you're doing some verification at all: the knm_number (or knm_numbers) module will do it anyways.

Is this related to the "moving numbers between account keeps app assignment" issue you mentioned?
This BTW sounds like a bug that can be properly resolved in knm_numbers:move/3 by adding a call to `knm_phone_number:set_used_by(_, undefined)` in there.

core/kazoo_number_manager/src/knm_number_states.erl
```diff
@@ -223,21 +223,19 @@ in_service_from_reserved_authorize(Number) ->
 
 -spec in_service_from_in_service_authorize(kn()) -> kn();
                                           (t()) -> t().
-in_service_from_in_service_authorize(T=#{todo := Ns, options := Options}) ->
```
fenollp: This is very good, but better yet:

```erlang
-spec in_service_from_in_service_authorize(t()) -> t().
in_service_from_in_service_authorize(T) ->
    knm_numbers:do_in_wrap(T, [fun knm_phone_number:is_authorized/1]).
```
Thanks for your time and great work!

applications/crossbar/src/modules/cb_modules_util.erl
```diff
@@ -563,29 +564,91 @@ remove_plaintext_password(Context) ->
...
+%% @end
+%%--------------------------------------------------------------------
+-spec validate_number_ownership(ne_binaries(), cb_context:context()) ->
+                                       cb_context:context().
+-spec validate_number_ownership(ne_binaries(), ne_binaries(), cb_context:context()) ->
+                                       cb_context:context().
+validate_number_ownership(Numbers, Context) ->
+    Filtered = [knm_converters:normalize(Num)
+                || Num <- Numbers,
+                   knm_converters:is_reconcilable(Num)
```
danielfinke: You're right - removed in latest push

applications/crossbar/src/modules/cb_modules_util.erl
```diff
@@ -563,29 +564,91 @@ remove_plaintext_password(Context) ->
...
+
+validate_number_ownership([], [], Context) -> Context;
+validate_number_ownership([], Unauthorized, Context) ->
+    Prefix = <<"unauthorized to use ">>,
+    Numbers = kz_binary:join(Unauthorized, <<", ">>),
+    Message = <<Prefix/binary, Numbers/binary>>,
+    cb_context:add_system_error(403, 'forbidden', Message, Context);
+validate_number_ownership([Number|Numbers], Unauthorized, Context) ->
+    Options = [{'auth_by', cb_context:auth_account_id(Context)}],
+    case knm_number:get(Number, Options) of
```
danielfinke: Alright, I've moved to using knm_numbers:get/2

applications/crossbar/src/modules/cb_modules_util.erl
```diff
@@ -563,29 +564,91 @@ remove_plaintext_password(Context) ->
...
--spec maybe_assign_to_port_number(ne_binary(), api_binary()) ->
+-spec maybe_assign_to_port_number(ne_binary(), api_binary(), ne_binary()) ->
                                          assignment_update().
-maybe_assign_to_port_number(DID, Assign) ->
+maybe_assign_to_port_number(DID, Assign, AccountId) ->
     Num = knm_converters:normalize(DID),
     case knm_port_request:get(Num) of
         {'error', _} ->
-            {DID, knm_number:assign_to_app(DID, Assign)};
+            maybe_assign_to_app(DID, Assign, AccountId);
```
danielfinke: Maybe you can validate that I approached this in the way you had in mind

applications/crossbar/src/modules/cb_modules_util.erl
```diff
@@ -563,29 +564,91 @@ remove_plaintext_password(Context) ->
...
+-spec maybe_assign_to_app(ne_binary(), api_binary(), ne_binary()) ->
+                                 assignment_update().
+maybe_assign_to_app(DID, Assign, AccountId) ->
+    case knm_number:get(DID) of
+        {'ok', Number} ->
+            PhoneNumber = knm_number:phone_number(Number),
+            AssignedTo = knm_phone_number:assigned_to(PhoneNumber),
+            maybe_assign_to_app(DID, Assign, AccountId, AssignedTo);
+        {'error', _}=E -> {DID, E};
+        {'dry_run', _, _} -> {DID, {'error', 'dry_run'}}
```
danielfinke: Removed

applications/crossbar/src/modules/cb_modules_util.erl
danielfinke: The problem here was that the existing code wasn't specifying the auth_by for the assignment, so it was just using system level and always succeeding. I've added that in and this now works as expected.

The circumstance was that if you left a number on a callflow, but reassigned the number to a different account, when it was removed from the callflow in the old account the assignment to 'undefined' was still successful. Authorization/validation should have prevented that. With auth_by set to cb_context:account_id(Context) now, it's all good.

danielfinke:
@fenollp Just pushed changes addressing all but your last comment. That is still coming once I get some more time to work on it.

I may take on the change to knm_port_request:get for bulk reads as part of a separate PR/task.

applications/crossbar/src/modules/cb_modules_util.erl
fenollp: Is this missing some commits?

danielfinke:
@fenollp yep, my cherry-pick missed the commit for your 2nd and 3rd comments. It has been pushed

applications/crossbar/src/modules/cb_modules_util.erl
```diff
@@ -563,29 +564,132 @@ remove_plaintext_password(Context) ->
...
+%% @end
+%%--------------------------------------------------------------------
+-spec validate_number_ownership(ne_binaries(), cb_context:context()) ->
+                                       cb_context:context().
+-spec validate_number_ownership([{knm_numbers:num(), knm_numbers:ko()}], ne_binaries(), cb_context:context()) ->
+                                       cb_context:context().
+validate_number_ownership(Numbers, Context) ->
+    Options = [{'auth_by', cb_context:auth_account_id(Context)}],
+    #{ko := KOs} = knm_numbers:get(Numbers, Options),
+    validate_number_ownership(maps:to_list(KOs), [], Context).
```
fenollp: Can you do a maps:fold instead of converting to list then doing a list fold?

applications/crossbar/src/modules/cb_modules_util.erl
```diff
@@ -563,29 +564,132 @@ remove_plaintext_password(Context) ->
...
+            {PRUpdates, [ToApply|NumUpdates]}
     end.
 
+-spec assign_to_port_number(port_req_assignments()) ->
+                                   assignment_updates().
+assign_to_port_number(PRUpdates) ->
+    lists:foldl(fun({Num, Assign, JObj}, Acc) ->
+                        [{Num, knm_port_request:assign_to_app(Num, Assign, JObj)}
+                         | Acc]
+                end, [], PRUpdates).
```
fenollp: That's lists:map/2. A list comprehension would be even better. Example:

```erlang
assign_to_port_number(PRUpdates) ->
    [{Num, knm_port_request:assign_to_app(Num, Assign, JObj)}
     || {Num, Assign, JObj} <- PRUpdates
    ].
```

applications/crossbar/src/modules/cb_modules_util.erl
```diff
@@ -563,29 +564,132 @@ remove_plaintext_password(Context) ->
...
+-type assign_to_groups() :: #{api_binary() => ne_binaries()}.
+
+-spec group_by_assign_to(assignments_to_apply()) -> assign_to_groups().
+group_by_assign_to(NumUpdates) ->
+    group_by_assign_to(NumUpdates, #{}).
+
+group_by_assign_to([], Groups) -> Groups;
+group_by_assign_to([{DID, Assign}|NumUpdates], Groups) ->
+    DIDs = maps:get(Assign, Groups, []),
+    Groups1 = Groups#{Assign => [DID|DIDs]},
```
fenollp: I just discovered maps:update_with/4 but I'm okay with this code.

applications/crossbar/src/modules/cb_modules_util.erl
```diff
@@ -563,29 +564,132 @@ remove_plaintext_password(Context) ->
...
+format_assignment_results(#{ok := OKs
+                           ,ko := KOs}) ->
+    format_assignment_oks(OKs) ++ format_assignment_kos(maps:to_list(KOs)).
+
+-spec format_assignment_oks(knm_number:knm_numbers()) -> assignment_updates().
+format_assignment_oks(Numbers) ->
+    lists:foldl(fun(Number, Acc) ->
+                        PN = knm_number:phone_number(Number),
+                        [{knm_phone_number:number(PN), {'ok', Number}} | Acc]
+                end, [], Numbers).
```
fenollp: Again, that's a list comprehension

core/kazoo_number_manager/src/knm_number_states.erl
fenollp: Please use this ^ code.

applications/crossbar/src/modules/cb_modules_util.erl
```diff
@@ -563,29 +564,132 @@ remove_plaintext_password(Context) ->
...
+group_by_assign_to([], Groups) -> Groups;
+group_by_assign_to([{DID, Assign}|NumUpdates], Groups) ->
+    DIDs = maps:get(Assign, Groups, []),
+    Groups1 = Groups#{Assign => [DID|DIDs]},
+    group_by_assign_to(NumUpdates, Groups1).
+
+-spec format_assignment_results(knm_numbers:ret()) -> assignment_updates().
+format_assignment_results(#{ok := OKs
+                           ,ko := KOs}) ->
+    format_assignment_oks(OKs) ++ format_assignment_kos(maps:to_list(KOs)).
```
fenollp: Please use maps:fold here too

applications/crossbar/src/modules/cb_modules_util.erl
danielfinke: @fenollp I have question about a list comprehension in this case.
Notice how I am assigning `PN = knm_number:phone_number(Number)`.
I can't seem to find that there is a list comprehension syntax to replicate that. I would be looking for something like:
`[{knm_phone_number(PN), {'ok', Number}} || Number <- Numbers, PN = knm_number:phone_number(Number)].`
Does such a thing not exist? It's not hugely necessary here, because I could do
`[{knm_phone_number:number(knm_number:phone_number(Number)), {'ok', Number}} || Number <- Numbers].`
But I could see that getting really clunky if I wanted to use the PN many times in the left side of my list comprehension.

applications/crossbar/src/modules/cb_modules_util.erl
danielfinke: @fenollp I went without maps:update_with/4 because it isn't available in R18

applications/crossbar/src/modules/cb_modules_util.erl
fenollp: You can do this in these cases:

```erlang
[a(Y) || X <- Xs, Y <- [b(X)]]
=:=  %% Semantically and compiled to the same code
[a(b(X)) || X <- Xs]
```

danielfinke:
@fenollp alright. I've added maps:folds and list comprehensions where suggested

applications/crossbar/src/modules/cb_modules_util.erl
```diff
@@ -572,27 +572,27 @@ remove_plaintext_password(Context) ->
...
+
+-spec validate_number_ownership_fold(knm_numbers:num(), knm_numbers:ko(), ne_binaries()) ->
+                                            ne_binaries().
+validate_number_ownership_fold(_, Reason, Unauthorized) when is_atom(Reason) ->
     %% Ignoring atom reasons, i.e. 'not_found' or 'not_reconcilable'
-    validate_number_ownership(KOs, Unauthorized, Context);
-validate_number_ownership([{Number, ReasonJObj}|KOs], Unauthorized, Context) ->
+    Unauthorized;
+validate_number_ownership_fold(Number, ReasonJObj, Unauthorized) ->
     case knm_errors:error(ReasonJObj) of
```
fenollp: Oh do note though that if a number is not reconcilable it fails with the atom `not_reconcilable` instead of a knm_errors JSON object. So you should rewrite the case to something like `case not is_atom(Reason) andalso knm_errors:error(Error) =:= <<"forbidden">> of true -> ...`

applications/crossbar/src/modules/cb_modules_util.erl
```diff
@@ -666,29 +665,26 @@ group_by_assign_to([{DID, Assign}|NumUpdates], Groups) ->
...
-format_assignment_kos([{Num, Reason}|KOs], Acc) when is_atom(Reason) ->
-    Acc1 = [{Num, {'error', Reason}} | Acc],
-    format_assignment_kos(KOs, Acc1);
-format_assignment_kos([{Num, ReasonJObj}|KOs], Acc) ->
+-spec format_assignment_kos_fold(knm_numbers:num(), knm_numbers:ko(), assignment_updates()) ->
+                                        assignment_updates().
+format_assignment_kos_fold(Number, Reason, Updates) when is_atom(Reason) ->
+    [{Number, {'error', Reason}} | Updates];
+format_assignment_kos_fold(Number, ReasonJObj, Updates) ->
     Error = knm_errors:error(ReasonJObj),
```
fenollp: I think you don't need to extract the `error` here, as your `ReasonJObj` was already the thing returned in `{error, ReasonJObj}` before the knm_numbers rewrite. Just returning the `error` actually changes the way the HTTP API works.

applications/crossbar/src/modules/cb_modules_util.erl
danielfinke: @fenollp this is captured by the previous validate_number_ownership/3 clause

danielfinke:
@fenollp Addressed the second comment in this commit. See my reply for the first comment

applications/crossbar/src/modules/cb_modules_util.erl
fenollp: Right! Not sure how I missed that!

fenollp:
Don't mind me, trying to trigger makebusy tests.

fenollp:
Thanks for your work and patience!

danielfinke:
@fenollp and thanks for your help!

PR #2980, opened 2016-12-13
fix friendly names of fax inbound error templates

lazedo:


applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -1068,12 +1068,14 @@ merge_available(AccountAvailable, Available) ->
 -spec merge_fold(kz_json:object(), kz_json:objects()) -> kz_json:objects().
 merge_fold(Overridden, Acc) ->
     Id = kz_doc:id(Overridden),
+    [Master] = [JObj || JObj <- Acc, kz_doc:id(JObj) =:= Id],
+    Filtered = [JObj || JObj <- Acc, kz_doc:id(JObj) =/= Id],
```
fenollp: You can use lists:partition/2 to go through it only once.

PR #2982, opened 2016-12-13
number services reconcile

lazedo:


core/kazoo_services/src/services/kz_service_phone_numbers.erl
```diff
@@ -57,6 +58,18 @@ reconcile(Services, PNs) ->
     S2 = kz_services:reset_category(?NUMBER_SERVICES, S1),
     update_numbers(S2, PNs).
 
+-spec reset(kz_services:services()) -> kz_services:services().
+reset(Services) ->
+    Categories = [?PHONE_NUMBERS
+                 ,?NUMBER_SERVICES
```
fenollp: What about the other two categories?
I'm thinking we can put the map as a macro and use it to get the list of categories to reset.
Also, the function just above shares the same body as this one

lazedo: @fenollp these were taken from reconcile/2 but feel free to update the PR.
note that, currently this is crashing when an account is created as reconcile gets called but there are no numbers to reconcile.

k-anderson:
@lazedo should this also go to 4.0?

lazedo:
@k-anderson no. its a sequence of https://github.com/2600hz/kazoo/commit/456141a978a623cac91b3c62e9e70585dfdc00d2

PR #2983, opened 2016-12-13
use notify to handle notifications if the account has "notifications"…

mark2600:
… configured


mark2600:
PR #2984 is the corresponding 4.0 PR

PR #2991, opened 2016-12-14
Docs target

jamesaimonetti:
Checks the mkdocs.yml for errors and fails the build if found

### DIFF
--- a/scripts/validate_mkdocs.py
+++ b/scripts/validate_mkdocs.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python2
 from __future__ import print_function
 
 import yaml, os.path, sys
--- a/scripts/validate_mkdocs.py
+++ b/scripts/validate_mkdocs.py
@@ -1,0 +1,46 @@
+#!/usr/bin/python
+from __future__ import print_function
+
+import yaml, os.path, sys
+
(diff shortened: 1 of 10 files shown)


scripts/validate_mkdocs.py
```diff
@@ -0,0 +1,46 @@
...
+        return parse_page_string(errors_detected, page)
+    else:
+        return errors_detected
+
+stream = open("doc/mkdocs/mkdocs.yml", 'r')
+mkdocs = yaml.load_all(stream)
+errors_detected = False
+
+for doc in mkdocs:
+    for k,v in doc.items():
```
fenollp: This looks like python3. Kazoo asks for python2 in the docs: should fix one of the two

jamesaimonetti: weird, I am using python 2.7.12+

jamesaimonetti:
@fenollp I think the circle.yml was misconfigured. It seems to have run properly now.

fenollp:
```
make docs
./scripts/validate_mkdocs.py
Traceback (most recent call last):
  File "./scripts/validate_mkdocs.py", line 4, in <module>
    import yaml, os.path, sys
ImportError: No module named yaml
make: *** [docs] Error 1
```

PR #2993, opened 2016-12-14
add optional view_json parameter to specify custom views

mark2600:


mark2600:
#2994 is the associated 4.0 pull request

mark2600:
Correction, #3013 is the associated 4.0 PR

mark2600:
NOTE, this is a TEMP fix and should not be used as it will be fgone shortly.

PR #2995, opened 2016-12-14
KAZOO-5242: delete unwinded knm_local numbers on deletion

fenollp:


k-anderson:
@lazedo did you file a ticket to move this to the carrier modules?

PR #2997, opened 2016-12-14
use kazoo bindings for handlers

lazedo:


applications/ecallmgr/src/ecallmgr_app.erl
```diff
@@ -72,3 +76,17 @@ declare_exchanges() ->
...
+    'ok'.
+
+-spec freeswitch_nodesup_unbind() -> 'ok'.
+freeswitch_nodesup_unbind() ->
+    _ = kazoo_bindings:unbind(<<"freeswitch.node.modules">>, ?MODULE, 'freeswitch_node_modules'),
+    'ok'.
+
+-spec freeswitch_node_modules() -> ne_binaries().
+freeswitch_node_modules() ->
+    application:get_env(?APP, 'node_modules', ?NODE_MODULES).
```
fenollp: Do we really need to fetch these when the application is not started?
Cause these node modules are now specified in 2 different places without even a comment saying "hey don't forget to keep this in sync with that other list"

applications/ecallmgr/src/ecallmgr_fs_node_sup.erl
```diff
@@ -137,7 +122,8 @@ init([Node, Options]) ->
 
     NodeB = kz_util:to_binary(Node),
     Args = [Node, Options],
-    Modules = ecallmgr_config:get(<<"modules">>, ?CHILDREN),
+    M = kazoo_bindings:map(<<"freeswitch.node.modules">>, []),
+    Modules = lists:foldl(fun(A, B) -> A ++ B end, [], M),
```
fenollp: How about `lists:flatten` http://erldocs.com/current/stdlib/lists.html#flatten%2F1 ?

applications/ecallmgr/src/ecallmgr_app.erl
lazedo: @fenollp not getting what you're saying. `?NODE_MODULES` is defined once in ecallmgr.hrl and referred once in here.

applications/ecallmgr/src/ecallmgr_fs_node_sup.erl
lazedo: @fenollp sorry to say your suggestion doesn't work.

fenollp: haha okay

PR #3002, opened 2016-12-15
HELP-26356: (KAZOO-5237) update an app's metada using SUP

fenollp:


fenollp:
#2956

lazedo:
@fenollp 4.0 is not for new features :)

PR #3003, opened 2016-12-15
HELP-26792: use reseller providers as opt-in from master

fenollp:


fenollp:
Fixes https://github.com/2600hz/kazoo/pull/2979 ( https://github.com/2600hz/kazoo/pull/2978 )

k-anderson:
@fenollp I cherry-picked the 4.0 version worried that this was out-of-date.

PR #3004, opened 2016-12-15
HELP-26792_4oh: use reseller providers as opt-in from master

fenollp:
https://github.com/2600hz/kazoo/pull/3003


k-anderson:
still in progress....

k-anderson:
@fenollp ok I think this is finally ready, just need the tests updated ;)

applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
```diff
@@ -512,17 +514,16 @@ view_account_phone_numbers(Context) ->
                                     ]),
     cb_context:set_resp_data(Context1, NewRespData).
 
-maybe_fix_available(NumJObj) ->
+fix_available(NumJObj) ->
```
k-anderson: @fenollp spec too (if I don't get to it first)

PR #3005, opened 2016-12-15
HELP-26693: better try-catch on is_number_billable call. Fixes CI on 4oh

fenollp:


core/kazoo_number_manager/test/knm_phone_number_test.erl
```diff
@@ -168,7 +168,7 @@ is_dirty2_test_() ->
     ,?_assertEqual(<<"in_service">>, kz_json:get_value(<<"pvt_state">>, NewJObj))
 
     ,?_assertEqual(undefined, kz_json:get_value(<<"pvt_is_billable">>, OldJObj))
-    ,?_assertEqual(false, kz_json:get_value(<<"pvt_is_billable">>, JObj))
+    ,?_assertEqual(true, kz_json:get_value(<<"pvt_is_billable">>, JObj))
     ,?_assertEqual(false, kz_json:get_value(<<"pvt_is_billable">>, NewJObj))
```
fenollp: this needs editing too BTW.
Also, the generated reduce view defaults to false on unknown carriers & should be edited accordingly.

PR #3008, opened 2016-12-16
Documenting storage

jamesaimonetti:


fenollp:
Refs and docs are not in sync

jamesaimonetti:
Storage ref and doc are. The others will be updated as everyone works on them

PR #3010, opened 2016-12-16
qf kz_att_http & kz_att_ftp

skorobkov:
Fix `kz_att_http:fetch_attachment/1`
Fix `default_format/0` in kz_att_http & kz_att_ftp
Fix parsing `field_list` from storage_plan in kz_att_http & kz_att_ftp

### DIFF
--- a/core/kazoo_attachments/src/kz_att_http.erl
+++ b/core/kazoo_attachments/src/kz_att_http.erl
@@ -104,7 +104,7 @@ fetch_attachment(HandlerProps, _DbName, _DocId, _AName) ->
 
 fetch_attachment(URL) ->
     case fetch_attachment(URL, 0, kz_json:new()) of
-        {'ok', _Url, Body, _Debug} -> {'ok', Body};
+        {'ok', Body} -> {'ok', Body};
         {'error', _} = Error -> Error
     end.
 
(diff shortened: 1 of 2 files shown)


core/kazoo_attachments/src/kz_att_http.erl
```diff
@@ -139,11 +139,15 @@ format_url(Fields, JObj, Args, Separator) ->
 format_url_field(JObj, Args, Fields, Acc)
   when is_list(Fields) ->
     [format_url(Fields, JObj, Args, <<>>) | Acc];
+format_url_field(JObj, Args, #{<<"arg">> := Arg}, Fields) ->
+    format_url_field(JObj, Args, {arg, Arg}, Fields);
```
lazedo: @skorobkov when is `Fields` a map item ?

core/kazoo_attachments/src/kz_att_http.erl
```diff
@@ -155,6 +159,6 @@ format_url_field(_JObj, _Args, Field, Fields) ->
 default_format() ->
     [{field, <<"pvt_account_id">>}
     ,{field, <<"owner_id">>}
-    ,{args, <<"id">>}
+    ,{arg, <<"id">>}
```
lazedo: good catch :)

core/kazoo_attachments/src/kz_att_http.erl
skorobkov: https://github.com/2600hz/kazoo/blob/master/core/kazoo_attachments/src/kz_att_http.erl#L29
```Erlang
Fields = maps:get('field_list', Params, default_format()),
```
Params is map, so this JSON in data plan will be converted to map and `field_list` will be parsed in `format_url_field/4`
```JSON
"attachments": {
       "modb-att-http": {
           "handler": "http",
           "settings": {
               "url": "http://ip:80",
               "verb": "put",
               "field_list": [
                   {
                       "field": "pvt_account_id"
                   },
                   {
                       "field": "owner_id"
                   },
                   {
                       "arg": "id"
                   },
                   {
                       "arg": "attachment"
                   }
               ]
           }
       }
...
```

lazedo: ah, good point. `kz_att_http` and `kz_att_ftp` were coded to provide backward compatibility for existing deployments for usage in  `kz_media_record`. using them as "real" att handlers in the `data plans`  is cool :)
thanks

skorobkov: I was planning to move call records files from CouchDB into plain files. Sometimes client wants to download their records not via API. So when I found this module, I thought it was made for this. о_О

lazedo:
@skorobkov awesome! thanks.

PR #3013, opened 2016-12-16
Mod views idea 4oh try2

mark2600:


lazedo:
@mark2600 i'm hesitating on this. i see a lot of view defs in memory.  i would prefer not to do this. if you have 100 requests in crossbar and each loads data from a view, we will have 100 x size of view in memory. and 100 is a very low estimate. maybe passing a callback to a function, or even have kz_data `guess` where it is (get calling app, get app priv/couch, get view from name)

mark2600:
NOTE, this is a TEMP fix and should not be used as it will be fgone shortly.

PR #3015, opened 2016-12-18
binding_fix: add needed clause to handle found failure

jamesaimonetti:
The long and short of it is that when you have a # and repeated
segments, the code was being too greedy with the # matching.

In matching <<"*.u.*.7.7.#">> against <<"i.u.e.7.7.7.a">>, the code
would match away 'a' and '7' but would also strip a '7' from the
routing key.

Uncomment dbg_test and run it with the above binding and routing keys
to see it in action.


jamesaimonetti:
@k-anderson found the issue and patched (and found a second one!).

PR #3016, opened 2016-12-19
qf-knm: remove kz_json:are_identical

fenollp:


core/kazoo_number_manager/src/kazoo_number_manager_maintenance.erl
```diff
@@ -98,7 +98,7 @@ update_number_services_view(?MATCH_ACCOUNT_ENCODED(_)=AccountDb) ->
                                  ]
                                 ,View
                                 ),
-    case kz_json:are_identical(View, NewView) of
+    case kz_json:are_equal(View, NewView) of
```
k-anderson: @fenollp If view is pulled from the DB then it will have _rev whereas NewView would not.  Please make sure this is ignored when checking if equal otherwise they never will be.

fenollp:
Don't merge: will be vampired by the `master` version of https://github.com/2600hz/kazoo/pull/3004

PR #3018, opened 2016-12-19
add 3 retry attempts to rendering, reduce timeout to 10 minutes, add …

mark2600:
…logging around erlydtl stuff


core/kazoo_templates/src/kz_template.erl
```diff
@@ -67,6 +68,7 @@ compile(Template, Module) ->
     compile(Template, Module, []).
 
 compile(Template, Module, CompileOpts) when is_binary(Template) ->
+    lager:info("starting compile1"),
```
icehess: why logging compile1, compile2? they either a template or a file

mark2600: Just to know which code path was used easily

PR #3054, opened 2016-12-29
KAZOO-5165 unlimited modb queries

jamhed:


applications/crossbar/src/crossbar_view.erl
```diff
@@ -102,10 +95,10 @@ ascending_end_key(Context, StartKey) ->
 
 page_size() -> ?PAGINATION_PAGE_SIZE.
 page_size(Context) -> page_size(Context, cb_context:api_version(Context)).
-page_size(_Context, ?VERSION_1) -> 'undefined';
+page_size(_Context, ?VERSION_1) -> undefined;
```
lazedo: @jamhed please do not perform unrelated changes. lots of these in the pr

jamhed: @lazedo, I rewrote a lot of it, so take it all as new.

fenollp:
Shouldn't this go to 4oh too?

lazedo:
only bugs go into 4.0 this is a changing behaviour

PR #3066, opened 2017-01-02
KAZOO-5260_4oh: fix number release of unknown carrier

fenollp:
Note: the catch-all was added so that binary_to_existing_atom failure is caught.

### DIFF
--- a/core/kazoo_number_manager/src/carriers/knm_carriers.erl
+++ b/core/kazoo_number_manager/src/carriers/knm_carriers.erl
@@ -325,7 +325,7 @@ disconnect(Number) ->
     try apply(Module, disconnect_number, [Number]) of
         Result -> Result
     catch
-        'error':'undef' ->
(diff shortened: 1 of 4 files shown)


core/kazoo_number_manager/src/carriers/knm_carriers.erl
```diff
@@ -325,7 +325,7 @@ disconnect(Number) ->
     try apply(Module, disconnect_number, [Number]) of
         Result -> Result
     catch
-        'error':'undef' ->
+        'error':_ ->
```
lazedo: @fenollp why this change ? what other values are you seeing here ?

fenollp: > Note: the catch-all was added so that binary_to_existing_atom failure is caught.

`binary_to_existing_atom` fails with `badarg`.

PR #3068, opened 2017-01-02
HELP-27160: correctly lookup account_db for finding attachments with …

mark2600:
…notify_outbound_fax_to_email

### DIFF
--- a/applications/notify/src/notify_fax_outbound_to_email.erl
+++ b/applications/notify/src/notify_fax_outbound_to_email.erl
@@ -35,7 +35,7 @@ handle_req(JObj, _Props) ->
 
     lager:debug("new outbound fax left, sending to email if enabled"),
 
-    AccountDb = kapi_fax:account_db(JObj),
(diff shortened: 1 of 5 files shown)


applications/notify/src/notify_fax_outbound_to_email.erl
```diff
@@ -115,7 +115,7 @@ create_template_props(Event, [FaxDoc | _Others]=_Docs, Account) ->
                  ,{<<"call_id">>, kz_json:get_value(<<"Call-ID">>, Event)}
                   | fax_values(kz_json:get_value(<<"Fax-Info">>, Event))
                  ]}
-    ,{<<"account_db">>, kz_doc:account_db(Account)}
+    ,{<<"account_db">>, kapi_fax:account_db(Event)}
```
lazedo: @mark2600 this works because `kapi_fax:account_db` looks for `Account-DB` in the payload, but the payload sent is actually a `notify`. maybe add the same lookup to `kapi_notifications` ?

PR #3072, opened 2017-01-03
send attachments to couch in chunks

lazedo:


core/kazoo_couch/src/kz_couch_attachments.erl
```diff
@@ -114,9 +114,25 @@ relay_stream_attachment(Caller, Ref, Msg) ->
...
+              ,{headers, [{<<"Transfer-Encoding">>, <<"chunked">>}
+                         ,{<<"Expect">>, <<"100-Continue">>}
+                         ]}
+              ,{content_length, erlang:byte_size(Contents0)}
+               | Options0
+              ],
     ?RETRY_504(couchbeam:put_attachment(Db, DocId, AName, Contents, Options)).
 
+-spec stream_att(binary()) -> {'ok', binary(), binary()} | 'eof'.
+stream_att(<<Data:16384/binary, Rest/binary>>) ->
```
fenollp: We should probably make this length configurable, shouldn't we?
Also I restarted the mk-bs build

PR #3073, opened 2017-01-03
KAZOO-5211 recursivly merge json objects

jamhed:


fenollp:
cc https://github.com/2600hz/kazoo/pull/3060

PR #3075, opened 2017-01-03
Typo fix for blackhole binding validation

tsromana:
the typo didnt let modules started with blachole in front of the names to be validated


fenollp:
Nice catch!
Can you also fix this one?

```
applications/crossbar/src/crossbar_bindings.erl:184:is_cb_module(<<"crossbar_", _binary>>) -> 'true';
```

fenollp:
Well I guess I spoke too soon! cc https://github.com/2600hz/kazoo/pull/3074

PR #3078, opened 2017-01-04
KAZOO-5262: Make number search timeout configurable

onnet:
I'm testing Telnyx as a number provider atm.
Sometimes Telnyx processes its number lookup slightly longer than 5 seconds.
Because of 5000 number search hardcoded timeout, Kazoo provides empty result...

### DIFF
--- a/core/kazoo_number_manager/src/knm_search.erl
+++ b/core/kazoo_number_manager/src/knm_search.erl
@@ -57,6 +57,8 @@
 
 -define(POLLING_INTERVAL, 5000).
 
+-define(NUMBER_SEARCH_TIMEOUT, kapps_config:get_integer(?KNM_CONFIG_CAT, <<"number_search_timeout_ms">>, 5000)).
+
 -define(EOT, '$end_of_table').
 
 -type state() :: #{}.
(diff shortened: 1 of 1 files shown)


core/kazoo_number_manager/src/knm_search.erl
```diff
@@ -57,6 +57,8 @@
 
 -define(POLLING_INTERVAL, 5000).
 
+-define(NUMBER_SEARCH_TIMEOUT, kapps_config:get_integer(?KNM_CONFIG_CAT, <<"number_search_timeout">>, 5000)).
```
fenollp: Can you suffix this to signify milliseconds? ie `"number_search_timeout_ms"`

fenollp: Can you suffix this to signify milliseconds? ie `"number_search_timeout_ms"`

fenollp:
I have experienced the same issue. Thanks for this patch! :)

onnet:
@fenollp Hi Pierre! There is another question raised with Telnyx module. Number lookup returns numbers with different activation charges and monthly fees costs. Since, at the moment, Kazoo charges numbers fees depending on configured service plans, there could be some impact from business point of view :)) Do you have any ideas about it already?

core/kazoo_number_manager/src/knm_search.erl
onnet: @fenollp Done.

onnet:
Look at the second line of the table:

telnyxnumbers

fenollp:
Hey I personally agree with your approach at https://github.com/2600hz/kazoo/pull/3081 
Let's hear @lazedo 's opinion on this?

PR #3079, opened 2017-01-04
HELP-27157-master: add support for changing put timeout

mark2600:


lazedo:
@mark2600 this is already supported thru connection properties, why adding another ?

PR #3081, opened 2017-01-04
KAZOO-5263: KNM_TELNYX: Select DID numbers with desired activation ch…

onnet:
…arge and monthly fee


fenollp:
Can you rebase & force-push this PR? I'd like to make sure I fixed our CircleCI config.

fenollp:
Nice it works!

onnet:
@fenollp @lazedo 

Hi guys,

Got back today to Telnyx module to test e911 number service.
Stumbled on the problem that "sandbox_provisioning" flag doesn't influences 911 provisioning behaviour.

So with "sandbox_provisioning" set to "true", Kazoo still tries to save 911 settings at Telnyx
This prevents me to test module and E911 feature itself in full.

Small correction resolves the problem:
https://github.com/onnet/kazoo_utils/blob/master/tmp40/telnyx_911.patch

But looks like we can solve this by adding a kind of a dummy e911 number service module (I cleaned up knm_telnyx_e911.erl):
https://github.com/onnet/onbill/blob/master/src/misc/knm_local_e911.erl

Does this make sense at all?
Maybe there is much easier way I just don't see?
Maybe PR?

Regards,
Kirill

fenollp:
I'll merge a PR for the first patch.

I like the second idea as it makes it clearer that only one E911 can be used per number.
I'd name it `knm_provider_e911`.
But I don't see anything calling actual provider code in that module. How do you use it?

onnet:
@fenollp Hi

Will do PR with first patch. Thanks for the comment

Regarding "knm_provider_e911". I didn't go so deep. So, it is your own cool idea :)

My use case is much easier:

Reseller wants to manage E911 addresses manually (therefore I called it "knm_local_e911").
Customer can manage 911 addresses at it's Kazoo account on its own. 
If E911 address is the new one, a proof of address document should be attached.
If address was already approved, it could be chosen from addresses list.
There could be two options after customer configured its 911 settings:
  - no calls to provider's API will be done (similar to dry run); reseller will be notified about E911 configuration attempts over email and will manage E911 configuration at provider's side manually;
   - changes will be passed through to provider's API; if new E911 address was used, reseller will still be notified over email with proof of address document  attached in order to check if everything was done correctly.

I created a small screen cast for better understanding: https://www.youtube.com/watch?v=riMk-qLyzxw

Regards,
Kirill

PR #3091, opened 2017-01-05
KAZOO-5118: forcefully set caller-id on transfered calls if specified

jamhed:
Forcefully set caller-id if specified.


fenollp:
Shouldn't that be on top of 4.0 too?

fenollp:
cc https://github.com/2600hz/kazoo/pull/3091

PR #3092, opened 2017-01-05
KAZOO-5270: read crossbar ip from system_config

icehess:
Check IP address for IPv6, if it failed check for IPv4.
Default to `::` if IPv6 is enabled, otherwise to bind to `0.0.0.0`.

### DIFF
--- a/applications/crossbar/src/crossbar_init.erl
+++ b/applications/crossbar/src/crossbar_init.erl
@@ -192,6 +192,7 @@ maybe_start_plaintext(Dispatch) ->
             %% Name, NbAcceptors, Transport, TransOpts, Protocol, ProtoOpts
             try
                 IP = get_binding_ip(),
+                lager:info("trying to bind to ~p:~b", [IP, Port]),
                 cowboy:start_http('api_resource', Workers
                                  ,[{'ip', IP}
                                   ,{'port', Port}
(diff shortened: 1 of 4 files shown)


lazedo:
@icehess looks good but needs dyalize fix and "ip" should NOT be required

lazedo:
also, make sure that binding to "::" also listens in ipv4, the previous value was {0,0,0,0,0,0,0,0} which did that.

icehess:
@lazedo wrt "::", `inet:parse_ipv6strict_address/1` resolve it to proper ip_address, so given "::" it would become `{0,0,0,0,0,0,0,0}`

But I don't know what you mean by "and "ip" should NOT be required" ?

icehess:
I mean, IP is looked up in `system_config/crossbar`, if it's not present it would set to default value of `0.0.0.0` or `::` based on availability of IPv6.

lazedo:
@icehess also, you need to default to ipv4 if the ip address in crossbar doc is ipv6 but ipv6 is not supported. and maybe check the other way around (ip is ipv4 but net is ipv6 only)

applications/crossbar/src/crossbar_init.erl
```diff
@@ -212,6 +216,36 @@ maybe_start_plaintext(Dispatch) ->
...
     end.
 
+-spec get_binding_ip() -> inet:ip_address().
+get_binding_ip() ->
+    Default = case is_ipv6_supported() of
+                  'true' -> <<"::">>;
+                  'false' -> <<"0.0.0.0">>
+              end,
+    %% expilicty convert to list to allow save the default value in human readable value
+    IP = kz_util:to_list(kapps_config:get_binary(?CONFIG_CAT, <<"ip">>, Default)),
```
lazedo: @icehess if you create a `get_default_binding` fun and use it here instead of `Default`, then `ip` will not be marked as required

icehess:
If user set a IPv6 address, but IPv6 is not supported, how we do know to bind to which IPv4 address? `0.0.0.0`?

I think it's better to stop here and just error out that IPv6 or IPv4 is not supported.

lazedo:
yes, bind to `Default` IPv4. crossbar config maybe used by different servers. also check the other way around

lazedo:
@icehess please apply the changes to #3093

sergey-safarov:
@icehess now is see ip value on couchdb like
```
       "ip": [
           48,
           46,
           48,
           46,
           48,
           46,
           48
       ]

```
Is it correct?
What I must write to bing '::'?
My dist supports IPv6 but on clean install crossbar bind to '0.0.0.0'

sergey-safarov:
installed 4.0.51

PR #3095, opened 2017-01-06
qf-gpt: fix missing view for global provisioner templates

fenollp:


lazedo:
@fenollp why this change ?

fenollp:
I un-mv'ed the file then squashed.

PR #3101, opened 2017-01-06
HELP-27139_master: fix searching numbers with v1 & ordering with Vitelity

fenollp:
cc https://github.com/2600hz/kazoo/pull/3100


fenollp:
Note: Travis tests actually pass. Only cover data upload is having issues.

> coverage/coveralls — First build on master at 6.916%

Yep that's not true!

PR #3102, opened 2017-01-06
httpc response headers are lower case

icehess:


lazedo:
@icehess please cherry-pick this to 4.0 branch

PR #3103, opened 2017-01-07
extract system_config parameters from record definition

icehess:
~~DO NOT MERGE YET!!~~

There are some modules that have system_config parameters inside their records definition. This PR extract such parameters.

Notes: During testing this, I founded because of the order of functions, the last expression that is calling one of the kapp_config get function replace the previous founded value for that parameters.

~~DO NOT MERGE YET!!~~


fenollp:
> Notes: During testing this, I founded because of the order of functions, the last expression that is calling one of the kapp_config get function replace the previous founded value for that parameters.

If you find a way to exit & report nicely on this kind of issue that would be very valuable: catches will-be race conditions on config flags.

icehess:
@fenollp I'll addressing reporting skipped config flags in another PR

core/kazoo_apps/src/kapps_speech.erl
```diff
@@ -245,7 +245,7 @@ attempt_asr_freeform(Content, ContentType, Locale, Options) ->
 attempt_asr_freeform(_, <<>>, _, _, _) -> {'error', 'no_content'};
 attempt_asr_freeform(<<"ispeech">>, Bin, ContentType, Locale, Opts) ->
     BaseUrl = kapps_config:get_string(?MOD_CONFIG_CAT
-                                     ,<<"asr_url">>
+                                     ,<<"asr_freeform_url">>
```
fenollp: Are you sure about this one?

icehess: why shouldn't I? :) The asr and asr_freedom api endpoint is different.

icehess:
It's ready btw!

core/kazoo_apps/src/kapps_speech.erl
fenollp: Cause it's the first time I see this string. Maybe that'll break some configs, no?

PR #3105, opened 2017-01-08
Update users.md to add `password` property

powerpbx:
Not in JSON so it would not be in documentation generated from that.

### DIFF
--- a/applications/crossbar/doc/users.md
+++ b/applications/crossbar/doc/users.md
@@ -69,6 +69,7 @@ Key | Description | Type | Default | Required
 `ringtones.internal` | The alert info SIP header added when the call is from external sources | `string(0..256)` |   | `false`
 `timezone` | User's timezone | `string` |   | `false`
 `username` | The GUI login username - alpha-numeric, dashes, at symbol, periods, plusses, and underscores allowed | `string(1..256)` |   | `false`
+`password` | The GUI login password
 `verified` | Determines if the user has been verified | `boolean` | `false` | `false`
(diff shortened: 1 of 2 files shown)


fenollp:
Hello!
Could you rebase your PR to latest master & force push please?

powerpbx:
I have now done that.

fenollp:
Okay in order to really add something here you need to update the schema.
For this file the schema is https://github.com/2600hz/kazoo/blob/3e6adafbb4748e00fe3b5582f21f75852a78aa97/applications/crossbar/priv/couchdb/schemas/users.json#L420

Please revert / remove your current commits, then modify that JSON file, then run `make apis` then push whatever files there are to push :)
Easy, right?

powerpbx:
As far as I can tell, password is a special case.  It is not in the schema because you do not save it anywhere in plain text.  So I am pretty sure you do not want it in the schema.

fenollp:
Why do you want to document this field? How have you been using it?

powerpbx:
To create/update accounts via API.

fenollp:
Okay so after discussion with the team this should be documented the way I described.
If you don't want to anymore, close this & I'll do it myself :)

In any case, please keep reporting things that you use & do not find in the documentation.
We are working on making it better every day :)

powerpbx:
So would this just be a placeholder in the scheme that doesn't do anything other than show up in autogenerated documents?  Is there an example of this somewhere in another scheme?

fenollp:
Hey @powerpbx I had a shot at this: https://github.com/2600hz/kazoo/pull/3148
Please close your PR if you're okay with mine ;)

PR #3106, opened 2017-01-09
Avoid infinite recursion when saving an empty list of docs

kalda341:


fenollp:
Thanks for this!
BTW if you experienced this on 4.0.X I encourage you to make the same PR against the 4.0 branch and mention this PR.

PR #3110, opened 2017-01-11
[4.0] HELP-26987: check service plan for ui_apps recursively to top reseller

icehess:
If the account does not have a service plan with apps defined then it lists all apps of the reseller, unless the reseller doesn't have apps defined in which case it list all system apps.

If the service plan defines the available apps then it gets the list from the service plan, regardless of the published state.

> NOTE: currently there is no logic for checking the `published` status, and **NONE** of the Monster apps have `published` set in their `metadata/app.json` file!

### DIFF
--- a/applications/crossbar/src/cb_apps_util.erl
+++ b/applications/crossbar/src/cb_apps_util.erl
@@ -1,5 +1,5 @@
 %%%-------------------------------------------------------------------
-%%% @copyright (C) 2012-2016, 2600Hz, INC
+%%% @copyright (C) 2012-2017, 2600Hz, INC
 %%% @doc
 %%%
 %%% @end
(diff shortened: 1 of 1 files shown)


applications/crossbar/src/cb_apps_util.erl
```diff
@@ -289,8 +333,9 @@ find_enabled_apps_fold(AppName, PlanApp, Acc) ->
     case kzd_item_plan:is_enabled(PlanApp)
         andalso find_app(AppId, PlanApp)
     of
-        TrueOrUndefined when TrueOrUndefined =:= 'true';
-                             TrueOrUndefined =:= 'undefined' ->
+        DisabledOrUndefined
+          when DisabledOrUndefined =:= 'false'
+               orelse DisabledOrUndefined =:= 'undefined' ->
```
fenollp: Not important but why switch to using `orelse` instead of `,` / `;`?

icehess: No specific reason.

lazedo:
@icehess master version ?

PR #3113, opened 2017-01-11
KAZOO-4791: block anonymous calls

jamhed:
check conditions for call to be anonymous:
- caller_id_number is set to two or more zeroes, maybe with plus sign
- caller_privacy_number is set to true
- caller_privacy_name is set to true

tests: https://github.com/2600hz/make-busy-callflow/commit/148a37c51dd196cc2dc15deed3cc588bfd8a9873

### DIFF
--- a/applications/stepswitch/src/stepswitch_inbound.erl
+++ b/applications/stepswitch/src/stepswitch_inbound.erl
@@ -228,7 +228,7 @@ maybe_add_prepend(NumberProps, JObj) ->
 
 -spec maybe_block_call(knm_number_options:extra_options(), kz_json:object()) -> kz_json:object().
 maybe_block_call(_, JObj) ->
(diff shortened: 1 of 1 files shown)


fenollp:
How about using `orelse` instead of `or`? Saves some CPU cycles & some log lines

applications/stepswitch/src/stepswitch_inbound.erl
```diff
@@ -282,6 +289,26 @@ transition_port_in(Number, JObj) ->
 %%
 %% @end
 %%--------------------------------------------------------------------
+
+%% string: two or more zeros -> true
+-spec is_zero(binary() | undefined) -> boolean().
```
fenollp: Note: this should be equivalent to `api_ne_binary()` (not super important, but for the future :) )

PR #3117, opened 2017-01-13
KAZOO-5274 correct schema names to match system_config

jamhed:


fenollp:
Hey since you got on there, how about documenting https://github.com/2600hz/kazoo/blob/master/applications/crossbar/doc/ref/configs.md in https://github.com/2600hz/kazoo/blob/master/applications/crossbar/doc/configs.md ❓

PR #3127, opened 2017-01-18
KAZOO-5278: create the number when adding a mobile device

icehess:
create number in func `add_mobile_mdn` in cb_devices

### DIFF
--- a/applications/crossbar/src/modules_v2/cb_devices_v2.erl
+++ b/applications/crossbar/src/modules_v2/cb_devices_v2.erl
@@ -807,32 +807,59 @@ maybe_add_mobile_mdn(Context) ->
 
 -spec add_mobile_mdn(cb_context:context()) -> cb_context:context().
 add_mobile_mdn(Context) ->
-    Normalized = knm_converters:normalize(get_mdn(Context)),
+    Normalized = get_mdn(Context),
+    AuthAccountId = cb_context:auth_account_id(Context),
+    Options = [{'assign_to', cb_context:account_id(Context)}
+              ,{'auth_by', AuthAccountId}
+              ,{'dry_run', not cb_context:accepting_charges(Context)}
+              ,{'module_name', ?CARRIER_MDN}
+              ],
+    case knm_number:create(Normalized, Options) of
+        {'error', Error} ->
+            lager:debug("unable to add mdn ~s to database: ~s"
+                       ,[Normalized, kz_json:encode(Error)]),
+            _ = crossbar_doc:delete(Context),
+            set_number_error_response(Error, Context);
+        _Else ->
+            lager:debug("created new mdn ~s", [Normalized]),
+            set_mobile_public_fields(Normalized, Context)
+    end.
+
+-spec set_mobile_public_fields(ne_binary(), cb_context:context()) -> cb_context:context().
+set_mobile_public_fields(Normalized, Context) ->
     AuthAccountId = cb_context:auth_account_id(Context),
     MobileField =
         kz_json:from_list(
           [{<<"provider">>, <<"tower-of-power">>}
-          ,{<<"authorizing">>, kz_json:from_list(
-                                 [{<<"account-id">>, AuthAccountId}
-                                 ])}
+          ,{<<"authorizing">>, kz_json:from_list([{<<"account-id">>, AuthAccountId}])}
           ,{<<"device-id">>, kz_doc:id(cb_context:doc(Context))}
           ]),
     PublicFields = kz_json:from_list([{<<"mobile">>, MobileField}]),
     Options = [{'auth_by', AuthAccountId}
-              ,{'assign_to', cb_context:account_id(Context)}
               ,{'dry_run', not cb_context:accepting_charges(Context)}
               ,{'public_fields', PublicFields}
-              ,{'module_name', ?CARRIER_MDN}
               ],
-    case knm_number:create(Normalized, Options) of
-        {'error', _R} ->
-            lager:debug("unable to add mdn ~s to database: ~p", [Normalized, _R]),
+    case knm_number:move(Normalized, cb_context:account_id(Context), Options) of
+        {'error', Error} ->
+            lager:debug("unable to update public fields on mdn ~s: ~s"
+                       ,[Normalized, kz_json:encode(Error)]),
             _ = crossbar_doc:delete(Context),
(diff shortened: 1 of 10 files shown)


applications/crossbar/src/modules_v2/cb_devices_v2.erl
```diff
@@ -838,6 +837,7 @@ set_mobile_public_fields(Normalized, Context) ->
     Options = [{'auth_by', AuthAccountId}
               ,{'dry_run', not cb_context:accepting_charges(Context)}
               ,{'public_fields', PublicFields}
+              ,{'state', ?NUMBER_STATE_IN_SERVICE}
```
fenollp: `move` already sets state to `in_service`. Setting this here initializes `Normalized` with a possibly wrong state, before going through transition checks.

jamhed:
Please rebase to latest master for makebusy to pass.

applications/crossbar/src/modules_v2/cb_devices_v2.erl
```diff
@@ -811,9 +811,8 @@ add_mobile_mdn(Context) ->
     Options = [{'assign_to', cb_context:account_id(Context)}
               ,{'dry_run', not cb_context:accepting_charges(Context)}
               ,{'module_name', ?CARRIER_MDN}
-              ,{'state', ?NUMBER_STATE_IN_SERVICE}
               ],
-    case knm_number:reconcile(Normalized, Options) of
+    case knm_number:create(Normalized, Options) of
```
fenollp: Note: reconcile came from a557c06cf4d15c837c8f2372c0ed82870ae655bd

applications/crossbar/src/modules_v2/cb_devices_v2.erl
```diff
@@ -829,28 +830,37 @@ set_mobile_public_fields(Normalized, Context) ->
...
     case knm_number:move(Normalized, cb_context:account_id(Context), Options) of
-        {'error', _R} ->
-            lager:debug("unable to update public fields on mdn ~s: ~p"
-                       ,[Normalized, _R]),
+        {'error', Error} ->
+            lager:debug("unable to update public fields on mdn ~s: ~s"
+                       ,[Normalized, kz_json:encode(Error)]),
             _ = crossbar_doc:delete(Context),
-            cb_context:add_system_error('datastore_fault', Context);
+            set_number_error_response(Error, Context);
```
fenollp: Why not directly use `set_response/2`?

applications/crossbar/src/modules_v2/cb_devices_v2.erl
```diff
@@ -807,50 +807,43 @@ maybe_add_mobile_mdn(Context) ->
...
               ,{'public_fields', PublicFields}
+              ,{'module_name', ?CARRIER_MDN}
               ],
     case knm_number:move(Normalized, cb_context:account_id(Context), Options) of
-        {'error', _R} ->
-            lager:debug("unable to update public fields on mdn ~s: ~p"
-                       ,[Normalized, _R]),
+        {'error', Error} ->
+            lager:debug("unable to add mdn ~s to database: ~s"
+                       ,[Normalized, kz_json:encode(Error)]),
```
fenollp: no need to log this error twice

applications/crossbar/src/modules_v2/cb_devices_v2.erl
```diff
@@ -870,15 +863,17 @@ remove_if_mobile(MDN, Context) ->
     Normalized = knm_converters:normalize(MDN),
     case knm_number:get(Normalized) of
         {'ok', Number} ->
+            PN = knm_number:phone_number(Number),
+            IsMdnCarrier = ?CARRIER_MDN =:= knm_phone_number:module_name(PN),
             case kz_json:get_ne_value(<<"mobile">>, knm_number:to_public_json(Number)) of
-                'undefined' -> Context;
+                'undefined' when not IsMdnCarrier -> Context;
                 Mobile ->
```
fenollp: what about the case when not undefined AND not IsMdnCarrier?

applications/crossbar/src/modules_v2/cb_devices_v2.erl
```diff
@@ -895,8 +890,12 @@ remove_if_mobile(MDN, Context) ->
 get_mdn(Context) ->
     case cb_context:req_value(Context, ?KEY_MOBILE_MDN) of
         'undefined' ->
-            kz_json:get_ne_value(?KEY_MOBILE_MDN, cb_context:fetch(Context, 'db_doc'));
-        MDN -> MDN
+            case kz_json:get_ne_value(?KEY_MOBILE_MDN, cb_context:fetch(Context, 'db_doc')) of
+                'undefined' -> 'undefined';
+                MDN -> knm_converters:normalize(MDN)
+            end;
+        <<>> -> 'undefined';
```
fenollp: How about using is_empty instead?

core/kazoo_number_manager/src/knm_numbers.erl
```diff
@@ -662,7 +662,7 @@ reconcile_number(T0, Options) ->
     pipe(do_in_wrap(F1, T0), [fun save_phone_numbers/1]).
 
 do_move_not_founds(Nums, Options) ->
-    pipe(new(Options, Nums)
+    pipe(new([{state, ?NUMBER_STATE_IN_SERVICE} | Options], Nums)
```
fenollp: Can you add a test in knm_create_new_number_test of a move of ?TEST_CREATE_NUM with carrier MDN and check that state is indeed in_service & assigned to reseller?

applications/crossbar/src/modules_v2/cb_devices_v2.erl
icehess: If mobile is not undefined, so device is mobile anyway. I'm checking module_name here only for cases that it's a mobile number but carrier sets to different carrier name because maybe it's an old number or something.

fenollp: No chance a mobile num has something else as a carrier. okay then :)

icehess:
@fenollp I'm backporting the commits to 4oh, we should address other issues for mdn in another PR maybe? This PR was only about to fix the crash during creating a mobile device.

fenollp:
@icehess Agreed. Moving stuff to KAZOO-5288.

PR #3132, opened 2017-01-20
KAZOO-5288: apply certain restrictions to MDNs

fenollp:
Note: still one thing not covered:

> Numbers of type MDN should not be able to be moved or deleted directly.  Only removing the mobile device should remove the number

I'm thinking of adding the number option `{mdn_run, true}` to knm_number calls in the concerned cb_devices_v2 clauses and have move&release fail on MDNs where this isn't set (unset := false).
That adds yet one more option but that's all I have.

Also, is it possible for a device to be deleted without going through the cb_devices_v2 path? Cause we'd need to handle that.


fenollp:
Ready for merge.

PR #3139, opened 2017-01-22
HELP-27667: support email and http headers

k-anderson:


applications/fax/src/fax_worker.erl
```diff
@@ -939,7 +939,13 @@ fetch_document_from_url(JObj) ->
 prepare_contents(JobId, RespHeaders, RespContent) ->
     lager:debug("preparing fax contents", []),
     TmpDir = kapps_config:get_binary(?CONFIG_CAT, <<"file_cache_path">>, <<"/tmp/">>),
-    case fax_util:normalize_content_type(props:get_value("content-type", RespHeaders, <<"application/octet-stream">>)) of
+    ContentType = props:get_first_defined(["content-type"
+                                          ,"Content-Type"
```
lazedo: @k-anderson are you sure about this ? we recently updated it to handle `content-type` in lower case because that's what  httpc returns. maybe the change is in line 915 https://github.com/2600hz/kazoo/pull/3139/files#diff-31927b00cdefb4dc32a54751fc331fedR915 to set to lower case

lazedo: wrong line. i think the fix should be in line 917 `Props = [{"Content-Type", ContentType}],`  to `Props = [{"content-type", ContentType}],`

lazedo:
@k-anderson if you have nothing against, i'll close this in favour of #3142

PR #3140, opened 2017-01-22
HELP-27667: support email and http headers

k-anderson:


lazedo:
@k-anderson if you have nothing against, i'll close this in favour of #3141

PR #3146, opened 2017-01-24
fake_trial_time_left

onnet:
'trial_time_left' is the flag one can use to assume the account is in trial/evaluation state.
Posting this flag to the account doc allows to keep it and pretend the account is trial despite it is not...


applications/crossbar/src/modules/cb_accounts.erl
```diff
@@ -816,7 +816,10 @@ leak_trial_time_left(Context) ->
     leak_trial_time_left(Context, JObj, kz_account:trial_expiration(JObj)).
 
 leak_trial_time_left(Context, _JObj, 'undefined') ->
-    Context;
+    RespData = kz_json:delete_key(<<"trial_time_left">>
+                                ,cb_context:resp_data(Context)
```
fenollp: Can you fix the indentation please?

onnet:
@fenollp Done.

PR #3148, opened 2017-01-24
qf-users_schema: document users.password schema property

fenollp:
cc https://github.com/2600hz/kazoo/pull/3105


powerpbx:
Thank you.  This is just a placeholder so that it shows up in generated docs correct?   Unless I am missing something it should normally be an empty value yes?  

Functionally, the actual password currently combined with username, hashed, and saved as `credentials` as defined in `user_auth` schema.  That all makes sense as is for security.

PR #2874, opened 2016-12-05
HotOrNot improvement

skorobkov:
https://groups.google.com/forum/#!topic/2600hz-dev/qzOzc9y_CPs

PR #3152, opened 2017-01-24
Dinkor's pimp my rate

jamesaimonetti:
Provides the base upgrades needed to start supporting multiple ratedecks within Kazoo.

Closes #2874 in favor of doc and code updates contained herein.


jamesaimonetti:
@lazedo please take a look and see if this branch works for you too. The CircleCI failure is due to a bug in the formatter and not to anything in the code.

@skorobkov please try this branch out as well, make sure it works for your use too.

@onnet want to try as well? :)

onnet:
@jamesaimonetti Certainly. Will give it a try. Many thanks!

applications/tasks/src/modules/kt_rates.erl
```diff
@@ -0,0 +1,405 @@
...
+    kz_datamgr:enable_change_notice(),
+    kzs_publish:maybe_publish_db(Db, <<"edited">>).
+
+%%%===================================================================
+%%% Internal functions
+%%%===================================================================
+
+-spec split_results(kz_json:objects()) -> {kz_json:objects(), kz_json:object()}.
+split_results([_|_] = JObjs) ->
+    {Head, [Last]} = lists:split(erlang:length(JObjs)-1, JObjs),
```
fenollp: Just length

applications/tasks/src/modules/kt_rates.erl
```diff
@@ -0,0 +1,405 @@
...
+is_allowed(Props) ->
+    AuthAccountId = props:get_value('auth_account_id', Props),
+    AccountId = props:get_value('account_id', Props),
+    {'ok', AccountDoc} = kz_account:fetch(AccountId),
+    {'ok', AuthAccountDoc} = kz_account:fetch(AuthAccountId),
+    kz_util:is_in_account_hierarchy(AuthAccountId, AccountId, 'true')
+    %% Serve request for reseller rates
+        andalso kz_account:is_reseller(AccountDoc)
+    %% or serve requests from SuperAdmin
+        orelse kz_account:is_superduper_admin(AuthAccountDoc).
```
fenollp: Can you move this check up? If it fails the worst case is better.

core/kazoo_services/src/services/kz_service_ratedeck.erl
```diff
@@ -0,0 +1,56 @@
+%%%-------------------------------------------------------------------
+%%% @copyright (C) 2012-2016, 2600Hz, INC
```
fenollp: year is wrong

core/kazoo_services/src/services/kz_service_ratedeck.erl
```diff
@@ -0,0 +1,56 @@
...
+%% @public
+%% @doc
+%%
+%% @end
+%%--------------------------------------------------------------------
+-spec reconcile(kz_services:services()) -> kz_services:services().
+-spec reconcile(kz_services:services(), api_binary()) -> kz_services:services().
+reconcile(Services) ->
+    PlanJObj = kz_services:service_plan_json(Services),
+    %% TODO: resolve conflict when there is more then one ratedeck
```
fenollp: than

core/kazoo_services/src/services/kz_service_ratedeck_name.erl
```diff
@@ -0,0 +1,56 @@
+%%%-------------------------------------------------------------------
+%%% @copyright (C) 2012-2016, 2600Hz, INC
```
fenollp: year is wrong

core/kazoo_services/src/services/kz_service_ratedeck_name.erl
```diff
@@ -0,0 +1,56 @@
...
+%% @public
+%% @doc
+%%
+%% @end
+%%--------------------------------------------------------------------
+-spec reconcile(kz_services:services()) -> kz_services:services().
+-spec reconcile(kz_services:services(), api_binary()) -> kz_services:services().
+reconcile(Services) ->
+    PlanJObj = kz_services:service_plan_json(Services),
+    %% TODO: resolve conflict when there is more then one ratedeck_name
```
fenollp: than

core/kazoo_tasks/Makefile
```diff
@@ -3,4 +3,12 @@ PROJECT = kazoo_tasks
 
 all: compile
 
+test: check_csvs
+
+check_csvs:
+	@stat -t -- test/*.csv >/dev/null 2>/dev/null ; if [[ $$? > 0 ]] ; then echo WARNING: run 'make csvs' for more test data ; fi
```
fenollp: How about directly running it?

core/kazoo_tasks/src/kz_csv.erl
```diff
@@ -55,16 +55,22 @@ count_rows(CSV) when is_binary(CSV) ->
...
+-spec throw_bad(row(), {integer(), non_neg_integer()}) -> {integer(), non_neg_integer()}.
 throw_bad(Header, {-1,0}) ->
     case lists:all(fun is_binary/1, Header) of
         %% Strip header line from total rows count
-        'true' -> {length(Header), 0};
-        'false' -> throw('bad_csv')
+        'true' ->
+            {length(Header), 0};
+        'false' ->
+            io:format("bad header: ~p~n", [Header]),
```
fenollp: Double these io:format calls with lager:debug calls.

core/kazoo_tasks/src/kz_csv.erl
```diff
@@ -26,7 +26,7 @@
 -endif.
 
 -type cell() :: ne_binary() | ?ZILCH.
--type row() :: [cell(), ...].
+-type row() :: [cell()].
```
fenollp: Why allow `[]` as a csv row? How is it useful?

core/kazoo_tasks/src/kz_csv.erl
```diff
@@ -90,32 +95,104 @@ fold(CSV, Fun, Acc)
...
+
+-spec split_field(binary(), field_terminator(), row()) ->
+                         {row(), binary()}.
+-spec split_field(binary(), field_terminator(), row(), [byte()]) ->
+                         {row(), binary()}.
+split_field(Row, EndChar, Fields) ->
+    split_field(Row, EndChar, Fields, []).
+
+split_field(<<$,>>, $,, Fields, FieldSoFar) ->
+    split_fields(<<>>, [<<>>, iolist_to_binary(lists:reverse(FieldSoFar)) | Fields]);
```
fenollp: That second `<<>>` should be `?ZILCH`, shouldn't it?

core/kazoo_tasks/src/kz_csv.erl
```diff
@@ -90,32 +95,104 @@ fold(CSV, Fun, Acc)
...
+split_field(<<EndChar>>, EndChar, Fields, FieldSoFar) ->
+    Field = iolist_to_binary(lists:reverse(FieldSoFar)),
+    split_fields(<<>>, [Field | Fields]);
+
+split_field(<<Char:1/binary>>, _EndChar, Fields, FieldSoFar) ->
+    Field = iolist_to_binary(lists:reverse([Char | FieldSoFar])),
+    split_fields(<<>>, [Field | Fields]);
+split_field(<<Char:8, Row/binary>>, EndChar, Fields, FieldSoFar) ->
+    split_field(Row, EndChar, Fields, [Char | FieldSoFar]).
```
fenollp: I don't get it: why accumulate a binary into a reversed list of binaries when you can just `<<FieldSoFar/binary, Char/binary>>`?

core/kazoo_tasks/src/kz_csv.erl
```diff
@@ -140,29 +217,40 @@ pad_row_to(_, Row) ->
...
+    end.
+
+-spec map_header_to_ordered_fields(ne_binaries(), ne_binaries()) ->
+                                          map().
+map_header_to_ordered_fields(CSVHeader, OrderedFields) ->
+    maps:from_list(
+      [{find_ordered_position(Header, OrderedFields, 1), CSVHeaderIndex}
+       || {CSVHeaderIndex,Header} <- lists:zip(lists:seq(1, length(CSVHeader)), CSVHeader)
+      ]).
+
```
fenollp: Again, did something change here above?

core/kazoo_tasks/src/kz_csv.erl
```diff
@@ -201,12 +289,12 @@ json_to_iolist(Records)
...
 %% @private
--spec find_position(A, [A], I) -> I when
-      A :: ne_binary(),
-      I :: pos_integer().
-find_position(Item, [Item|_], Pos) -> Pos;
-find_position(Item, [_|Items], N) ->
-    find_position(Item, Items, N+1).
+-spec find_ordered_position(ne_binary(), ne_binaries(), non_neg_integer()) ->
+                                   non_neg_integer().
+find_ordered_position(_Item, [], _Pos) -> 0;
```
fenollp: Why return 0 now?

core/kazoo_tasks/test/kz_csv_tests.erl
```diff
@@ -0,0 +1,113 @@
...
+add_split_row_assertions({Row, Split}, Tests) ->
+    {Fields, _} = kz_csv:split_row(Row),
+    [{binary_to_list(Row), ?_assertEqual(Split, Fields)} | Tests].
+
+files_test_() ->
+    filelib:fold_files("test/", "\.csv\$", 'false', fun gen_file_tests/2, []).
+
+gen_file_tests(File, Tests) ->
+    {'ok', CSV} = file:read_file(File),
+    [{File, ?_assert(0 < (catch kz_csv:count_rows(CSV)))} | Tests].
```
fenollp: Remove the catch or explicitly expect a throw.

applications/crossbar/src/modules/cb_tasks.erl
```diff
@@ -284,6 +290,13 @@ put(Context) ->
             crossbar_util:response_400(<<"bad request">>, Reason, Context)
     end.
 
+-spec task_account_id(cb_context:context()) -> api_ne_binary().
+task_account_id(Context) ->
+    case cb_context:account_id(Context) of
+        'undefined' -> cb_context:auth_account_id(Context);
```
jamesaimonetti: I wanted to hit /v2/tasks as super duper admin to upload a ratedeck, which isn't tied to an account

applications/tasks/src/modules/kt_rates.erl
```diff
@@ -296,7 +296,7 @@ generate_row([?VARS], Props) ->
     Update = [{<<"pvt_type">>, <<"rate">>}
              ,{<<"rate_name">>, maybe_generate_name(Name, Prefix, ISO, Direction)}
              ,{<<"weight">>, maybe_generate_weight(Weight, Prefix, Cost, IntCost)}
-             ,{<<"routes">>, [<<"^\\+?", Prefix/binary, ".+$">>]}
+             ,{<<"routes">>, [<<"^\\+?", Prefix/binary, ".+\$">>]}
```
jamesaimonetti: I didn't want to go this route because it is confusing and I, frankly, think your formatter is broken.

I use the exact same emacs function to format Erlang buffers automatically when saving them:

```emacs-lisp
(add-hook 'erlang-mode-hook                             
          (lambda ()                     
            (add-hook 'before-save-hook 'erlang-indent-current-buffer nil 'make-it-local)))
```
AFAICT, your [fmt.el](https://github.com/fenollp/erlang-formatter/blob/master/fmt.el#L14) does the same thing. I diffed my erlang.el against your repo's version and they are identical. Something else is causing this; maybe emacs 24 (I'm using 25), maybe you're not bringing along enough elisp to do this? Perhaps edts is helping fix this issue in my installation and raw erlang.el has the issue?

Not sure, but I think this commit should be backed out and the formatter fixed.

applications/hotornot/doc/README.md
```diff
@@ -1,2 +1,95 @@
...
+    "_id": "f36cac061205f7a0f7fde6b4d3370bf9",
+    "_rev": "1-1b67fb507c69c35200178a9b1e0e6cfc",
+    "routes": [
+        "^\\+?1.+$"
+    ],
+    "weight": "1",
+    "rate_name": "outbound_US_1",
+    "pvt_type": "rate",
+    "prefix": "1",
+    "rate_cost": "1.27",
```
jamesaimonetti: currency, which will be converted to units. It is also made-up

applications/hotornot/src/hon_util.erl
```diff
@@ -103,53 +107,73 @@ build_keys(<<D:1/binary, Rest/binary>>, Prefix, Acc) ->
...
-
-matching_rates(Rates, DID, Direction, RouteOptions) ->
-    E164 = knm_converters:normalize(DID),
-    [Rate || Rate <- Rates,
-             matching_rate(Rate, E164, Direction, RouteOptions)
-    ].
+matching_rates(Rates, ReqJObj) ->
+    FilterList = kapps_config:get(?APP_NAME, <<"filter_list">>, ?DEFAULT_FILTER_LIST),
+    lists:foldl(fun(Filter, Acc) ->
+                        lists:filter(fun(R) -> matching_rate(R, Filter, ReqJObj) end, Acc)
```
jamesaimonetti: There could be N matching rates by prefix. `FilterList` I believe then allows you to filter those rates that matched by prefix by more criteria. Then, once you have the final list of "matching" rates, you sort them and take the first one in the list.

applications/crossbar/src/modules/cb_tasks.erl
fenollp: I don't see the link. Since you auth as superduper you have an account.
This root endpoint is not documented and furthermore no special authentication step ensures only superduper has rights to this endpoint (see kz_rates).
I'd say either fix these 2 points and remove is_allowed in kz_rates, or just use the normal endpoint.

applications/hotornot/src/hon_util.erl
```diff
@@ -161,14 +185,23 @@ sort_rate(RateA, RateB) ->
...
 
+-spec sort_rate_by_cost(kz_json:object(), kz_json:object()) -> boolean().
+sort_rate_by_cost(RateA, RateB) ->
+    PrefixA = byte_size(kz_json:get_binary_value(<<"prefix">>, RateA)),
+    PrefixB = byte_size(kz_json:get_binary_value(<<"prefix">>, RateB)),
+
+    case PrefixA =:= PrefixB of
+        'true' ->
+            kz_json:get_float_value(<<"rate_cost">>, RateA, 0) >
+                kz_json:get_float_value(<<"rate_cost">>, RateB, 0);
```
jamesaimonetti: I'm not sure I follow this reasoning - if "rate_cost" is undefined in either, it will default to 0. The comparison can only ever be a float value or 0.

applications/tasks/src/kz_tasks_scheduler.erl
```diff
@@ -233,9 +243,22 @@ handle_call({'start_task', TaskId}, _From, State) ->
             end;
         [#{started := Started}]
           when Started /= 'undefined' ->
+            lager:info("task ~s exists already", [TaskId]),
+            ?REPLY(State, {'error', 'already_started'})
+    end;
+handle_call({'restart_task', TaskId}, _From, State) ->
```
jamesaimonetti: I'll defer to @skorobkov to answer :)

applications/hotornot/doc/README.md
fenollp: Then maybe it should be in units? Or if it's USD it should say so, shouldn't it?

applications/tasks/priv/couchdb/views/tasks.json
```diff
@@ -6,6 +6,7 @@
             "map": [
                 "function(doc) {",
                 "  if (doc.pvt_type != 'task' || doc.pvt_deleted) return;",
+                "  var csvs = undefined; ('_attachments' in doc) ? csvs = Object.keys(doc['_attachments']) : csvs = undefined;",
```
fenollp: We have to do without `Object.keys`. Instead try

```js
var csvs = [], attachments = doc._attachments || {};
for (var k in attachments)
  if (attachments.hasOwnProperty(k))
    csvs.push(k);
```

and below

```js
csvs: (csvs === []) ? undefined : csvs,
```

fenollp:
Careful, I workedaroun ERL204 again

applications/tasks/priv/couchdb/views/tasks.json
```diff
@@ -6,7 +6,11 @@
             "map": [
                 "function(doc) {",
                 "  if (doc.pvt_type != 'task' || doc.pvt_deleted) return;",
-                "  var csvs = undefined; ('_attachments' in doc) ? csvs = Object.keys(doc['_attachments']) : csvs = undefined;",
+                "  var csvs;",
+                "  if ('_attachments' in doc) {",
+                "    csvs = []; ",
+                "    for (var k in doc['_attachments']) { csvs.push(k); } ",
```
fenollp: This doesn't work. Need hasOwnProperty.

applications/crossbar/src/modules/cb_tasks.erl
```diff
@@ -178,11 +178,12 @@ content_types_provided(Context, _TaskId, _CSV) ->
 
 -spec ctp(cb_context:context()) -> cb_context:context().
 ctp(Context) ->
-    cb_context:add_content_types_provided(Context, [{'to_csv', ?CSV_CONTENT_TYPES}
-                                                   ,{'to_json', ?JSON_CONTENT_TYPES}
+    cb_context:add_content_types_provided(Context, [{'to_json', ?JSON_CONTENT_TYPES}
+                                                   ,{'to_csv', ?CSV_CONTENT_TYPES}
```
fenollp: Order is important here. Why the change?

applications/tasks/src/kz_tasks_scheduler.erl
fenollp: @skorobkov ping

applications/tasks/src/modules/kt_rates.erl
```diff
@@ -0,0 +1,405 @@
...
+                         ,[props:get_value('account_id', Props)
+                          ,props:get_value('auth_account_id', Props)
+                          ]
+                         ),
+            {<<"task execution is forbidden">>, 'stop'}
+    end;
+export(_Props, 'stop') -> 'stop';
+export(_Props, State) ->
+    Db = props:get_value('db', State),
+    Options = props:get_value('options', State),
```
fenollp: ping?

applications/tasks/src/modules/kt_rates.erl
fenollp: ping?

core/kazoo_tasks/Makefile
fenollp: How big is this set of csv files? Does it get updated often? How about commiting it to Kazoo?

core/kazoo_tasks/src/kz_csv.erl
fenollp: ping?

fenollp:
Please go over unanswered comments

applications/tasks/src/kz_tasks_scheduler.erl
skorobkov: Yes. I am here.
Was on vacation.
I will look and answer tomorrow.

skorobkov: I don't remember I change anything in "tasks" app internals.
"kt_rates" is the only thing I added.

applications/tasks/src/modules/kt_rates.erl
skorobkov: I don't mind.
While writing code it was more convenient  for me.to use proplist (for debug).
But this is not my branch, should I create pull request into this branch?

applications/tasks/src/modules/kt_rates.erl
skorobkov:
```Erlang
kz_account:is_superduper_admin(AuthAccountDoc)
orelse kz_util:is_in_account_hierarchy(AuthAccountId, AccountId, 'true')
andalso kz_account:is_reseller(AccountDoc).
```
?

applications/tasks/src/kz_tasks_scheduler.erl
fenollp: Okay. @jamesaimonetti can you revert adding this function + formatting changes to tasks app's code?

applications/tasks/src/modules/kt_rates.erl
fenollp: I'm rewriting some things in tasks for a client & Props will now be a map. The variable amount of arguments an input task takes will also be merged into one map (one map per csv row).

applications/tasks/src/modules/kt_rates.erl
fenollp: hm disregard that.

applications/tasks/src/modules/kt_rates.erl
skorobkov: CSV rows as maps is great!

applications/tasks/src/kz_tasks_scheduler.erl
jamesaimonetti: @fenollp I found it useful from the command line to restart failing tasks while i was testing importing various CSVs.

applications/tasks/src/modules/kt_rates.erl
jamesaimonetti: @skorobkov nah, don't worry about it. I'll merge changes @fenollp makes and update kt_rates

core/kazoo_tasks/Makefile
jamesaimonetti: Well, they're not our CSVs so it felt unnecessary to commit them.

core/kazoo_tasks/src/kz_csv.erl
jamesaimonetti: No, tests fail if not the empty binary

core/kazoo_tasks/src/kz_csv.erl
jamesaimonetti: Habit of accumulating and reversing. I'll have to check the impact of appending the binaries vs accumulating a list

core/kazoo_tasks/Makefile
fenollp: hm maybe there can be some licensing issues? but if they're not too big I'd rather we include them in kazoo_csv/test or kazoo_csv/priv/fixtures

core/kazoo_tasks/src/kz_csv.erl
fenollp: But in kz_csv a cell can either be ?ZILCH or ne_binary(). For cells declared as `""` I think the expected value is `<<"\'\'">>` even. No?

core/kazoo/src/kz_binary.erl
```diff
@@ -28,10 +28,10 @@
         ,remove_white_spaces/1, remove_white_spaces/2
         ]).
 
--export([binary_md5/1]).
+-export([md5/1]).
 -export([pad/3, pad_left/3
         ,join/1, join/2
-        ,binary_reverse/1
+        ,reverse/1
```
fenollp: Can you move all this renaming to a dedicated PR?

applications/crossbar/src/modules/cb_tasks.erl
fenollp: @jamesaimonetti ^

core/kazoo_json/src/kz_json.erl
```diff
@@ -1183,15 +1183,15 @@ is_private_key(<<"_", _/binary>>) -> 'true';
 is_private_key(<<"pvt_", _/binary>>) -> 'true';
 is_private_key(_) -> 'false'.
 
--type composite_key() :: binary() | [binary()].
+-type composite_key() :: key() | [keys()].
```
fenollp: a list of keys? do you mean a path?

core/kazoo_csv/test/kz_csv_test.erl
```diff
@@ -152,3 +154,31 @@ json_to_iolist_test_() ->
...
+           ,{<<"\"0.1651\",\"ZAMBIA, MOBILE\",\"ZAMBIA, MOBILE-26094\",\"ZAMBIA, MOBILE\",\"26094\",0">>
+            ,[<<"0.1651">>, <<"ZAMBIA, MOBILE">>, <<"ZAMBIA, MOBILE-26094">>, <<"ZAMBIA, MOBILE">>, <<"26094">>, <<"0">>]
+            }
+           ,{<<"0.1651,\"ZAMBIA, MOBILE\",\"ZAMBIA, MOBILE-26094\",\"ZAMBIA, MOBILE\",\"26094\",\"0\"">>
+            ,[<<"0.1651">>, <<"ZAMBIA, MOBILE">>, <<"ZAMBIA, MOBILE-26094">>, <<"ZAMBIA, MOBILE">>, <<"26094">>, <<"0">>]
+            }
+           ,{<<",">>, [<<>>, <<>>]}
+           ,{<<"test,">>,[<<"test">>,<<>>]}
+           ,{<<"test,,">>,[<<"test">>,<<>>,<<>>]}
+           ,{<<"test,,foo bar">>,[<<"test">>,<<>>,<<"foo bar">>]}
```
fenollp: These `<<>>` should be `?ZILCH` instead. You cannot revert existing behavior.

core/kazoo_csv/test/kz_csv_test.erl
```diff
@@ -152,3 +154,31 @@ json_to_iolist_test_() ->
...
+add_split_row_assertions({Row, Split}, Tests) ->
+    {Fields, _} = kz_csv:split_row(Row),
+    [{binary_to_list(Row), ?_assertEqual(Split, Fields)} | Tests].
+
+files_test_() ->
+    filelib:fold_files("test/", "\\.csv\$", 'false', fun gen_file_tests/2, []).
+
+gen_file_tests(File, Tests) ->
+    {'ok', CSV} = file:read_file(File),
+    [{File, ?_assert(0 < (catch kz_csv:count_rows(CSV)))} | Tests].
```
fenollp: This is terrible. `0` is always inferior to tuples.

core/kazoo_documents/include/kzd_ratedeck.hrl
```diff
@@ -0,0 +1,24 @@
...
+
+-define(ENCODE_RATEDECK_DB(RatedeckId)
+       ,<<"ratedeck%2F", RatedeckId/binary>>
+       ).
+-define(encode_RATEDECK_DB(RatedeckId)
+       ,<<"ratedeck%2f", RatedeckId/binary>>
+       ).
+-define(UNENCODED_RATEDECK_DB(RatedeckId)
+       ,<<"ratedeck/", RatedeckId/binary>>
+       ).
```
fenollp: I don't know anything about ratedeck ids nor dbs, but is it at all possible to explicitly match a byte_size here?
It can help a lot (like ?MATCH_ACCOUNT_stuff)

core/kazoo_json/include/kazoo_json.hrl
```diff
@@ -29,6 +29,9 @@
 -type object() :: {json_proplist()} | ?EMPTY_JSON_OBJECT.
 -type objects() :: [object()] | [].
 
+-type flat_object() :: {[{[key()], json_term()}]} | ?EMPTY_JSON_OBJECT.
```
fenollp: `keys() :: [key()]`

scripts/export_auth_token.bash
```diff
@@ -0,0 +1,63 @@
...
+
+usage() { echo 'Usage: eval $('"$0"' [-c {CREDENTIALS_HASH}] [-a {ACCOUNT_NAME}] [-p {PHONE_NUMBER}] [-r {ACCOUNT_REALM}])' 1>&2;}
+
+function authenticate() {
+    local C="$1"
+    local TYPE="$2"
+    local ID="$3"
+    AUTH_RESP=$(curl -s -X PUT http://localhost:8000/v2/user_auth -d "{\"data\":{\"credentials\":\"$C\", \"$TYPE\":\"$ID\"}}")
+
+    STATUS=$(echo $AUTH_RESP | jq -r '.status')
```
fenollp: We already require python2. How about using it?
there's a json formatter in the same folder too

applications/crossbar/doc/tasks.md
```diff
@@ -376,7 +389,8 @@ curl -v -X GET \
             "node": "whistle_apps@qwd",
             "start_timestamp": 63633924909,
             "status": "executing",
-            "success_count": 50
+            "success_count": 50,
+            "csvs":["in.csv"]
```
fenollp: Explain what this is for?
What happens if this list contains more than one element? none? the same element more than once? a wrong name?

There is already a way to fetch input & output CSVs. `file_name` contains the name of the input CSV.

applications/crossbar/doc/tasks.md
```diff
@@ -469,15 +483,17 @@ curl -v -X PATCH \
 }
 ```
 
+#### Retrieve a task's CSVs
 
-#### Retrieve a task's input CSV
+When you `GET /v2/accounts/{ACCOUNT_ID}/tasks/{TASK_ID}`, the JSON will include a `"csvs":[...]" array with input and output CSVs as appropriate. Use the name(s) in the array to specify which you would like to receive.
 
-> GET /v2/accounts/{ACCOUNT_ID}/tasks/{TASK_ID}/input
+> GET /v2/accounts/{ACCOUNT_ID}/tasks/{TASK_ID}
```
fenollp: While you are free to ADD documentation, you CANNOT remove documentation for EXISTING endpoints.

applications/crossbar/doc/tasks.md
```diff
@@ -110,11 +119,15 @@ curl -v -X GET \
...
 
+#### Add a new system task
+
 
 #### Add a new task
 
 > PUT /v2/accounts/{ACCOUNT_ID}/tasks
 
+Note: There are tasks that run against system resources, only for use by the superduper admin (like rate uploading), which can omit `/accounts/{ACCOUNT_ID}` from the URI. Leaving the account in the URI should have no impact.
+
```
fenollp: Other tasks will profit from this:

1. Move the "is super admin" check logic into `kz_tasks:new`
1. Really document this `/v2/tasks` endpoint (curl PUT with success & error cases)
1. Add a boolean `<<"needs_admin">>` to tasks requiring superadmin rights to be run, where `true` means returning `{error, unauthorized}` when not a superadmin in `kz_tasks:new`.
1. Update tasks with `is_allowed/1` code to only rely on that boolean instead

Because as it is this /v2/tasks API makes no sense. Plus, it violates the `extra_args()` type.

core/kazoo_json/src/kz_json.erl
jamesaimonetti: No, path() includes array indicies (integers). Flatten doesn't go deeper if its not a json object.

applications/crossbar/src/modules/cb_tasks.erl
```diff
@@ -153,19 +163,36 @@ cta(Context, _) ->
...
+ctp(Context) ->
+    cb_context:add_content_types_provided(Context, [{'to_json', ?JSON_CONTENT_TYPES}
+                                                   ,{'to_csv', ?CSV_CONTENT_TYPES}
+                                                   ]).
+
+-spec to_csv({cowboy_req:req(), cb_context:context()}) ->
+                    {cowboy_req:req(), cb_context:context()}.
+to_csv({Req, Context}) ->
+    Filename = requested_attachment_name(Context),
+    Headers = props:set_values([{<<"content-type">>, <<"application/octet-stream">>}
```
jamesaimonetti: This is similar to what we do for CDR csv downloads.

applications/crossbar/src/modules/cb_tasks.erl
```diff
@@ -352,20 +391,113 @@ set_db(Context) ->
...
+read_doc_or_attachment(TaskId, Context, AccountId, [?MEDIA_VALUE(<<"text">>, <<"comma-separated-values">>, _, _, _)|_]) ->
+    read_attachment(TaskId, Context, AccountId);
+read_doc_or_attachment(TaskId, Context, AccountId, [?MEDIA_VALUE(<<"application">>, <<"octet-stream">>, _, _, _)|_]) ->
+    read_attachment(TaskId, Context, AccountId);
+read_doc_or_attachment(TaskId, Context, AccountId, [_Accept|Accepts]) ->
+    lager:debug("failed to handle accept value ~p", [_Accept]),
+    read_doc_or_attachment(TaskId, Context, AccountId, Accepts);
+read_doc_or_attachment(TaskId, Context, AccountId, []) ->
+    lager:info("failed to find valid accept value"),
+    read_doc(TaskId, Context, AccountId).
```
jamesaimonetti: I mean, validate calls validate_tasks which calls read which validates Accept header values. So it is a sub-function of validate...

applications/hotornot/src/hon_rater.erl
```diff
@@ -12,49 +12,55 @@
...
--spec handle_req(kz_json:object(), kz_proplist()) -> 'ok'.
-handle_req(JObj, _Props) ->
-    'true' = kapi_rate:req_v(JObj),
-    _ = kz_util:put_callid(JObj),
+-spec init() -> 'ok'.
+init() ->
+    _ = [kapps_maintenance:refresh(kzd_ratedeck:format_ratedeck_db(Ratedeck))
+         || Ratedeck <- hotornot_config:ratedecks()
+        ],
+    'ok'.
```
jamesaimonetti: Seems like its 6 one way, half a dozen the other. Not sure one is better than the other in this particular case.

applications/tasks/src/kz_tasks_scheduler.erl
```diff
@@ -85,6 +86,15 @@ start_link() ->
 start(TaskId=?NE_BINARY) ->
     gen_server:call(?SERVER, {'start_task', TaskId}).
 
+-spec restart(kz_tasks:id()) -> {'ok', kz_json:object()} |
+                                {'error'
+                                ,'not_found' |
+                                 'already_started' |
+                                 any()
```
jamesaimonetti: This is identical to `start/1` and `handle_call_start_task/2`

applications/tasks/src/modules/kt_rates.erl
```diff
@@ -0,0 +1,422 @@
...
+            {<<"task execution is forbidden">>, 'stop'}
+    end;
+export(_ExtraArgs, 'stop') -> 'stop';
+export(_ExtraArgs, State) ->
+    Db = props:get_value('db', State),
+    Options = props:get_value('options', State),
+    Limit = props:get_value(['options', 'limit'], State),
+    case kz_datamgr:get_results(Db, ?RATES_VIEW, Options) of
+        {'ok', []} -> 'stop';
+        {'ok', Results} when length(Results) >= Limit ->
```
jamesaimonetti: No, I think this was his attempt to paginate since he sets limit to bulk_limit + 1 earlier.

core/kazoo_csv/src/kz_csv.erl
```diff
@@ -88,31 +95,104 @@ fold(CSV, Fun, Acc)
 %% @doc
 %% @end
 %%--------------------------------------------------------------------
--spec take_row(csv()) -> {row(), csv()} |
-                         eof.
-take_row(<<>>) -> eof;
+-spec take_row(binary()) -> {row(), binary()} |
+                            'eof'.
```
jamesaimonetti: I think this handles a trailing newline in a csv, no?

core/kazoo_csv/src/kz_csv.erl
```diff
@@ -225,10 +304,11 @@ map_io_indices(Header, CSVHeader) ->
     lists:foldl(MapF, #{}, IndexToCSVHeader).
 
 %% @private
--spec cell_to_binary(cell()) -> csv().
+-spec cell_to_binary(cell()) -> binary().
```
jamesaimonetti: A cell != a csv right? So why would the return type be `csv()` (ignoring that `csv()` is equivalent to `binary()`). Seems misleading to say that converting a cell to binary returns a csv.

core/kazoo_csv/src/kz_csv.erl
```diff
@@ -225,10 +304,11 @@ map_io_indices(Header, CSVHeader) ->
...
 cell_to_binary(?ZILCH) -> <<>>;
+cell_to_binary(<<>>) -> <<>>;
```
jamesaimonetti: They are allowed in rows: 
```csv
"Number","First","Middle","Last"
17,"James","","Aimonett"`
```

core/kazoo_csv/test/kz_csv_test.erl
jamesaimonetti: They aren't ?ZILCH though, they are empty binaries and not 'undefined'.

core/kazoo_documents/include/kzd_ratedeck.hrl
jamesaimonetti: ratedeck ids are arbitrary strings so can't match them precisely

scripts/export_auth_token.bash
jamesaimonetti: What's wrong with jq; seems ideally suited to what I'm doing so I used it.

applications/crossbar/doc/tasks.md
jamesaimonetti: 1. super admin is a crossbar construct, meaning the auth account id is flagged super duper. cb_tasks takes care of sending the auth account id to new/7 when there is no account id on the url path
1. The GET is added and this note should suffice. If people include the account id on the url, that's fine. This just hints they can leave it off if they want to save keystrokes.
1. Task module authors can add it. With rates, the eventual goal may be to allow resellers to upload their own ratedecks.

`extra_args()` is fine because new/7 is called with account IDs

applications/crossbar/src/modules/cb_tasks.erl
fenollp: Except this API returns text/csv since its inception. Use text/csv.

core/kazoo_json/src/kz_json.erl
fenollp: okay

applications/crossbar/src/modules/cb_tasks.erl
fenollp: @jamesaimonetti [your function here](https://github.com/2600hz/kazoo/blob/5bb8459006a859611aa647ec6048c85e27e229c8/applications/crossbar/src/modules/cb_tasks.erl#L425-L443) acts only on its 4th argument. I'm asking that you rewrite it to an arity 1 function, or to something that actually makes sense.

applications/hotornot/src/hon_rater.erl
fenollp: 6 what? lists:foreach brings explicit semantics & skips creating an unused list. Use it

applications/tasks/src/kz_tasks_scheduler.erl
fenollp: ok

applications/tasks/src/modules/kt_rates.erl
fenollp: Whatever `Limit`'s value, it cannot be more than itself. `>=` should be `=:=`

core/kazoo_csv/src/kz_csv.erl
fenollp: empty rows are not allowed by kz_csv

core/kazoo_csv/src/kz_csv.erl
fenollp: you're right

core/kazoo_csv/src/kz_csv.erl
fenollp: No they are not. You want them to, that's the difference.
`""` is `?ZILCH`.

core/kazoo_csv/test/kz_csv_test.erl
fenollp: not before your changes. see comment above.

scripts/export_auth_token.bash
fenollp: `jq` is not installed by default on most systems, plus doc/installation.md lists packages required by Kazoo.
What's wrong with Python

applications/crossbar/doc/tasks.md
fenollp: 1. "super admin is a crossbar construct" this is false. "cb_tasks takes care of sending the auth account id to new/7 when there is no account id on the url path" your code does account_id = auth_account_id = cb_context:auth_account_id, which violates the concept.
2. Either you justify a usage for it, either you remove `/v2/tasks`

fenollp:
Please add the little examples on quoting the RFC has as tests to kazoo_csv app.

https://tools.ietf.org/html/rfc4180#section-2

applications/hotornot/src/hon_rater.erl
jamesaimonetti: It is an expression, meaning there's no difference. See the [efficiency guide](http://erlang.org/doc/efficiency_guide/listHandling.html#id68285) to see that the compiler doesn't create a list when the result isn't used.

core/kazoo_csv/src/kz_csv.erl
jamesaimonetti: Why be so restrictive though. If someone leaves an empty line at the end of the file, you want to punish the request?

scripts/export_auth_token.bash
jamesaimonetti: jq isn't listed because this script isn't necessarily part of normal kazoo installation/operation. I've included it because it was handy for me, when testing, to regenerate the auth token and still be able to re-run curl commands by using the env variables instead of the literal values. That's the whole point of the script, to set auth_token and account_id into env variables.

applications/crossbar/doc/tasks.md
jamesaimonetti: As I have said, `/v2/tasks` makes a lot of sense when uploading the "global" ratedecks (not tied to an account). It will make sense to include `/accounts/{ACCOUNT_ID}` when we enable creating ratedecks specifically for resellers.

Crossbar and blackhole appear to be the only applications that care about superduper admin status of an account, with the exception of kt_rates, which it needs at the moment.

If there is no account id on the URL, then it uses the auth_account_id instead; why is this a problem? If kz_tasks didn't require an account_id and auth_account_id we could send in undefined for account_id to those task modules that don't care about accounts. As-is, this works just fine.

lazedo:
things to improve

- ratedeck ids could/should be system generated uuid and use MATCH_ACCOUNT like in guards

applications/hotornot/src/hon_rater.erl
fenollp: ah didn't know about that for LCs. (it being an expression has nothing to do with it though)

core/kazoo_csv/src/kz_csv.erl
fenollp: [This is already handled](https://github.com/2600hz/kazoo/blob/master/core/kazoo_csv/test/kz_csv_test.erl#L108)
What you are adding is allowing empty rows in THE MIDDLE of a CSV file.

scripts/export_auth_token.bash
fenollp: I don't care about the whole point of the script. I'm saying that it would make more sense to rewrite it to use stuff that we know people using Kazoo will already have installed.

applications/crossbar/doc/tasks.md
fenollp: what about this?

fenollp:
things to improve
* revert allowing empty rows in the middle of csvs
* revert allowing <<>> instead of just undefined in csv cells
* revert csvs field of explain what it doesn't duplicates
* revert undocumenting /input and /output API endpoints
* refactor read_doc_or_attachment so that it doesn't quarry arguments for no purpose

applications/hotornot/src/hon_rater.erl
jamesaimonetti: I was referring to the expression: "6 one way, half a dozen the other". +1 for English ambiguity :)

PR #3155, opened 2017-01-25
Follow up to Kazoo-5277: released numbers not fully deleting when they should

danielfinke:
Kazoo-5277 patch solved a portion of this, but the circumstance remained that numbers that are released would still remain in the database when the should_delete option was on. The change to release/2 makes sure that permanently deleted numbers don't get passed to save_phone_numbers/1 but still do return in the ok list. Removing the set_assigned_to/2 call in knm_phone_number.erl both ensures that the now-available number matches the state of an available number that is freshly added AND ensures that the knm_phone_number:delete/1 call made on permanently deleted numbers can also delete the number from the assigned_to account.

### DIFF
--- a/core/kazoo_number_manager/src/knm_numbers.erl
+++ b/core/kazoo_number_manager/src/knm_numbers.erl
@@ -339,13 +339,21 @@ release(Nums) ->
     release(Nums, knm_number_options:default()).
 
 release(Nums, Options) ->
-    ret(pipe(do_get_pn(Nums, Options)
-            ,[fun knm_phone_number:release/1
-             ,fun knm_number:new/1
-             ,fun knm_providers:delete/1
-             ,fun unwind_or_disconnect/1
-             ,fun save_phone_numbers/1
-             ])).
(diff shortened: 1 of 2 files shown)


fenollp:
Hi Daniel, 
What is your issue exactly?
If you release a knm_local number, are you still able to find it in number db or account db?
I don't understand your change to not save deleted PNs (since they are skipped in knm_phone_number:save anyway).
Maybe add some eunit tests?

danielfinke:
@fenollp Yeah, when they are released, they still remain in both the number/account dbs. I hadn't noticed that knm_phone_number should be skipping those deleted numbers in save_to_number_db, so I am going to revisit this and see why that isn't catching my issue. I'll update the PR with my findings.

danielfinke:
@fenollp I'm going to close this. Somehow the branch that I cherry-picked these commits from (and was testing on) was poorly merged and missed some critical changes from commit a993ca9.

fenollp:
Ah makes sense! Thanks :)

PR #3156, opened 2017-01-25
KAZOO-5268: allow to control anonymizing CID if carrier handling privacy

icehess:
Add a system config option to allow disabling caller id anonymous and just setting appropriate privacy/screen flags on outbound calls if the carrier honoring the privacy flags.

* Introduce a new module to handling the anonymizing logic in a core
* Handle Inbound calls to the an endpoint if the privacy flags are set
* Handle loopback calls
* Can define the privacy settings in a Resource and if not found use system_config (previously it was in Account's document only and Trunkstore calls)
* Works with calling extension->extension


jamhed:
I have strong suspicion MakeBusy fails here for a reason.

icehess:
@jamhed make-busy test should be passing now, it was for blocking anonymous call not looking for the right CCV headers.

applications/callflow/src/module/cf_privacy.erl
```diff
@@ -18,35 +18,48 @@ handle(Data, Call) ->
...
-                ,kapps_config:get_non_empty(?CF_CONFIG_CAT
-                                           ,<<"privacy_number">>
-                                           ,kz_util:anonymous_caller_id_number()
-                                           )
-                }
+               ,{fun kapps_call:set_custom_channel_vars/2, CCVs}
                ],
     cf_exe:set_call(kapps_call:exec(Routines, Call)).
+
+-spec ccvs_by_privacy_mode(ne_binary()) -> kz_proplist().
```
jamesaimonetti: api_ne_binary()

icehess:
@jamhed probably some makebusy tests regarding anonymous calls will failed or have not correct result after the commit 6011442 due to system config `block_anonymous_caller_id` is moved from `stepswitch` to `privacy`. I'll issue PR for fix them if I can.

applications/crossbar/priv/couchdb/schemas/system_config.callflow.json
```diff
@@ -64,14 +64,6 @@
             "description": "callflow post answer delay",
             "type": "integer"
         },
-        "privacy_name": {
```
jamesaimonetti: @icehess you will need to add a migrate clause to move existing callflow settings for privacy to the privacy system_config doc

jamesaimonetti: Something along the lines of https://github.com/2600hz/kazoo/blob/master/core/kazoo_media/src/kazoo_media_maintenance.erl#L22

icehess: @jamesaimonetti I already did that, also if you notice I'm calling `kapps_config:migrate/1` in `kapps_controller` during start up to initiate config migration.

icehess: @jamesaimonetti check `kapps_config:migrate`

PR #3173, opened 2017-01-31
KAZOO-5302: Updated tollfree regex

sergey-safarov:


fenollp:
I'd say that regexp could be extended to support all of what's in https://github.com/2600hz/kazoo/blob/0def604c6f8d142bac2e4a2d076c48130fa8adef/core/kazoo_number_manager/src/knm.hrl#L15-L31 (that's from wikipedia).
What do you think @lazedo ?

fenollp:
Once done, run `make apis`. That's what's bothering circleci.

sergey-safarov:
Regex extended

sergey-safarov:
make apis results here
http://pastebin.com/dT8zcq3S

fenollp:
Thanks! Let me try to merge this.

PR #3174, opened 2017-01-31
KAZOO-5303: Fixed parsing error in Vitelity carrier module

sergey-safarov:


fenollp:
Wow can't believe this, thanks for the patch!
Can you edit `vitelity.hrl` to wrap the fixture calls with `list_to_binary`?
Cause right now tests mock Vitelity's server's reply as strings when they should use binaries!
Thanks again

sergey-safarov:
@fenollp my skills not enough to do this

sergey-safarov:
Please, make this changes by self and reject this pull request.

fenollp:
Sure. Will do :)

fenollp:
Thanks again!

PR #3186, opened 2017-01-31
Pivot docs refresh

jamesaimonetti:


fenollp:
@jamesaimonetti wrt circleci's xref: no idea how, but it's broken just for your PR.

PR #3188, opened 2017-02-01
HELP-27022_master: ensure views exist on MoDb creation

fenollp:


lazedo:
@fenollp what are you trying to solve here ?

fenollp:
That when you run `create` on an empty modb it ensures to populate the views. Can happen sometimes. (it just did today on my laptop).
It also happened on prod today but views got there eventually. I think the account crawler put them there.

lazedo:
i think this is related to https://github.com/2600hz/kazoo/pull/3193

fenollp:
It isn't harmful though? views are added only if they are missing, right?

lazedo:
but if you have 10 views, you still have 10 HEAD operations on the database.
if you create the db, then you create the views. no need for extr db operations (even HEAD requests)

fenollp:
> if you create the db, then you create the views. no need for extr db operations (even HEAD requests)

Hm that was obviously not the case on my laptop. This call is not done very frequently (as opposed to the ping to couch root every 5 seconds or so).
Also, #3193 will not fix this for me (and other devs) as the VM may be restarted/shutdown at midnight.

I don't see the harm here

lazedo:
`Hm that was obviously not the case on my laptop` its related to cache not being cleared.
didn't understand the second, if the VM is restarted no cache issues, right ?

fenollp:
I don't know if it's a cache issue. I had an empty modb, with nothing in there and errors in the logs.

PR #3190, opened 2017-02-01
add from_list_recursive to kz_json

lazedo:


fenollp:
Tests?

core/kazoo_json/src/kz_json.erl
```diff
@@ -245,6 +245,26 @@ are_equal(JObj1, JObj2) ->
...
+    recursive_from_list(L).
+
+-spec recursive_from_list(list()) -> object().
+recursive_from_list([First | _]=List)
+  when is_list(List)
+       andalso is_tuple(First)  ->
+    set_values([{K, recursive_from_list(V)} || {K,V} <- List], new());
+recursive_from_list(X) when is_float(X) -> X;
+recursive_from_list(X) when is_integer(X) -> X;
+recursive_from_list(X) when is_atom(X) -> X;
```
fenollp: A test example: try a nested get in the created jobj for an atom key & the same binary key. That would be interesting to test/document.

core/kazoo_json/src/kz_json.erl
```diff
@@ -245,6 +245,26 @@ are_equal(JObj1, JObj2) ->
...
+-spec from_list_recursive(json_proplist()) -> object().
+from_list_recursive([]) -> new();
+from_list_recursive(L)
+  when is_list(L) ->
+    recursive_from_list(L).
+
+-spec recursive_from_list(list()) -> object().
+recursive_from_list([First | _]=List)
+  when is_list(List)
+       andalso is_tuple(First)  ->
```
fenollp: I think this could use `?JSON_WRAPPER(...)` instead?

core/kazoo_json/src/kz_json.erl
```diff
@@ -245,6 +245,26 @@ are_equal(JObj1, JObj2) ->
...
+    set_values([{K, recursive_from_list(V)} || {K,V} <- List], new());
+recursive_from_list(X) when is_float(X) -> X;
+recursive_from_list(X) when is_integer(X) -> X;
+recursive_from_list(X) when is_atom(X) -> X;
+recursive_from_list(X) when is_list(X) -> X;
```
fenollp: A test should show that lists are properly handled.

core/kazoo_json/src/kz_json.erl
```diff
@@ -245,6 +245,26 @@ are_equal(JObj1, JObj2) ->
...
+recursive_from_list(X) when is_binary(X) -> X;
+recursive_from_list({_Y, _M, _D}=Date) -> kz_time:iso8601({Date, {0,0,0}});
+recursive_from_list({{_, _, _}, {_, _, _}}=DateTime) -> kz_time:iso8601(DateTime);
+recursive_from_list(_Else) -> null.
```
fenollp: I would use `undefined` here, since almost nothing in Kazoo checks for `null`.

core/kazoo_json/src/kz_json.erl
lazedo: no, here we're checking for KV in the form {somekey, somevalue}

core/kazoo_json/src/kz_json.erl
lazedo: null is used to remove values

core/kazoo_json/test/kz_json_test.erl
```diff
@@ -557,3 +557,21 @@ sum_test_() ->
...
+
+from_list_recursive_test() ->
+    List = [{<<"fax">>,
+             [{<<"id">>,<<"201702-1b6cfec4e3ab0b7bb97ced78bf431f39">>},
+              {<<"info">>,
+               [{<<"fax_hangup_code">>,200},
+                {<<"fax_hangup_cause">>,<<"NORMAL_CLEARING">>}]},
+              {<<"notifications">>,
+               {[{<<"email">>,
+                  {[{<<"send_to">>,
```
fenollp: This and above are literal jobjs. I thought the idea was that this should be a proplist that would get transformed into jobjs by your function?

lazedo: check again, one is the other isn't

lazedo: proplist to jobj, not  jobjs

PR #3193, opened 2017-02-01
publish database create/delete

lazedo:
datamgr caching of db_exists requires it
explicit fails db_exists for 'other' dbs

### DIFF
--- a/core/kazoo_data/src/kzs_publish.erl
+++ b/core/kazoo_data/src/kzs_publish.erl
@@ -8,8 +8,7 @@
 -module(kzs_publish).
 
 
--export([maybe_publish_db/2
-        ,maybe_publish_doc/3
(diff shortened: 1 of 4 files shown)


core/kazoo_data/src/kzs_publish.erl
```diff
@@ -53,14 +53,10 @@ maybe_publish_doc(Db, Doc, JObj) ->
         'false' -> 'ok'
     end.
 
--spec maybe_publish_db(ne_binary(), kapi_conf:action()) -> 'ok'.
+-spec maybe_publish_db(ne_binary(), kapi_conf:action()) -> boolean().
 maybe_publish_db(DbName, Action) ->
```
fenollp: Maybe remove "maybe_" then?

PR #3199, opened 2017-02-02
URI-encode prompt ids for uploading media

danielfinke:
Restore uri-encoding that can be found in 3.22. Without it, doc ids contain the '/' and cb_media APIs cannot retrieve the doc contents (return 404s).

### DIFF
--- a/core/kazoo_media/src/kz_media_util.erl
+++ b/core/kazoo_media/src/kz_media_util.erl
@@ -449,7 +449,7 @@ prompt_id(<<"/system_media/", PromptId/binary>>, Lang) ->
 prompt_id(PromptId, 'undefined') -> PromptId;
 prompt_id(PromptId, <<>>) -> PromptId;
 prompt_id(PromptId, Lang) ->
-    <<Lang/binary, "/", PromptId/binary>>.
+    kz_util:uri_encode(<<Lang/binary, "/", PromptId/binary>>).
 
 -spec get_prompt(ne_binary()) -> api_binary().
 -spec get_prompt(ne_binary(), api_binary() | kapps_call:call()) ->


jamesaimonetti:
@danielfinke manually merged but github doesn't like that i reworded the commit to ref the pr number, so...

lazedo:
@jamesaimonetti i think there will be double encodings errors because of this. needs verification.

lazedo:
@danielfinke i don't think the problem is here and we may revert this. what was your use case so i can try it out ? give specific examples. the problem may be in cb_media / crossbar or even in the client making the call to crossbar.

danielfinke:
@lazedo where do you think there might be double encoding?

So our use case is an upload of prompts via PUT to /v1/accounts/<account_id>/media. The payload looks like:

`{"data":{"language":"es","prompt_id":"agent-already_logged_in","description":"C:\\fakepath\\uploaded_file_63651824866.mp3","name":"es%2Fagent-already_logged_in","ui_metadata":{"ui":"kazoo-ui"}},"verb":"PUT"}`

Notice that the id is not specified ahead of time. cb_media generates the prompt id in maybe_add_prompt_fields/1. If it publishes the doc with the '/' instead of '%2F', GETs, POSTs, and DELETEs for the doc will 404.

The reason for modifying the kz_media_util:prompt_id/2 function is twofold. First, the only places in the code base this function is used is by cb_media above and by the kazoo_media_maintenance import_prompt command. They should be consistent in their prompt id formatting. Second, the 3.22 version of prompt_id/2 has cow_qs doing the uri-encode and for some reason the encode is not present in 4.0.

The change doesn't affect the ability for prompts to be retrieved by kz_media_map and it does ensure that doc contents are retrievable by the various cb_media APIs.

lazedo:
@danielfinke `name":"es%2Fagent-already_logged_in"`  why is this encoded ? maybe this is your issue ?
the problem with this change is prompt_id is encoding a part of url

danielfinke:
@lazedo neither the couch view nor the cb_media code use the name field. I agree they should match up, but it won't solve the issue to remove encoding there.

Couch is fine with fetching the doc when its id contains a '/', but the issue is that the tokenization of crossbar URLs can't cleanly handle requests to a doc with slashes because it's the separator for URL tokens. So the natural solution is to have couch double-encode the URL for fetches (which it can do just fine) and single-encode the id for crossbar requests.

danielfinke:
@lazedo oh also one more thing, regarding kz_media_map: it doesn't have a problem with the encoded URLs as they are fetched via the couch view, which works fine either way.

lazedo:
@danielfinke if its there maybe is because its receiving it that way. what is the exact request url you're sending in the request? 
couch has nothing to do here. the issue is either crossbar tokenization or cb_media.

jamesaimonetti:
@lazedo `prompt_id` was encoded in 3.22 this way; I couldn't find why it was removed in 4.x so felt it was better to go back to a known working state (confirmed by @danielfinke since he did the PR).

lazedo:
@jamesaimonetti and it led to double encoding names in couchdb, we don't see it because of big rename.

danielfinke:
@lazedo the request URL as logged by api_resource:
`api_resource:151 (<0.15631.1>) PUT: /v1/accounts/043a546e37a2e2caa151f91c50d701f5/media?`

lazedo:
@danielfinke you were saying that `name` was not used, it seems it is used. have you tried not to encode it ?

danielfinke:
@lazedo yes, it doesn't have an effect. The name is used but not for document fetches.

danielfinke:
Here is a doc that is successfully uploaded with the id encoded:
`{"_id":"es%2Fagent-already_logged_in","_rev":"2-eb07667d9ddd8af0a82e993f9b1e1abb","prompt_id":"agent-already_logged_in","language":"es","description":"C:\\fakepath\\uploaded_file_63651824866.mp3","name":"es%2Fagent-already_logged_in","ui_metadata":{"ui":"kazoo-ui"},"media_source":"upload","streamable":true,"tts":{"voice":"female/en-US"},"pvt_type":"media","pvt_vsn":"1","pvt_account_id":"043a546e37a2e2caa151f91c50d701f5","pvt_account_db":"account%2F04%2F3a%2F546e37a2e2caa151f91c50d701f5","pvt_created":63653363429,"pvt_modified":63653363429,"pvt_request_id":"c95896bf804bb6804acc54f72e04b3d4","pvt_auth_user_id":"254bee867ee06a5fb0f18e31075aa2c7","pvt_auth_account_id":"66ed30a1ceb6edee352c7105c5733792","pvt_is_authenticated":true,"_attachments":{"uploaded_file_63653363429.mp3":{"content_type":"audio/mp3","revpos":2,"digest":"md5-JbpMCdzJhiXHTJ2VU+/pcQ==","length":39215,"stub":true}}}`

Couch fetches it via the url: hostname:5984/account%2F04%2F3a%2F546e37a2e2caa151f91c50d701f5/es%252Fagent-already_logged_in

danielfinke:
With that doc, you can get it using crossbar:
GET /v1/accounts/043a546e37a2e2caa151f91c50d701f5/media/es%2Fagent-already_logged_in

If you have the doc without the id encoded, you can try to get via:
GET /v1/accounts/043a546e37a2e2caa151f91c50d701f5/media/es/agent-already_logged_in
but it will fail because there is no matching route for tokens [accounts, <account_id>, media, es, <media_id>]

lazedo:
@danielfinke thanks for posting the complete doc. you are encoding the `_id`. trying remove the patch your side and not encoding `es%2Fagent-already_logged_in` at all.

danielfinke:
@lazedo As mentioned in my last comment, you cannot GET that doc without it encoded.

lazedo:
@danielfinke the second GET is wrong because it is not encoded, so i would expect that

danielfinke:
@lazedo if you try the first GET (encoded) with the doc (id unencoded), it's a 404, because it looks specifically for the encoded id in couch (couch request is for es%252Fagent-already_logged_in == double encoded).

danielfinke:
Just to double check, I did what you suggested:
Doc contents:
`{
  "_id": "de-gr/agent-already_logged_in",
  "_rev": "1-5859b0627bb761b16cd03c77f8db43f9",
  "language": "de-gr",
  "prompt_id": "agent-already_logged_in",
  "description": "C:\\fakepath\\uploaded_file_63651824866.mp3",
  "name": "de-gr/agent-already_logged_in",
  "ui_metadata": {
    "ui": "kazoo-ui"
  },
  "media_source": "upload",
  "streamable": true,
  "tts": {
    "voice": "female/en-US"
  },
  "pvt_type": "media",
  "pvt_vsn": "1",
  "pvt_account_id": "043a546e37a2e2caa151f91c50d701f5",
  "pvt_account_db": "account%2F04%2F3a%2F546e37a2e2caa151f91c50d701f5",
  "pvt_created": 63653365023,
  "pvt_modified": 63653365023,
  "pvt_request_id": "9f4e77b0c80f9181c410b9ed0b506b4e",
  "pvt_auth_user_id": "254bee867ee06a5fb0f18e31075aa2c7",
  "pvt_auth_account_id": "66ed30a1ceb6edee352c7105c5733792",
  "pvt_is_authenticated": true
}`

As I was saying, both media API URL formats are a 404:
curl -H "X-Auth-Token: $AUTH_TOKEN" hostname:8000/v1/accounts/043a546e37a2e2caa151f91c50d701f5/media/de-gr/agent-already_logged_in
`{"data":"not found","error":"404"...`

curl -H "X-Auth-Token: $AUTH_TOKEN" hostname:8000/v1/accounts/043a546e37a2e2caa151f91c50d701f5/media/de-gr%2Fagent-already_logged_in
`{"data":"bad identifier","error":"404"...`

lazedo:
@danielfinke the changed will break this http://image.prntscr.com/image/954fc73f4251498ab4176ad664ce0b5a.png
how does your documents look like in futton ?

danielfinke:
@lazedo they are also broken:
`{
  "_id": "en-us/agent-already_logged_in",
  "_rev": "2-e4261e4641b74105366bac1665af515c",
  "name": "en-us/agent-already_logged_in",
  "prompt_id": "agent-already_logged_in",
  "description": "System prompt in en-us for agent-already_logged_in",
  "content_length": 122156,
  "language": "en-us",
  "content_type": "audio/x-wav",
  "source_type": "whistle_media_maintenance",
  "streamable": true,
  "pvt_type": "media",
  "pvt_created": 63628582802,
  "pvt_modified": 63628582802,
  "pvt_account_db": "system_media",
  "_attachments": {
    "agent-already_logged_in.wav": {
      "content_type": "audio/x-wav",
      "revpos": 2,
      "digest": "md5-BfbqaJthCwtyQGEXQLBotQ==",
      "length": 122156,
      "stub": true
    }
  }
}`

curl -H "X-Auth-Token: $AUTH_TOKEN" devapi.voxter.com:8000/v1/media/prompts/en-us/agent-already_logged_in
`{"data":"not found","error":"404"...`

 curl -H "X-Auth-Token: $AUTH_TOKEN" devapi.voxter.com:8000/v1/media/prompts/en-us%2Fagent-already_logged_in
`{"data":[],"revision":"undefined"...`
^^ This is empty because no prompts match due to %2F =/= /.

This PR also makes sure they are uploaded encoded, so that the latter curl can work. Probably, we should add a migration to re-id those docs. Would require a flush of media maps afterwards.

lazedo:
@danielfinke what do you mean with `add a migration to re-id those docs`, what docs ? the ones i posted ?

danielfinke:
@lazedo yep, those docs. They are also broken because the id is not url-encoded. Similarly to the prompt overrides, the system prompts are url-encoded in 3.22.

lazedo:
@danielfinke that's exactly what we don't want to, they work fine as they are.

danielfinke:
@lazedo I'm explicitly telling you they do not.

danielfinke:
I don't know how much more evidence of this I can give you.

danielfinke:
I respect your concern about the impact that this could have if it wasn't thoroughly considered. The back-and-forth identified that a migration would be a necessary part of a complete solution in this PR. I do not see where the issue lies as a result of these changes outside of the aesthetic change of prompt doc ids being url-encoded. I believe this is a minor concern. I would like to hear what you believe would be an additional impact so we can discuss it. Could you could attempt the various examples I provided here and observe the behaviour that I am targeting, specifically that the media docs are not retrievable via API? I feel that you may have a different perspective on this PR if you do so.

lazedo:
@danielfinke i think i identified the issue. its in cb_media. i'll make a PR that will reverse this and update cb_media. for your reference the problem is here https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/modules/cb_media.erl#L249 and another part where kz_http_util:urlencode(MediaId) should not be there.

danielfinke:
@lazedo I did a test with the url encodes in cb_media removed and it looks workable. Just a few updates required on the client end. Thanks!

lazedo:
https://github.com/2600hz/kazoo/pull/3217

PR #3201, opened 2017-02-02
HELP-27959: bulk functions for managing number carrier modules

k-anderson:


fenollp:
cc https://github.com/2600hz/kazoo/pull/3202

PR #3203, opened 2017-02-03
Revert KAZOO-5107 change as that actually /broke/ it in the way the ticket describes

danielfinke:
The change to andalso in KAZOO-5107 actually causes the caller ID to never appear for forwarded external callers. We just ran into this after a recent merge and reverting the change fixes the issue. Not directly reverting commit so as to keep lager line english update ;)

### DIFF
--- a/core/kazoo_endpoint/src/kz_endpoint.erl
+++ b/core/kazoo_endpoint/src/kz_endpoint.erl
@@ -1439,7 +1439,7 @@ maybe_retain_caller_id({_Endpoint, _Call, 'undefined', _JObj}=Acc) ->
 maybe_retain_caller_id({Endpoint, Call, CallFwd, JObj}) ->
     {Endpoint, Call, CallFwd
     ,case kapps_call:inception(Call) =:= 'undefined'
-         andalso kz_json:is_true(<<"keep_caller_id">>, CallFwd)
+         orelse kz_json:is_true(<<"keep_caller_id">>, CallFwd)
      of
          'true' ->
              lager:info("call forwarding will retain caller id"),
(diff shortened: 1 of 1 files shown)


lazedo:
@danielfinke you are proposing here that every `onnet`  call will keep caller.

danielfinke:
@lazedo So this PR would cause the outbound caller id on an internal user/device/account to be used as opposed to what the carrier would send. Am I getting that right?

The way I'm reading it then, we don't need to check the inception at all. It doesn't matter if the call is onnet or offnet if you want to keep caller id and if you don't want to keep caller id, it should just always be off. What do you think?

lazedo:
@danielfinke @jamesaimonetti found the `inception` thing . https://github.com/voxter/kazoo/commit/218c612977ada22694198fca712f49b3f935d023#diff-c66ede63de147b68cc20ebb4f2bc4d36R1082
maybe you can comment why it was added in the first place.
thanks

danielfinke:
@lazedo https://2600hz.atlassian.net/browse/KAZOO-3529
sort of looks like the original issue was that the extension calling another extension with a cell device did not have an outbound CID set. So, it would try to use the extension as CID (internal CID) instead of the account level outbound CID. Somehow I don't think this is an issue anymore as kz_endpoint explicitly requests the external CID: https://github.com/voxter/kazoo/blob/7f5cd3a7052e9243bc84390a327e744cbac89935/core/kazoo_endpoint/src/kz_endpoint.erl#L1216

lazedo:
@danielfinke thanks for the update. so it seems that the change set should drop the `inception` part. will you update it ?
@icehess thoughts ?

danielfinke:
@lazedo sure thing, I'll make this change a little later today

danielfinke:
@lazedo I've updated this PR

lazedo:
@danielfinke thanks

PR #3205, opened 2017-02-03
KAZOO-5313: port-in numbers must not have features & have state port_in

fenollp:
cc https://github.com/2600hz/kazoo/pull/3204


fenollp:
Ready.

PR #3208, opened 2017-02-03
qf-travis: see if starting lager solves lager crash reports

fenollp:
Not for merging (yet, at least)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ script:
   - # make eunit
   - make deps
   - ERL_LIBS="$HOME/proper" ERLC_OPTS='-DPROPER' make compile-test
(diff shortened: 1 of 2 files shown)


fenollp:
Need help here:
This job passed without error: https://api.travis-ci.org/jobs/198075269/log.txt?deansi=true
This one failed: https://api.travis-ci.org/jobs/198075268/log.txt?deansi=true
They are based on the same code, one runs on OTP18, the other OTP19.

See a difference?

lazedo:
@fenollp i think this was related to travis having problems with its worker

fenollp:
@lazedo this pr is older than your comment about travis...
anyway, this didn't fix anything, closing.

PR #3214, opened 2017-02-03
HELP-27910: remove our name from the default template

k-anderson:


fenollp:
cc https://github.com/2600hz/kazoo/pull/3213

PR #3215, opened 2017-02-03
check if modb is created then try to do operation, otherwise return failed

icehess:
Return boolean if the modb is actually created if modb createed, try the operation again otherwise return error


lazedo:
@icehess why this pr ?

icehess:
@lazedo I see a lot of error `modb ~p is exists, not creating` in the logs which they are mostly from `kz_media_recording`

lazedo:
@icehess seems a large change set just for this. how about removing the log ?

PR #3218, opened 2017-02-04
Update cb_rates.erl

powerpbx:
Need to reduce the minimum increments for 6sec/6sec billing.  Not sure why these are here.  Perhaps you can consider removing them completely.

### DIFF
--- a/applications/crossbar/priv/api/swagger.json
+++ b/applications/crossbar/priv/api/swagger.json
@@ -6240,14 +6240,14 @@
                 "rate_increment": {
                     "default": 60,
                     "description": "The time slice, in seconds, to bill in.",
-                    "minimum": 10,
+                    "minimum": 6,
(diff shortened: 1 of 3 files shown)


jamesaimonetti:
We're working hard a a tasks-based uploader for ratedecks. I hope to have it ready this weekend, for next week. This work will deprecate the cb_rates uploading mechanism.

I'll add some tests for doing 6/6 billing too and see what happens.

jamesaimonetti:
Hi @powerpbx, please take a look at the [ratedeck uploader docs](https://github.com/2600hz/kazoo/blob/master/doc/blog/understanding_rates.md) and see if your use-case is addressed now. There shouldn't be any limitations on increment or minimum.

Thanks for the PR but at this time I think the issue it addresses has been fixed. Please keep them coming though!

PR #3219, opened 2017-02-04
qf-list_port_in: @kanderson this ignores unconfirmed and canceled. Wo…

fenollp:
…rks for you?

https://github.com/2600hz/kazoo/pull/3205 was merged without some debated changes from https://github.com/2600hz/kazoo/pull/3204


lazedo:
@fenollp why all these changes when the issue is only the view ?

PR #3223, opened 2017-02-06
kz_json:diff, expand, flatten/2

jamhed:
- calculate difference between json objects
- new api for flatten (second argument determines the way to join keys)
- new function expand to convert from flatten representation

### DIFF
--- a/applications/crossbar/src/api_resource.erl
+++ b/applications/crossbar/src/api_resource.erl
@@ -876,7 +876,7 @@ accept_override(Context) ->
 
 -spec flatten_jobj(cb_context:context()) -> iolist().
 flatten_jobj(Context) ->
-    JObj = kz_json:flatten(cb_context:resp_data(Context)),
+    JObj = kz_json:flatten(cb_context:resp_data(Context), binary_join),
     Routines = [fun check_integrity/1
                ,fun create_csv_header/1
                ,fun json_objs_to_csv/1
--- a/core/kazoo_json/src/kz_json.erl
+++ b/core/kazoo_json/src/kz_json.erl
@@ -91,7 +91,7 @@
 -export([decode/1, decode/2]).
 -export([unsafe_decode/1, unsafe_decode/2]).
 
--export([flatten/1]).
+-export([flatten/1, flatten/2, expand/1, diff/2]).
 
 -export([sum/2, sum/3]).
 -export_type([sumer/0]).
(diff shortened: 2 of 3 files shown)


fenollp:
Please add tests in core/kazoo_json/test/kz_json_test.erl

fenollp:
Coverage report is actually pretty useful https://coveralls.io/builds/10011151/source?filename=core%2Fkazoo_json%2Fsrc%2Fkz_json.erl#L1163

jamhed:
@fenollp, how is it useful? It's ignores test suite.

lazedo:
@jamhed what other options are there for flatten/2 ? you should handle the default there and change flatten/1 to call flatten/2 with default.
also, why the change in api_resource ?

jamhed:
@lazedo, there are only one option for flatten/2, binary_keys denoting join keys as binaries with "_" as separator.

> you should handle the default there

No, i shouldn't. flatten/1 is generic, flatten/2 is specific.

> why the change in api_resource ?

Because api_resource uses specific version of flatten, flatten/2 namely.

lazedo:
@jamhed api_resource does not use specific flatten until now, why the change ?
change flatten/1 to call flatten/2. handle default in flatten/2.

jamhed:
@lazedo, in fact api_resource uses specific version of flatten (namely the way keys are joined is specific). There is no way to deduce generic version of flatten from specific version of flatten as information is lost in join operation, therefore the change you're requesting is misleading: generic version should not be presented as more specific.

lazedo:
@jamhed you are changing behaviour in api_resource, right ? why this change ?

applications/crossbar/src/api_resource.erl
```diff
@@ -876,7 +876,7 @@ accept_override(Context) ->
 
 -spec flatten_jobj(cb_context:context()) -> iolist().
 flatten_jobj(Context) ->
-    JObj = kz_json:flatten(cb_context:resp_data(Context)),
+    JObj = kz_json:flatten(cb_context:resp_data(Context), binary_join),
```
lazedo: why this change ?

core/kazoo_json/src/kz_json.erl
```diff
@@ -1159,20 +1159,44 @@ is_private_key(<<"_", _/binary>>) -> 'true';
...
+-type composite_key() :: binary() | [binary()].
+
 -spec flatten(object() | objects()) -> object() | objects().
 flatten(L) when is_list(L) -> [ flatten(JObj) || JObj <- L ];
 flatten(?JSON_WRAPPER(L)) when is_list(L) ->
     from_list(lists:flatten([ flatten_key(K,V) || {K,V} <- L ])).
 
--spec join_keys(binary(), binary()) -> binary().
-join_keys(K1, K2) when is_binary(K1), is_binary(K2) -> <<K1/binary, "_", K2/binary>>.
+-spec flatten(object() | objects(), binary_join) -> object() | objects().
```
lazedo: handle the generic case from flatten/1 here. change flatten/1 to call flatten/2

jamhed:
@lazedo, because api_resource uses specific version of flatten, and there were no generic version of flatten prior to this commit. Therefore api_resource is changed to reflect this.

applications/crossbar/src/api_resource.erl
jamhed: because api_resource uses specific version of flatten,

lazedo:
@jamhed you're still not saying why you want this change ? and flatten/1 => flatten/2 with default is the way we do for expanding arities. no options arity calls options arity with default

applications/crossbar/src/api_resource.erl
lazedo: it doesn't until now. why this change ?

jamhed:
@lazedo, flatten/1 is generic, flatten/2 is specific, there are no defaults for both of them. what are you trying to achieve?

applications/crossbar/src/api_resource.erl
jamhed: because this use is incompatible with generic/specific versions of flatten.

lazedo:
@jamhed if flatten/2 is specific then it should be renamed. flatten should have one meaning not 2 meanings.
what is it that your not understanding in the change request?
flatten(X) -> flatten(X, default).
flatten(X, binary_join) -> ...
flatten(X, default) -> ....

still need to understand why you want this change ?
note: diff & expand are ok

applications/crossbar/src/api_resource.erl
lazedo: where's the incompatibility ? where else is flatten called ?

jamhed: see pr. the use is in diff.

jamhed:
flatten(X, default) -> .... == flatten/1
flatten(X, whatever) = whatever(flatten/1).

if you don't like specificity by argument surely flatten/2 could be renamed to flatten_to_binary_keys/1, with the same semantic.

core/kazoo_json/src/kz_json.erl
```diff
@@ -1202,6 +1202,7 @@ join_keys(K1, K2) when is_list(K1) -> K1 ++ [K2];
 join_keys(K1, K2) -> [K1, K2].
 
 -spec flatten_key(composite_key(), any()) -> [{composite_key(), any()}].
+flatten_key(K, {[]}) -> [ {K, {[]}} ];
```
lazedo: use `?JSON_WRAPPER`

core/kazoo_json/test/kz_json_test.erl
```diff
@@ -558,6 +558,7 @@ sum_test_() ->
     ,?_assertEqual(?CHARGES_DOUBLE, kz_json:sum(?CHARGES_SIMPLE, ?CHARGES_SIMPLE))
     ].
 
+<<<<<<< HEAD
```
lazedo: rebase problem

lazedo:
@jamhed the requested changes were not implemented as discussed. accepting this as an exception as other work needs to be committed.

PR #3224, opened 2017-02-06
cb_configs reworked

jamhed:
- http get returns actual account config (as a merge of account, reseller and system), or new empty
- http post/put/patch validates merged config against schema and stores diff document to account
- merge is done recursively in this order: account, reseller, system, new empty

### DIFF
--- a/applications/crossbar/src/modules/cb_configs.erl
+++ b/applications/crossbar/src/modules/cb_configs.erl
@@ -1,9 +1,6 @@
 %%%-------------------------------------------------------------------
 %%% @copyright (C) 2011-2017, 2600Hz INC
 %%% @doc
-%%%
-%%% Listing of all expected v1 callbacks
-%%%
 %%% @end
 %%% @contributors:
 %%%   Karl Anderson
@@ -12,27 +9,20 @@
 -module(cb_configs).
 
 -export([init/0
-        ,allowed_methods/1
-        ,resource_exists/0, resource_exists/1
-        ,validate/2
-        ,post/2
-        ,delete/2
+        ,allowed_methods/1, allowed_methods/2
+        ,resource_exists/0, resource_exists/1, resource_exists/2
+        ,validate/2, validate/3
+        ,post/2, post/3
+        ,delete/2, delete/3
         ]).
 
--export([validate_schema/2]).
-
 -include("crossbar.hrl").
+-define(DEFAULT_NODE, <<"default">>).
 
 %%%===================================================================
 %%% API
 %%%===================================================================
 
-%%--------------------------------------------------------------------
-%% @public
-%% @doc
-%% Initializes the bindings this module will respond to.
-%% @end
-%%--------------------------------------------------------------------
 -spec init() -> 'ok'.
 init() ->
     _ = crossbar_bindings:bind(<<"*.allowed_methods.configs">>, ?MODULE, 'allowed_methods'),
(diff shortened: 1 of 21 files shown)


applications/crossbar/src/modules/cb_configs.erl
```diff
@@ -15,198 +12,164 @@
...
-    OnSuccess = fun(C) -> crossbar_doc:load_merge(Id, C, ?TYPE_CHECK_OPTION(<<"account_config">>)) end,
-    cb_context:validate_request_data(make_schema_name(Config), Context, OnSuccess).
+-spec save_diff(cb_context:context(), path_token(), kz_json:object()) -> cb_context:context().
+save_diff(Context, ConfigId, Parent) ->
+    JObj = cb_context:doc(Context),
+    JObjDiff = kz_json:diff(JObj, Parent),
+    StoredDocument = kz_json:private_fields(kapps_account_config:get(cb_context:account_id(Context), ConfigId)),
+    Document = kz_json:merge_recursive(StoredDocument, JObjDiff),
+    flush(crossbar_doc:save(Context, Document, [])),
+    crossbar_doc:handle_datamgr_success(set_id(ConfigId, JObj), Context).
```
lazedo: why use save/3 and not save/2 ?
also `flush` seems redundant as `kapps_account_config` would auto invalidate the cache when the doc is saved.
using it here invalidates every config document and not only the one being saved here.

applications/crossbar/src/modules/cb_configs.erl
```diff
@@ -15,198 +12,164 @@
...
-%% @public
-%% @doc
-%% If the HTTP verib is PATCH, execute the actual action, usually a db save
-%% (after a merge).
-%% @end
-%%--------------------------------------------------------------------
--spec patch(cb_context:context(), path_token()) -> cb_context:context().
-patch(Context, _) ->
-    crossbar_doc:save(Context).
-
```
lazedo: we need support for `patch`

applications/crossbar/src/modules/cb_configs.erl
```diff
@@ -15,198 +12,164 @@
...
-%%--------------------------------------------------------------------
-%% @public
-%% @doc
-%% If the HTTP verib is DELETE, execute the actual action, usually a db delete
-%% @end
-%%--------------------------------------------------------------------
+post(Context, ConfigId) ->
+    Parent = kapps_config_util:get_reseller_config(cb_context:account_id(Context), ConfigId),
+    save_diff(Context, ConfigId, Parent).
```
lazedo: if you set the context doc with diff as part of validation then you can just do crossbar_doc:save/2

applications/crossbar/src/modules/cb_configs.erl
```diff
@@ -15,198 +12,164 @@
...
+-spec validate_schema(api_binary(), kz_json:object()) -> no_schema_present | valid | {invalid_document, term()}.
+validate_schema(Name, JObj) ->
+    case kz_json_schema:load(Name) of
+        {ok, Schema} ->
+            case kz_json_schema:validate(Schema, JObj) of
+                {ok, _} -> valid;
+                {error, Error} -> {invalid_document, Error}
+            end;
+        _ -> no_schema_present
+    end.
```
lazedo: why not use `context:validate_request_data/2' ?

applications/crossbar/src/modules/cb_configs.erl
```diff
@@ -15,198 +12,164 @@
...
+              _ -> set_id(ConfigId, kz_json:new())
+          end,
+    pass_validation(Context, Doc);
+
+validate(Context, ?HTTP_PUT, ConfigId) ->
+    validate(Context, ?HTTP_POST, ConfigId);
+
+validate(Context, ?HTTP_PATCH, ConfigId) ->
+    Parent = kapps_config_util:get_config(cb_context:account_id(Context), ConfigId),
+    save_diff(validate_diff(Context, ConfigId, Parent), ConfigId, Parent);
```
lazedo: performing the actual save as part of validation should not be done. implement the patch fun

applications/crossbar/src/modules/cb_configs.erl
jamhed: > using it here invalidates every config document and not only the one being saved here.

this is not true, flush/1 defined flushes only modified config.

applications/crossbar/src/modules/cb_configs.erl
jamhed: well, thing is patch is not called. i have no idea why.

applications/crossbar/src/modules/cb_configs.erl
lazedo:
```
flush(Context) ->
    kapps_account_config:flush(cb_context:account_id(Context)),
```

jamhed: > kapps_account_config would auto invalidate the cache when the doc is saved.

what do you mean by that? document is saved to db, kapps_account_config will try to get it from cache, and will get the previous version.

applications/crossbar/src/modules/cb_configs.erl
lazedo: you have the correct binding but you removed the implementation of patch

jamhed: it was there, but wasn't called after successful validation.

applications/crossbar/src/modules/cb_configs.erl
lazedo: no, when the doc is saved the cache is invalidated and will get the new version

applications/crossbar/src/modules/cb_configs.erl
lazedo: maybe because of the way it was validated. were you calling `crossbar_doc:handle_datamgr_success` like in validate for get ? check the other modules that implement patch.

jamhed: well, the validation works for post, and the exactly the same validation doesn't work for patch. how is that? next, i've checked system_configs and found the same approach there.

lazedo: check `cb_menus` or `cb_vmboxes`

applications/crossbar/src/modules/cb_configs.erl
jamhed: it doesn't allow to validate a json object, only context document, what is not the case here.

applications/crossbar/src/modules/cb_configs.erl
jamhed: during the validation complete config document is formed (e.g. with all parents merged), and this complete document is validated. but we need to save a difference to this document, therefore save_diff/3 instead of just save/2.

jamhed: diff object is saved, but complete (merged) object is returned. this requires to save both of them to context, what will complicate things. other option is to save in validate, what is semantically wrong.

PR #3227, opened 2017-02-06
KAZOO-5276: bulk writes for KNM

fenollp:
This significantly speeds up number collection operations. Not number Tasks unfortunately: they use a different API.

KNM migration may be able to use the collections API, but this would be another PR.


fenollp:
now mkbusy fails with the same issue it did before I my fix...

lazedo:
@fenollp the fail maybe related to the changes in datamgr

PR #3230, opened 2017-02-06
KAZOO-5283: add crossbar auth configuration

icehess:
Intermediate PR adding settings to kazoo_auth.

Right now it's only checking the authenticator module is enabled, if it is enabled generate token otherwise return unauthorized.


lazedo:
@icehess i would prefer to move these to crossbar_auth. let's talk about this.

PR #3236, opened 2017-02-07
Update presence on call_forward toggle via phone or crossbar

kalda341:
When a user sets call forwarding to on the BLF should show busy. 
When a user sets call forwarding to off the BLF should show available.
This is configurable on the account level, defaulting to off.

I would have preferred to have implemented this in a separate app listening to user document changes to reduce code duplication, but I don't know of a way to figure out if a field was toggled this way.

### DIFF
--- a/applications/callflow/src/cf_util.erl
+++ b/applications/callflow/src/cf_util.erl
@@ -45,6 +45,8 @@
         ,event_listener_name/2
         ]).
 
+-export([update_presence/2]).
+
 -include("callflow.hrl").
 -include_lib("kazoo_json/include/kazoo_json.hrl").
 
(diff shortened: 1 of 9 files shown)


kalda341:
Any updates?

noahmehl:
We're +1 for this as well.  Thanks!  @kalda341  Let me know if you need anything :)

grahamsnz:
@kalda341 If I understand correctly, enabled this feature in an account would allow for Call Forward to show busy BLF state. I could see this being useful for some clients. 

The scenario I imagine is someone who's stepped out for lunch and wants to discourage staff in the office to think they are available for regular calls at their desk. However, if a call does happen to reach get though to the forwarded number, they'd like to have the option of answering. 

Is this similar to the use case you imagined?

kalda341:
@grahamsnz Yes - this is the use case we imagined.

lazedo:
@kalda341 did you try to create a callflow that would chain `manual presence` after `call forward` ?

kalda341:
@lazedo This would work when call forwarding is toggled via the callflow app, but not when it is toggled through the API.

lazedo:
@kalda341 then the way to go is to allow `manual presence` to bet set from the api, and then call both endpoints. we should try not to mix features and adding dependencies between them. this will lead to maintenance nightmares.

kalda341:
Fair enough - I definitely understand that. I will add this functionality to cb_presence sometime in the next few days.

PR #3237, opened 2017-02-07
KAZOO-5335: Teletype customer_update enhancement

onnet:
As a kapp developer I would like to be able to send different updates form my app to customer(s) with the information generated in my kapp.

In order to do that I need to be able to:

- manage multiple templates for notification emails of customer_update type, e.g. "new billing period approaching", "Happy New Year", etc;
- provide these templates with data generated by my kapp during template compilation;
- use existing customer_update mechanism in order to not to create teletype modules again and again

### DIFF
--- a/applications/crossbar/doc/notifications.md
+++ b/applications/crossbar/doc/notifications.md
@@ -586,3 +586,32 @@ You can send a message with changed subject, html and plain text templates by pr
     }
 }
 ```
+
+To send an update to a customer from your kapp, you can build payload including you apps data (see <<"DataBag">> field) and send it over amqp using predefined particular template (see <<"Template-ID">> field) or your own hardcoded templates (see <<"HTML">> and <<"Text">> fields):
+
+```
+-spec send_account_update(ne_binary()) -> 'ok'.
+send_account_update(AccountId) ->
+    case kz_amqp_worker:call(build_customer_update_payload(AccountId)
+                            ,fun kapi_notifications:publish_customer_update/1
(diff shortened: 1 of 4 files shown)


lazedo:
@onnet thanks for the pr. i understand the need for this, but i don't think this is the correct approach (sending the templates to render). let's work on this to fit your (and other's) need for this.

onnet:
@lazedo Hi

If I understood you correctly, under "sending the templates to render" you mean sending <<"html">> and <<"txt">> variables in customer_update payload.
This is already existing feature. I'm personally using it for sending updates over API.
I exposed it in this docs update just as an example.

The new feature is an ability to send template_id and databag in payload.
So, having template_id and databag we can render chosen template with generated by our kapp data.

This way our payload will look like this:

    props:filter_empty(
      [{<<"Account-ID">>, kz_services:find_reseller_id(AccountId)}
      ,{<<"Recipient-ID">>, AccountId}
      ,{<<"Template-ID">>, <<"customer_update_billing_period">>}
      ,{<<"DataBag">>, {[{<<"field1">>,<<"value1">>},{<<"field2">>,{[{<<"subfield1">>,<<"subvalue1">>},{<<"subfield2">>,<<"subvalue2">>}]}}]}}
      ]).


...which looks to be lightweight enough.

Please let me know if I understood your concerns right.

Thanks!

onnet:
@lazedo Another thing to mention is attachments. 
Send a reference to a doc in Couch?

PR #3239, opened 2017-02-07
HELP-27796 check response before releasing job 4.0

lazedo:


icehess:
@lazedo I have no idea why MakBusy BlockPID test failed here.

lazedo:
@icehess all 4.0 builds are failing on this test because of the privacy pr.

PR #3241, opened 2017-02-07
Allow number port transition from submitted to scheduled

kalda341:
This is disabled by default. Can be enabled by setting "allow_port_transition_from_submitted_to_scheduled" in system_config/number_manager to true.


fenollp:
Hello there, what is the issue you're fixing here? What is your use case?

kalda341:
Pending is not a relevant number port state in New Zealand, so it would be useful to be able to skip it.

fenollp:
Interesting. Thanks for making Kazoo better for everyone :)

PR #3242, opened 2017-02-08
Allow escape from voicemail to next callflow element

kalda341:
Some customers want the ability for a caller to have the option to contact a secretary or to leave a message. Currently the only way to achieve that is to put a menu in front of the voicemail which means the caller hears two messages. This change allows a button to be pressed to skip the voicemail element (# by default).

### DIFF
--- a/applications/callflow/src/module/cf_voicemail.erl
+++ b/applications/callflow/src/module/cf_voicemail.erl
@@ -460,8 +460,7 @@ compose_voicemail(#mailbox{keys=#keys{login=Login
                         {'error', _R} -> record_voicemail(tmp_file(Ext), Box, Call)
                     end;
                 Continue ->
-                    lager:info("caller chose to continue to the next element in the callflow"),
-                    'ok';
+                    lager:info("caller chose to continue to the next element in the callflow");
                 _Else ->
                     lager:info("caller pressed unbound '~s', skip to recording new message", [_Else]),
                     record_voicemail(tmp_file(Ext), Box, Call)
@@ -550,11 +549,13 @@ record_voicemail(AttachmentName, #mailbox{max_message_length=MaxMessageLength}=B
             lager:info("error while attempting to record a new message: ~p", [_R])
     end.
 
--spec maybe_review_recording(ne_binary(), mailbox(), kapps_call:call(), ne_binary() | 'undefined', integer(), boolean()) -> 'ok'.
+-spec maybe_review_recording(ne_binary(), mailbox(), kapps_call:call(), api_ne_binary(), integer(), boolean()) -> 'ok'.
 maybe_review_recording(AttachmentName, #mailbox{}=Box, Call, _Digit, Length, _IsCallUp='false') ->
     new_message(AttachmentName, Length, Box, Call);
-maybe_review_recording(_AttachmentName, #mailbox{keys=#keys{continue=Digit}}, _Call, Digit, _Length, _IsCallUp) ->
+maybe_review_recording(AttachmentName, #mailbox{keys=#keys{continue=Digit}}=Box, Call, Digit, Length, _IsCallUp) ->
     lager:info("caller chose to continue to the next element in the callflow"),
+    new_message(AttachmentName, Length, Box, Call),
+    _ = kapps_call_command:prompt(<<"vm-saved">>, Call),
     'ok';
 maybe_review_recording(AttachmentName, #mailbox{media_extension=Ext}=Box, Call, _Digit, Length, _IsCallUp) ->
     case review_recording(AttachmentName, 'true', Box, Call) of
--- a/applications/callflow/src/module/cf_voicemail.erl
+++ b/applications/callflow/src/module/cf_voicemail.erl
@@ -131,7 +131,7 @@
               ,next = <<"6">>
               ,delete = <<"7">>
 
-              %% Greeting, instructions or leaving a voicemail message
+              %% Greeting or instructions
               ,continue = <<"#">>
          }).
 -type vm_keys() :: #keys{}.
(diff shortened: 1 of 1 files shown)


lazedo:
@kalda341 hey, thanks for this pull request (and the others). can't you achieve this by setting the skip_greeting in vm action parameters ?

kjcsb1:
The use-case is as follows: 
1. caller arrives at the voicemail box
2. message says "please leave a message or to speak with my personal assistant please press #"
3. if the caller presses # then the call proceeds to the next step in the callflow

This is common PBX functionality that our customers desired. The alternative, to put a menu in front of the voicemail element, is a hack and not easily compatible with SmartPBX whereas with this change it is easy to modify SmartPBX to allow this (just add a Next Action step to users).

The # is configurable per-account.

Kazoo also has similar functionality to dial 0 but this only allows one destination per account whereas it is common for our customers to have personal assistants for a group of people (common in law firms for example) and therefore there needs to be multiple escape destinations.

grahamsnz:
@lazedo We see this as a helpful extension of existing dial-0-for-operator feature, but with more flexible routing. 

We've had similar user requests for this feature from:
- executives with personal assistants (same example as @kjcsb1)
- property management services firms with maintenance escalation where their caller is a tenant who has encountered a voicemail box, but has a need to escalate maintenance emergency contact.
- tech support firms who have escalation procedures, similar to property management firms.

lazedo:
@kjcsb1 @grahamsnz thanks for the comments.

@kalda341 i'm ok with adding it, however i do see a problem. if i understood it correctly, the `#` is meant to skip the voicemail and go to next step.
i see the `#` being defaulted and enabled to every account. we should avoid these but i don't actually see a problem with that.
but, i believe the default play message ends with something like "...press any key or stop talking to end the recording", and the `#` is also handled after the recording, which may lead a user pressing the `#` to end the recording and the vm not being saved.

thoughts / comments ?

applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -455,6 +459,9 @@ compose_voicemail(#mailbox{keys=#keys{login=Login
                         {'ok', Flow} -> {'branch', Flow};
                         {'error', _R} -> record_voicemail(tmp_file(Ext), Box, Call)
                     end;
+                Continue ->
+                    lager:info("caller chose to continue to the next element in the callflow"),
+                    'ok';
```
icehess: no need for `'ok` here since `lager:info` result in `'ok'` anyway

applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -539,27 +544,33 @@ record_voicemail(AttachmentName, #mailbox{max_message_length=MaxMessageLength
...
-                    _ = kapps_call_command:prompt(<<"vm-saved">>, Call),
-                    {'branch', Flow}
-            end;
+            Digit = kz_json:get_value(<<"Terminator">>, Msg),
+            maybe_review_recording(AttachmentName, Box, Call, Digit, Length, IsCallUp);
         {'error', _R} ->
             lager:info("error while attempting to record a new message: ~p", [_R])
     end.
 
+-spec maybe_review_recording(ne_binary(), mailbox(), kapps_call:call(), ne_binary() | 'undefined', integer(), boolean()) -> 'ok'.
```
icehess: `ne_binary() | 'undefined'`  -> `api_ne_binary()`

applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -539,27 +544,33 @@ record_voicemail(AttachmentName, #mailbox{max_message_length=MaxMessageLength
...
+maybe_review_recording(AttachmentName, #mailbox{}=Box, Call, _Digit, Length, _IsCallUp='false') ->
+    new_message(AttachmentName, Length, Box, Call);
+maybe_review_recording(_AttachmentName, #mailbox{keys=#keys{continue=Digit}}, _Call, Digit, _Length, _IsCallUp) ->
+    lager:info("caller chose to continue to the next element in the callflow"),
+    'ok';
```
icehess: Why not saving the already recorded voicemail and then continue?
Also no need for 'ok'.

applications/callflow/src/module/cf_voicemail.erl
kalda341: I was aware of this, however decided to be explicit. Have changed according to your suggestion.

applications/callflow/src/module/cf_voicemail.erl
kalda341: Much nicer - I wasn't aware of this

applications/callflow/src/module/cf_voicemail.erl
kalda341: Have modified the code to do this.,

applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -539,27 +543,35 @@ record_voicemail(AttachmentName, #mailbox{max_message_length=MaxMessageLength
...
+            Digit = kz_json:get_value(<<"Terminator">>, Msg),
+            maybe_review_recording(AttachmentName, Box, Call, Digit, Length, IsCallUp);
         {'error', _R} ->
             lager:info("error while attempting to record a new message: ~p", [_R])
     end.
 
+-spec maybe_review_recording(ne_binary(), mailbox(), kapps_call:call(), api_ne_binary(), integer(), boolean()) -> 'ok'.
+maybe_review_recording(AttachmentName, #mailbox{}=Box, Call, _Digit, Length, _IsCallUp='false') ->
+    new_message(AttachmentName, Length, Box, Call);
+maybe_review_recording(AttachmentName, #mailbox{keys=#keys{continue=Digit}}=Box, Call, Digit, Length, _IsCallUp) ->
```
lazedo: if `#` is used to provide an exit to next callflow element before the recording it shouldn't get here as mentioned in pr comment

kalda341:
The original intent is that pressing # while listening to the voicemail greeting or while leaving a message will take the caller to the next callflow element. The voicemail message would not be stored.

I have now modified this to store the message when the # is pressed while leaving a message, but also proceed  to the next callflow element. This would mean that the prompt instructing the user to press any key to end the recording would still be correct, and our functionality work as intended.

lazedo:
@kalda341 next callflow action is called by default.

kalda341:
@lazedo My mistake. I have reverted back to the original recording and review code.

lazedo:
@kalda341 thanks for the update. 
the only objection is the default for `continue` as it introduces a change in behaviour

kalda341:
I picked # as it was the least used key, and also corresponds to exit in the main menu. I'm fine with any other default though.

lazedo:
@kalda341 you could have picked any other key, that's not the issue. the issue is that, this should be  `opt-in` by account and you're creating the default in a way that it would be working for every account and they would have to `opt-out`.

lazedo:
@kalda341 thanks for the pr

PR #3244, opened 2017-02-09
KAZOO-5320: add a system option to diable prepend & forward

icehess:


applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -97,6 +97,14 @@
                         )
        ).
 
+-define(DEFAULT_FORWARD_TYPE(AccountId)
+       ,kapps_account_config:get_global(AccountId
```
k-anderson: @icehess this should not be an account based config, as it requires the system to support it.  Lets change this to just kapps_config please ;)

fenollp:
Linking PRs https://github.com/2600hz/kazoo/pull/3278

PR #3251, opened 2017-02-10
account_config: generate account schemas

jamhed:


fenollp:
Does the source part really need to be stored in the JSON? Can't it be guessed from the module name?

jamhed:
@fenollp, what module name?

jamhed:
this will fail circle-ci due to unmerged #3223

lazedo:
@jamhed rebase to fix ci

icehess:
@jamhed what about parsing `kapps_account_config:get/2,3,4` as well?

PR #3256, opened 2017-02-12
Allow additional teletype notifications w/o modifying existing code

kalda341:
We would like to be able to send our own notifications through teletype, without modifying any existing 2600Hz code. This allows us to achieve this, by defining additional bindings and responders in the system_config.

### DIFF
--- a/applications/teletype/src/teletype_shared_listener.erl
+++ b/applications/teletype/src/teletype_shared_listener.erl
@@ -170,8 +170,8 @@
 -spec start_link() -> startlink_ret().
 start_link() ->
     gen_listener:start_link(?SERVER
-                           ,[{'bindings', ?BINDINGS}
-                            ,{'responders', ?RESPONDERS}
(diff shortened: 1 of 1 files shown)


jamesaimonetti:
I like this idea and wonder...why not just have it load all teletype templates at startup? Then it will load "default" templates but still allow others to add their own template modules as well.

That's my first impression, anyway. Make it the general loader for all templates, default and custom. Thoughts?

kalda341:
Sounds good to me! What I had in mind too, but thought it might be too drastic a change to be accepted.
Should I develop this further, or leave it with you?

kalda341:
@jamesaimonetti Any updates?

jamesaimonetti:
@kalda341 sorry for the delay, didn't get Github notifications in email! Yes, please refactor to be a general loader of templates.

kalda341:
Functionality improved in https://github.com/2600hz/kazoo/pull/3363

PR #3257, opened 2017-02-12
Include branched callflow id + group id in CCVs

kalda341:
We would like more useful information to be stored in CDRs. I have added Group-ID and Branched-Callflow-ID to the Custom-Channel-Vars.
The accessors are used by existing code that I have not yet created pull requests for.

### DIFF
--- a/applications/callflow/src/module/cf_callflow.erl
+++ b/applications/callflow/src/module/cf_callflow.erl
@@ -31,5 +31,48 @@ handle(Data, Call) ->
         {'ok', JObj} ->
             lager:info("branching to new callflow ~s", [Id]),
             Flow = kzd_callflow:flow(JObj, kz_json:new()),
-            cf_exe:branch(Flow, Call)
+            cf_exe:branch(Flow, update_call(JObj, Call))
     end.
+
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%% Add information to CCVs, making call logs more useful
+%% @end
+%%--------------------------------------------------------------------
+-spec update_call(kz_json:object(), kapps_call:call()) -> kapps_call:call().
+update_call(JObj, Call) ->
+    Updaters = [fun remove_old_values/1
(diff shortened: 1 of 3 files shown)


lazedo:
@kalda341 thanks for this pull request. please include the other accessors at the time of your other pull request. also, `monster_group_id` is not an appropriate name, the api should be client agnostic.

kalda341:
I agree that the name is poor, but I did not want it to be confused with the many other types of groups. Do you have any suggestions as to what to name it? Just `group_id`?

kalda341:
@lazedo I have renamed monster_group_id to group_id.

applications/callflow/src/module/cf_callflow.erl
```diff
@@ -31,5 +31,48 @@ handle(Data, Call) ->
...
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%% Set the group_id field in the CCVs if the callflow is a baseGroup
+%% @end
+%%--------------------------------------------------------------------
+-spec maybe_add_group_id(kz_json:object(), kapps_call:call()) -> kapps_call:call().
+maybe_add_group_id(JObj, Call) ->
+    case kz_json:get_value(<<"type">>, JObj) of
+        <<"baseGroup">> ->
```
jamesaimonetti: Where is "baseGroup" coming from? This seems very specific to your operation and not generally useful.

kalda341: In Monster UI's VOIP app a baseGroup is the callflow that contains a ring group. We would like the group_id field to be included if the callflow element is a baseGroup.
You are correct - this is only useful for callflows created with Monster.

kalda341:
@jamesaimonetti Any updates?

lazedo:
@kalda341 this approach is very tied to a specific use case `baseGroup`. we need to figure out a different way of doing this without overloading CCVs.

kalda341:
I agree. Maybe something like what you've done with custom SIP headers?

PR #3259, opened 2017-02-13
Allow user to log in or out of ring group

kalda341:
This pull request includes two changes.
The first is to allow endpoints in a ring group to log out of the group until a certain time.
The second is a callflow module that allows a user to log in and out of a ring group.

The works as follows:

1. The user sets up two callflows containing the log_in_out_ring_group element, one to log in ("login": true) and one to log out ("login": false) for a ring group. They could be on the same callflow with a menu. This callflow is linked to the ring group by the "callflow_id" field.
2. The user dials into the logout callflow. Kazoo recursively searches the callflow for ring groups which contain the user. The endpoints for the user are disabled until a long time in the future. The user is told that they are logged out.

### DIFF
--- a/applications/callflow/src/module/cf_log_in_out_ring_group.erl
+++ b/applications/callflow/src/module/cf_log_in_out_ring_group.erl
@@ -1,0 +1,91 @@
+%%%-------------------------------------------------------------------
+%%% @doc
+%%% Logs a user in or out of a ring group
+%%% @end
+%%% @contributors
+%%%   Max Lay
+%%%-------------------------------------------------------------------
+-module(cf_log_in_out_ring_group).
+
+-behaviour(gen_cf_action).
+
+-include("callflow.hrl").
+
+-export([handle/2]).
+
+-define(ON_VAL, 0).
+-define(DEFAULT_OFF_VAL, 66269664000).
+
+%%--------------------------------------------------------------------
+%% @public
+%% @doc
+%% Entry point for this module, attempts to call an endpoint as defined
+%% in the Data payload.  Returns continue if fails to connect or
+%% stop when successfull.
+%% @end
+%%--------------------------------------------------------------------
+-spec handle(kz_json:object(), kapps_call:call()) -> 'ok'.
+handle(Data, Call) ->
+    _ = kapps_call_command:answer(Call),
+    CallflowId = kz_json:get_value(<<"callflow_id">>, Data),
+    DisableUntil = case kz_json:is_true(<<"login">>, Data) of
+                       'true' -> ?ON_VAL;
+                       'false' -> ?DEFAULT_OFF_VAL
+                   end,
+    AccountDb = kapps_call:account_db(Call),
+    {'ok', Callflow} = kz_datamgr:open_doc(AccountDb, CallflowId),
+    UpdatedFlow = find_and_change_ring_group(kz_json:get_value(<<"flow">>, Callflow), Call, DisableUntil),
+    kz_datamgr:save_doc(AccountDb, kz_json:set_value(<<"flow">>, UpdatedFlow, Callflow)),
+    lager:debug("updated flow from ~p to ~p", [kz_json:encode(Callflow), kz_json:encode(UpdatedFlow)]),
+    cf_exe:continue(Call).
+
+-spec find_and_change_ring_group(kz_json:object(), kapps_call:call(), integer()) -> kz_json:object().
+find_and_change_ring_group(Module, Call, DisableUntil) ->
+    NewModule = case kz_json:get_value(<<"module">>, Module) of
+                    <<"ring_group">> -> alter_endpoints(Module, Call, DisableUntil);
+                    _ -> Module
+                end,
+    case kz_json:get_value(<<"children">>, NewModule) of
+        'undefined' -> NewModule;
+        Children -> kz_json:set_value(<<"children">>, maybe_modify_children(Children, Call, DisableUntil), NewModule)
+    end.
+
+-spec maybe_modify_children(kz_json:object(), kapps_call:call(), integer()) -> kz_json:object().
+maybe_modify_children(Children, Call, DisableUntil) ->
+    lists:foldl(fun(Key, JObj) ->
+                        SubModule = find_and_change_ring_group(kz_json:get_value(Key, JObj), Call, DisableUntil),
+                        kz_json:set_value(Key, SubModule, JObj)
+                end, Children, kz_json:get_keys(Children)).
+
+-spec alter_endpoints(kz_json:object(), kapps_call:call(), integer()) -> kz_json:object().
+alter_endpoints(Module, Call, DisableUntil) ->
+    EndpointsPath = [<<"data">>, <<"endpoints">>],
+    OldEndpoints = kz_json:get_value(EndpointsPath, Module),
+    {Found, EndPoints} = lists:foldl(fun(Endpoint, {Found, UpdatedEndpoints}) ->
+                                             {ThisFound, Res} = maybe_alter_disable_until(kz_json:get_value(<<"endpoint_type">>, Endpoint), Endpoint, Call, DisableUntil),
+                                             {Found or ThisFound, [Res | UpdatedEndpoints]}
+                                     end, {'false', []}, OldEndpoints),
+    Media = case {Found, DisableUntil} of
+                {'false', _} ->
+                    %% User not found
+                    kz_media_util:get_prompt(<<"agent-invalid_choice">>, Call);
+                {'true', ?ON_VAL} ->
+                    kz_media_util:get_prompt(<<"agent-logged_in">>, Call);
+                {'true', _} ->
+                    kz_media_util:get_prompt(<<"agent-logged_out">>, Call)
+            end,
+    kapps_call_command:play(Media, Call),
+    kz_json:set_value(EndpointsPath, EndPoints, Module).
+
+-spec maybe_alter_disable_until(ne_binary(), kz_json:object(), kapps_call:call(), integer()) -> {boolean(), kz_json:object()}.
+maybe_alter_disable_until(<<"user">>, Endpoint, Call, DisableUntil) ->
+    lager:debug("comparing owner ~p, to endpoint ~p", [kapps_call:owner_id(Call), Endpoint]),
+    case kapps_call:owner_id(Call) =:= kz_json:get_value(<<"id">>, Endpoint) of
+        'true' ->
+            {'true', kz_json:set_value([<<"disable_until">>], DisableUntil, Endpoint)};
+        'false' ->
+            {'false', Endpoint}
+    end;
+
+maybe_alter_disable_until(_EndpointType, Endpoint, _Call, _DisableUntil) ->
+    {'false', Endpoint}.
--- a/applications/callflow/src/module/cf_ring_group.erl
+++ b/applications/callflow/src/module/cf_ring_group.erl
@@ -134,8 +134,8 @@ start_builder(EndpointId, Member, Call) ->
       end
      ).
 
--spec member_active(kz_json:object()) -> boolean().
-member_active(Member) ->
+-spec is_member_active(kz_json:object()) -> boolean().
+is_member_active(Member) ->
     DisableUntil = kz_json:get_value(<<"disable_until">>, Member, 0),
     calendar:datetime_to_gregorian_seconds(calendar:universal_time()) > DisableUntil.
 
@@ -148,7 +148,7 @@ member_active(Member) ->
 -spec resolve_endpoint_ids(kz_json:object(), kapps_call:call()) -> endpoints().
 resolve_endpoint_ids(Data, Call) ->
     Members = kz_json:get_value(<<"endpoints">>, Data, []),
-    FilteredMembers = lists:filter(fun member_active/1, Members),
+    FilteredMembers = lists:filter(fun is_member_active/1, Members),
     lager:debug("filtered members of ring group ~p", [FilteredMembers]),
     ResolvedEndpoints = resolve_endpoint_ids(FilteredMembers, [], Data, Call),
 
--- a/applications/callflow/src/module/cf_ring_group.erl
+++ b/applications/callflow/src/module/cf_ring_group.erl
@@ -134,6 +134,11 @@ start_builder(EndpointId, Member, Call) ->
       end
      ).
 
+-spec member_active(kz_json:object()) -> boolean().
+member_active(Member) ->
+    DisableUntil = kz_json:get_value(<<"disable_until">>, Member, 0),
+    calendar:datetime_to_gregorian_seconds(calendar:universal_time()) > DisableUntil.
+
 -type endpoint() :: {ne_binary(), kz_json:object()}.
 -type endpoints() :: [endpoint()].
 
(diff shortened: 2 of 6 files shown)


grahamsnz:
@kalda341 I see this enhancement of ring groups being very useful. Lots of user don't want full agent/queue functionality, they simple want to be able to join and leave a ring group.

lazedo:
@kalda341 how about using `ac/dc` ?

grahamsnz:
@lazedo Not to speak on @kalda341 's behalf. However, in my experience, the use cases are very different between ring groups and ACDC (with agents and queues). 

Ring groups are a simple approach to group call distribution. This PR provides users with more control joining or leaving a ring group. This is helpful to users as they don't have to wait for an admin user to make the change for them.

lazedo:
@grahamsnz how different are they ? `group call distribution` is another way of saying `queues`. this pull request is creating another type of queue system. i would like to see what this pr adds that is not already possible with  `ac/dc`.

grahamsnz:
@lazedo By definition **ring groups** are a form of **group call distribution** in that they distribute calls to a group of phones or users. 

Although they are both forms of **group call distribution**, there are very few similarities between **ring groups** and **ACDC**.  A good example of these differences starts with **ACDC**'s use of agents and queues, of which **ring groups** has no concept. The functionality provided by agents and queues allows for a true call centre. Whereas, **ring groups** is far to simplistic and casual to ever be used for a call centre.

remcovanvugt:
@lazedo This allows for static configuration of ring groups (instead of logging on/logging off and managing queues) with the option of parallel calls (i.e. call waiting functionality) but still allow users to opt-out of a ring group for a give time. It goes nice with SmartPBX (as opposed to acdc). As a user, if you are busy and not willing to accept group calls, but still want to be available for transferred calls going DND is not an option. That is the best solution in this use case imo.

noahmehl:
@lazedo I'm +1 on this feature.  The difference between a queue and a ring group is that of simplicity and immediacy.  I have a client that would love this.  They have a front desk with two desk phones and a cordless phone.  Right now, the cordless phone rings everytime someone calls, even though it's right next to the desk phone.  It would be useful for them to be able to dial something to log into the distribution list when they step away from the desk, etc.  They don't want queues with hold times, etc, they just want to dynamically update which stations ring when a call comes in (from a device).  Make sense?

kjcsb1:
@lazedo ac/dc is complex, broken, and not supported by 2600hz (although there are some fixes being submitted that will improve it). cubicle is commercial. This change is part of a number of enhancements to groups that provide some of the basic functionality required by small call centres who don't need the functionality of ac/dc or cubicle.

applications/callflow/src/module/cf_log_in_out_ring_group.erl
```diff
@@ -0,0 +1,91 @@
...
+-module(cf_log_in_out_ring_group).
+
+-behaviour(gen_cf_action).
+
+-include("callflow.hrl").
+
+-export([handle/2]).
+
+-define(ON_VAL, 0).
+-define(DEFAULT_OFF_VAL, 66269664000).
```
jamesaimonetti: This is an interesting number... :)

applications/callflow/src/module/cf_log_in_out_ring_group.erl
```diff
@@ -0,0 +1,91 @@
+%%%-------------------------------------------------------------------
+%%% @doc
+%%% Logs a user in or out of a ring group
+%%% @end
+%%% @contributors
+%%%   Max Lay
+%%%-------------------------------------------------------------------
+-module(cf_log_in_out_ring_group).
```
jamesaimonetti: Naming is hard! `"module":"log_in_out_ring_group"` just seems...verbose. What about `cf_ring_group_toggle`?

applications/callflow/src/module/cf_ring_group.erl
```diff
@@ -134,6 +134,11 @@ start_builder(EndpointId, Member, Call) ->
       end
      ).
 
+-spec member_active(kz_json:object()) -> boolean().
+member_active(Member) ->
```
jamesaimonetti: Generally functions that return booleans are prefixed with `is_` or `should_`. In this case, `is_member_active/1`.

applications/callflow/src/module/cf_ring_group.erl
```diff
@@ -134,6 +134,11 @@ start_builder(EndpointId, Member, Call) ->
...
+    DisableUntil = kz_json:get_value(<<"disable_until">>, Member, 0),
+    calendar:datetime_to_gregorian_seconds(calendar:universal_time()) > DisableUntil.
```
jamesaimonetti: Does this take into account the user's timezone? How is `disable_until` set?

applications/callflow/src/module/cf_log_in_out_ring_group.erl
kalda341: cf_ring_group_toggle is definitely a better name.

applications/callflow/src/module/cf_log_in_out_ring_group.erl
```diff
@@ -0,0 +1,91 @@
...
+                    _ -> Module
+                end,
+    case kz_json:get_value(<<"children">>, NewModule) of
+        'undefined' -> NewModule;
+        Children -> kz_json:set_value(<<"children">>, maybe_modify_children(Children, Call, DisableUntil), NewModule)
+    end.
+
+-spec maybe_modify_children(kz_json:object(), kapps_call:call(), integer()) -> kz_json:object().
+maybe_modify_children(Children, Call, DisableUntil) ->
+    lists:foldl(fun(Key, JObj) ->
```
jamesaimonetti: Why not
```erlang
kz_json:foldl(fun(Key, Value, JObj) ->
    SubModule = find_and_change_ring_group(Value, Call, DisableUntil),
    kz_json:set_value(Key, SubModule, JObj)
end
,Children
,Children
)
```

applications/callflow/src/module/cf_log_in_out_ring_group.erl
kalda341: It corresponds to 2100 00:00:00, however any large number would work.

applications/callflow/src/module/cf_ring_group.erl
kalda341: Thanks for the tip!

applications/callflow/src/module/cf_log_in_out_ring_group.erl
```diff
@@ -0,0 +1,91 @@
...
+-spec handle(kz_json:object(), kapps_call:call()) -> 'ok'.
+handle(Data, Call) ->
+    _ = kapps_call_command:answer(Call),
+    CallflowId = kz_json:get_value(<<"callflow_id">>, Data),
+    DisableUntil = case kz_json:is_true(<<"login">>, Data) of
+                       'true' -> ?ON_VAL;
+                       'false' -> ?DEFAULT_OFF_VAL
+                   end,
+    AccountDb = kapps_call:account_db(Call),
+    {'ok', Callflow} = kz_datamgr:open_doc(AccountDb, CallflowId),
```
jamesaimonetti: `kz_datamgr:open_cache_doc/2` should be used unless there's a real need.

applications/callflow/src/module/cf_log_in_out_ring_group.erl
```diff
@@ -0,0 +1,91 @@
...
+    UpdatedFlow = find_and_change_ring_group(kz_json:get_value(<<"flow">>, Callflow), Call, DisableUntil),
+    kz_datamgr:save_doc(AccountDb, kz_json:set_value(<<"flow">>, UpdatedFlow, Callflow)),
```
jamesaimonetti: Want to handle if the save was temporarily unsuccessful? Or use `kz_datamgr:ensure_saved/2`?

applications/callflow/src/module/cf_log_in_out_ring_group.erl
```diff
@@ -0,0 +1,91 @@
...
+    lager:debug("updated flow from ~p to ~p", [kz_json:encode(Callflow), kz_json:encode(UpdatedFlow)]),
+    cf_exe:continue(Call).
+
+-spec find_and_change_ring_group(kz_json:object(), kapps_call:call(), integer()) -> kz_json:object().
+find_and_change_ring_group(Module, Call, DisableUntil) ->
```
jamesaimonetti: `Module` should really be named `Action` since its the callflow action being stored

applications/callflow/src/module/cf_log_in_out_ring_group.erl
```diff
@@ -0,0 +1,91 @@
...
+handle(Data, Call) ->
+    _ = kapps_call_command:answer(Call),
+    CallflowId = kz_json:get_value(<<"callflow_id">>, Data),
+    DisableUntil = case kz_json:is_true(<<"login">>, Data) of
+                       'true' -> ?ON_VAL;
+                       'false' -> ?DEFAULT_OFF_VAL
+                   end,
+    AccountDb = kapps_call:account_db(Call),
+    {'ok', Callflow} = kz_datamgr:open_doc(AccountDb, CallflowId),
+    UpdatedFlow = find_and_change_ring_group(kz_json:get_value(<<"flow">>, Callflow), Call, DisableUntil),
```
jamesaimonetti: Use `kzd_callflow:flow/1` and `kzd_callflow:set_flow/2`

applications/callflow/src/module/cf_ring_group.erl
kalda341: No timezone is taken into account. Should it instead use the callee's timezone?
`disable_until` is currently only used to toggle the user on and off through `log_in_out_ring_group`, however as a potential enhancement it could be used to log a user out temporarily.

applications/callflow/src/module/cf_log_in_out_ring_group.erl
kalda341: Thanks for that suggestion!

applications/callflow/src/module/cf_log_in_out_ring_group.erl
kalda341: Whoops, thanks for the improvement!

applications/callflow/src/module/cf_ring_group.erl
jamesaimonetti: Ok, I see now. Is the plan to allow users to customize how long to disable? Otherwise it seems like this should be a binary decision and not a temporal one. Thoughts?

applications/callflow/src/module/cf_log_in_out_ring_group.erl
jamesaimonetti: Also, I don't see the generated callflow schema in the commits. Please add it, along with filling out the descriptions if possible

applications/callflow/src/module/cf_log_in_out_ring_group.erl
kalda341: Will do

applications/callflow/src/module/cf_log_in_out_ring_group.erl
```diff
@@ -0,0 +1,91 @@
...
+%% @doc
+%% Entry point for this module, attempts to call an endpoint as defined
+%% in the Data payload.  Returns continue if fails to connect or
+%% stop when successfull.
+%% @end
+%%--------------------------------------------------------------------
+-spec handle(kz_json:object(), kapps_call:call()) -> 'ok'.
+handle(Data, Call) ->
+    _ = kapps_call_command:answer(Call),
+    CallflowId = kz_json:get_value(<<"callflow_id">>, Data),
```
jamesaimonetti: If you use 'kz_json:get_ne_binary_value/2` it indicates the expected type and our AST parsers can infer the type for the JSON schema and add it.

applications/callflow/src/module/cf_ring_group.erl
kalda341: Correct - currently only the API can be used to disable the endpoint for a specified time via the `disable_until` field. `log_in_out_ring_group` uses a binary on/off - but achieves this through the same field.

applications/callflow/src/module/cf_log_in_out_ring_group.erl
kalda341: I didn't want to run the risk of overwriting changes to the callflow made elsewhere, but didn't know how to handle the potential failure to save the document. Any suggestions?

applications/callflow/src/module/cf_log_in_out_ring_group.erl
kalda341: Cool - thanks!

applications/callflow/src/module/cf_log_in_out_ring_group.erl
jamesaimonetti: I guess if `{'error', 'conflict'}` is returned, you'll want/need to refetch the callflow doc and re-run `find_and_change_ring_group` again. You could then check for other error conditions and retry the save (with a max retry count).

applications/callflow/src/module/cf_log_in_out_ring_group.erl
kalda341: Agreed.

lazedo:
the arguments here are in the route to provide one way or the other a simple call queue mechanism, i'm not against, but what i'm trying to avoid here is duplication of code and making sure that we're are not creating code that duplicates already existing features.

as commented by @k-anderson earlier today, there seems to nothing in this pr that couldn't be done by using hotdesking.

kjcsb1:
@lazedo @k-anderson Hotdesking provides an association between a device and a user and is boolean. This PR allows a user to log in/out of a group - one user may be a member of many groups but perhaps only needs to be logged in (or out) of one group for a period of time. It also allows the possibility of a wrap-up time (since the active member is determined based on time). Hope that makes sense.

lazedo:
@kjcsb1 sorry, it seems you are confusing things a bit. hotdesking allows an arbitrary extension to have a log in/out from a user, ie, associate an endpoint id with an extension.
when you write about timed logouts and wrap-up timers, you're talking about queues, right ?

kalda341:
I have made the discussed changes. I also changed the `login` field to be called `action` (which can be `login` or `logout`). This will allow additional functionality such as timed logouts to be implemented in the future.

kjcsb1:
@lazedo Hotdesking does not duplicate the functionality in this PR. 
1. Hotdesking is either logged in or logged out (boolean). This PR allows for a time to be set in the future where the user will become an active member of a group.
2. This PR allows one user to be associated with some groups and not others, in an ad hoc way. For example I may be logged into the sales group and logged out of the support group. That is not possible with hotdesking. To achieve something similar I would need to have a sales device and a separate support device which is not a practical solution for most users.

jamesaimonetti:
@kalda341 fix the formatting issue that circle complains about (look at the `make fmt` target to see how to run this locally if you need - specify the file you want to format) and I'll give the PR a test.

jamesaimonetti:
Also, please add a doc in `applications/callflow/doc` for this module so we can start getting those fleshed out too. I'll be adding a similar builder as I did for Crossbar to stub out all the callflow actions but certainly any new actions should come with docs by default.

And thanks for persevering through this!

kalda341:
@jamesaimonetti I've added the documentation. What's next?

kalda341:
@jamesaimonetti Any updates?

jamesaimonetti:
@kalda341 I manually rebased/merged the PR (got approval from on high for it). Thanks for sticking with it while everyone got on the same page.

@grahamsnz @remcovanvugt @noahmehl @kjcsb1 thanks for your input here too everyone!

PR #3260, opened 2017-02-13
Tst5276

fenollp:
do not merge me


fenollp:
same idea as https://github.com/2600hz/kazoo/pull/3245 to solve makebusy failure of https://github.com/2600hz/kazoo/pull/3227

PR #3262, opened 2017-02-13
KAZOO-5344: Fix - Vitelity e911 API calls response cannot be parsed

sergey-safarov:


fenollp:
Hi! thanks for the patch!
I'm seeing a trend now https://github.com/2600hz/kazoo/pull/3181
Could you take a look at Vitelity's CNAM provider too?

I looked deep in the history, looks like some behavior in kz_http changed (binary from string). Lots of things may be broken...

sergey-safarov:
Hello @fenollp
I tested CNAM - issue also exists.
Pull request now is updated.

fenollp:
Thanks!

PR #3264, opened 2017-02-13
Tst5276 2

fenollp:
same as https://github.com/2600hz/kazoo/pull/3260


fenollp:
Looks like issue is in KNM, not in kazoo_couch nor kazoo_data

icehess:
@fenollp add description and write better title for the PR

PR #3266, opened 2017-02-13
KAZOO-5341_4oh: fix KNM's is_dirty

fenollp:
cc https://github.com/2600hz/kazoo/pull/3263


k-anderson:
@fenollp umm, using functions that don't exist on 4.0...

PR #3268, opened 2017-02-13
KAZOO-5344_4oh: Fix - Vitelity CNAM and e911 API response handling

fenollp:
cc https://github.com/2600hz/kazoo/pull/3262


k-anderson:
@fenollp no master version?

fenollp:
@k-anderson see 1st comment

PR #3274, opened 2017-02-14
KAZOO-5347: ensure dst gets written properly

fenollp:


core/kazoo_config/src/kapps_config.erl
```diff
@@ -853,7 +853,7 @@ migrate_config_setting(UpdatedFrom, Removed, To) ->
     case add_config_setting(To, Removed) of
         {'ok', UpdatedTo} ->
             {'ok', _} = kz_datamgr:save_doc(?KZ_CONFIG_DB, UpdatedTo),
-            {'ok', _} = kz_datamgr:save_doc(?KZ_CONFIG_DB, UpdatedFrom),
+            {'ok', _} = kz_datamgr:ensure_saved(?KZ_CONFIG_DB, UpdatedFrom),
```
lazedo: no, if someone updated before, we are overriding it.

fenollp: my testing says the opposite :7
it removed old field, kept the right one

lazedo: what tests did you make ?
client 1 -> get
client 2 -> get
client 1 -> set
client 2 -> set
value set from client 1 is lost

fenollp: what are you talking about? It's the config migration code.
Anyway, I'll implement the process I described on slack

lazedo: ah! you're right. i freak out every time i see `ensure_save`

PR #3275, opened 2017-02-14
KAZOO-5347_4oh: ensure dst gets written properly

fenollp:
master: https://github.com/2600hz/kazoo/pull/3274


lazedo:
@fenollp `ensure_saved` is not safe to use

PR #3277, opened 2017-02-15
[master] kvm copy leftover add missing spec and a generic way to extract media id

icehess:


fenollp:
cc https://github.com/2600hz/kazoo/pull/3276

PR #3279, opened 2017-02-15
HELP-27791: the ts worker might be checked in then have stepswitch se…

k-anderson:
…nd a targeted message for a previous call id


fenollp:
Please explicitly link your sibling PRs, like this: https://github.com/2600hz/kazoo/pull/3280

PR #3284, opened 2017-02-15
HELP-28223: fix header map duplicating headers

k-anderson:


fenollp:
cc https://github.com/2600hz/kazoo/pull/3283

PR #3287, opened 2017-02-15
Install guide cleanup

jamesaimonetti:
The intent is to get Kazoo cloned and built on a local machine.

### DIFF
--- a/doc/.org/installation.org
+++ b/doc/.org/installation.org
@@ -13,22 +13,10 @@
 #+CREATOR: Emacs 25.1.91.1 (Org mode 9.0.4)
 
 * Kazoo Installation Guide
-  :PROPERTIES:
-  :CUSTOM_ID: kazoo-installation-guide
-  :END:
-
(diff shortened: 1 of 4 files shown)


doc/.org/installation.org
```diff
@@ -0,0 +1,116 @@
...
+** Dependencies
+*** Packages Required
+#+BEGIN_SRC shell
+sudo apt-get install build-essential libxslt-dev \
+     zip unzip expat zlib1g-dev libssl-dev curl \
+     libncurses5-dev git-core libexpat1-dev \
+     htmldoc
+#+END_SRC
+
+Note: =htmldoc= is required only if [[./announcements.md#company-directory-pdf][you want to be able to download PDFs]].
```
fenollp: dead link

fenollp:
So you removed most of the text, and the rest you copied to an .org file.
But this org file is not used in the docs website, only the markdown version is?

What is going on here?

PR #3295, opened 2017-02-17
KAZOO-5351_4oh: flush cache while AMQP publishes change

fenollp:
master: https://github.com/2600hz/kazoo/pull/3294


lazedo:
don't merge yet

PR #3296, opened 2017-02-17
KAZOO-5222 include node, version and timestamp in crossbar response

lazedo:


applications/crossbar/src/api_util.erl
```diff
@@ -1266,13 +1266,19 @@ do_create_resp_envelope(Context) ->
                    [{<<"auth_token">>, cb_context:auth_token(Context)}
                    ,{<<"status">>, <<"success">>}
                    ,{<<"request_id">>, cb_context:req_id(Context)}
-                   ,{<<"revision">>, kz_term:to_binary(cb_context:resp_etag(Context))}
+                   ,{<<"node">>, kz_term:to_binary(node())}
```
jamesaimonetti: Are we worried about leaking infrastructure details here? Like, should people really know which API server processed the request when there could be a bunch sitting behind a load balancer? Kind of like we don't necessarily want to expose FreeSWITCH details when behind Kamailio?

lazedo:
don't merge this yet

core/kazoo_globals/src/kz_nodes.erl
```diff
@@ -772,14 +802,13 @@ whapp_from_json(Key, JObj) ->
...
-        V ->
-            #whapp_info{startup=V}
-    end.
+    whapp_info_from_json(#whapp_info{}, kz_json:get_values(JObj)).
+
+whapp_info_from_json(Info, {[], []}) -> Info;
+whapp_info_from_json(Info, {[V | V1], [<<"Roles">> | K1]}) ->
+    whapp_info_from_json(Info#whapp_info{roles=V}, {V1, K1});
+whapp_info_from_json(Info, {[V | V1], [<<"Startup">> | K1]}) ->
+    whapp_info_from_json(Info#whapp_info{startup=V}, {V1, K1}).
```
mark2600: seeing a lot of 'whapp' stuff here, curious why we don't change that ?

core/kazoo_globals/src/kz_nodes.erl
```diff
@@ -78,11 +83,16 @@
 -define(SIMPLE_ROW_STR, ?HEADER_COL ": ~s~n").
 -define(SIMPLE_ROW_NUM, ?HEADER_COL ": ~B~n").
 
+-define(NODE_GEN_SECRET, kz_binary:rand_hex(16)).
```
mark2600: This is not useful at all really, if we have to fallback to the default and its random we will have no way of knowing what the correlation is.  We should change this to something that we can determine at a later point in time if we need to.

core/kazoo_globals/src/kz_nodes.erl
lazedo: well, pr is welcomed

core/kazoo_globals/src/kz_nodes.erl
lazedo: why do you want to do that ? this is a way to provide immediate correlation not time basis.

mark2600: Perhaps I should re-phrase :)

I would think that something where we can determine the node in question after the fact without additional information would be better i.e md5(node())

lazedo: that makes sense :)

PR #3297, opened 2017-02-17
HELP-28312: include active port in numbers by default

k-anderson:


fenollp:
master: https://github.com/2600hz/kazoo/pull/3298

PR #3300, opened 2017-02-20
KAZOO-5355: KNM fix set_created

fenollp:


core/kazoo_number_manager/src/knm_phone_number.erl
```diff
@@ -1241,10 +1242,12 @@ created(#knm_phone_number{created = undefined}) -> kz_time:current_tstamp();
 created(#knm_phone_number{created = Created}) -> Created.
 
 -spec set_created(knm_phone_number(), gregorian_seconds()) -> knm_phone_number().
+set_created(PN=#knm_phone_number{created = undefined}, undefined) ->
```
k-anderson: @fenollp these are stored as the string undefined, should have two function clauses here... one for the atom and one for the string.

PR #3304, opened 2017-02-20
Add UML diagram to document porting states flow

azefiel:


lazedo:
@azefiel hey, is this link permanent ? maybe we should import it @jamesaimonetti ?

azefiel:
@lazedo: it should be, as long as the image hosting service exists.

If you guys come up with an internal hosting solution, let me know. We have a bunch of images for `monster.ui` helpers in the UI documentation that are currently hosted using the same service.

PR #3306, opened 2017-02-21
Add Billing-Number custom sip header

kalda341:
We require this field to bill our calls. This allows us to report calls by extension at the carrier side.

### DIFF
--- a/applications/callflow/src/cf_route_win.erl
+++ b/applications/callflow/src/cf_route_win.erl
@@ -269,6 +269,7 @@ update_ccvs(Call) ->
               [{<<"Hold-Media">>, kz_attributes:moh_attributes(<<"media_id">>, Call)}
               ,{<<"Caller-ID-Name">>, CIDName}
               ,{<<"Caller-ID-Number">>, CIDNumber}
+              ,{<<"X-Billing-Number">>, CIDNumber}
                | get_incoming_security(Call)
               ]),
     kapps_call:set_custom_channel_vars(Props, Call).
--- a/applications/ecallmgr/src/ecallmgr.hrl
+++ b/applications/ecallmgr/src/ecallmgr.hrl
@@ -303,6 +303,7 @@
                               ,{<<"Conference-Exit-Sound">>, <<"conference_exit_sound">>}
                               ,{<<"SIP-Refer-To">>, <<"sip_refer_to">>}
                               ,{<<"SIP-Referred-By">>, <<"sip_h_Referred-By">>}
+                              ,{<<"X-Billing-Number">>, <<"sip_h_X-Billing-Number">>}
                               ]).
 
 %% [{FreeSWITCH-App-Name, Kazoo-App-Name}]


jamesaimonetti:
I wonder if we can instead make a `cf_custom_sip_headers` action that would allow you to insert SIP headers like this with marcos for the values. So this particular one could be:

```json
{"module":"custom_sip_headers",
 "data":{"X-Billing-Number":"{caller_id_number}"},
 "children":{...}
}
```

Then you could create a preflow that gets run all the time, or per-callflow if you want to run it only when going to carriers, etc. Thoughts about that approach?

lazedo:
the first problem i see with this is that it will be filtered by kamailio.

jamesaimonetti:
But if its going FreeSWITCH->Carrier without proxying through Kamailio, it should be okay, right @lazedo ?

lazedo:
true for now but, with `BYOC` and `topology hiding` it will soon go thru kamailio. but i like your idea. also, maybe another use case for outbound formatters at the carrier document ?

grahamsnz:
This enhancement would be a big help for for the same reasons that @kalda341 mentioned. Additionally, the `"X-Billing-Number":` SIP header would allow our external SIP auditing tools to provide much more accurate reports.

@jamesaimonetti @lazedo Would it be possible to change Kamailio behaviours to preserve this and other customer SIP headers? Would outbound carrier document formatters allow for specific customer SIP headers to be preserved?

lazedo:
@grahamsnz as is now ( `X-Billing-Number` ) , i don't think it will be included. but in a larger change set that would serve everyone's interest with flexible options like the one @jamesaimonetti mentioned and/or `formatters` i think it can be integrated.

grahamsnz:
@lazedo Thanks for these details. This discussion is helping us understand your design goals for `BYOC`.  These more flexible carrier formatter options would be a big help when there are many resellers with many independent carriers.

In the current PR which @kalda341 has provided, this custom header would be helpful for folks who connect directly from FS nodes to their carriers. I believe the majority of the self-hosted Kazoo users operate using their own carriers as global resources in this fashion. 

It would be great to include this PR for those of us who would benefit from it. Can we also continue with the more versatile carrier formatters that @jamesaimonetti suggested? Should a Jira ticket be opened for this? Can we help?

jamesaimonetti:
What I think needs to happen is:
1. `cf_resources` sets `custom_sip_headers` in the `data` object with the key/value pairs:
  a. Use stand-in values like `{caller_id_number}` to be replaced at runtime with the value(s) from the context of the call
2. `stepswitch` runs formatters on SIP headers in the offnet AMQP payload (set in `cf_resources` in this case) and from the resource document (merged to prefer resource's version).
  a. Add support for a whitelist of macros in the [formatter properties](https://docs.2600hz.com/dev/core/kazoo_endpoint/doc/formatters/)

I think 2a is the work to be done. Looking at the kz_formatters code, it would need to include the Call record for more context for substitutions, and detect when the `Value` is a macro - if it is a macro, replace the macro with the runtime value *then* run the formatter(s); otherwise run the formatters.

Does this make sense? I think that would support the Kazoo side getting the headers to FreeSWITCH (see `ecallmgr_fs_xml:sip_headers_fold/3`).

kalda341:
@jamesaimonetti I didn't think of something like that, but would be an awesome addition to Kazoo.

jamesaimonetti:
@kalda341 @grahamsnz we talked about this on our daily core chat and think we have a way forward.

We'll create a whitelist of macros, like `{caller_id_name}`, `{caller_id_number}`, `{account_id}`, `{reseller_id}`, that can be used in custom_sip_headers as the value for the header.

Under the hood, we'll take care of replacing these values with the call-specific values and sent to FreeSWITCH on the bridge string. In the future, we'll be able to make sure these headers are respected by Kamailio too.

I'm hoping to get a PR up today with an implementation that can hopefully make it to 4.1

jamesaimonetti:
[Ticket here](https://2600hz.atlassian.net/browse/KAZOO-5382) for work

jamesaimonetti:
@kalda341 @grahamsnz see if https://github.com/2600hz/kazoo/pull/3380/ makes sense and test if you can; make sure it works for you!

grahamsnz:
Thank you @jamesaimonetti :) 

Looks perfect for my use case. 

I'll let @kalda341 comment on his use case.

kalda341:
Sounds great to me. Closing ticket in favour of this approach.

PR #3312, opened 2017-02-21
Merged config document for account with system defaults.

jamhed:
merged config for account (account/reseller/system/system_default)

PR #3320, opened 2017-02-22
cb_configs using cb_context, crossbar_doc

jamhed:


jamhed:
This will fail tests and ci as it depends on #3319, #3318, #3312

applications/crossbar/src/modules/cb_configs.erl
```diff
@@ -15,198 +12,111 @@
...
+
+validate(Context, ?HTTP_PATCH, ConfigId) ->
+    Parent = kapps_config_util:get_config(cb_context:account_id(Context), ConfigId),
+    validate_with_parent(Context, ConfigId, Parent);
+
+validate(Context, ?HTTP_POST, ConfigId) ->
+    Parent = kapps_config_util:get_reseller_config(cb_context:account_id(Context), ConfigId),
+    validate_with_parent(Context, ConfigId, Parent).
+
+validate_with_parent(Context, ConfigId, Parent) ->
```
k-anderson: @jamhed missing spec

k-anderson:
@jamhed missing documentation as well :)

applications/crossbar/src/modules/cb_configs.erl
```diff
@@ -15,198 +12,112 @@
...
+    Stored = kz_json:private_fields(kapps_account_config:get(cb_context:account_id(Context), ConfigId)),
+    case {cb_context:req_data(Context), kz_doc:revision(Stored)} of
+        {?EMPTY_JSON_OBJECT, undefined} -> skip;
+        {?EMPTY_JSON_OBJECT, _} ->
+            crossbar_doc:delete(cb_context:set_doc(Context, Stored));
+        {Diff, _} ->
+            crossbar_doc:save(Context, kz_json:merge_recursive(Stored, Diff), [])
+    end,
+    set_config_to_context(ConfigId, Context).
+
```
lazedo: can we move this common handling to a new fun (save, save_config, or other name) and call that fun from here, from put and from patch.

PR #3323, opened 2017-02-23
Send fax failure notification in more circumstances

kalda341:
Currently no failure message is sent by teletype if a fax fails for the following reasons:
- sender not authorised
- no attachment
- the number is invalid

In each case the sender should be notified by return email to explain why their fax has failed.

### DIFF
--- a/applications/fax/src/fax_smtp.erl
+++ b/applications/fax/src/fax_smtp.erl
@@ -280,6 +280,7 @@ maybe_system_report(#state{faxbox='undefined', account_id='undefined'}=State) ->
         'false' -> 'ok'
     end;
 maybe_system_report(State) ->
+    send_outbound_smtp_fax_error(State),
     case kapps_config:get(?CONFIG_CAT, <<"report_faxbox_system_errors">>, 'true') of
         'true' -> system_report(State);
         'false' -> 'ok'
@@ -891,3 +892,14 @@ write_tmp_file(Filename, Extension, Body) ->
             lager:debug("error writing file ~s : ~p", [Filename, Error]),
             Error
     end.
+
+send_outbound_smtp_fax_error(State) ->
+    Message = props:filter_empty(
+                [{<<"Account-ID">>, State#state.account_id}
+                ,{<<"Fax-From-Email">>, State#state.from}
+                ,{<<"Fax-To-Email">>, State#state.to}
+                ,{<<"Errors">>, State#state.errors}
+                 | kz_api:default_headers(?APP_NAME, ?APP_VERSION)
+                ]),
+    %% Do not crash if fields were undefined
+    catch kapi_notifications:publish_fax_outbound_smtp_error(Message).
--- a/applications/fax/src/fax_smtp.erl
+++ b/applications/fax/src/fax_smtp.erl
@@ -39,7 +39,7 @@
                ,from :: api_ne_binary()
                ,to :: api_ne_binary()
                ,doc :: kz_json:object()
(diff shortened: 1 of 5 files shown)


applications/teletype/src/templates/teletype_fax_outbound_smtp_error_to_email.erl
```diff
@@ -0,0 +1,98 @@
...
+
+-define(TEMPLATE_ID, <<"fax_outbound_smtp_error_to_email">>).
+-define(MOD_CONFIG_CAT, <<(?NOTIFY_CONFIG_CAT)/binary, ".", (?TEMPLATE_ID)/binary>>).
+-define(FAX_CONFIG_CAT, <<(?NOTIFY_CONFIG_CAT)/binary, ".fax">>).
+
+-define(TEMPLATE_MACROS, kz_json:from_list([])).
+
+-define(TEMPLATE_TEXT, <<"Error sending fax to faxbox: {{error}}">>).
+-define(TEMPLATE_HTML, <<"<html><body>Error sending fax to faxbox: {{error}}</body></html>">>).
+-define(TEMPLATE_SUBJECT, <<"Error Sending Fax">>).
```
fenollp: Maybe the default template should include the fax id?

applications/fax/src/fax_smtp.erl
```diff
@@ -892,3 +893,14 @@ write_tmp_file(Filename, Extension, Body) ->
             lager:debug("error writing file ~s : ~p", [Filename, Error]),
             Error
     end.
+
+send_outbound_smtp_fax_error(State) ->
+    Message = props:filter_empty(
+                [{<<"Account-ID">>, State#state.account_id}
+                ,{<<"Fax-From-Email">>, State#state.from}
+                ,{<<"Fax-To-Email">>, State#state.to}
+                ,{<<"Errors">>, State#state.errors}
```
fenollp: How about including more information?
peer_ip, proxy, numbers, ... would this help?

@lazedo ideas?

lazedo:
@kalda341 i understand the need but don't agree with the implementation. the changes makes every system error to be relayed to user and we already send a system notification.

kalda341:
@lazedo I'm also not convinced it is the best implementation, what do you suggest?
It doesn't work for every system error, and this is deliberate. If state.account_id isn't set - which it isn't for some errors then publish will crash. This means that individual accounts can turn on/off the error.

lazedo:
i need some time to think about it. if we fail to attach a file to a document, the user would still get the notification and that's an operational issue that the user should not be notified of.

kalda341:
@lazedo Any thoughts about how to proceed?

lazedo:
@kalda341 hey, i'll try to run thru this today and let you know

lazedo:
@kalda341 lets wait for the rework in `teletype` to revisit this.

jamesaimonetti:
@kalda341 can you rebase now that the other teletype work is merged?

kalda341:
@jamesaimonetti Done :)

fenollp:
circle ci failed: Maybe try `make apis` and see if that fixes anything ;)

kalda341:
@jamesaimonetti / @lazedo  What is the status of this?

jamesaimonetti:
@kalda341 rebased and cleaned up the PR a bit. Teletype now has text/html files for the email templates instead of being included in the source, for instance. See #3687 for those updates. Once checks pass, will get it merged.

PR #3328, opened 2017-02-23
KAZOO-5368: retrieve cached docs in order

fenollp:


core/kazoo_json/src/kz_json.erl
```diff
@@ -402,6 +406,22 @@ sum(?JSON_WRAPPER(_)=JObj1, Value, Sumer, Keys)
...
+%% @doc
+%% Reorder JSON objects according to the given list of binaries.
+%% Given Path MUST resolve to all distinct values in the given objects.
+%% These resolved values MUST all be in the list of binaries too.
+%% List of binaries MUST NOT contain duplicates.
+%% Both lists MUST be of same size.
+%% @end
+%%--------------------------------------------------------------------
+-spec order_by(path(), ne_binaries(), objects()) -> objects().
+order_by(Path, Ids, JObjs)
```
jamesaimonetti: I'm confused why you want/need a list of IDs? Shouldn't you order the JSON objects by the path?

Also, process dictionary seems unnecessary.

Untested but what about:

```erlang
order_by(Path, JObjs) ->
    WithValues = [{get_value(Path, JObj), JObj} || JObj <- JObjs],
    [JObj || {_, JObj} <- lists:keysort(1, WithValues)].
```

Does this do the same job?

fenollp: I think you misunderstood this PR. Look into kz_datamgr.

jamesaimonetti: Please ELI5 then.

fenollp: Big boy you probably got stuck on this functions name.

Anyway, this PR is about retrieving cached docs in the order they were queried,
thus mimicking the behavior of open_docs.

WRT using p.dict: what is wrong with using the right tool for the job?
Did you find another Erlang data structure with O(1) random access? If so please share.

lazedo: @jamesaimonetti open_cached_docs allows to replace the use of  `kz_datamgr:get_results` with `keys` option as in the case of  `/accounts/:Id/tree` when caching is desirable.

jamesaimonetti: @lazedo @fenollp In this case, it seems like a specific feature to kz_datamgr and not a "generic" JSON object operation. I would put it there and maybe name it `sort_by_ids/3` maybe? As always, naming is hard.

fenollp:
@lazedo I did not see how mapfold could be used with better time complexity here. I reworked things a bit to save some copying. No benchmarks though

PR #3330, opened 2017-02-23
HELP-27041: make it clear that the recording will be stored by other …

k-anderson:
…means


lazedo:
@k-anderson check this https://github.com/2600hz/kazoo/pull/3332

PR #3338, opened 2017-02-24
Auto-generate Callflow docs based on schemas

jamesaimonetti:


applications/callflow/src/cf_flow.erl
```diff
@@ -14,12 +14,11 @@
...
--record(pattern, {
-          flow_id :: ne_binary(),
-          has_groups :: boolean(),
-          names = [] :: ne_binaries(),
-          regex :: re:mp()
-         }).
+-record(pattern, {flow_id :: ne_binary(),
+                  has_groups :: boolean(),
+                  names = [] :: ne_binaries(),
+                  regex :: re:mp()
```
fenollp: leading ,s

applications/callflow/src/module/cf_acdc_agent.erl
```diff
@@ -183,9 +183,12 @@ send_new_status(Call, AgentId, Data, PubFun, Timeout) ->
                ]),
     PubFun(Update).
 
-presence_id(JObj) -> kz_device:presence_id(JObj).
-presence_state(JObj) ->
-    format_presence_state(kz_json:get_value(<<"presence_state">>, JObj)).
+-spec presence_id(kz_json:object()) -> api_binary().
```
fenollp: api_ne_binary

applications/callflow/src/module/cf_call_waiting.erl
```diff
@@ -87,7 +87,7 @@ maybe_get_doc(_, 'undefined') ->
 maybe_get_doc('undefined', _) ->
     {'error', 'no_account_db'};
 maybe_get_doc(AccountDb, Id) ->
-    kz_datamgr:open_doc(AccountDb, Id).
+    kz_datamgr:open_cache_doc(AccountDb, Id).
```
fenollp: isn't this dangerous / outside this pr's scope?

applications/callflow/src/module/cf_cidlistmatch.erl
```diff
@@ -32,7 +32,7 @@
 -spec handle(kz_json:object(), kapps_call:call()) -> 'ok'.
 handle(Data, Call) ->
     CallerIdNumber = kapps_call:caller_id_number(Call),
-    ListId = kz_doc:id(Data),
+    ListId = kz_json:get_ne_binary_value(<<"id">>, Data),
```
fenollp: what's the rationale behind all these acccessors removals?

applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -154,27 +154,27 @@
                  ,temporary_unavailable_media_id :: api_binary()
                  ,name_media_id :: api_binary()
                  ,pin = <<>> :: binary()
-                 ,timezone :: ne_binary()
+                 ,timezone :: api_binary()
```
fenollp: api_ne_binary

core/kazoo_ast/src/cb_api_endpoints.erl
```diff
@@ -31,12 +33,18 @@
...
     lists:foreach(fun api_to_ref_doc/1, ?MODULE:get()).
 
 to_ref_doc(CBModule) ->
+    ensure_ref_dir(),
     Path = code:which(CBModule),
     api_to_ref_doc(hd(process_module(Path, []))).
 
+-spec ensure_ref_dir() -> 'ok'.
+ensure_ref_dir() ->
+    'ok' = filelib:ensure_dir(filename:join([?REF_PATH, ".placeholder"])).
```
fenollp: ref/'s files are part of the git tree. this call is useless.

core/kazoo_ast/src/cf_data_usage.erl
```diff
@@ -22,18 +22,45 @@
...
     'ok'.
 
 -spec to_schema_doc(module()) -> 'ok'.
 to_schema_doc(M) ->
+    ensure_ref_dir(),
     to_schema_doc(M, process(M)).
 
+-spec ensure_ref_dir() -> 'ok'.
+ensure_ref_dir() ->
+    'ok' = filelib:ensure_dir(filename:join([code:lib_dir('callflow'), "doc", "ref", ".placeholder"])).
```
fenollp: now that you added files there this should be removed.

applications/callflow/src/module/cf_call_waiting.erl
jamesaimonetti: No, there's little cause to *not* use open_cache_doc and this isn't one.

core/kazoo_ast/src/cb_api_endpoints.erl
jamesaimonetti: When using `ensure_dir` it will create the parent directory only - if the path ends with a directory - so `ensure_dir(<<"/a/b/c">>)` will create `/a/b`. This is just a sanity check to make sure the ref path exists (could be someone built a release and wants docs for some reason?). Doesn't hurt to have it.

fenollp: > could be someone built a release and wants docs for some reason?

that doesn't make sense.

> Doesn't hurt to have it

lots of things can go by that comment... I thought we were not supposed to bring in dead code?

jamesaimonetti: Its a sanity check at the beginning of the process - hardly dead code. Yes, the likely-hood of the directory not existing is low - but with a few milliseconds at the beginning, we can take that likely-hood to 0. That sounds like a win!

fenollp: The likelyhood of `applications/crossbar/doc/ref` to not exist is `0`

PR #3340, opened 2017-02-25
KAZOO-5371: Relaxed server maxLength restrictions in resources.json s…

sergey-safarov:
…chema


fenollp:
hey there, can you run `make apis` and push the results? Thanks

sergey-safarov:
Done, but some checks is not passed.

lazedo:
thanks @sergey-safarov

PR #3341, opened 2017-02-27
json schema validation extensions

lazedo:


fenollp:
Is it possible to not have to add extended-validation? My rationale is for the json schemas to contain all the required info, that way newcomers/users can get up to speed faster just reading from the "familiar" json files.

applications/crossbar/priv/couchdb/schemas/callflows.action.json
```diff
@@ -13,11 +13,14 @@
             "type": "object"
         },
         "data": {
+            "default": {},
             "description": "The data/arguments of the callflow module",
+            "extended-validation": true,
```
jamesaimonetti: JSON schema uses camelCase - any reason not to use it too for consistency?

lazedo: i could have added `extendedValidation` but wanted to make sure this is our extension.
feel free to change it to a better name, just make sure it doesn't collide with schema handled names  :)

lazedo: maybe `kazoo-extended-validation` or `kazooExtendedValidation` ?

jamesaimonetti: `kzExtendedValidation` :) I don't know though, was more just commenting on following the camelCase convention.

PR #3343, opened 2017-02-27
HELP-28466: fix state picking when transitioning port request

fenollp:


core/kazoo_number_manager/src/knm_numbers.erl
```diff
@@ -626,6 +629,14 @@ take_not_founds(T=#{ko := KOs}) ->
     Nums = [Num || {Num,not_found} <- NumsNotFound],
     {T#{ko := maps:from_list(NewKOs)}, Nums}.
 
+maybe_set_ported_in(T=#{todo := PNs, options := Options}) ->
```
k-anderson: @fenollp spec

PR #3345, opened 2017-02-27
HELP-27463 media proxy ip configuration - 4.0

lazedo:


core/kazoo_media/src/kz_media_proxy.erl
```diff
@@ -113,3 +119,53 @@ find_file(File, Root) ->
...
             end
     end.
+
+-spec get_binding_ip() -> inet:ip_address().
+get_binding_ip() ->
+    IsIPv6Enabled = is_ip_family_supported("localhost", 'inet6'),
+    IsIPv4Enabled = is_ip_family_supported("localhost", 'inet'),
+
+    %% expilicty convert to list to allow save the default value in human readable value
+    IP = kz_util:to_list(kapps_config:get_binary(?CONFIG_CAT, <<"ip">>, default_ip())),
```
k-anderson: @lazedo howdy, (this applies to both PRs) but as we just discussed this should be called 'proxy_ip' to match the corresponding 'proxy_port'.  Additionally, it is preferring ipv6 as the default and should be ipv4.  The change for the preference should also be applied to the logic in crossbar where this originated.

PR #3346, opened 2017-02-27
HELP-27320: handle errors when can't get current balance

icehess:
return error tuple back to caller to let them handle error logic


k-anderson:
@icehess lets modify the migration to add the wht_util module to system_config/modb routines and remove it from kazoo_services.

k-anderson:
@icehess also this needs a 4.0 PR please :)

PR #3347, opened 2017-02-28
HELP-28353: schema for e911 region field

fenollp:

PR #3348, opened 2017-02-28
HELP-28353_4oh: schema for e911 region field

fenollp:
master: https://github.com/2600hz/kazoo/pull/3347

### DIFF
--- a/applications/crossbar/priv/couchdb/schemas/phone_numbers.json
+++ b/applications/crossbar/priv/couchdb/schemas/phone_numbers.json
@@ -84,6 +84,8 @@
                 },
                 "region": {
                     "description": "The region (state) where the number is in service",
+                    "maxLength": 2,
+                    "minLength": 2,
                     "type": "string"
                 },
                 "status": {
--- a/core/kazoo_number_manager/src/providers/knm_providers.erl
+++ b/core/kazoo_number_manager/src/providers/knm_providers.erl
@@ -260,9 +260,14 @@ legacy_provider_to_feature(Else) -> Else.
 requested_modules(Number) ->
     PhoneNumber = knm_number:phone_number(Number),
     AccountId = knm_phone_number:assigned_to(PhoneNumber),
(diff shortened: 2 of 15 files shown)


fenollp:
don't merge yet

fenollp:
ready for review

core/kazoo_number_manager/src/providers/knm_providers.erl
```diff
@@ -328,15 +333,21 @@ exec(Number, Action=delete) ->
     lager:debug("requested number features: ~p", [RequestedModules]),
     exec(Number, Action, RequestedModules);
 exec(Number, Action) ->
+    {AllowedRequests, DeniedRequests} = split_requests(Number),
+    lager:debug("allowing number features ~p", [AllowedRequests]),
+    case DeniedRequests =:= [] of
+        true -> exec(Number, Action, AllowedRequests);
+        false ->
+            lager:debug("denied number features ~p", [DeniedRequests]),
+            knm_errors:unauthorized()
```
fenollp: @k-anderson note WRT numbers which somehow got denied features activated
* before this: the allowed features would get saved then the denied features would get deleted automatically
* with this PR: it fails here, before attempting any providers/feature operations (ie returns 403 instead of save-then-delete)

applications/crossbar/priv/couchdb/schemas/phone_numbers.json
```diff
@@ -84,6 +84,8 @@
                 },
                 "region": {
                     "description": "The region (state) where the number is in service",
+                    "maxLength": 2,
+                    "minLength": 2,
```
fenollp: @k-anderson

fenollp: yeah so about this: am I right in assuming "every" e911 provider uses the 2-letter format for states?
I couldn't find info on Dash's handling of Canada regions (and we need this)

PR #3352, opened 2017-03-01
HELP-27956: rollover maintenance command improvements and low balance…

k-anderson:
… 0 recheck

### DIFF
--- a/core/kazoo_modb/src/kazoo_modb_maintenance.erl
+++ b/core/kazoo_modb/src/kazoo_modb_maintenance.erl
@@ -114,7 +114,7 @@ current_rollups([Account|Accounts]) ->
             Else ->
                 io:format("[~p] account ~s error getting monthly rollup ~p~n", [Remaining, AccountId, Else])
         end,
-     current_rollups(Accounts).
+    current_rollups(Accounts).
 
 -spec verify_rollups() -> 'ok'.
 verify_rollups() ->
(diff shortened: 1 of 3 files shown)


k-anderson:
I need to create a PR for master as well.

applications/tasks/src/kz_account_crawler.erl
```diff
@@ -326,7 +326,13 @@ maybe_test_for_low_balance(AccountId, AccountJObj) ->
 
 -spec test_for_low_balance(ne_binary(), kz_account:doc()) -> 'ok'.
 test_for_low_balance(AccountId, AccountJObj) ->
-    CurrentBalance = wht_util:current_balance(AccountId),
+    CurrentBalance = case wht_util:current_balance(AccountId) of
+                         Balance when Balance == 0 ->
+                             kazoo_modb_maintenance:fix_rollup(AccountId),
+                             timer:sleep(5000),
```
icehess: With changes in #3346 I don't think this is necessary here. Also it's making crawler slower.

core/kazoo_modb/src/kazoo_modb_maintenance.erl
```diff
@@ -86,67 +90,112 @@ do_archive_modbs(MODbs, AccountId) ->
...
 -spec fix_rollup(ne_binary()) -> 'ok'.
 fix_rollup(Account) ->
+    {Year, Month, _} = erlang:date(),
+    fix_rollup(Account, Year, Month).
+
+-spec fix_rollup(ne_binary(), ne_binary(), ne_binary()) -> 'ok'.
+fix_rollup(Account, Year, Month) ->
     AccountId = kz_util:format_account_id(Account, 'raw'),
-    {Y, M, _} = erlang:date(),
+    Y = kz_util:to_integer(Year),
```
icehess: no need for convert to integer since `kazoo_modb_util:prev_year_month` is already doing that

PR #3353, opened 2017-03-01
qf-coveralls: upgrade coveralls-erl

fenollp:


fenollp:
Issue is that tests will pass but upload of coverage data to coveralls fails at some point.

Hm this upgrade doesn't hurt, but issue may be just due to travis-ci's API issues since S3 outage yesterday.

PR #3355, opened 2017-03-01
State of docs index

jamesaimonetti:
Adds all but three docs to mkdocs.yml. Not sure what to do with them


jamesaimonetti:
I, too, like TYPING IN ALL CAPS because who uses **bold** for emphasis these days...

I appreciate the zealousness with which you want docs to be ideal. However, what's in the PR is an improvement over what was there prior. I'd rather have a basic listing of requests over nothing. When someone does work on the endpoints, they should fill in the docs as it will be relevant.

fenollp:
> When someone does work on the endpoints, they should fill in the docs as it will be relevant.

There is a misunderstanding / laziness creeping in here and I was going to repeat a good part of my previous comment.

Instead let's emphasize this quote, since your reply doesn't mention it
document 
> Otherwise this is just lying to scripts/state-of-docs

What this means is that this script counts/finds endpoints in ref/ and doc/. It returns the percentage of ref endpoints that are in doc/. I'm not being overzealous when I ask that you either do more than c/p or that you remove the c/p, I am saying that by doing just c/p you are falsifying the output of this script.

Now, it is true that there are already cases like these in doc/ (and I "fought" against them) and you may disregard this review as you have done before.

However, you know lots about Kazoo, you built most of it, hell you're even in charge of documenting it now.
How much time do you think it would take you to paste the JSON response of the what 7 API enpoints you pasted in this PR?
Is it really a lot to ask for?

jamesaimonetti:
@fenollp Yes it is a lot, because its not just throwing some cURL commands at Crossbar and pasting the resultant JSON. There's context around these APIs, crafting JSON request bodies that are reasonable, highlighting potential error responses that would be "normal" to receive, etc.

I have to prioritize this among all the other documentation efforts, in Kazoo and across the company. Pragmatically speaking, copy-pasting the auto-generated endpoint calls is an improvement on the existing doc (which had nothing of use). The hope is that the next developer/community member to use the endpoint will see fit to improve the docs.

Additionally, these endpoints are not well-used so spending the time to properly document isn't the highest priority. If there are complaints from community or support tickets about how to use them, then we know time is worth spending to flesh these out. Until then, there are better things to do with our time.

PR #3360, opened 2017-03-02
Fax attachment was being checked before saving

tsromana:
When trying to save a fax attachement it was first checked if there was an attachement if not than it was saved by always outputting a warning in the logs, this change will first save the attachment and than will check if it exists or not and if not than output an error and try to save again.


jamesaimonetti:
Thanks @tsromana, merged and I'm going to backport to 4.0 too

jamesaimonetti:
@k-anderson backported to 4.0

PR #3363, opened 2017-03-03
Modular teletype

kalda341:
This allows teletype modules to be added without modifications to teletype_shared_listener.
Modules can be enabled and disabled at runtime, using `start_module` and `stop_module` sup commands, much like existing kazoo applications such as crossbar.

`teletype_bindings` is used to distribute notifications to the relevant modules. It does not use Kazoo bindings as we do not want to fold through registered modules. All existing teletype modules have been updated to use `teletype_bindings`.

`teletype_shared_listener` has become a registered process. This is to allow teletype modules to listen to notifications send with modules other than `kapi_notifications`. This is essential for us, as we wish to add custom notification events without modifying `kapi_notifications`.

I welcome suggestions as to improvements in both design and implementation.


applications/teletype/src/teletype_bindings.erl
```diff
@@ -0,0 +1,119 @@
...
+    lager:debug("finished dispatching notification {~s, ~s}", [EventCategory, EventName]),
+    'ok'.
+
+-spec start_modules() -> 'ok'.
+start_modules() ->
+    start_modules(?AUTOLOAD_MODULES).
+
+-spec start_modules([module()]) -> 'ok'.
+start_modules([Module | Remaining]) ->
+    teletype_maintenance:start_module(Module),
```
kalda341: What are people's thoughts about running this in a new process? Or is startup time not a concern?

applications/teletype/src/teletype_bindings.erl
```diff
@@ -0,0 +1,119 @@
+-module(teletype_bindings).
```
kalda341: What would you like me to do regarding copyright notices?

applications/teletype/src/teletype_bindings.erl
```diff
@@ -0,0 +1,119 @@
...
+        ]).
+
+-include("teletype.hrl").
+
+-define(SERVER, ?MODULE).
+
+-type event() :: {api_binary(), api_binary()}.
+-type responder() :: {module(), atom()}.
+%% Set would be better suited, but lists are more convenient
+-type responders() :: [responder()].
```
kalda341: Would it be preferable to use a set? We don't want duplicates, but lists are much easier to iterate over and append to.

lazedo:
@kalda341 we like the idea of moving teletype modules to kazoo_bindings. we don't understand the need of not using kazoo_bindings (see crossbar, blackhole). as for the shared_listener / custom notification, you should take the approach of creating a custom message inside kapi_notifications as @onnet did for custom templates.

another note, check crossbar, blackhole as a pattern for implementing the modules configuration. we configure `users` not `cb_users`

kalda341:
What I had seen of kazoo_bindings was only to fold over bindings, not execute them in parallel. Looking closer I can see map, which seems to be well suited. I'll change it to use this.

What has @onnet done for custom templates? We would really like to avoid having to modify kapi_notifications to add custom notifications, as we would like to avoid running any custom versions of your code.

@lazedo What do you mean by the modules configuration?

lazedo:
@kalda341 configuration should not include the prefix. check how crossbar and blackhole works
```
                "autoload_modules": {
 +                    "default": [
 +                        "account_zone_change",
 +                        "cnam_request",
 +                        "customer_update",
 +                        "denied_emergency_bridge",
```` 
instead of
```
                "autoload_modules": {
 +                    "default": [
 +                        "teletype_account_zone_change",
 +                        "teletype_cnam_request",
 +                        "teletype_customer_update",
 +                        "teletype_denied_emergency_bridge",
````

kalda341:
@lazedo I have switched to kazoo_bindings.

From what I can see, both the DEFAULT_MODULES in crossbar.hrl and the system_config database include the prefixes. Perhaps I'm misunderstanding you?

lazedo:
@kalda341 forget my comment about modules, this never passed of the intention of doing that way.

jamesaimonetti:
@kalda341 I am running this branch locally and testing. Will get you updates (or merge) when I can (am traveling today so tomorrow most likely).

applications/teletype/src/teletype_bindings.erl
jamesaimonetti: I'd rather do it as you are, makes sure teletype is loaded properly before accepting any notification AMQP events.

jamesaimonetti:
@kalda341 merged manually after applying a few updates. Thanks for the patience and persistence!

lazedo:
@kalda341 FYI, the registered process `teletype_shared_listener` will be reverted as with the `add_binding`  when we move to pools of listeners with consistent hashing. we will provide a way to create `custom notifications` by that time.

kalda341:
Thank you @lazedo! I was not particularly happy with my approach, I just couldn't think of a better solution.

PR #3373, opened 2017-03-03
qf-knm-dirty: fix set_created is_dirty-ness

fenollp:
4.0: https://github.com/2600hz/kazoo/pull/3372


fenollp:
Same as https://github.com/2600hz/kazoo/pull/3379

PR #3375, opened 2017-03-03
qf-datamgr: export kz_datamgr:get_result_keys/2

fenollp:
Very important to have this for debug


fenollp:
https://github.com/2600hz/kazoo/pull/3370 time to take a break

PR #3377, opened 2017-03-03
PROD-123: allow master account to set module name when creating num

icehess:
When creating a number the API should honor a carrier module name, defaulting to local if not present or the authorizing user does not belong to the master account.

### DIFF
--- a/applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
+++ b/applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
@@ -355,12 +355,15 @@ put(Context, ?COLLECTION) ->
     CB = fun() -> ?MODULE:put(cb_context:set_accepting_charges(Context), ?COLLECTION) end,
     set_response(Results, Context, CB);
 put(Context, Number) ->
+    ReqData = cb_context:req_data(Context),
+    ModuleName = kz_json:get_ne_value(<<"module_name">>, ReqData),
     Options = [{'auth_by', cb_context:auth_account_id(Context)}
               ,{'assign_to', cb_context:account_id(Context)}
               ,{'dry_run', not cb_context:accepting_charges(Context)}
-              ,{'public_fields', cb_context:doc(Context)}
+              ,{'public_fields', kz_json:delete_key(<<"module_name">>, ReqData)}
+              ,{'module_name', ModuleName}
               ],
-    Result = knm_number:create(Number, Options),
+    Result = knm_number:create(Number, props:filter_undefined(Options)),
     CB = fun() -> ?MODULE:put(cb_context:set_accepting_charges(Context), Number) end,
     set_response(Result, Context, CB).
 
@@ -976,12 +979,15 @@ numbers_action(Context, ?ACTIVATE, Numbers) ->
               ],
     knm_numbers:move(Numbers, cb_context:account_id(Context), Options);
 numbers_action(Context, ?HTTP_PUT, Numbers) ->
+    ReqData = cb_context:req_data(Context),
+    ModuleName = kz_json:get_ne_value(<<"module_name">>, ReqData),
     Options = [{'auth_by', cb_context:auth_account_id(Context)}
               ,{'assign_to', cb_context:account_id(Context)}
               ,{'dry_run', not cb_context:accepting_charges(Context)}
-              ,{'public_fields', cb_context:req_data(Context)}
+              ,{'public_fields', kz_json:delete_key(<<"module_name">>, ReqData)}
+              ,{'module_name', ModuleName}
               ],
-    knm_numbers:create(Numbers, Options);
+    knm_numbers:create(Numbers, props:filter_undefined(Options));
 numbers_action(Context, ?HTTP_POST, Numbers) ->
     Options = [{'auth_by', cb_context:auth_account_id(Context)}
               ,{'assign_to', cb_context:account_id(Context)}
--- a/core/kazoo_number_manager/src/knm_numbers.erl
+++ b/core/kazoo_number_manager/src/knm_numbers.erl
@@ -262,20 +262,20 @@ do_get_pn(Nums, Options, Error) ->
 %%--------------------------------------------------------------------
 -spec create(ne_binaries(), knm_number_options:options()) -> ret().
 create(Nums, Options) ->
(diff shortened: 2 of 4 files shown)


icehess:
Please do not merge this yet, as we need to discuss an issue here with `cb_jobs_listener`

core/kazoo_number_manager/src/knm_numbers.erl
```diff
@@ -262,20 +262,20 @@ do_get_pn(Nums, Options, Error) ->
 %%--------------------------------------------------------------------
 -spec create(ne_binaries(), knm_number_options:options()) -> ret().
 create(Nums, Options) ->
-    ?MATCH_ACCOUNT_RAW(AccountId) = knm_number_options:assign_to(Options), %%FIXME: can crash
```
fenollp: This ensures one cannot create numbers that would not be assigned to anyone. 
I realize now that this line is missing a test case.

core/kazoo_number_manager/src/knm_numbers.erl
```diff
@@ -637,6 +637,53 @@ maybe_set_ported_in(T=#{todo := PNs, options := Options}) ->
...
+                                 ret().
+state_create_option(Nums, Options) ->
+    case knm_number_options:assign_to(Options) of
+        ?MATCH_ACCOUNT_RAW(AccountId) ->
+            ToState = knm_number:state_for_create(AccountId, Options),
+            lager:debug("picked state ~s for ~s for ~p", [ToState, AccountId, Nums]),
+            [{'state', ToState} | Options];
+        _ ->
+            Error = knm_errors:to_json(assign_failure, undefined, field_undefined),
+            ret(new(Options, [], Nums, Error))
```
fenollp: I still think this test should happen at the very beginning. Also this should be using T1 instead of new.

core/kazoo_number_manager/test/knm_create_new_number_test.erl
```diff
@@ -50,6 +50,7 @@ create_new_number_test_() ->
 reseller_new_number_test_() ->
     Props = [{'auth_by', ?RESELLER_ACCOUNT_ID}
             ,{'assign_to', ?RESELLER_ACCOUNT_ID}
+            ,{'module_name', <<"knm_bandwidth2">>}
```
fenollp: Is this needed for the new behavior?

core/kazoo_number_manager/test/knm_create_new_number_test.erl
```diff
@@ -282,6 +283,22 @@ create_non_existing_mobile_number_test_() ->
      }
     ].
 
+create_number_with_module_name_test_() ->
+    Props = [{'auth_by', ?MASTER_ACCOUNT_ID}
```
fenollp: Needs a test that checks for failure too

core/kazoo_number_manager/src/knm_numbers.erl
```diff
@@ -637,6 +637,53 @@ maybe_set_ported_in(T=#{todo := PNs, options := Options}) ->
...
+    end.
+
+%% @private
+-spec module_name_create_option(ne_binaries(), knm_number_options:options()) -> knm_number_options:options().
+module_name_create_option(_Nums, Options) ->
+    AuthBy = knm_number_options:auth_by(Options),
+    case kz_term:is_not_empty(AuthBy)
+        andalso {props:get_ne_binary_value('module_name', Options), kapps_util:get_master_account_id()}
```
fenollp: There's module_name/1,2

core/kazoo_number_manager/src/knm_numbers.erl
```diff
@@ -262,20 +262,20 @@ do_get_pn(Nums, Options, Error) ->
...
 -spec create(ne_binaries(), knm_number_options:options()) -> ret().
 create(Nums, Options) ->
-    ?MATCH_ACCOUNT_RAW(AccountId) = knm_number_options:assign_to(Options), %%FIXME: can crash
     T0 = do_get_pn(Nums, Options, knm_errors:to_json(not_reconcilable)),
     case take_not_founds(T0) of
         {#{ok := []}, []} -> T0;
         {T1, NotFounds} ->
-            ToState = knm_number:state_for_create(AccountId, Options),
-            lager:debug("picked state ~s for ~s for ~p", [ToState, AccountId, Nums]),
-            NewOptions = [{'state', ToState} | Options],
```
fenollp: How about just adding `{module_name, pick_module(Options)}`? Should default to knm local.

applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
```diff
@@ -976,12 +979,15 @@ numbers_action(Context, ?ACTIVATE, Numbers) ->
...
     knm_numbers:move(Numbers, cb_context:account_id(Context), Options);
 numbers_action(Context, ?HTTP_PUT, Numbers) ->
+    ReqData = cb_context:req_data(Context),
+    ModuleName = kz_json:get_ne_value(<<"module_name">>, ReqData),
     Options = [{'auth_by', cb_context:auth_account_id(Context)}
               ,{'assign_to', cb_context:account_id(Context)}
               ,{'dry_run', not cb_context:accepting_charges(Context)}
-              ,{'public_fields', cb_context:req_data(Context)}
+              ,{'public_fields', kz_json:delete_key(<<"module_name">>, ReqData)}
+              ,{'module_name', ModuleName}
```
lazedo: @icehess shouldn't this be restricted to `superduper` ?

icehess: @lazedo idk, ticket says master account

fenollp:
Yeah so this looks a lot like KAZOO-5290 https://github.com/2600hz/kazoo/pull/3254
Same behavior can be reproduced already by just PUTing a number with that feature on.

Can you replace this PR with just a number creation test that uses the feature?
Take a look at the end of core/kazoo_number_manager/test/knm_telnyx_test.erl for help.

fenollp:
Note https://github.com/2600hz/kazoo/pull/3364 for import task too.

applications/crossbar/src/modules_v2/cb_phone_numbers_v2.erl
lazedo: but where is `master account` checked ?

PR #3379, opened 2017-03-04
HELP-28542: initial setting of used_by should mark the record as dirty

k-anderson:


fenollp:
This is superseeded by the forward port of https://github.com/2600hz/kazoo/pull/3384 I am working on.

PR #3382, opened 2017-03-04
Fax abnormal termination fix

tsromana:
Bug fix when job was released without adding the status to the state which caused the subsequent channel destroy to release the job again which caused fax termination and caused no email delivery of the status of the fax.

### DIFF
--- a/applications/fax/src/fax_worker.erl
+++ b/applications/fax/src/fax_worker.erl
@@ -313,14 +313,16 @@ handle_cast({'fax_status', <<"result">>, JobId, JObj}
     Data = kz_call_event:application_data(JObj),
     {Resp, Doc} = case kz_json:is_true([<<"Fax-Success">>], Data) of
                       'true' ->
+                          lager:debug("fax sucessfully sent"),
                           send_status(State, <<"Fax Successfuly sent">>, ?FAX_END, Data),
                           release_successful_job(JObj, Job);
                       'false' ->
+                          lager:debug("error sending fax"),
                           send_status(State, <<"Error sending fax">>, ?FAX_ERROR, Data),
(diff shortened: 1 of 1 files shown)


lazedo:
@tsromana are you saying that you're getting (fax_status, result) without (fax_status, negociateresult) or (fax_status, pageresult) ?

tsromana:
@lazedo after tx_response it goes straight to fax_status result which causes it to crash

tsromana:
should the status be changed to api_binary as suggested by circleci ?

lazedo:
@tsromana can you provide a way to reproduce this ?

tsromana:
@lazedo sorry didn't see your comment, as for reproducing i am not entirely sure as it happened when i was testing fax when trying to use a fax machine once it said the call was disconnected prematurely, i received it once more when after applying the fix but that time it worked perfectly so i guess its just a rare issue, and if it poses a problem to other scenarios i guess it best to not include it in then as it will be very rare chance i guess :) as i am not an expert in this and don't have much experience with fax so can't be 100% sure of the reason why it happened my suspicion is that its because of a crappy carrier as the one used for testing is not very reliable.

lazedo:
@tsromana just trying to understand the issue as it seems the function clause should be elsewhere. i'll try to find a way to reproduce it.

jamesaimonetti:
@tsromana can you rebase and check if this PR is still needed please?

tsromana:
Sure, sorry i forgot about this, will try to recreate the issue.

lazedo:
@tsromana check https://github.com/2600hz/kazoo/pull/3760

tsromana:
@lazedo will test that out and will report back if i find any issue and will reopen this pr :)

PR #3386, opened 2017-03-06
qf-kpn: fix Dialyzer warnings in KNM that come from OTP19.2

fenollp:


lazedo:
@mark2600 @fenollp didn't we talked about this ? i'm really  against this merge

fenollp:
@lazedo Really? I amended as discussed. Adding an exception to the Dialyzer script was failing in lots of places.

lazedo:
@fenollp i believe that we discussed about adding the exception handling in the script and you didn't mentioned it was complex to added there, that's the reason for the comment. but i guess i reacted too soon, sorry for that. i'm ok with the commit.

PR #3387, opened 2017-03-06
Update haproxy.md

powerpbx:
Required on v1.5.18 for check to work.


powerpbx:
Nevermind.  I was using an older version.  The latest one seems to work fine.

PR #3388, opened 2017-03-07
[master] properly return an error when exception raised in account creation

icehess:
When account creation failed, we don't return an error to UI, so UI doesn't know that account has not created, which result in UI continues to make other request to put limits, callflow an etc to the non-existent account.

### DIFF
--- a/applications/crossbar/src/modules/cb_accounts.erl
+++ b/applications/crossbar/src/modules/cb_accounts.erl
@@ -318,15 +318,15 @@ put(Context, PathAccountId) ->
             case cb_context:is_context(C) of
                 'true' -> delete(C, NewAccountId);
                 'false' ->
-                    C = cb_context:add_system_error('unspecified_fault', Context),
-                    delete(C, NewAccountId)
+                    _ = delete(Context, NewAccountId),
+                    cb_context:add_system_error('unspecified_fault', <<"internal error, unable to create the account">>, Context)
             end;
         _E:_R ->
             ST = erlang:get_stacktrace(),
             lager:debug("unexpected failure when creating account: ~s: ~p", [_E, _R]),
             kz_util:log_stacktrace(ST),
-            C = cb_context:add_system_error('unspecified_fault', Context),
-            delete(C, NewAccountId)
+            _ = delete(Context, NewAccountId),
+            cb_context:add_system_error('unspecified_fault', <<"internal error, unable to create the account">>, Context)
     end.
 
 put(Context, AccountId, ?RESELLER) ->


applications/crossbar/src/modules/cb_accounts.erl
```diff
@@ -1342,7 +1342,7 @@ set_notification_preference(Context, Preference) ->
 
 -spec create_account_mod(ne_binary()) -> any().
 create_account_mod(AccountId) ->
-    Db = kz_util:format_account_mod_id(AccountId),
+    Db = kz_util:format_account_mod_id(AccountId, raw),
```
lazedo: this will fail, revert this line as the 2nd parameter is time related

icehess: yeah, fixed, just did that for test :)

applications/crossbar/src/modules/cb_accounts.erl
```diff
@@ -1342,8 +1342,7 @@ set_notification_preference(Context, Preference) ->
 
 -spec create_account_mod(ne_binary()) -> any().
 create_account_mod(AccountId) ->
-    Db = kz_util:format_account_mod_id(AccountId),
-    kazoo_modb:create(Db).
+    kazoo_modb:create(kazoo_modb:get_db(AccountId)).
```
lazedo: why this change ? was it not working ?

icehess: it's related to wht_util, `kazoo_modb:create` was expecting raw account_modb which would be passed to create_routines

icehess: I know you fixed that in wht_util, but the main cause was here actually

lazedo: and wasn't that fixed in wht_util ?

icehess: yes, but with formating account_modb there is no need to handle encoded format in modb routines, since `cb_accounts` is the only place that calls `kazoo_modb:create`, and all the maybe_create functions in kazoo_modb converts the modb to raw. I would like to be consistence anywhere to not confused in the future.

lazedo: wht_util had an an error that was fixed, the problem was not here, kz_services_modb handled the modb formats.

this pr is about returning a proper message in case account creation fails, lets keep the scope to that. 

to keep consistence, you can create another pr to remove usage of format_account_mod_id in all base code and deprecate it for 4.1 (add a note in announcements).

icehess: ok, i'll open another pr for this later.

PR #3389, opened 2017-03-07
[4.0] properly return an error when exception raised in account creation

icehess:
When account creation failed, we don't return an error to UI, so UI doesn't know that account has not created, which result in UI continues to make other request to put limits, callflow an etc to the non-existent account.

backport of (#3388)

### DIFF
--- a/applications/crossbar/src/modules/cb_accounts.erl
+++ b/applications/crossbar/src/modules/cb_accounts.erl
@@ -318,15 +318,15 @@ put(Context, PathAccountId) ->
             case cb_context:is_context(C) of
                 'true' -> delete(C, NewAccountId);
                 'false' ->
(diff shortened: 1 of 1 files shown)


applications/crossbar/src/modules/cb_accounts.erl
```diff
@@ -1340,8 +1340,7 @@ set_notification_preference(Context, Preference) ->
 
 -spec create_account_mod(ne_binary()) -> any().
 create_account_mod(AccountId) ->
-    Db = kz_util:format_account_mod_id(AccountId),
-    kazoo_modb:create(Db).
+    kazoo_modb:create(kazoo_modb:get_modb(AccountId)).
```
lazedo: why this change ? was it not working ?

PR #3397, opened 2017-03-08
Help 27871 master

k-anderson:
see #3396

### DIFF
--- a/applications/tasks/src/modules/kt_numbers.erl
+++ b/applications/tasks/src/modules/kt_numbers.erl
@@ -364,23 +364,27 @@ import(ExtraArgs, init, Args) ->
     kz_datamgr:suppress_change_notice(),
     IterValue = sets:new(),
     import(ExtraArgs, IterValue, Args);
-import(#{account_id := Account}, AccountIds, #{<<"e164">> := E164
-                                              ,<<"account_id">> := AccountId0
-                                              ,<<"carrier_module">> := Carrier
-                                               %%TODO: use all the optional fields
-                                              ,<<"port_in">> := _PortIn
-                                              ,<<"previously_assigned_to">> := _PrevAssignedTo
-                                              ,<<"created">> := _Created
-                                              ,<<"modified">> := _Modified
-                                              ,<<"used_by">> := _UsedBy
-                                              ,<<"cnam.inbound">> := CNAMInbound0
-                                              ,<<"cnam.outbound">> := CNAMOutbound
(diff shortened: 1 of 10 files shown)


applications/tasks/src/modules/kt_numbers.erl
```diff
@@ -403,7 +403,7 @@ import(#{account_id := Account}, AccountIds, #{<<"e164">> := E164
                   ])),
     PublicFields = cnam(CNAMInbound, CNAMOutbound) ++ E911
         ++ additional_fields_to_json(Args),
-    Options = [{auth_by, ?KNM_DEFAULT_AUTH_BY}
+    Options = [{auth_by, AccountId}
```
fenollp: Did you mean to use auth_account_id like the other knm_number functions in this file use?

core/kazoo_number_manager/src/carriers/knm_inventory.erl
```diff
@@ -0,0 +1,148 @@
...
+%%--------------------------------------------------------------------
+-spec find_numbers(ne_binary(), pos_integer(), knm_search:options()) ->
+                          {'ok', knm_number:knm_numbers()} |
+                          {'error', any()}.
+find_numbers(Prefix, Quantity, Options) ->
+    case knm_carriers:account_id(Options) of
+        'undefined' -> {'error', 'not_available'};
+        AccountId ->
+            Offset = knm_carriers:offset(Options),
+            QID = knm_search:query_id(Options),
```
fenollp: but knm_search calls find_numbers...?

core/kazoo_number_manager/src/knm_phone_number.erl
```diff
@@ -326,6 +326,10 @@ save(T0=#{todo := PNs, options := Options}) ->
 save(PN=#knm_phone_number{is_dirty = false}) ->
     lager:debug("not dirty, skip saving ~s", [number(PN)]),
     PN;
+save(PN=#knm_phone_number{state=?NUMBER_STATE_AVAILABLE
+                         ,assigned_to=AssignedTo
+                         }) when AssignedTo =/= 'undefined' ->
+    save(PN#knm_phone_number{assigned_to='undefined'});
```
fenollp: ehhh this should happen in unwind/disconnect or knm_number_states instead of here.
Also, you can find existing tests and for every number getting into the available state check for this field's value.
I would especially be interested in a test about the release of a non local number with reserve history.

PR #3414, opened 2017-03-09
only track assignments for reconcilable numbers

lazedo:


applications/crossbar/src/modules/cb_callflows.erl
```diff
@@ -416,33 +416,40 @@ maybe_reconcile_numbers(Context) ->
 track_assignment('post', Context) ->
     NewNums = kz_json:get_value(<<"numbers">>, cb_context:doc(Context), []),
     OldNums = kz_json:get_value(<<"numbers">>, cb_context:fetch(Context, 'db_doc'), []),
+    AccountId = cb_context:account_id(Context),
 
     Unassigned = [{Num, 'undefined'}
                   || Num <- OldNums,
                      not lists:member(Num, NewNums),
-                     Num =/= <<"undefined">>
+                     Num =/= <<"undefined">>,
```
fenollp: these can be replaced with the line below

PR #3425, opened 2017-03-10
cb_system_configs with defaults

jamhed:
A system configuration document is a JSON document of certain
syntetic type (each node value must be a JSON object of type
system_config.$config_name). REST API allows to address both
documents and and docuemnt nodes. Default node of the
configuration document is treated specially, see below.

In order to avoid confusion configuration document nodes
we call document sections, each such section is a JSON
object representing a configuration object for Kazoo
node or zone.

General idea is to return (with with_defaults=true URI
option) the effective values for configuration, therefore
a complex mergedown is performed: document values are
merged with document default section, and then merged
with schema defaults.

Therefore on save actions (PUT/POST/PATCH) a reverse
operation is performed: a difference from defaults
is calculated and stored to the document (if any).
Thus by setting a value to schema default will effectively
remove it from the document on save action.

Documents
---------

GET always returns either an empty document or stored
document. The document always has a default section included,
populated with default values deduced from schema.

POST expects to receive a complete configuration document.
The document must be valid, each section of the document
must pass the validation against system_config.$config_name
schema. Before saving the difference with provided default
is calculated, and then the difference with schema defaults
is calculated, and then the result of this is stored, meaning
if section value is equal to default section value it will
not be stored to the document. The actual (stored) values
are then returned as POST results.

PATCH will merge provided changes with stored document,
and then the resulting document is valid then the same
store procedure is applied as in POST. The actual stored
values are then returned as PATCH result.

DELETE deletes the document (or wipes a section). Attempt
to delete non-existing configuration object will generate
an HTTP 404 error.

Sections
--------

GET always returns either an empty document or stored
values.

POST expects to receive a valid configuration document
(against schema system_config.$config_name). If the
document is valid, then a difference is calculated
with configuration document default section, and
then the difference is calculated with schema defaults,
and then the result is stored. For default section
only the difference from schema defaults are calculated.
The actual (stored) values are returned as result of
POST request.

PATCH will merge provided changes with stored section,
and then the result is validated against schema. If
result is valid, then the same procedure is applied
as in POST. The actual (stored) values are returned
as PATCH result.

DELETE deletes the section specified. If configuration
document doesn't exist (or secion in the document), then
the request will fail with HTTP 404.

Defaults
--------

In order to have effective configuration values one must
provide `with_defaults=true` URI parameter for GET requests
(either to get document or document section). With this
parameter provided a configuration mergedown as described
above is performed.

### DIFF
--- a/applications/crossbar/src/modules/cb_system_configs.erl
+++ b/applications/crossbar/src/modules/cb_system_configs.erl
@@ -18,13 +18,13 @@
         ,validate/1, validate/2, validate/3
         ,put/2
         ,post/2, post/3
+        ,patch/2, patch/3
         ,delete/2, delete/3
         ]).
 
 -include("crossbar.hrl").
 
 -define(DEFAULT, <<"default">>).
-
 
 %%%===================================================================
 %%% API
@@ -41,12 +41,13 @@ init() ->
     _ = kz_datamgr:db_create(?KZ_CONFIG_DB),
     _ = kz_datamgr:revise_doc_from_file(?KZ_CONFIG_DB, 'crossbar', <<"views/system_configs.json">>),
 
-    _ = crossbar_bindings:bind(<<"*.authorize">>, ?MODULE, 'authorize'),
+    _ = crossbar_bindings:bind(<<"*.authorize.system_configs">>, ?MODULE, 'authorize'),
     _ = crossbar_bindings:bind(<<"*.allowed_methods.system_configs">>, ?MODULE, 'allowed_methods'),
     _ = crossbar_bindings:bind(<<"*.resource_exists.system_configs">>, ?MODULE, 'resource_exists'),
     _ = crossbar_bindings:bind(<<"*.validate.system_configs">>, ?MODULE, 'validate'),
     _ = crossbar_bindings:bind(<<"*.execute.put.system_configs">>, ?MODULE, 'put'),
     _ = crossbar_bindings:bind(<<"*.execute.post.system_configs">>, ?MODULE, 'post'),
+    _ = crossbar_bindings:bind(<<"*.execute.patch.system_configs">>, ?MODULE, 'patch'),
     _ = crossbar_bindings:bind(<<"*.execute.delete.system_configs">>, ?MODULE, 'delete').
 
 %%--------------------------------------------------------------------
@@ -56,12 +57,20 @@ init() ->
 %% allowed to access the resource, or false if not.
 %% @end
 %%--------------------------------------------------------------------
--spec authorize(cb_context:context()) -> boolean().
--spec authorize(cb_context:context(), path_token()) -> boolean().
--spec authorize(cb_context:context(), path_token(), path_token()) -> boolean().
-authorize(Context) -> cb_context:is_superduper_admin(Context).
-authorize(Context, _Id) -> cb_context:is_superduper_admin(Context).
-authorize(Context, _Id, _Node) -> cb_context:is_superduper_admin(Context).
+-type authorize_return() :: boolean() | {'halt', cb_context:context()}.
+
+-spec authorize(cb_context:context()) -> authorize_return().
+-spec authorize(cb_context:context(), path_token()) -> authorize_return().
+-spec authorize(cb_context:context(), path_token(), path_token()) -> authorize_return().
+authorize(Context) ->
+    case cb_context:is_superduper_admin(Context)
+        andalso cb_context:req_nouns(Context) of
+        'false' -> {'halt', cb_context:add_system_error('forbidden', Context)};
+        [{<<"system_configs">>, _}] -> 'true';
+        _ -> {'halt', cb_context:add_system_error('bad_identifier', Context)}
+    end.
+authorize(Context, _Id) -> authorize(Context).
+authorize(Context, _Id, _Node) -> authorize(Context).
 
 %%--------------------------------------------------------------------
 %% @public
@@ -76,9 +85,9 @@ authorize(Context, _Id, _Node) -> cb_context:is_superduper_admin(Context).
 allowed_methods() ->
     [?HTTP_GET].
 allowed_methods(_SystemConfigId) ->
-    [?HTTP_GET, ?HTTP_POST, ?HTTP_DELETE, ?HTTP_PUT].
+    [?HTTP_GET, ?HTTP_POST, ?HTTP_DELETE, ?HTTP_PUT, ?HTTP_PATCH].
 allowed_methods(_SystemConfigId, _Node) ->
-    [?HTTP_GET, ?HTTP_POST, ?HTTP_DELETE].
+    [?HTTP_GET, ?HTTP_POST, ?HTTP_DELETE, ?HTTP_PATCH].
 
 %%--------------------------------------------------------------------
 %% @public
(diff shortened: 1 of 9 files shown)


fenollp:
are the docs up to date? https://github.com/2600hz/kazoo/blob/master/applications/crossbar/doc/system_configs.md

jamhed:
@fenollp, not really. I'd like @lazedo to review this pr first though.

applications/crossbar/src/modules/cb_system_configs.erl
```diff
@@ -280,6 +284,17 @@ validate_with_parent(Context, ConfigId, Node, Parent) ->
                                      end
                                     ).
 
+-spec validate_request(cb_context:context(), ne_binary(), ne_binary(), kz_json:object()) -> cb_context:context().
+validate_request(Context, Id, Schema, Parent) ->
+    RequestData = strip_id(cb_context:req_data(Context)),
+    FullConfig = kz_json:merge_recursive(strip_id(kz_json:public_fields(Parent)), RequestData),
+    cb_context:validate_request_data(Schema, cb_context:set_req_data(Context, FullConfig),
+                                     fun(Ctx) ->
+                                             Doc = maybe_set_id(Id, kz_json:diff(RequestData, Parent)),
```
lazedo: you should use `cb_context:doc(Ctx)` instead of `RequestData` because it will have the `defaults` applied

applications/crossbar/src/modules/cb_system_configs.erl
```diff
@@ -293,3 +308,16 @@ get_system_config(Config, Node) ->
...
     SystemNode = kz_json:get_value(Node, System, kz_json:new()),
     kz_json:merge_recursive(Default, SystemNode).
+
+-spec make_schema(ne_binary()) -> kz_json:object().
+make_schema(Id) ->
+    Flat = [
+        {[<<"patternProperties">>, <<".+">>, <<"$ref">>], kapps_config_util:system_schema_name(Id)}
+        ,{[<<"patternProperties">>, <<".+">>, <<"type">>], <<"object">>}
+        ,{<<"type">>, <<"object">>}
+    ],
```
lazedo: can we change this to
```
    Flat = [
            {<<"$schema">>,<<"http://json-schema.org/draft-04/schema#">>}
           ,{<<"id">>, <<"system_config">>}
           ,{<<"type">>, <<"object">>}
           ,{<<"default">>, kz_json:new()}
           ,{[<<"properties">>, <<"default">>, <<"$ref">>], kapps_config_util:system_schema_name(Id)}
           ,{[<<"properties">>, <<"default">>, <<"default">>], kz_json:new()}
           ,{[<<"properties">>, <<"default">>, <<"type">>], <<"object">>}
           ,{[<<"patternProperties">>, <<".+">>, <<"$ref">>], kapps_config_util:system_schema_name(Id)}
           ,{[<<"patternProperties">>, <<".+">>, <<"type">>], <<"object">>}
           ],
```
it will ensure the `default` is present and with defaults

applications/crossbar/src/modules/cb_system_configs.erl
```diff
@@ -324,14 +255,69 @@ update_db(Context) ->
...
+strip_id(JObj) -> kz_json:delete_key(<<"id">>, JObj, prune).
+
+-spec validate_with_parent(cb_context:context(), ne_binary(), ne_binary(), kz_json:object()) -> cb_context:context().
+validate_with_parent(Context, ConfigId, Node, Parent) ->
+    RequestData = strip_id(cb_context:req_data(Context)),
+    FullConfig = kz_json:merge_recursive(Parent, RequestData),
+    Schema = kapps_config_util:system_schema_name(ConfigId),
+    cb_context:validate_request_data(Schema, cb_context:set_req_data(Context, FullConfig),
+                                     fun(Ctx) ->
+                                             Doc = kz_json:set_value(Node, kz_json:diff(RequestData, Parent),
```
lazedo: see below

applications/crossbar/src/modules/cb_system_configs.erl
```diff
@@ -324,14 +255,69 @@ update_db(Context) ->
...
+    SystemNode = kz_json:get_value(Node, System, kz_json:new()),
+    kz_json:merge_recursive(Default, SystemNode).
+
+-spec make_schema(ne_binary()) -> kz_json:object().
+make_schema(Id) ->
+    Flat = [
+            {[<<"patternProperties">>, <<".+">>, <<"$ref">>], kapps_config_util:system_schema_name(Id)}
+           ,{[<<"patternProperties">>, <<".+">>, <<"type">>], <<"object">>}
+           ,{<<"type">>, <<"object">>}
+           ],
```
lazedo: can we chsnge this to
```
    Flat = [
            {<<"$schema">>,<<"http://json-schema.org/draft-04/schema#">>}
           ,{<<"id">>, <<"system_config">>}
           ,{<<"type">>, <<"object">>}
           ,{<<"default">>, kz_json:new()}
           ,{[<<"properties">>, <<"default">>, <<"$ref">>], kapps_config_util:system_schema_name(Id)}
           ,{[<<"properties">>, <<"default">>, <<"default">>], kz_json:new()}
           ,{[<<"properties">>, <<"default">>, <<"type">>], <<"object">>}
           ,{[<<"patternProperties">>, <<".+">>, <<"$ref">>], kapps_config_util:system_schema_name(Id)}
           ,{[<<"patternProperties">>, <<".+">>, <<"type">>], <<"object">>}
           ],
```
it will ensure the `default` object is present and with defaults.

applications/crossbar/src/modules/cb_system_configs.erl
```diff
@@ -59,9 +61,17 @@ init() ->
...
+authorize(Context) ->
+    cb_context:is_superduper_admin(Context)
+        andalso
+        is_system_request(Context).
+authorize(Context, _Id) -> authorize(Context).
+authorize(Context, _Id, _Node) -> authorize(Context).
+
+-spec is_system_request(cb_context:context()) -> boolean().
+is_system_request(Context) ->
+    [{Mod, _}|_] = cb_context:req_nouns(Context),
```
lazedo: this will crash if used with /v2/accounts/xxx/system_configs

jamhed: @lazedo, why?
Nouns are of type
-type cb_mod_with_tokens() :: {ne_binary(), path_tokens()}.
-type cb_mods_with_tokens() :: [cb_mod_with_tokens()].

lazedo: humm, missed the `|_` . ignore this

jamhed:
@lazedo, latest changes were discussed with @k-anderson.

lazedo:
@jamhed this doesn't work as expected. we'll discuss this later

jamhed:
@lazedo, expected by who? It works as described.

applications/crossbar/src/modules/cb_system_configs.erl
```diff
@@ -18,7 +18,8 @@
         ,validate/1, validate/2, validate/3
         ,put/2
         ,post/2, post/3
-        ,delete/2, delete/3
+        ,patch/2, patch/3
+        ,delete/2, delete/3, default/1
```
lazedo: why export `default/1` ?

jamhed: @lazedo, remnants from tests.

lazedo:
@jamhed only one use case to look at.
the `post`  for a specific node is checking the diff against the `schema default` and it should be done against `default` section with `schema defaults applied`.

```
schema "X" ; default 10
schema "Y" ; default 20
schema "W" ; default 30
schema "Z" ; default 40
"default" : { "X" :  15}
"nodea" : {"Y" : 25}
nodea with defaults  => "nodea" : {  "X" :  15}, {"Y" : 25},{"W" : 30},{"Z" : 40}} => OK
if we post "nodea" : {"X" :  15}, {"Y" : 25},{"W" : 30},{"Z" : 45}}
it saves {"Y" : 25}, {"Z" : 45}, and incorrectly {"X" :  15} to nodea
```
so it seems the diff is being made against the schema and not the `default` with schema defaults

lazedo:
@jamhed also, `mkbusy` would probably fail because of system_configs url that was fixed.
/v2/system_configs/xxxx OK
/v2/accounts/xxxx/system_configs/yyyyy will fail

lazedo:
@jamhed the `applications/crossbar/doc/ref/system_configs.md` should be merged to `applications/crossbar/doc/system_configs.md` and the text added in pr description should be added to that file as documentation.

lazedo:
@jamesaimonetti hey James, can you be the 2nd pair of eyes ? thanks

core/kazoo_config/src/kapps_config_doc.erl
```diff
@@ -0,0 +1,122 @@
...
+-spec config_with_defaults(ne_binary()) -> kz_json:object().
+config_with_defaults(Id) ->
+    apply_schema_defaults(Id, maybe_insert_default_node(apply_default_node(get_config(Id)))).
+
+-spec default_config(ne_binary(), [ne_binary()]) -> kz_json:object().
+default_config(Id, Keys) ->
+    DefaultNode = schema_defaults(Id),
+    lists:foldl(fun(Key,A) -> kz_json:set_value(Key, DefaultNode, A) end, kz_json:new(), [?DEFAULT|Keys]).
+
+-spec stored_config(ne_binary(), kz_json:object()) -> kz_json:object().
```
jamesaimonetti: I don't think `Keys` is a JSON object. Maybe `kz_json:keys()` ?

core/kazoo_config/src/kapps_config_util.erl
```diff
@@ -80,3 +81,14 @@ account_schema(Config) when is_binary(Config) ->
     Name = account_schema_name(Config),
     {'ok', Schema} = kz_json_schema:load(Name),
     Schema.
+
+-spec system_config_document_schema(ne_binary()) -> kz_json:object().
+system_config_document_schema(Id) ->
+    Flat = [
```
jamesaimonetti: I think flattened JSON objects should have keys as lists of keys. I would change the raw binary keys to [<<"key">>], personally. Can you also update the `json_proplist()` type in `kazoo_json.hrl` to allow `key() | keys()`? Makes dialyzer happier when using the `scripts/dialyze-changed.bash` to dialyze.

jamhed: > objects should have keys as lists of keys

Well, it works both ways (e.g. one can have a key as binary() or [binary()], I'm not quite sure we need to restrict this. And this behaviour comes from how set_value/3 is defined.

jamesaimonetti: I think using the `[Key,...]` when constructing a `flat_object()` helps differentiate it from a normal `object()`. `{[{[<<"a">>, <<"b">>], <<"val">>}]}` is not a valid JSON object:

```erlang
> kz_json:encode({[{[<<"a">>, <<"b">>], <<"val">>}]}).
** exception throw: {error,{invalid_object_member_key,[<<"a">>,<<"b">>]}}
   in function  jiffy:encode/2 (src/jiffy.erl, line 93)
```

So I think, when constructing a `flat_object()`, we should opt to use `[Key]` over `Key`.

Thoughts?

jamhed: Agreed.

PR #3437, opened 2017-03-14
knm-paramig_4un: parallel migration for KNM

fenollp:
renamed branch of https://github.com/2600hz/kazoo/pull/3436
(I need to sleep more)


fenollp:
master -> 4.1 merge down'll happen some time later

PR #3442, opened 2017-03-15
document how Kazoo auto-generates docs/schemas

jamesaimonetti:


jamesaimonetti:
This is in prep for PDX Erlang meetup tomorrow!

PR #3445, opened 2017-03-15
KAZOO-5412: start cowlib as well as cowboy applications

k-anderson:


lazedo:
@k-anderson master and 4.0 please

PR #3446, opened 2017-03-15
Prioritise device over user over account dialplans

kalda341:
This seems far more logical, as it means users/devices based in locations other than the main location of the account can have different dial plans.

### DIFF
--- a/core/kazoo_endpoint/src/kz_endpoint.erl
+++ b/core/kazoo_endpoint/src/kz_endpoint.erl
@@ -288,6 +288,15 @@ merge_attribute(<<"caller_id">> = Key, Account, Endpoint, Owner) ->
             CallerId = kz_json:set_value(L, Number, Merged),
             kz_json:set_value(Key, CallerId, Endpoint)
     end;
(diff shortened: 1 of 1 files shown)


lazedo:
@kalda341 hi, thanks for your suggestion.
as of now, the merge is `Account <- Device <- User` and you want to change to `Account <- User <- Device` which means that Users/Devices already have this ability (supersede account values).

the question is, in which use cases should we change the default order.

you need to enumerate some valid reasons here.

- as a user, i want my dialplan to always be the same

kalda341:
@lazedo That's a good justification, and makes complete sense to me. This pull request didn't even actually fix the problems we were having anyway, it just temporarily seemed like it did. We didn't even necessarily want `User <- Device`, we wanted `Account <- (Device | User)`, which was actually already happening. Basically we completely misunderstood how dialplans work.

All our dialplans had different keys, meaning that we would always get all the dialplans on the device, user and account. This led to really unpredictable things happening when multiple regexes matched the number dialed. This was my fix for what I thought was the problem.

Sorry to waste your time!

PR #3448, opened 2017-03-16
QF Crossbar log & kz_network_utils:verify_cidr/2

skorobkov:
Fix kz_network_utils:verify_cidr/2 (it never match)
Display IPv4 addres when it is mapped in IPv6 (0:0:0:0:0:ffff:0:0/96 [RFC](https://tools.ietf.org/html/rfc4038#section-4.2))

### DIFF
--- a/core/kazoo/src/kz_network_utils.erl
+++ b/core/kazoo/src/kz_network_utils.erl
@@ -29,6 +29,7 @@
         ,srvtuple_to_binary/1
         ,naptrtuple_to_binary/1
         ,mxtuple_to_binary/1
+        ,maybe_mapped_ipv4/1
         ]).
 -export([pretty_print_bytes/1]).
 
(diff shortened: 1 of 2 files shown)


core/kazoo/src/kz_network_utils.erl
```diff
@@ -151,7 +152,8 @@ verify_cidr(IP, CIDR) when is_binary(CIDR) ->
     verify_cidr(IP, kz_term:to_list(CIDR));
 verify_cidr(IP, CIDR) ->
     Block = inet_cidr:parse(CIDR),
-    inet_cidr:contains(Block, inet:parse_address(IP)).
+    {'ok', IPTuple} = inet:parse_address(IP),
```
skorobkov: I thought about it too.
It minimize changes, then the simpliest way is change [kz_network_utils:iptuple_to_binary/1](https://github.com/2600hz/kazoo/blob/master/core/kazoo/src/kz_network_utils.erl#L309) then it always return IPv4 address when it found 0:0:0:0:0:ffff:0:0/96 address.

skorobkov:
I thought about it too.
It minimize changes, then the simpliest way is change [kz_network_utils:iptuple_to_binary/1](https://github.com/2600hz/kazoo/blob/master/core/kazoo/src/kz_network_utils.erl#L309) then it always return IPv4 address when it found 0:0:0:0:0:ffff:0:0/96 address.

skorobkov:
@jamhed ???

jamhed:
@skorobkov, nm, making test suite work.

lazedo:
@jamhed lets try mkbusy tests with other pr (maybe use your personal repo) , merging this.

PR #3451, opened 2017-03-16
KAZOO-5333 4one: add logging when ecallmgr_config changes something

mark2600:


lazedo:
@mark2600 no need for this one for now

PR #3454, opened 2017-03-16
HELP-28755: add users.qubicle schema

mark2600:


k-anderson:
@mark2600 there is an unrelated change that I think would be removed if you rebase.  Also, have you checked that the dangling ref when qubicle is not present does not cause any issues?

k-anderson:
4.0 PR #3455

PR #3462, opened 2017-03-20
Process preflow children

koctep:


lazedo:
@koctep hi, what is the intent for this pr ? thanks

koctep:
@lazedo I was expected that would be executed full callflow because in document https://github.com/2600hz/kazoo/blob/master/applications/callflow/doc/preflow.md written that would be executed a callflow, not first callflow element.
Anyway, it may be useful to execute several modules as a preflow.

jamesaimonetti:
@koctep We discussed today and have an alternative approach that would avoid needing to walk the preflow tree and should be cleaner. Let me write up a PR and I'll ping you to test it locally and see if it makes sense for your use cases.

PR #3469, opened 2017-03-22
Fix endpoint for getting API key

andrewreedy:


applications/crossbar/doc/api_authentication.md
```diff
@@ -12,9 +12,9 @@ Get your API key for your account:
 * It can also be accessed by system administrators directly from the database by using `curl` to request the account doc from CouchDB:
 
 ```shell
-curl -v -X PUT \
+curl -v -X GET \
-    http://localhost:15984//accounts/{ACCOUNT_ID}
+    http://localhost:15984/accounts/{ACCOUNT_ID}/api_key
```
fenollp: You are right this should be a GET instead of a PUT, however this call is made against CouchDB not Kazoo so the URI is just `/accounts/{ACCOUNT_ID}`

andrewreedy: Can't this call be made directly to the Kazoo API as well?

fenollp: @k-anderson this needs reverting.

Get api_key from Couch:
`curl http://localhost:15984/accounts/{ACCOUNT_ID} | grep pvt_api_key`

Get it from Kazoo:
`curl -H 'X-Auth-Token: {AUTH_TOKEN}' http://localhost:8000/v2/accounts/{ACCOUNT_ID}/api_key`

as documented in
https://github.com/2600hz/kazoo/blob/master/applications/crossbar/doc/accounts.md#fetch-the-accounts-api-key
https://github.com/2600hz/kazoo/blob/master/applications/crossbar/doc/api_authentication.md#getting-your-api-key-from-the-api-must-already-authenticate-as-a-user

andrewreedy: @fenollp Would you like to make this change or would you like me to submit a new PR.

andrewreedy: nm this looks good now https://docs.2600hz.com/dev/applications/crossbar/doc/api_authentication/

PR #3476, opened 2017-03-23
qf-dash: return e911 outputed by Dash instead of the one inputed by t…

fenollp:
…he user


fenollp:
Don't merge yet: need to update public feature too

fenollp:
ready to merge

PR #3481, opened 2017-03-23
qf-knmbuy: skip reconciling undefined accounts

fenollp:


fenollp:
This does not need a 4.0 backport.

PR #3482, opened 2017-03-24
qf-ladder: fix dialogue reordering

fenollp:
~~Note that timeout clause in the maintenance module?
Ideas what I could do in order to remove that clause?~~


fenollp:
Ready to merge

applications/call_inspector/src/parsers/ci_parsers_util.erl
```diff
@@ -25,11 +27,13 @@
 
 %% API
 
--spec timestamp() -> api_number().
+-spec timestamp() -> float().
 timestamp() ->
-    timestamp(os:timestamp()).
+    {_, _, Micro} = os:timestamp(),
+    1.0e-6 * kz_term:to_integer(Micro) +
```
icehess: No magic number

PR #3485, opened 2017-03-27
extend private keys to auth apps

lazedo:


koctep:
@lazedo hello!
I don't know purposes of this PR and how it's going to be used but as it superseeds mine PR I want you to know that it breaks mine vision of how mine application will use kazoo core fuctions.
I wanted to use `kz_auth_jwt:encode/2 (Claims, {KeyId, Key})` for creating JWT token.

Also key should be signed by certificate so it's going to be generate by openssl in shell.
This key belongs to an account and account administrator should be able to maintain it using crossbar API.

App entity is not needed but may be useful. Anyway, issuer cannot be an app name in our case.

lazedo:
@koctep this is part of oauth. you should create a provider (if it doesn't exist already) , a app (consumer in any oauth provider) that binds to provider, and load the private key for the app if you want to use jwt flow. 
when you create a `oauth consumer` you'll get an identifier (`AppId` in kazoo) , `secret`, and more info. this information goes into `app`.

not sure what you mean with `key`. `key` is a rsa private key. the requests will be signed by that key.
in salesforce for example you have to upload or generate your certificate to your application.

PR #3501, opened 2017-03-28
KAZOO-5446: bulk add should continue after dry_run so that numbers th…

fenollp:
…at can be created are created


fenollp:
This is 4.1 only

PR #3503, opened 2017-03-29
metaflow.callflow: should "id" field be the mandatory one?

onnet:
Hi! 

Since we've got an option to look for a callflow based on captures, it seems "id" field shouldn't be strictly demanded...
https://github.com/2600hz/kazoo/blob/master/applications/konami/src/module/konami_callflow.erl#L32-L38

Regards,


lazedo:
@onnet Thanks! i had to revert because i merged it too soon, sorry. can you please reopen the pull request but also run `make apis` ?

PR #3508, opened 2017-03-29
migrate ref doc content to actual doc

jamesaimonetti:


applications/callflow/doc/ring_group.md
```diff
@@ -1,51 +1,44 @@
...
+Key | Description | Type | Default | Required
+--- | ----------- | ---- | ------- | --------
+`endpoints` | Endpoint IDs (devices, users, etc) included in the ring group | `array(object)` | `[]` | `true`
+`endpoints.[].delay` | How long to delay ringing the endpoint, in seconds | `integer` | `0` | `false`
+`endpoints.[].endpoint_type` | The type (device, user, etc) of endpoint | `string('device', 'user', 'group')` |   | `true`
+`endpoints.[].id` | The ID of the endpoint | `string(1..128)` |   | `true`
+`endpoints.[].timeout` | How long to ring the endpoint, in seconds | `integer` | `20` | `false`
+`endpoints.[].weight` | Weight of endpoint, different usage in various strategies | `integer` |   | `false`
+`fail_on_single_reject` | If any leg rejects the call, cancel all other legs | `boolean` |   | `false`
+`ignore_forward` | Whether to ring forwarded endpoints | `boolean` | `true` | `false`
```
k-anderson: @jamesaimonetti This description is not accurate. :) 

This parameter, if set to 'true', will ignore any SIP redirect requests when ringing the group.  For example, if a phone is configured with call forwarding on the device (not in Kazoo) and is part of the ring group then when the ring group is issued that phone will cause the devices to stop ringing and forward the call.  Setting this to 'true' will ignore the forward request from the device and keep ringing the other devices.

jamesaimonetti: Its what's in the schema. Have a succinct description to put there @k-anderson ?

jamesaimonetti: How about "If true, will ignore SIP redirect requests for call-forwarded devices"

jamesaimonetti:
@lazedo no word from @k-anderson, mind reviewing and merging if you don't find any faults at the moment?

PR #3509, opened 2017-03-29
looks like a typo

onnet:
https://github.com/2600hz/kazoo/blob/master/applications/konami/src/module/konami_callflow.erl#L36-L37


lazedo:
@onnet i don't think it is a typo, `captures` is correct

onnet:
@lazedo Exactly! It was 'captured' previously..

onnet:
@lazedo typo in docs, not in code..

lazedo:
@onnet need to sleep more :)

PR #3513, opened 2017-03-30
qf-tasks_4oh: retry uploading output CSVs harder

fenollp:
master: https://github.com/2600hz/kazoo/pull/3512


applications/crossbar/priv/api/swagger.json
```diff
@@ -11443,6 +11443,11 @@
                     "description": "tasks aging expiry in days",
                     "type": "integer"
                 },
+                "attempt_upload_output_times": {
```
jamesaimonetti: You can add "minimum":1 to enforce the not-zero requirement

applications/tasks/src/kz_tasks_scheduler.erl
```diff
@@ -143,25 +147,31 @@ worker_finished(TaskId=?NE_BINARY, TotalSucceeded, TotalFailed, Output=?NE_BINAR
...
     _ = gen_server:call(?SERVER, {'worker_finished', TaskId, TotalSucceeded, TotalFailed}),
     {'ok', CSVOut} = file:read_file(Output),
-    AName = ?KZ_TASKS_ANAME_OUT,
-    attempt_upload(TaskId, AName, CSVOut, Output, 3).
+    Max = ?UPLOAD_ATTEMPTS,
+    attempt_upload(TaskId, ?KZ_TASKS_ANAME_OUT, CSVOut, Output, Max, Max).
 
-attempt_upload(_TaskId, _AName, _, _, 0) ->
-    lager:error("failed saving ~s/~s: 3rd conflict", [_TaskId, _AName]),
+attempt_upload(_TaskId, _AName, _, _, 0, _) ->
```
jamesaimonetti: Why not start Retries at 1 and go until `Max, Max)` here? Then you don't have to do math at line 159 below

applications/tasks/src/kz_tasks_scheduler.erl
```diff
@@ -143,25 +147,31 @@ worker_finished(TaskId=?NE_BINARY, TotalSucceeded, TotalFailed, Output=?NE_BINAR
...
+    lager:error("failed saving ~s/~s: last failing attempt", [_TaskId, _AName]),
     {error, conflict};
-attempt_upload(TaskId, AName, CSVOut, Output, Retries) ->
-    lager:debug("attempt #~p to save ~s/~s", [3-Retries+1, TaskId, AName]),
+attempt_upload(TaskId, AName, CSVOut, Output, Retries, Max) ->
+    Retries =/= Max
+        andalso pause_between_upload_attempts(TaskId),
```
jamesaimonetti: Why not pause in the `{error, _R}` clause after logging?

applications/crossbar/priv/api/swagger.json
fenollp: I am very wary about manually adding things to swagger.json for 2 reasons:
1. no one uses this file
2. chances are these manual edits will get deleted by the next `make apis`

applications/tasks/src/kz_tasks_scheduler.erl
fenollp: What's wrong with some math?

applications/crossbar/priv/api/swagger.json
jamesaimonetti: Sorry, meant to put that in the schema - it will auto-populate here.

fenollp: `make apis` won't let me do that...

jamesaimonetti:
@fenollp added it, will forward-port to master. Was small enough and needed doing to support the schema change, so I would argue its a relevant change.

PR #3514, opened 2017-03-30
Auth client

lazedo:


core/kazoo/src/kz_maps.erl
```diff
@@ -0,0 +1,70 @@
...
+
+%% ====================================================================
+%% API functions
+%% ====================================================================
+-export([merge/2]).
+-export([get/2, get/3]).
+-export([keys_to_atoms/1, keys_to_atoms/2]).
+
+-spec merge(map(), map()) -> map().
+merge(Map1, Map1) -> Map1;
```
jamesaimonetti: Can you add unit tests so we can see the intended usages/results? I think there may be more efficient ways to do this merge but I wouldn't want to break functionality, so tests would be great to make sure there isn't a regression.

lazedo: @jamesaimonetti added the test

jamesaimonetti:
@lazedo i rebased and added perf testing for the merge functions. Want to try locally `make perf` in `core/kazoo` to make sure you see similar speed differences?

core/kazoo/src/kz_maps.erl
```diff
@@ -0,0 +1,125 @@
...
+%% ====================================================================
+-export([merge/2
+        ,merge_r/2
+        ,merge_l/2
+        ]).
+-export([get/2, get/3]).
+-export([keys_to_atoms/1, keys_to_atoms/2]).
+
+-spec merge(map(), map()) -> map().
+merge(Map1, Map1) -> Map1;
```
fenollp: Can you add an is_map(Map1) on this line?

PR #3522, opened 2017-04-02
Added cb_me endpoint to alias /accounts/{AUTH_ACCOUNT_ID}

kalda341:
It would be useful to be able to use /me instead of /accounts/{AUTH_ACCOUNT_ID}. This mimics many other APIs, and would make it easier to integrate Crossbar with other services. It also increases the usability of Crossbar as this is what most uses want for every API request.

This currently doesn't work with the autogenerated API documentation as it calls methods in cb_accounts as part of allowed_methods. Would you prefer I hardcoded the values obtained from cb_accounts, or kept it as it is and manually wrote the documentation?

### DIFF
--- a/applications/crossbar/src/modules/cb_account.erl
+++ b/applications/crossbar/src/modules/cb_account.erl
@@ -5,7 +5,7 @@
 %%% @contributors:
 %%%   Max Lay
 %%%-------------------------------------------------------------------
--module(cb_me).
+-module(cb_account).
 
 -export([init/0
         ,allowed_methods/0, allowed_methods/1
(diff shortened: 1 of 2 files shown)


lazedo:
@kalda341 thanks for this. i need to discuss this with the team as the `me` endpoint was kind of reserved for oauth server integration and targeted at the user.

lazedo:
@kalda341 hey, the `me` will be reserved for the user `me => accounts/xxx/users/yyy`, ex: `/v2/me/devices` or `/v2/faxes` or `/v2/recordings`.  

as for the `shortcut` / `alias` you are asking, we suggest `account`, which will allow `/v2/account/devices` or `/v2/account/users`. 

our goal is to have something `/v2/devices` or `/v2/users` work without issues but there's a lot of work to be done to make it possible.

thoughts ?

kalda341:
Sounds sensible to me!

kalda341:
@lazedo I've updated it to use /account

lazedo:
@kalda341 thanks. can you rebase with latest master so CI errors go away ?

kalda341:
@lazedo Done :)

lazedo:
@kalda341 thanks again.

PR #3528, opened 2017-04-03
qf-ladder: properly accumulate filtered callids + interesting macro fix

fenollp:


applications/crossbar/src/modules/cb_call_inspector.erl
```diff
@@ -188,30 +188,38 @@ send_chunked_cdrs([Db | Dbs], {Req, Context}) ->
     {Context3, CDRIds} = cb_cdrs:maybe_paginate_and_clean(Context2, Ids),
     send_chunked_cdrs(Dbs, cb_cdrs:load_chunked_cdrs(Db, CDRIds, {Req, Context3})).
 
--spec get_cdr_ids(ne_binary(), ne_binary(), crossbar_doc:view_options()) ->
-                         kz_proplist().
-get_cdr_ids(Db, View, ViewOptions) ->
-    {'ok', Ids} = cb_cdrs:get_cdr_ids(Db, View, ViewOptions),
-    filter_cdr_ids(Ids).
+-define(MATCH_CDR_ID(YYYYMM, CallId), <<YYYYMM:6/binary, "-", CallId/binary>>).
```
mark2600: This macro is never called with a first parameter that is not '_'.  Is that parameter needed at all ?

fenollp: It's not needed for this crossbar module but since this data comes from cb_cdrs it is probably used over there

PR #3534, opened 2017-04-04
HELP-29236: try to save voicemail at least 3 times

icehess:
Try save document at 3 times if error is `timeout` or `conflict`.

Also archive the prepend message if merge audio files was unsuccessful, so it can be recovered later by sysadmin.

### DIFF
--- a/core/kazoo_voicemail/src/kvm_message.erl
+++ b/core/kazoo_voicemail/src/kvm_message.erl
@@ -758,6 +758,7 @@ retry_conflict(Call, JObj, DieAnotherDay) ->
         {'error', _}=Error -> Error
     end.
 
+-spec check_for_collision(kapps_call:call(), kz_json:object(), kz_json:object(), fun((kz_json:object()) -> db_ret())) -> db_ret().
 check_for_collision(Call, JObj, SavedJObj, DieAnotherDay) ->
     OrigPublic = kz_json:public_fields(JObj),
     SavedPublic = kz_json:public_fields(SavedJObj),
@@ -781,6 +782,7 @@ check_for_collision(Call, JObj, SavedJObj, DieAnotherDay) ->
             DieAnotherDay(kz_doc:set_id(NewJObj, NewId))
     end.
 
+-spec give_me_another_id(ne_binary()) -> ne_binary().
 give_me_another_id(?MATCH_MODB_PREFIX(Year, Month, _)) ->
     ?MODB_MSG_ID(Year, Month, kz_binary:rand_hex(16)).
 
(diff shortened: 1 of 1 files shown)


core/kazoo_voicemail/src/kvm_message.erl
```diff
@@ -337,6 +354,10 @@ do_copy(AccountId, FromId, OldBoxId, NBId, NBoxJ) ->
                          ,[FromDb, FromId, ToDb, ToId]
                          ),
             NotFound;
+        {'error', 'timeout'} ->
+            do_copy(AccountId, FromId, OldBoxId, NBId, NBoxJ, Loop - 1);
+        {'error', 'conflict'} ->
```
mark2600: What would cause a conflict in this situation ?  Also what would resolve it ? Unless it is something external to this process retrying the same operation would never succeed, correct ?

icehess: I don't what would cause this but I know this was happened before. This is not resolving why conflict is happening.

mark2600: So then why retry 3 times when we know it will just continue to fail ?

core/kazoo_voicemail/src/kvm_message.erl
```diff
@@ -637,3 +740,63 @@ update_metadata(AccountId, BoxId, MessageId, UpdateFuns) ->
...
+-spec retry_conflict(kapps_call:call() | 'undefined', kz_json:object(), fun((kz_json:object()) -> db_ret())) -> db_ret().
+retry_conflict(Call, JObj, DieAnotherDay) ->
+    case fetch(kz_doc:account_id(JObj), kz_doc:id(JObj)) of
+        {'ok', SavedJObj} -> check_for_collision(Call, JObj, SavedJObj, DieAnotherDay);
+        {'error', 'timeout'} -> DieAnotherDay(JObj);
+        {'error', 'not_found'} -> DieAnotherDay(JObj);
+        {'error', _}=Error -> Error
+    end.
+
+check_for_collision(Call, JObj, SavedJObj, DieAnotherDay) ->
```
k-anderson: @icehess missing spec

core/kazoo_voicemail/src/kvm_message.erl
```diff
@@ -637,3 +740,63 @@ update_metadata(AccountId, BoxId, MessageId, UpdateFuns) ->
...
+                               ),
+            lager:critical(Msg),
+            Subject = <<"Document ID collision detected">>,
+            send_system_alert(Call, kz_doc:account_id(JObj), Subject, kz_term:to_binary(Msg)),
+            NewId = give_me_another_id(kz_doc:id(JObj)),
+            NewJObj = kzd_box_message:update_media_id(NewId, JObj),
+            DieAnotherDay(kz_doc:set_id(NewJObj, NewId))
+    end.
+
+give_me_another_id(?MATCH_MODB_PREFIX(Year, Month, _)) ->
```
k-anderson: @icehess and here (spec)

PR #3535, opened 2017-04-04
HELP-29219: walk to reseller then system_config in cb_notification

icehess:
Walk up in account up to the first reseller and then system_config for getting templates in crossbar notifications. same logic as how templates are fetching in teletype.

Advantage is users see the same teletype that they're getting from teletype.

### DIFF
--- a/applications/crossbar/src/modules/cb_notifications.erl
+++ b/applications/crossbar/src/modules/cb_notifications.erl
@@ -724,9 +724,9 @@ read_account(Context, Id, LoadFrom) ->
          }
     of
         {404, 'error'} when LoadFrom =:= 'system'; LoadFrom =:= 'system_migrate' ->
-            maybe_read_from_parent(Context, Id, LoadFrom, cb_context:fetch(Context, 'reseller_id'));
+            maybe_read_from_parent(Context, Id, LoadFrom, cb_context:reseller_id(Context));
         {_Code, 'success'} ->
-            lager:debug("loaded ~s from account database", [Id]),
+            lager:debug("loaded ~s from account database ~s", [Id, cb_context:account_db(Context)]),
             NewRespData = note_account_override(cb_context:resp_data(Context1)),
             cb_context:set_resp_data(Context1, NewRespData);
         {_Code, _Status} ->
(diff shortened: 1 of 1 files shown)


fenollp:
What about using `kapps_account_config:get_from_reseller/4`?

icehess:
@fenollp it's not walking account tree to reseller, it jumps to reseller.

applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -97,6 +99,19 @@ allowed_methods(?SMTP_LOG, _SMTPLogId) ->
...
+-spec authorize(cb_context:context()) -> boolean().
+authorize(Context) ->
+    authorize(Context, cb_context:req_nouns(Context), cb_context:account_id(Context)).
+
+-spec authorize(cb_context:context(), req_nouns(), api_binary()) -> boolean().
+authorize(_Context, [{<<"notifications">>, [?NE_BINARY=_Id]}], 'undefined') ->
+    lager:debug("allowing system notifications request"),
+    'true';
+authorize(Context, [{<<"notifications">>, _}, {<<"accounts">>, [AccountId]}], AccountId) ->
+    cb_simple_authz:authorize(Context);
```
k-anderson: @icehess I don't think this is required, the simple_authz will preform it independently due to the map (or fold).  Can you test without this?

applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -724,17 +739,35 @@ read_account(Context, Id, LoadFrom) ->
...
-            lager:debug("loaded ~s from account database", [Id]),
+            lager:debug("loaded ~s from account database ~s", [Id, cb_context:account_db(Context)]),
             NewRespData = note_account_override(cb_context:resp_data(Context1)),
             cb_context:set_resp_data(Context1, NewRespData);
         {_Code, _Status} ->
             lager:debug("failed to load ~s: ~p", [Id, _Code]),
             Context1
     end.
 
+maybe_read_from_parent(Context, Id, LoadFrom, 'undefined') ->
```
k-anderson: @icehess missing spec

PR #3540, opened 2017-04-05
[KAZOO-5359] Allow manual presence to be set for user or device through crossbar

kalda341:


lazedo:
@kalda341 why did made it as a path token and not extend the already existing post ? presently post does reset but could be extended.

kalda341:
The reset style endpoint like an odd way of doing it as there would never be any other data in the body. Additionally, you would need to prevent the case where reset and the state is set. This follows the pattern in cb_port_requests. IMO the reset should work in the same way, but I didn't want to create any breaking changes.

lazedo:
we're moving away from that pattern. check the channels api

kalda341:
I've updated the code to be more like cb_channels.

PR #3544, opened 2017-04-06
erlang_terms: move modules to stdlib

jamesaimonetti:
modules that deal with manipulating data structures are moved to
kazoo_stdlib (much like lists and such are in Erlang's stdlib). This
also has the nice property of removing any remote function calls to
core applications, simplifying the dependency graphs for Kazoo
applications.


jamesaimonetti:
@lazedo not sure why makebusy is not reporting back here...but everything else looks good afaict.

jamhed:
@jamesaimonetti i think this will pass makebusy after rebase.

k-anderson:
Hey all, hold of on accepting this until we can finalize the 4.1 merge down.  Nothing wrong with the PR just keeping it for 4.2

jamesaimonetti:
@jamhed looks like system config API tests are failing? Otherwise, @k-anderson, this should be ready for 4.2 / master when the merge down is done

k-anderson:
Hey, nm ya this should be in 4.1

jamhed:
@jamesaimonetti, makebusy should pass against master greater than 87d1f1052

PR #3547, opened 2017-04-06
Repeat service reconciles: Remove unused/bad return type

danielfinke:
The {'error', 'no_change'} return value of maybe_save isn't used, but causes an error such as in https://github.com/2600hz/kazoo/blob/master/core/kazoo_number_manager/src/knm_services.erl#L135. The return value is used as if it were a services record. Making the change in kz_services:maybe_save/1 because nothing uses the error tuple.


fenollp:
Thanks! (I thought I fixed that, apparently not so many thanks!)

PR #3546, opened 2017-04-06
KAZOO-5467 support url parameters

lazedo:

PR #3552, opened 2017-04-07
KAZOO-5467 support url parameters 4.0

lazedo:


lazedo:
master https://github.com/2600hz/kazoo/pull/3546

PR #3554, opened 2017-04-08
KAZOO-5467 improved support url parameters

lazedo:


lazedo:
4.0 https://github.com/2600hz/kazoo/pull/3555

kalda341:
This doesn't seem to work for our case. I would fix it and create a pull request myself, but I don't understand what has been done here well enough to be confident in making that change. 
The call recording URL is set up as http://DOMAIN/recording.php?ID=c94d302eb2c8d900962b756f9ae29c39. 

The URL that the recording is sent to is 23:05:09.738 debug kz_http.218 1005.32ms: put http://DOMAIN/recording.php?ID=c94d302eb2c8d900962b756f9ae29c39/&recording=call_recording_4b26352e-25d7c87%40192.168.1.116.mp3 ...

My approach would have been to remove the existing query parameters from the URL, and add them to field list, but as I said, I don't really understand what is being done here.

Thanks :)

lazedo:
@kalda341 the problem with ´/´ is here (needs to check if there are params and only add the ´/´ if not) https://github.com/2600hz/kazoo/blob/master/core/kazoo_attachments/src/kz_att_http.erl#L33

some legacy have the url as `http://somewhere/mypscrip.php?recording=` and we need to support it.

you url would also require extra work before since the `/` was there before recording, right ?
http://domain/recording.php?ID=c94d302eb2c8d900962b756f9ae29c39/recording=call_recording_4b26352e-25d7c87%40192.168.1.116.mp3?otherparametershere

lazedo:
@kalda341 https://github.com/2600hz/kazoo/pull/3576 should fix your usage. thanks for reporting

PR #3558, opened 2017-04-10
check conflict when saving parked calls - 4.0

lazedo:


fenollp:
MB tests failed, restarting.

lazedo:
@fenollp 4.0 MB will fail according to @jamhed

PR #3569, opened 2017-04-11
Fix error by unifying return types

jamhed:


lazedo:
@jamhed duplicate from https://github.com/2600hz/kazoo/pull/3565 ? i asked for your revision there.

jamhed:
@lazedo, right, I've missed it.

PR #3571, opened 2017-04-11
move hangup to tail of ecallmgr execution queue

lazedo:
giving time for other actions in ecallmgr execution queue to complete


fenollp:
Dialyzer is asking for #state.queue to be api_ne_binary()

PR #3578, opened 2017-04-12
temporary fix for mkbusy tests

lazedo:


lazedo:
@jamhed since the `say` is not failing with this commit, we may need more time to detect the `message_menu` . take this in consideration if you change the language to en-mb

jamhed:
Superceded by latest MakeBusy changes.

PR #3580, opened 2017-04-12
move compaction to a task, configurable per-node

jamesaimonetti:


core/kazoo_couch/src/kz_couch_view.erl
```diff
@@ -39,15 +39,21 @@ design_info(#server{}=Conn, DBName, Design) ->
...
     Db = kz_couch_util:get_db(Conn, DBName),
-    Filter = [{'startkey', <<"_design/">>}
-             ,{'endkey', <<"_design0">>}
-              | Options
-             ],
+    Filter = props:set_values([{'startkey', <<"_design/">>}
+                              ,{'endkey', <<"_design0">>}
+                              ]
+                             ,Options
+                             ),
```
lazedo: this change doesn't work. `props:set_values/2` will remove `single` value options like `include_docs`.
either revert here or support preservation of `single` value in `props:set_value/3` (preferred).
```
-spec set_values(kz_proplist(), kz_proplist()) -> kz_proplist().
set_values([], Props) -> Props;
set_values([{K, V}|KVs], Props) ->
    set_values(KVs, set_value(K, V, Props)).

-spec set_value(kz_proplist_key(), kz_proplist_value(), kz_proplist()) ->
                       kz_proplist().
set_value(K, V, Props) ->
    [{K, V} | [KV || {Key, _}=KV <- Props, K =/= Key]].
```

jamesaimonetti: @lazedo all set, added proper testing and updated set_value(s) to handle it

applications/tasks/doc/README.md
```diff
@@ -120,3 +116,40 @@ Be careful as a rate too low may corrupt a task's state.
...
+    _ = tasks_bindings:bind(?TRIGGER_ALL_DBS, ?MODULE, 'handle_database').
+```
+
+Find the trigger macros in [the tasks header](../src/tasks.hrl). This particular example will bind the module's `handle_database/1` to be run each time a database is processed by the [`kz_tasks_trigger`](../src/kz_tasks_trigger.erl) process.
+
+Cron triggers will need an arity-0 function to callback to; database triggers call an arity-1 function.
+
+### Callback function
+
+All modules bound to a particular trigger will run in serial, so be mindful of that. If your operation is quick, do it directly. If it has the potential to take a while, consider spawning the work so the other modules can get to their business too.
```
fenollp: It's probably time to add kazoo_bindings:pmap/2

jamesaimonetti: Agreed

PR #3595, opened 2017-04-18
Updated app schema to allow for masqueradable field

JRMaitre:


applications/crossbar/priv/couchdb/schemas/app.json
```diff
@@ -63,6 +63,12 @@
             "required": true,
             "type": "string"
         },
+        "masqueradable": {
```
jamesaimonetti: Should be "masquerade-able", non? Or "can_be_masqueraded" or "masquerading_is_allowed"

PR #3596, opened 2017-04-18
use configured doc name when saving the parked calls doc

jamesaimonetti:


jamesaimonetti:
@lazedo I don't think this is an issue but just in case, it would be wise to use the configured doc name as the id/type since that's what `cb_parked_calls` uses.

PR #3598, opened 2017-04-18
reseller_bookkeeper_lookup

onnet:
Hi,
At the moment only kz_bookkeeper_local could be used to bill non master reseller's children.
This reduces the benefit of reseller tree feature.
I would like to propose a little step forward in order subresellers could also have fully featured billing solution within their Kazoo platform tenant  
Regards,
Kirill

### DIFF
--- a/core/kazoo_services/src/kz_services.erl
+++ b/core/kazoo_services/src/kz_services.erl
@@ -556,8 +556,8 @@ activation_charges(CategoryId, ItemId, Account=?NE_BINARY) ->
 %% @end
 %%--------------------------------------------------------------------
 -spec commit_transactions(services(), kz_transactions:kz_transactions()) -> atom().
-commit_transactions(#kz_services{billing_id=BillingId}, Activations) ->
-    Bookkeeper = select_bookkeeper(BillingId),
+commit_transactions(#kz_services{billing_id=BillingId}=Services, Activations) ->
+    Bookkeeper = select_bookkeeper(Services),
     Transactions = [Activation
                     || Activation <- Activations
                            ,kz_transaction:amount(Activation) > 0
@@ -571,8 +571,8 @@ commit_transactions(#kz_services{billing_id=BillingId}, Activations) ->
 %% @end
 %%--------------------------------------------------------------------
 -spec charge_transactions(services(), kz_transactions:kz_transactions()) -> kz_json:objects().
-charge_transactions(#kz_services{billing_id=BillingId}, Activations) ->
-    Bookkeeper = select_bookkeeper(BillingId),
+charge_transactions(#kz_services{billing_id=BillingId}=Services, Activations) ->
+    Bookkeeper = select_bookkeeper(Services),
     Transactions = [kz_transaction:to_json(Activation)
                     || Activation <- Activations
                            ,kz_transaction:amount(Activation) > 0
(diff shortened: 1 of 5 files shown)


k-anderson:
@onnet howdy, I added a comment as a PR review and it didn't flag in the list that there is a comment.  So I am adding this in case you didn't see the prior.   Thanks!

onnet:
@k-anderson Hi Karl, I can't see it anyway. Dashboard (top left) says "No reviews".
Could you please put it as a comment over here. Thanks in advance.
Regards,

core/kazoo_services/src/kz_services.erl
```diff
@@ -592,7 +592,11 @@ select_bookkeeper(BillingId) ->
     ResellerId = get_reseller_id(BillingId),
     {'ok', MasterAccountId} = kapps_util:get_master_account_id(),
     case ResellerId =/= MasterAccountId of
-        'true' -> 'kz_bookkeeper_local';
+        'true' ->
+            case ?KZ_RESELLER_BOOKKEEPER_LOOKUP of
+                'true' -> ?KZ_ACCOUNT_BOOKKEEPER(BillingId);
```
k-anderson: @onnet Howdy, this is a very nice addition, however I would expect KZ_ACCOUNT_BOOKKEEPER to look up the from the ResellerId rather than the BillingId. Using the BillingId would allow each subaccount to have a different bookkeeper, but it would be our intention that the bookkeeper be configured at the reseller level.

IE: If an account is flagged as a reseller then they can set the bookkeeper they would like to use on their account, and it would implicitly apply to all sub-accounts. As it stands it would have to be configured on each of the resellers subaccounts.

Let us know what you think, thanks!

onnet: @k-anderson agreed, will do it asap. Was thinking about BillingId as an AccountId of another child of the same reseller which means the same bookkeeper at the end of the day...
And yep, setting BillingId as an AccountId from another sub/resellers thread will make the game much more complex :)))
Thanks for an explanation, Karl!

onnet:
@k-anderson 
So, here is the new logic:
- if BillingId is direct master_account's child, we are OK to use master_account's bookkeeper, do we? :)
- if BillingId and AccountId are children of the same reseller, we chasing for bookkeeper vertically with help of kapps_account_config:get_global/4 starting from ResellerId 
- in case BookkeeperId is From Somewhere Out There, we do not bother and provide 'kz_bookkeeper_local'

What do you think?

onnet:
@k-anderson Hi Karl. Ping... :) I'm afraid I can't see your comment as it was previously. Was there any?
Regards, Kirill

onnet:
@k-anderson ping...

onnet:
MRP - stands for Monthly Recurrent Ping :))

k-anderson:
@onnet Howdy!

I am so sorry, its inexcusable :(  I just completed a QA of this and looks good!


In the initial version I could visualize the ramifications of the change and had one update but the resulting change was more than I expected.  Since its a more complex code path I needed to find time to fully ensure that all was well.  Additionally, (as you hinted on the other PR) we are about to work on billing to decouple from Braintree which I was hoping that assignee would look at this then, but that got delayed too.  With all this I kept thinking I would get to then not, and apologize.

onnet:
@k-anderson Hi Karl, thanks for spending time for that. Can move forward now to tame billing unicorn... :) And no need to apologize, it's your, guys, sandbox and your rules, I was prepared to wait much longer :)  Thanks again!

PR #3599, opened 2017-04-18
support dialplan for refers/redirects

lazedo:


lazedo:
4.0 https://github.com/2600hz/kazoo/pull/3604

PR #3602, opened 2017-04-19
try to catche in kazoo binding for map

icehess:
useful for debugging crashes (e.g. crashes in teletype modules)

### DIFF
--- a/core/kazoo_bindings/src/kazoo_bindings.erl
+++ b/core/kazoo_bindings/src/kazoo_bindings.erl
@@ -651,7 +651,7 @@ fold_bind_results([#kz_responder{module=M
                  ,Route
                  ,RespondersLen
                  ,ReRunResponders) ->
-    try apply_responder(Responder, Payload) of
+    try erlang:apply(M, F, Payload) of
         'eoq' ->
             lager:debug("putting ~s to eoq", [M]),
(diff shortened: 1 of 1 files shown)


core/kazoo_bindings/src/kazoo_bindings.erl
```diff
@@ -829,19 +829,29 @@ pmap_processor_fold(#kz_binding{binding_parts=BParts
...
         ++ Acc.
 
+-spec try_map_responders(map_responder_fun(), any()) -> any().
+try_map_responders(MapFun, Responder) ->
+    try MapFun(Responder)
+    catch
+        _T:_E ->
+            ST = erlang:get_stacktrace(),
+            lager:error("exception: ~s: ~p", [_T, _E]),
+            kz_util:log_stacktrace(ST)
```
fenollp: 1. this breaks the expected return value (it now returns `ok` on crashes)
1. not all binding crashes need to be logged.

Ideas?

jamesaimonetti: To add an example, crossbar modules may bind for `content_types_provided` bindings but only define the function when the path includes an ID (so `/things/{ID}` but not `/things`). We don't care that this crashes when running the binding for `/things`; its an "expected" crash.

jamhed:
@icehess see http://docker.2600hz.com/status.php?ref=601bd7f5eb&type=kazoo

PR #3609, opened 2017-04-19
Test please ignore

jamhed:


UzairM:
Hi ,
 I am still testing your docker and it is working .But the issue i am facing is that when i am testing a call ,it enables a connection but it cant send voice.

PR #3612, opened 2017-04-19
ktn-DYWTKM: get more information & power out of number tasks

fenollp:
4.0 version coming up


core/kazoo_number_manager/src/providers/knm_failover.erl
```diff
@@ -42,19 +39,19 @@ save(Number, _State) ->
 %% @end
 %%--------------------------------------------------------------------
 -spec delete(knm_number:knm_number()) -> knm_number:knm_number().
-delete(Number) ->
-    case feature(Number) =:= 'undefined' of
-        'true' -> Number;
-        'false' -> knm_services:deactivate_feature(Number, ?KEY)
+delete(N) ->
```
jamesaimonetti: This hunk seems unnecessary. Number is clearer than N, there's no change to functionality, so why the change?

core/kazoo_number_manager/src/providers/knm_failover.erl
```diff
@@ -63,16 +60,12 @@ feature(Number) ->
...
-        'false' when NotChanged ->
-            Number;
-        'false' ->
-            knm_services:activate_feature(Number, {?KEY, Failover})
+update_failover(N) ->
+    Private = feature(N),
+    Public = kz_json:get_ne_value(?KEY, knm_phone_number:doc(knm_number:phone_number(N))),
+    NotChanged = kz_json:are_equal(Private, Public),
+    case kz_term:is_empty(Public) of
+        true -> knm_services:deactivate_feature(N, ?KEY);
```
jamesaimonetti: Atoms are still to be tickied, changing them here is not helpful in that regard. Please revert

core/kazoo_number_manager/src/knm_number.erl
```diff
@@ -30,6 +30,7 @@
         ,charges/2, set_charges/3
         ,to_public_json/1
         ,is_number/1
+        ,force_outbound_feature/1
```
jamesaimonetti: How is this related to tasks cleanup?

core/kazoo_csv/download_csvs.py
```diff
@@ -1,19 +0,0 @@
-#!/usr/bin/env python2
-
-import requests, sys, json
-
-print("version: ", requests.__version__)
-csvs_url = 'https://api.github.com/repos/maxogden/csv-spectrum/contents/csvs'
```
jamesaimonetti: I disagree with removing this script/step from testing. Just dumping the CSVs into the repo removes attribution from where they came from.

core/kazoo_csv/Makefile
```diff
@@ -3,12 +3,4 @@ PROJECT = kazoo_csv
 
 all: compile
 
-test: check_csvs
-
-check_csvs:
-	@stat -t -- test/*.csv >/dev/null 2>/dev/null ; if [[ $$? > 0 ]] ; then echo WARNING: run 'make csvs' for more test data ; fi
```
jamesaimonetti: Revert this change too

core/kazoo_stdlib/src/kz_binary.erl
```diff
@@ -213,3 +213,20 @@ ucfirst(<<F:8, Bin/binary>>) -> <<(kz_term:to_upper_char(F)):8, Bin/binary>>.
...
+
+-spec pos(char(), binary()) -> non_neg_integer() | -1.
+pos(Char, Bin) ->
+    pos(Char, Bin, 0).
+pos(_, <<>>, _) -> -1;
+pos(Char, <<Char:8, _/binary>>, N) -> N;
+pos(Char, <<_:8, Bin/binary>>, N) ->
+    pos(Char, Bin, N + 1).
+
+-spec closests([char(),...], binary()) -> [{char(), non_neg_integer()}].
```
jamesaimonetti: This function is unused in the PR. What is its purpose? Seems like clutter at this point. Move it to another PR that shows its utility to the project.

applications/tasks/src/tasks_maintenance.erl
```diff
@@ -110,13 +108,6 @@ start(TaskId) ->
         {'error', Reason} -> print_error(Reason)
     end.
 
--spec restart(text()) -> 'no_return'.
-restart(TaskId) ->
```
jamesaimonetti: I disagree with removing this. I find it quite useful when debugging tasks that failed to run because of errors in the code. Avoids having to create/start the job all over again, and I can rerun them many times as needed.

applications/tasks/src/modules/kt_numbers.erl
```diff
@@ -259,6 +330,30 @@ account_id(_) -> 'false'.
 carrier_module(Data) ->
     lists:member(Data, knm_carriers:all_modules()).
 
+-spec ported_in(ne_binary()) -> boolean().
+ported_in(Cell) -> is_cell_boolean(Cell).
+
+-spec 'cnam.inbound'(ne_binary()) -> boolean().
+'cnam.inbound'(Cell) -> is_cell_boolean(Cell).
+
+-spec 'prepend.enabled'(ne_binary()) -> boolean().
```
jamesaimonetti: Typically functions that return booleans are named to suggest that. `is_prepend_enabled` would be better IMO. `is_cnam_inbound_enabled` above. For someone that doesn't like ticky-ing atoms, its an odd choice to create function names that would require the tickies! ;)

core/kazoo_number_manager/src/providers/knm_failover.erl
fenollp: This is KNM nomenclature. Number is very ambiguous.

core/kazoo_number_manager/src/knm_number.erl
fenollp: Is this PR about tasks cleanup? No.
It's about (see title) and this function is needed for people to be able to set this feature from a CSV job.

core/kazoo_csv/download_csvs.py
fenollp: Attribution for 327 bytes of text?
1. Attribution is right there in our git history
2. These tests were never run by CI
3. These tests did not pass CI

I want them here so that they are actually used.

core/kazoo_csv/Makefile
fenollp: See above

core/kazoo_stdlib/src/kz_binary.erl
fenollp: > What is its purpose? Seems like clutter at this point.

Right, because tests are so hard to read.
This function was introduced for the CSV parser but is no longer used.
I did not remove it because I think it can be useful in the future.

applications/tasks/src/tasks_maintenance.erl
fenollp: As we discussed multiple times before, the task scheduler was never meant to support restarting tasks. If this was really intended as a debugging command then its implementation would be within 
ifdef(TEST)...endif boundaries.

applications/tasks/src/modules/kt_numbers.erl
fenollp: For someone that implemented a task, you should know why this is named that way.
BTW tickies *are* needed here and this is the only justifiable usage of tickies possible: when removing tickies induces a parse error.

core/kazoo_number_manager/src/providers/knm_failover.erl
fenollp: It's way harder for me to read though: I'm colorblind.

core/kazoo_number_manager/src/providers/knm_failover.erl
jamesaimonetti: I thought `Number` is knm_number and `PhoneNumber` is knm_phone_number`?

core/kazoo_csv/download_csvs.py
jamesaimonetti: Yes, attribution is important. Why force people to dig through the git history to find out where the files came from. We didn't create them; they did.

If CI isn't running them, CI should be updated to run them. If they don't pass, that's a regression cause tests passed when I added them

core/kazoo_stdlib/src/kz_binary.erl
jamesaimonetti: Well, the utility of the function is unclear, there's nothing showing why its a useful function, its unused AKA dead code.

applications/tasks/src/tasks_maintenance.erl
jamesaimonetti: "Never meant to" doesn't mean "never will". I am showing you a use case where I start a job via API, it fails because the code didn't work properly, I adjust the code and restart the task. Its really quite simple, doesn't impact existing functionality, only exists for admins via SUP...

applications/tasks/src/modules/kt_numbers.erl
jamesaimonetti: I have implemented a couple tasks now and have yet to see a reason for 'function.name' style naming. Why are the alternatives suggested not reasonable?

And read my last sentence again - I fully understand why the tickies are needed; just pointing out that its ironic an opponent of the ticky for atoms would purposefully choose a function name that requires the ticky to work (esp in light of there being alternatives that don't require tickies to be compiled).

fenollp: The function name has to match the header name.

core/kazoo_number_manager/src/providers/knm_failover.erl
jamesaimonetti: https://github.com/2600hz/kazoo/blob/master/CONTRIBUTING.md: "Never introduce a single letter variable."

core/kazoo_number_manager/src/providers/knm_failover.erl
jamesaimonetti: https://github.com/2600hz/kazoo/blob/master/CONTRIBUTING.md "Tickie all atoms "

lazedo:
@jamesaimonetti the repo for csv looks dead (4 years last update). single letter variable can be addressed in a different PR. tickies can be back to normal with tool to add/help developers. to rethink, function naming in task

PR #3613, opened 2017-04-19
add displayName and photoUrl to profile properties

lazedo:


lazedo:
monster-ui https://github.com/2600hz/monster-ui/pull/86

PR #3614, opened 2017-04-20
PATCH fix

tsromana:
while doing patch request the doc never got saved as the patch was made in reverse order as the db doc was being patched on top of request doc


lazedo:
@tsromana good catch. i think the problem lies elsewhere, namely `kz_json:merge` which defaults to `merge_left`. we will debate this internally and give you feedback.

tsromana:
awesome, thanks

PR #3618, opened 2017-04-20
PISTON-421: allow teletype to use system-level templates as fallback

danielfinke:
It seems useful to allow accounts to modify the notification docs for a teletype notification while using the system level template. This allows us to make a change to the template in system_config and have it apply to any accounts that have overridden fields such as "from" in the notification doc but not provided their own template. (That's our use case)
This is currently possible by doing a POST to /v[12]/accounts/<account_id>/notifications/<notification_id> with just the metadata fields and no attachment. Without the modifications in this PR, the notification doesn't work when a POST is done this way. With them, if a notification doc doesn't have attachments going up the tree, it will fall back to the system_config template attachments.

### DIFF
--- a/applications/crossbar/src/modules/cb_notifications.erl
+++ b/applications/crossbar/src/modules/cb_notifications.erl
@@ -744,8 +744,9 @@ read_account(Context, Id, LoadFrom) ->
             maybe_read_from_parent(Context, Id, LoadFrom, cb_context:reseller_id(Context));
         {_Code, 'success'} ->
             lager:debug("loaded ~s from account database ~s", [Id, cb_context:account_db(Context)]),
-            NewRespData = note_account_override(cb_context:resp_data(Context1)),
-            cb_context:set_resp_data(Context1, NewRespData);
+            Context2 = maybe_merge_ancestor_attachments(Context1, Id),
+            NewRespData = note_account_override(cb_context:resp_data(Context2)),
+            cb_context:set_resp_data(Context2, NewRespData);
         {_Code, _Status} ->
             lager:debug("failed to load ~s: ~p", [Id, _Code]),
             Context1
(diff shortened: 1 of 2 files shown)


applications/teletype/src/teletype_templates.erl
```diff
@@ -163,20 +163,38 @@ templates_source(_TemplateId, 'undefined', _ResellerId) ->
...
+            parent_templates_source(TemplateId, AccountId, ResellerId);
+        {'ok', Template} ->
+            templates_source_has_attachments(TemplateId, AccountId, ResellerId, Template);
         {'error', _E} -> 'undefined'
     end.
 
+-spec templates_source_has_attachments(ne_binary()
+                                      ,ne_binary()
+                                      ,ne_binary()
+                                      ,kz_json:object()) -> api_binary().
```
fenollp: no need to take that much space for just a spec ;)

icehess:
Thank you for your your contribution. I'm agree with this change.

I tried to sync up how we fetch notification in both `teletype` app and `crossbar:cb_notifications`, so people can see what they would actually receive from teletype.
Can you check `cb_notifications` and change if the account's notification doesn't have the template attachment it would fetch it from its parent? This needs to support both fetching/changing a template

lazedo:
@icehess `This needs to support both fetching/changing a template`, only the `fetch`, changing will be saved to the account that is customizing it.

applications/teletype/src/teletype_templates.erl
danielfinke: @fenollp happy to minimize it. I was aiming to keep an 80-or-less character line length. Is there any style standard when it comes to specs (or anything else for matter) that I can adhere to?

danielfinke:
@icehess @lazedo OK, cb_notifications will return provide the content types and attachments up the account hierarchy to system_config if applicable, matching the behaviour of teletype now. This includes the "template" object in the metadata GET ({"<attachment_name>": {"length": <length>}}). I noticed that the attachment fetch already wasn't matching btw - it was going straight to the system_config db if the attachment wasn't found in the account. That's fixed too.

applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -837,6 +838,78 @@ maybe_hard_delete(Context, Id) ->
...
+            Doc = kz_json:set_value(<<"_attachments">>, Attachments, cb_context:doc(Context)),
+            cb_context:setters(Context
+                              ,[{fun cb_context:store/3, 'db_doc', Doc}
+                               ,{fun cb_context:set_doc/2, Doc}
+                               ,{fun cb_context:store/3, 'attachments_db', ?KZ_CONFIG_DB}])
+    end;
+%% Not yet at reseller, try parent account
+merge_ancestor_attachments(Context, Id, AccountId, ResellerId) ->
+    AccountContext = masquerade(Context, AccountId),
+    AccountContext1 = crossbar_doc:load(AccountId, AccountContext, ?TYPE_CHECK_OPTION(kz_account:type())),
```
icehess: You can use `get_parent_account_id/1` (look at `maybe_read_from_parent/4`)

applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -960,8 +1025,8 @@ read_account_attachment(Context, DocId, Name) ->
          }
     of
         {404, 'error'} ->
-            lager:debug("~s not found in account, reading from master", [DocId]),
-            read_system_attachment(Context, DocId, Name);
+            lager:debug("~s not found in account, checking for ancestor attachments", [DocId]),
+            maybe_read_other_account_attachment(Context, DocId, Name);
```
icehess: good catch

danielfinke:
@icehess alright, change made to use get_parent_account_id/1

applications/teletype/src/teletype_templates.erl
icehess: @danielfinke can you please make it one line? I don't thing we have a specific code style for line length (code or spec).
If we want to break long line, we usually will do it for just the code not specs, for long specs we usually put the return type into new line. 
For this spec specifically there's no need to break it into multi line.

danielfinke:
@icehess shrunk the spec

icehess:
@danielfinke thank you, approved and merged.

PR #3625, opened 2017-04-25
Kazoo speech

jamesaimonetti:
break out TTS/ASR into interface and provider modules


fenollp:
Hey there, I see `"media.tts_cache"` is not replaced by anything. Is this expected?

PR #3628, opened 2017-04-25
perf testing merge vs merge_recursive

jamesaimonetti:
For 100K iterations:
kz_json_test:merge in 0.176618s
kz_json_test:merge_recursive in 0.595275s

Property testing merge left/right variants

test confirming merge_recursive is merge(right)


core/kazoo_services/src/kz_service_plans.erl
```diff
@@ -81,7 +81,7 @@ public_json([#kz_service_plans{plans=Plans}|ServicePlans], JObj) ->
 
 -spec merge_plans(kzd_service_plan:doc(), kz_json:object()) -> kz_json:object().
 merge_plans(SerivcePlan, JObj) ->
-    kz_json:merge(JObj, kzd_service_plan:plan(SerivcePlan)).
+    kz_json:merge(fun kz_json:merge_left/2, JObj, kzd_service_plan:plan(SerivcePlan)).
```
lazedo: @jamesaimonetti this one was changed to merge but was previously merge_recursive. can you please check if we should  change here?

PR #3631, opened 2017-04-25
HELP-29204_master: add API to query search prefix maximum length

fenollp:
4.0: https://github.com/2600hz/kazoo/pull/3624


fenollp:
Hold on this version for a sec. I'm adding a side thing that's only needed for 4.1.

PR #3639, opened 2017-04-26
fix schema to allow both string and array

icehess:


applications/callflow/doc/ref/missed_call_alert.md
```diff
@@ -9,5 +9,5 @@ Validator for the missed_call_alert callflow's data object
 Key | Description | Type | Default | Required
 --- | ----------- | ---- | ------- | --------
 `recipients` | One or more specific email addresses, Kazoo user ids or a combination of both | `array(object)` |   | `false`
-`recipients.[].id` | The email address/user ID or the list of email addresses/user IDs based on sepecified type | `string` |   | `false`
+`recipients.[].id` | The email address/user ID or the list of email addresses/user IDs based on sepecified type | `string, array()` |   | `false`
```
icehess: any idea to how let the type be `array(string)` here?

fenollp: Have a look at https://github.com/2600hz/kazoo/pull/3625/files

PR #3643, opened 2017-04-27
knm-auth: streamline knm_number_states auth logic

fenollp:
No 4.0 equivalent.

This is dangerous: it's possible that not enough test cases are written & so needs cocooning.

Please merge https://github.com/2600hz/kazoo/pull/3641 first!


lazedo:
@fenollp rebase

fenollp:
I lied, here's a 4.0 version: https://github.com/2600hz/kazoo/pull/3697

PR #3649, opened 2017-04-27
add media_extention to vmbox schema

icehess:


applications/crossbar/priv/couchdb/schemas/vmboxes.json
```diff
@@ -44,6 +44,11 @@
             "required": false,
             "type": "object"
         },
+        "media_extension": {
+            "default": "mp3",
+            "description": "Voicemail audio format",
+            "type": "string"
```
lazedo: please a `enum` with `wav`, `mp3`, `mp4`

icehess: done

PR #3650, opened 2017-04-28
process /account alias as /accounts/{AUTH_ACCOUNT_ID}

jamesaimonetti:


jamesaimonetti:
@lazedo I tried to move the authenticate earlier, but it relies on req nouns existing. At this point, it feels too invasive to try to get that untangled. Worth the effort or leave it as is until we want to add `/user`?

jamesaimonetti:
@lazedo just kidding, i think i fixed it.

lazedo:
@jamesaimonetti awesome! just one comment,  no need to validate the token twice (early and authenticate)

jamesaimonetti:
@lazedo rebased

PR #3654, opened 2017-05-01
KAZOO-5498: node metrics with SUP

fenollp:


core/kazoo_perf/src/kz_metrics.erl
```diff
@@ -0,0 +1,118 @@
+%%
+%% Simple (and mostly naive) script that connects to a remote Erlang node,
+%% fetches memory and process statistics and prints them in a Sensu compatible way
+%% for metrics gathering (great for Graphite)
+%%
+-module(kz_metrics).
```
k-anderson: @fenollp modules for sup should be _maintenance... please rename to kazoo_pref_maintenance.  Thanks!

For clarity I think the functions should also be renamed to:
graphite_metrics
json_metrics

core/kazoo_perf/src/kazoo_perf_maintenance.erl
```diff
@@ -0,0 +1,119 @@
...
+    ,{output_io_bytes, Output}
+    ];
+to_props(scheduler_reductions, {TotalReductions, _ReductionsSinceLastCall}) ->
+    [{reductions, TotalReductions}
+    ];
+to_props(processes_in_run_queue_of_each_schedulers, RunQueue) ->
+    [{run_queue, RunQueue}
+    ];
+to_props(ets_tables_sizes, Tabs) ->
+    [{"tab_" ++ kz_term:to_list(Tab), Size}
```
jamesaimonetti: Why the prefix on the table?

fenollp: @swysor says that graphite needs it but yes I wi;; remove it from the json output.

lazedo:
@fenollp what was the criteria to what metrics to deliver ? why not bring back and finish the stats app ?

PR #3658, opened 2017-05-02
KAZOO-5500_4oh: allow picking number state on creation

fenollp:
master: https://github.com/2600hz/kazoo/pull/3655


k-anderson:
@fenollp please address the circleci error

PR #3666, opened 2017-05-03
rel-migrate: test release migration function

fenollp:


fenollp:
wait on this: testing circle

core/kazoo_apps/src/kapps_maintenance.erl
```diff
@@ -146,7 +146,7 @@ migrate(Pause, Databases) ->
     io:format("removing depreciated databases...~n"),
     _  = remove_depreciated_databases(Databases),
 
-    kazoo_bindings:map(binding('migrate'), Accounts),
+    kazoo_bindings:map(binding('migrate'), [Accounts]),
```
fenollp: This would otherwise error in crossbar_maintenance receiving an account db instead of a list of account db.

core/sup/src/sup.erl
```diff
@@ -49,19 +50,24 @@ main(CommandLineArgs, Loops) ->
...
-            Verbose
+            IsVerbose
                 andalso stdout("Running ~s:~s(~s)", [Module, Function, string:join(Args, ", ")]),
-            case rpc:call(Target, Module, Function, [list_to_binary(Arg) || Arg <- Args], Timeout) of
+
+            case rpc:call(Target, ?MODULE, in_kazoo, [Module, Function, Arguments], Timeout) of
                 {'badrpc', {'EXIT',{'undef', _}}} ->
                     print_invalid_cli_args();
+                {badrpc, {'EXIT', {timeout_value,[{Module,Function,_,_}|_]}}} ->
+                    stderr("Command failed: timeout", []),
```
fenollp: Adding this kinda breaks compat: before on timeouts it would return the stack trace (which only has the given M F As cause that's all the ST there is)

core/sup/src/sup.erl
```diff
@@ -72,6 +78,13 @@ main(CommandLineArgs, Loops) ->
             end
     end.
 
+-spec in_kazoo(module(), atom(), binaries()) -> no_return().
+in_kazoo(M, F, As) ->
+    kz_util:put_callid(<<"sup_", (kz_binary:rand_hex(8))/binary>>),
+    lager:notice("~s: ~s ~s ~s", [?MODULE, M, F, kz_util:iolist_join($,, As)]),
+    R = apply(M, F, As),
+    lager:notice("~s result: ~p", [?MODULE, R]),
```
fenollp: Now isn't this great!?

fenollp:
This is will now
1. take less time to run (no more hard coded overlapping sleeps)
1. set a callid on the running sup job
1. log start and end time of sup command
1. check for non-normal errors in error.log and fail accordingly
1. check for any crash dumps and fail if one exists

PR #3667, opened 2017-05-03
PISTON-425: use parent teletype config subject/to as defaults

danielfinke:
As an extension to https://github.com/2600hz/kazoo/pull/3618, we are finding it useful to allow POST updates to .../notifications/<notification_id> to omit a couple of the required fields (subject/to) and merge them in from the notification inherited from up the account tree. They are still required as per the schema - retains compatibility with teletype logic.

We use this to provide control of the "from" of a notification, but automatically fill in "subject" (without having to explain the macros) and avoid potentially unwanted configuration of the "to".

### DIFF
--- a/applications/crossbar/priv/api/descriptions.system_config.json
+++ b/applications/crossbar/priv/api/descriptions.system_config.json
@@ -169,6 +169,7 @@
     "crossbar.maximum_range": "crossbar maximum range",
     "crossbar.media.normalization_format": "crossbar.media normalization format",
     "crossbar.media.normalize_media": "crossbar.media normalize media",
+    "crossbar.notifications.inherit_default_values": "crossbar.notifications use parent notifications doc as default values, to be overridden by request data (prior to validation)",
     "crossbar.notifications.notification_timeout_ms": "crossbar.notifications notification timeout in milliseconds",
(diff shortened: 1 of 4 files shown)


lazedo:
@danielfinke this is very restrictive and not intended, you can merge the upper doc andalso  make this configurable `Opt-In`

danielfinke:
@lazedo OK, I made this behaviour configurable.
What did you mean about it being restrictive?

lazedo:
@danielfinke you're only handling the `to` and `subject`. if you get existing document `self` or `parent` , you can merge the posted data into the retrieved data and save it. no need to restrict fields and it just works for other fields

danielfinke:
@lazedo moved to merging all public fields. Do you think this is equivalent to a kz_json:merge/2? If so I'll use that instead.

jamesaimonetti:
@danielfinke `kz_json:merge/2` merges whatever is in the second arg into the first (uses `kz_json:merge_right/2`). You can alternatively use `kz_json:merge(fun kz_json:merge_left/2, J1, J2)` to merge `J1` into `J2`. So I would guess you want to `kz_json:merge(Defaults, ReqData)` ?

danielfinke:
Cleaned this up a lot by moving to kz_json:merge/2.

PR #3665, opened 2017-05-03
remove maintenance code duplicating notify's maintenance code

fenollp:
Reported by @swysor  after crash:

```
<9847.5234.0> (21/21) refreshing database 'system_config'
archiving to /tmp/system_config.json
   archived 83 docs
<9847.5234.0> (20/21) refreshing database 'system_data'
<9847.5234.0> (19/21) refreshing database 'offnet'
<9847.5234.0> (18/21) refreshing database 'accounts'
Stderr Command failed: {'EXIT',{badarg,[{erlang,list_to_binary,[[{error,bad_name},"/couchdb/",<<"views/notify.json">>]],[]},{kapps_util,get_view_json,2,[{file,"src/kapps_util.erl"},{line,530}]},{kapps_maintenance,refresh,1,[{file,"src/kapps_maintenance.erl"},{line,315}]},{kapps_maintenance,refresh,3,[{file,"src/kapps_maintenance.erl"},{line,244}]},{kapps_maintenance,migrate,2,[{file,"src/kapps_maintenance.erl"},{line,141}]},{kapps_maintenance,migrate,1,[{file,"src/kapps_maintenance.erl"},{line,126}]},{rpc,'-handle_call_call/6-fun-0-',5,[{file,"rpc.erl"},{line,206}]}]}}
```

PR #3668, opened 2017-05-03
remove maintenance code duplicating notify's maintenance code (master)

fenollp:
4.0: https://github.com/2600hz/kazoo/pull/3665


lazedo:
@fenollp this is actually interesting. if the `app` doesn't exist, there's a crash because of the malformed path. but the `ACCOUNTS_AGG_NOTIFY_VIEW_FILE` does not exist in `notify` and we don't say nothing. we should `error loudly` if we don't find the file.

PR #3670, opened 2017-05-04
[4.0] use ping to detect ip family is supported or not

icehess:
backport of (#3669)


lazedo:
@icehess this doesn't seem to work. check the CI run, check also https://github.com/2600hz/kazoo/pull/3676 that adds the fun clause when ipv4 and ipv6 are not detected

PR #3671, opened 2017-05-04
Build metadata for all id or *_id in action data

jamesaimonetti:


core/kazoo_data/src/kz_datamgr.erl
```diff
@@ -110,7 +110,7 @@
 %%--------------------------------------------------------------------
 -spec update_doc_from_file(ne_binary(), atom(), nonempty_string() | ne_binary()) ->
                                   {'ok', kz_json:object()} |
-                                  data_error().
+                                  kazoo_data:data_error().
```
lazedo: @jamesaimonetti i think this was previously discussed. `data_error()` is part of the library (local type). prepending `kazoo_data:` gives you the feeling of an external type. thoughts ?

jamesaimonetti: @lazedo I think we need something exported so calling modules don't need to use `{'error', any()}`. It could be `kz_datamgr:data_error()` or we could make specific ones for `open_doc` vs `get_results` if desired.

lazedo: if exporting data_error() from kz_datamgr works for you, i'm ok with that

jamesaimonetti:
@lazedo updated types to `kz_datamgr`

PR #3674, opened 2017-05-04
KAZOO-5504: use chunked bulk reads in knm_numbers

fenollp:


core/kazoo_data/src/kz_datamgr.erl
```diff
@@ -661,6 +665,50 @@ open_docs(DbName, DocIds, Options) ->
...
+
+open_docs_chunked(DbName, DocIds) ->
+    open_docs_chunked(DbName, DocIds, []).
+open_docs_chunked(DbName, DocIds, Options) ->
+    read_chunked(fun open_docs/3, DbName, DocIds, Options).
+
+read_chunked(Opener, DbName, DocIds, Options) ->
+    read_chunked(Opener, DbName, DocIds, Options, []).
+read_chunked(Opener, DbName, DocIds, Options, Acc) ->
+    %% try lists:split(max_bulk_read(), DocIds) of
```
lazedo: forgot this ?

fenollp: indeed

fenollp:
I am thinking this is more of a security thing on the DB side. It can return a possibly very large amount of data indeed.
There is a greater chance the DB will not like reading too many documents but a much lesser chance an Erlang node will not.
The intent here is not to provide full support for too-larger queries, it is to gracefully handle just-a-bit-too-large queries.
Code calling these newly added functions should be careful not to use too much memory. I think this is an okay compromise.

lazedo:
i agree with you. so it would be better to handle that not as a new exported function but as an option in `open_doc(s)` and `open_cached_doc(s)` arity/3 where we pre-process the options and check for `chunk_size` (or other name)  option and add a default if absent, and then internally do the chunk_reads.
thoughts ?

fenollp:
That's what I wanted from the beginning :)

fenollp:
Still working on this.

k-anderson:
@fenollp whats the story here ? ;)

fenollp:
Waiting on me to move from "adding new functions" to just "adding an option with good defaults"

fenollp:
Ready

PR #3675, opened 2017-05-05
KAZOO-5455: SUP output error codes

fenollp:


fenollp:
Once merged I'll do the 4.0 version

core/sup/src/sup.erl
```diff
@@ -154,9 +155,10 @@ get_host() ->
             print_unresolvable_host(Host)
     end.
 
--spec my_name() -> atom().
+-spec my_name() -> ne_binary().
```
lazedo: wrong spec

PR #3677, opened 2017-05-05
[master] set password recovery link for notify

icehess:


fenollp:
4.0: https://github.com/2600hz/kazoo/pull/3678

PR #3681, opened 2017-05-09
PropEr testing

jamesaimonetti:
Creates a framework for building PropEr based testing of Crossbar APIs, with basic testing of `/phone_numbers` as a start.

### DIFF
--- a/applications/crossbar/src/modules/cb_search.erl
+++ b/applications/crossbar/src/modules/cb_search.erl
@@ -70,7 +70,6 @@ allowed_methods(?MULTI) ->
 -spec resource_exists() -> 'true'.
 -spec resource_exists(path_token()) -> 'true'.
 resource_exists() -> 'true'.
-
 resource_exists(?MULTI) -> 'true'.
 
 %%--------------------------------------------------------------------
@@ -100,12 +99,7 @@ authorize(Context, ?MULTI) ->
 -spec validate(cb_context:context(), path_token()) -> cb_context:context().
 validate(Context) ->
     Type = cb_context:req_value(Context, <<"t">>),
-    Ctx = case cb_context:account_id(Context) of
-              'undefined' -> cb_context:set_account_db(Context, ?KZ_ACCOUNTS_DB);
-              _AccountId -> Context
-          end,
-    validate_search(Ctx, Type).
-
+    validate_search(Context, Type).
 
 validate(Context, ?MULTI) ->
     Type = cb_context:req_value(Context, <<"t">>),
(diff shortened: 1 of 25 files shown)


applications/crossbar/src/modules/cb_search.erl
```diff
@@ -411,7 +410,7 @@ get_end_key(Context, Type, Value) ->
 %%--------------------------------------------------------------------
 -spec next_binary_key(binary()) -> ne_binary().
 next_binary_key(<<>>) ->
-    <<"\ufff0">>;
+    <<0>>;
```
fenollp: Are you sure about this one? It's supposed to be the biggest possible string. Isn't `<<0>>` considered the smallest?
Actually is this valid?

core/kazoo_number_manager/src/kazoo_number_manager_maintenance.erl
```diff
@@ -212,10 +212,9 @@ update_number_services_view(?MATCH_ACCOUNT_ENCODED(_)=AccountDb) ->
                                 ,View
                                 ),
     case kz_json:are_equal(View, NewView) of
-        true -> ?LOG("View is up to date.", []);
+        true -> 'ok';
         false ->
-            true = kz_datamgr:db_view_update(AccountDb, [{ViewName, NewView}]),
-            ?LOG("View updated!", [])
```
fenollp: Please keep the above log line. Helps a lot.

core/kazoo_proper/src/pqc_cb_phone_numbers.erl
```diff
@@ -0,0 +1,298 @@
...
+                        'true' -> 'reserved' =:= pqc_cb_response:number_state(APIResult);
+                        'false' -> 500 =:= pqc_cb_response:error_code(APIResult)
+                    end
+            end
+    end;
+postcondition(Model
+             ,{'call', _, 'remove_number', [_API, AccountId, Number]}
+             ,APIResult
+             ) ->
+    case pqc_kazoo_model:does_account_exist(Model, AccountId) of
```
fenollp: If Model was a map() or if we could use macros to match this property from the model, pre/post-condition code would get much less lengthier. What do you think? I believe that with time we would be checking more and more properties and these clauses will get monstruous.

core/kazoo_stdlib/test/kz_json_test.erl
```diff
@@ -19,21 +19,28 @@
...
+     ,[{atom_to_list(F)
+       ,fun () ->
+                ?assert(proper:quickcheck(?MODULE:F(), [{'to_file', 'user'}
+                                                       ,{'numtests', 500}
+                                                       ]))
+        end
+       }
+       || {F, 0} <- ?MODULE:module_info('exports'),
+          F > 'prop_',
+          F < 'prop`'
```
fenollp: Pretty sure this is not doing what you think it's doing: needs conversion to a string type.

core/kazoo_proper/src/pqc_cb_phone_numbers.erl
jamesaimonetti: Agreed. I tried to keep it simple to make sure failing runs could be debugged easily while I worked out the kinks (hence one account, one number for the time being). Not married to how predicates are defined for post-conditions or next_state transitions; will let them evolve as this moves forward.

core/kazoo_stdlib/test/kz_json_test.erl
jamesaimonetti: Missed the '`' in the second one to get the sorting. Worked in the shell, copy/pasted wrong apparently.

jamesaimonetti:
@fenollp At the moment, this runs locally. Enter the master account API key, compile, and away you go. The README in `doc/` should have the sequential and parallel test initiation commands to run in the shell. I wanted to get some eyes on it and extend it a bit more before setting up any CI testing.

jamesaimonetti:
@fenollp also, with respect to the SDK, that would be great to have.

There's going to be trickiness once we start testing POST and others that require the data from a previous GET/PUT. Numbers was easy with mostly PUTs and DELETEs that didn't require tracking an ID. I was going to do `/accounts` next and see what types of race conditions there lurking in name/realm uniqueness!

fenollp:
Cool!
Yes trickiness will ensue but it's definitely doable. It's a state machine after all :)

lazedo:
@fenollp  what changes are you requiring here ?

fenollp:
Just some Dialyzer issues and we're good to go!

fenollp:
@jamesaimonetti props PropEr tests failing

jamesaimonetti:
@fenollp i constrained the tests because we can't distinguish list-as-key vs list-as-path and trying to generate nested proplists got caught in a recursive loop trap or something.

PR #3682, opened 2017-05-09
directory lookup should have X-headers

lazedo:

PR #3683, opened 2017-05-09
directory lookup should have X-headers

lazedo:


lazedo:
4.0 https://github.com/2600hz/kazoo/pull/3682

PR #3686, opened 2017-05-09
qf-get: use typed kapps_config:get calls when possible

fenollp:


lazedo:
@fenollp xref & tests. other than that, looks good

PR #3687, opened 2017-05-09
Opentelecom smtp fax error notifications

jamesaimonetti:
includes and resolves #3323

### DIFF
--- a/applications/fax/src/fax_smtp.erl
+++ b/applications/fax/src/fax_smtp.erl
@@ -280,6 +280,7 @@ maybe_system_report(#state{faxbox='undefined', account_id='undefined'}=State) ->
         'false' -> 'ok'
     end;
 maybe_system_report(State) ->
+    send_outbound_smtp_fax_error(State),
     case kapps_config:get(?CONFIG_CAT, <<"report_faxbox_system_errors">>, 'true') of
         'true' -> system_report(State);
         'false' -> 'ok'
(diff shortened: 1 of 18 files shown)


applications/fax/src/fax_smtp.erl
```diff
@@ -280,6 +280,7 @@ maybe_system_report(#state{faxbox='undefined', account_id='undefined'}=State) ->
         'false' -> 'ok'
     end;
 maybe_system_report(State) ->
+    send_outbound_smtp_fax_error(State),
```
lazedo: @jamesaimonetti fax attachment error will be sent here. this is not desirable. you can put behind a configuration flag (false by default) and would prefer to put that in the `maybe_faxbox_log` clauses

jamesaimonetti: @lazedo noted, will update

jamesaimonetti:
@lazedo updated, what do you think? @kalda341 thoughts? care to test it on your setup?

lazedo:
@jamesaimonetti i think the solution you found with the `has_smtp_errors` is great and would make the `report_smtp_errors` defaulting to `true` possible

kalda341:
Looks good to me! Finding time to test is quite tricky for me at the moment, so the PR might have to wait a while if you need it tested. I am happy with it if it works for you though :)

jamesaimonetti:
@lazedo ok

@kalda341 sounds good, I think we'll go for it.

PR #3694, opened 2017-05-10
qf-free: skip free-ing an account's numbers if something is wrong

fenollp:
Found during failed creation of the master account

### DIFF
--- a/core/kazoo_number_manager/src/knm_numbers.erl
+++ b/core/kazoo_number_manager/src/knm_numbers.erl
@@ -506,8 +506,11 @@ account_listing(AccountDb=?MATCH_ACCOUNT_ENCODED(_,_,_)) ->
             [{kz_doc:id(JObj), kz_json:get_value(<<"value">>, JObj)}
              || JObj <- JObjs
             ];
+        {'error', 'not_found'=_R} ->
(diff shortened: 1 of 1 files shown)


core/kazoo_number_manager/src/knm_numbers.erl
```diff
@@ -506,8 +506,11 @@ account_listing(AccountDb=?MATCH_ACCOUNT_ENCODED(_,_,_)) ->
             [{kz_doc:id(JObj), kz_json:get_value(<<"value">>, JObj)}
              || JObj <- JObjs
             ];
+        {'error', 'not_found'=_R} ->
+            lager:error("error listing numbers for ~s: ~p", [AccountDb, _R]),
+            [];
         {'error', _R} ->
-            lager:debug("error listing numbers for ~s: ~p", [AccountDb, _R])
+            lager:error("error listing numbers for ~s: ~p", [AccountDb, _R])
```
jamesaimonetti: I'm confused why you don't either throw an error or return an empty list too? It will crash further up (or the caller) but it would be nice to have it immediately crash here instead, no? Stacktrace should then point here as the cause.

fenollp:
@jamesaimonetti updated both PRs

PR #3699, opened 2017-05-11
change order of children for proper shutdown

lazedo:
also sets debug mode for CI

### DIFF
--- a/rel/ci-config.ini
+++ b/rel/ci-config.ini
@@ -3,3 +3,8 @@
 [amqp]
 uri = "amqp://guest:guest@127.0.0.1:5672"
 pool_size = 10
+
+[log]
+syslog=debug
+console=debug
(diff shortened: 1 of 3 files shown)


rel/ci-config.ini
```diff
@@ -3,3 +3,8 @@
 [amqp]
 uri = "amqp://guest:guest@127.0.0.1:5672"
 pool_size = 10
+
+[log]
+syslog=debug
+console=debug
```
fenollp: What do you think about setting console to info and if an error is detected cat the syslog logs?
This way we have smaller and colored output in most cases, but a full output on failures.

lazedo: i prefer `debug` for now, wouldn't caught up the order thing without it. lets keep it for now

PR #3701, opened 2017-05-12
Teletype dual email fix

tsromana:
when there is no post url or when there's an error in the post request than the email is sent twice, so this change will send email as default as well as send post to a url but when the email is turned off and the post request fails only than send email as a backup option.


icehess:
Thanks for your contribution.

PR #3710, opened 2017-05-14
KAZOO-5505: As a provisioner, I want kazoo to send an API request to …

tsromana:
…update provisioner when devices are PATCHed

This change will make sure that the processing for PATCH is same as for POST, the doc after patching goes through the same process as POST so why not during saving of the doc.


jamesaimonetti:
@tsromana thanks! merged manually into master

PR #3713, opened 2017-05-15
add more logging around routing/authz

jamesaimonetti:


applications/ecallmgr/src/ecallmgr_fs_authz.erl
```diff
@@ -71,28 +71,38 @@ kill_channel(<<"outbound">>, _, CallId, Node) ->
 maybe_authorize_channel(Props, Node) ->
     CallId = kzd_freeswitch:call_id(Props),
     case kzd_freeswitch:channel_authorized(Props) of
-        <<"true">> -> 'true';
-        <<"false">> -> 'false';
+        <<"true">> ->
+            lager:debug("channel is authorized by FreeSWITCH"),
```
k-anderson: @jamesaimonetti these are a bit confusing as it wasn't FS the auth'd but rather a previous kazoo process (maybe: channel is already authorized)

PR #3715, opened 2017-05-15
KAZOO-5426 & KAZOO-5352: support ranges in ledgers + better data accumulation

fenollp:
1. Accumulates total credits/debits for any ledger on /ledger, within the given or default time range (changes to kz_ledgers)
1. Fixes multi-databases modb crossbar view queries (see crossbar_doc's introduction of the should_paginate boolean)

### DIFF
--- a/applications/crossbar/src/crossbar_doc.erl
+++ b/applications/crossbar/src/crossbar_doc.erl
@@ -395,8 +395,8 @@ view_sort_direction(Options) ->
         'undefined' -> 'ascending'
     end.
 
-load_view(#load_view_params{dbs=[]
-                           ,context=Context
+load_view(#load_view_params{dbs = []
+                           ,context = Context
                            }) ->
     case cb_context:resp_status(Context) of
         'success' ->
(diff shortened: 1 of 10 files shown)


applications/crossbar/src/crossbar_doc.erl
```diff
@@ -395,33 +398,34 @@ view_sort_direction(Options) ->
...
+load_view(#load_view_params{page_size = PageSize
+                           ,context = Context
+                           ,should_paginate = true
+                           })
+  when is_integer(PageSize), PageSize =< 0 ->
     lager:debug("page_size exhausted: ~p", [PageSize]),
-    case cb_context:resp_status(Context) of
-        'success' -> handle_datamgr_success(cb_context:doc(Context), Context);
-        _Status -> Context
+    case success =:= cb_context:resp_status(Context) of
```
icehess: the whole propose of this change was removing the atom tickies I assume?

fenollp: yes :D
no this was caught in the middle of debugging and it looks like this after reversing some of my changes here

PR #3724, opened 2017-05-17
KAZOO-5490-master: prune null values from provision.combo_keys on dev…

mark2600:
…ices


fenollp:
fails makebusy!

k-anderson:
@mark2600 this needs to prune feature keys as well as combo keys.

PR #3725, opened 2017-05-17
KAZOO-5490-4oh: prune null values from provision.combo_keys on devices

mark2600:


applications/crossbar/src/modules_v2/cb_devices_v2.erl
```diff
@@ -258,16 +258,30 @@ post(Context, DeviceId) ->
...
                           _ = provisioner_util:maybe_sync_sip_data(Context1, 'device')
                   end),
-            maybe_add_mobile_mdn(Context2)
+            maybe_add_mobile_mdn(Context3)
     end.
 
+-spec prune_null_provisioner_fields(cb_context:context()) -> cb_context:context().
+prune_null_provisioner_fields(Context) ->
+    Filter = fun({_, Val}) ->
+                     case Val of 'null' -> 'false'; _ -> 'true' end
```
fenollp: How about function clauses instead of case clauses?

mark2600: Any compelling reason for that ?

PR #3726, opened 2017-05-17
KAZOO-5492: declare exchange to handle check_sync messages

fenollp:


lazedo:
@fenollp the only real change here is adding kapi_switch:declare_exchange to ecallmgr start but if that was the issue then any media save (voicemail, fax, call recordings) would not be working.

lazedo:
@fenollp revisited this and this pr is not changing anything. the `kapi_switch:declare_exchange` adds nothing here because the exchange is already declared (kazoo_cache does it in init fun).
the problem seems to be that we're getting the flush registration before the sync (not controlled, depends on rabbitmq amd amqp library) when we post to devices. in ecallmgr when we lookup the registration to send the sync message the device is no longer registered. you may get some requests that work and others that don't.

resuming, this is working if we post to `/accounts/x/devices/w/sync` but not to `/accounts/x/devices/w` (if user/password changes) see above.

more on this, i would like to see kapi_switch with its own exchange so that kazoo_cache won't get extra unnecessary messages.

fenollp:
Without this change the POST to `/accounts/x/devices/w/sync` doesn't work.
I agree however that kapi_switch should have its own exchange (and not use `configuration`).
Will amend.

lazedo:
@fenollp i don't know how you are testing but it works without this PR.

fenollp:
@lazedo I have no idea what you are testing. I'm not sure that's in line with the actual ticket...
Just
1. go on sandbox's Advanced Provisioner
1. pick a registered device
1. click the settings wheel
1. click Restart (this calls the /sync API which I understand is what the HELP tickets are about)
1. watch your device not restart

Can't really remove the exchange from SB right now though.
Are you sure your rabbit exchanges are all cleaned up?

lazedo:
@fenollp i'm very sure that this PR doesn't do anything, the exchange is declared from kazoo_cache and the `sync` is catched by ecallmgr_fs_notify and the NOTIFY is sent from freeswitch and relayed from kamailio. you may want to look at crossbar devices/users/utils. there is a issue when the password/user/owner changes, the registration is deleted before the `sync` so its not sent.

PR #3736, opened 2017-05-19
qf-schemas: remove validation of non existing schemas

fenollp:


lazedo:
@fenollp as for the cb_google_auth, you can just delete it.

PR #3739, opened 2017-05-19
Toll us classifier typo

tsromana:
toll_us typo fix


fenollp:
I'm sorry, what does this solve?

tsromana:
It doesn't try to solve something just i wanted to say that the var name should be changed as Tollfree are ones which are free numbers but Toll numbers are which are premium call numbers, while looking at the code i got confused for a bit so that's why thought it might be a mistake so made changes according to that :)

fenollp:
Okay then :) Thanks!

PR #3742, opened 2017-05-20
gen-swag: make sdks

fenollp:


scripts/make-swag.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash -e
+
+# Generate Kazoo SDKs using the Sagger file
```
jamesaimonetti: Sagger? :)

jamesaimonetti:
I wonder if this should be a `kazoo-client-sdks` or `kazoo-crossbar-sdks` repo instead? I don't see the value, personally, in having it part of the `kazoo` repo. Let's discuss on the next standup with the team but my 2 cents.

fenollp:
I agree: next step after this is to push generated code to different repos (and there have another CI verifying that stuff works, and generate tags).
I do think though that it's easier to generate this from this repo and have CI push to the SDK repos when master builds pass: it provides some more test coverage.
But we do have lots of tests to verify our schemas and swagger file so we can do without.
Also, having to push to kazoo just to fix some SDKs that live in other repos would get annoying quickly...
I don't know. Open to suggestions :)

PR #3743, opened 2017-05-20
crossbar: added logging IP address of failed auth request

sergey-safarov:


lazedo:
@sergey-safarov thanks

PR #3744, opened 2017-05-22
directory lookup should have X-headers

k-anderson:


k-anderson:
please don't accept yet

k-anderson:
ready! (and #3746)

PR #3745, opened 2017-05-22
fix harmless typo

fenollp:


fenollp:
Note: 4.1 version of this will be in the PR for https://github.com/2600hz/kazoo/compare/KAZOO-5525

PR #3752, opened 2017-05-24
qf-prov: fix provisioning devices

fenollp:
Multiple fixes at once:

* Before the PRs fixing schemas the provisioner_v5.json schema would not be valid and any data going through it would be accepted.
* Turns out the GET to `/devices/{MAC}` should instead be to `/devices/search/{MAC}` as the first endpoint does not exist!
* Among schema fixes: realms where asked to have the `"uri"` format which has nothing to do with realms.

Karl says no need for a 4.0 version of this PR


plau:
Saving a device that exists in Kazoo but not Provisioner does not create a device in Provisioner. 

Reproducing steps:
Delete device in Advanced Provisioner -> Save device in Smart PBX
The device doesn't show up.

fenollp:
@plau Tried after last commit and it works as it used to: device shows up :)

PR #3753, opened 2017-05-24
KAZOO-5404 and KAZOO-5405: send notifications reliably

icehess:


applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -390,7 +390,6 @@ compose_voicemail(#mailbox{exists='false'}, _, Call) ->
 compose_voicemail(#mailbox{max_message_count=MaxCount
                           ,message_count=Count
                           ,mailbox_id=VMBId
-                          ,mailbox_number=VMBN
```
fenollp: Wow how did that ever compile?

applications/tasks/src/kz_notify_resend.erl
```diff
@@ -0,0 +1,451 @@
...
+                  ,{'limit', ?READ_LIMIT}
+                  ,'include_docs'
+                  ],
+    case kz_datamgr:get_results(?KZ_PENDING_NOTIFY_DB, <<"pending_notify/list_by_modified">>, ViewOptions) of
+        {'ok', []} -> {'noreply', State, ?TIME_BETWEEN_CYCLE};
+        {'ok', Pendings} ->
+            process_then_next_cycle([kz_json:get_value(<<"doc">>, J) || J <- Pendings]),
+            {'noreply', State#state{running=Pendings}};
+        {'error', 'not_found'} ->
+            kapps_maintenance:refresh(?KZ_PENDING_NOTIFY_DB),
```
icehess: This case clause returns `{'noreply', State, ?TIME_BETWEEN_CYCLE};`.
The clause above after processing will call `next()` which initiate the timer.

applications/tasks/src/notify_resend_maintenance.erl
```diff
@@ -0,0 +1,356 @@
...
+%% @public
+%% @doc
+%% Prints a limited amount of notifications publish which maxed their tries.
+%% Options:
+%%   ShowCount: amount of notificatiosn to show
+%%   Details: Whether or not to show notifications details
+%% @end
+%%--------------------------------------------------------------------
+-spec failed(ne_binary(), ne_binary()) -> 'no_return'.
+failed(ShowCount, <<"true">>) ->
```
icehess: it's a maintenance command

applications/callflow/src/module/cf_voicemail.erl
icehess: act of toleration?

applications/tasks/src/kz_notify_resend.erl
fenollp: Yeah I still think that the refresh should also have a spawn.

PR #3754, opened 2017-05-25
KAZOO-5518-master: set data.message to message in case of error, or r…

mark2600:
…eturn validation failed


fenollp:
4.0 version: https://github.com/2600hz/kazoo/pull/3755

PR #3756, opened 2017-05-25
Ignore the cdr conflict

tsromana:
Ignore the confict if the cdr is tried to be saved again, 

If the cdr is tried to save again the conflict error is ignored and the value is given as retries error which causes a error log to be generated in cdr_util


fenollp:
Hi there. Can you reproduce this reliably? What steps did you take?

tsromana:
sorry i couldn't reply to your question in time :)

PR #3757, opened 2017-05-27
KAZOO-5540: rewrite called number

tsromana:
Rewrite the called number can have multiple uses when a number needs to routed according to a certain pattern to a certain number

e.g. 611 can route to reseller's main number. The ui can create the callflow needed while provisioning an account, the number is not hard programmed to only accept reseller's number as that type of restrictions can be added by ui according to different use cases.

### DIFF
--- a/applications/callflow/doc/number_rewrite.md
+++ b/applications/callflow/doc/number_rewrite.md
@@ -1,0 +1,52 @@
+## Number Rewrite
+
+### About Number Rewrite
+
+The `number_rewrite` callflow enables you to rewrite the number called by user.
+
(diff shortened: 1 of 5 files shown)


lazedo:
@tsromana thanks for the PR. `cf_resources` (called from `cf_offnet`) already has a `to_did`. what does this add ?

tsromana:
@lazedo sorry i forgot what extra thing i had in mind which wasn't in cf_resources and i think i may have not looked completely in cf_resources, will reopen the issue if i find another use case of this pr.

PR #3758, opened 2017-05-27
Update connectivity.json

powerpbx:
Cannot add PBX in MonsterUI PBX Connector with the caller_id required.

### DIFF
--- a/applications/crossbar/priv/couchdb/schemas/connectivity.json
+++ b/applications/crossbar/priv/couchdb/schemas/connectivity.json
@@ -164,10 +164,6 @@
                                         "type": "string"
                                     }
                                 },
-                                "required": [
-                                    "cid_name",
-                                    "cid_number"
-                                ],
                                 "type": "object"
                             },
                             "delay": {


lazedo:
@powerpbx what are you trying to resolve here ? the schema is ok as is.

powerpbx:
schema is broken.

Very simple to prove.  Go to MonsterUI PBX connector and try create a PBX.

lazedo:
@powerpbx schema is not broken. you can open a pull request for https://github.com/2600hz/monster-ui-pbxs to fix the empty `caller_id` object being sent.

powerpbx:
???

The MonsterUI developer has confirmed the problem and says it should be fixed in the scheme.  Instead of closing pull requests so quickly maybe you guys should figure these things out between yourselves first.

jamesaimonetti:
@powerpbx the issue isn't in the schema. The 'cid_*' fields are required *if* you include the caller id object. If the object is omitted from the request data, then validation will succeed. So @lazedo is saying that the UI team needs to stop sending the empty `caller_id` object as it is the cause of failed validation.

@JRMaitre, thoughts on adjusting the PBX Connector code?

JRMaitre:
@powerpbx I'm the UI dev you talked to. I'm sorry I led you to push stuff to Kazoo while it might have been an UI issue, I made the assumption that this would be fixed in the schema. It seems to be an issue we need to talk about internally so we'll update you when we decide of the right fix. Thanks for your contribution anyway.

k-anderson:
@lazedo howdy, why do you consider this required?  For the PBX connector this is really not required, its supposed to be a pass-through by default, setting the value to required defaults this primary use-case.  I think this PR is good, but doesn't go far enough!  There is a second caller id object that is requiring name/number and it should be removed as well.

Thoughts?

lazedo:
@k-anderson the `caller_id` object is not required

k-anderson:
@lazedo indeed but that would not be retroactive.  As in, all existing pbx connector documents already have an empty object.  We could have the UI team stop adding that but we need to have backward compatibility here.

jamesaimonetti:
@k-anderson I think when the schema was changed from v3 to v4, the validation step was corrected to rightly error on the empty `caller_id` object. So the schema should have been erroring before but the v3 validator apparently didn't do that properly (afaict anyway). As-is, existing connectivity documents will work for call processing but if changes are made to the doc, the empty `caller_id` object will have to be rectified.

k-anderson:
@jamesaimonetti that is the issue tho, something that worked before and doesn't now is not backward compatible.  Since it didn't require it before we can't now :)

k-anderson:
@powerpbx Howdy, we are still working on this but to address the CI and the other required parameter I have moved this PR to:
https://github.com/2600hz/kazoo/pull/3770

PR #3763, opened 2017-05-30
KAZOO-5538 handle case where we attempt to update a non-existant…

mark2600:
… view


fenollp:
4.0: https://github.com/2600hz/kazoo/pull/3762

PR #3768, opened 2017-05-30
[master] KAZOO-5530: do not return priority for MX DNS look up

icehess:


fenollp:
4.0: https://github.com/2600hz/kazoo/pull/3769

PR #3770, opened 2017-05-30
Powerpbx fix connectivity.json

k-anderson:


k-anderson:
Original PR and discussion: 
https://github.com/2600hz/kazoo/pull/3758

PR #3773, opened 2017-05-31
HELP-29745_4oh: fix MOVEing call to a user's other device

fenollp:
…th it


fenollp:
master: https://github.com/2600hz/kazoo/pull/3772

PR #3776, opened 2017-05-31
[master] KAZOO-5523: add block anonymous to blacklist

icehess:


applications/crossbar/priv/couchdb/schemas/blacklists.json
```diff
@@ -14,6 +14,10 @@
             "description": "Map of caller id number to block",
             "properties": {},
             "type": "object"
+        },
+        "should_block_anonymous": {
+            "description": "Should block Anonymous call",
+            "type": "boolean"
```
fenollp: Shouldn't this have a default value that reflects the default behavior?

PR #3778, opened 2017-05-31
HELP-29120: ignore query string filters when paginating

jamesaimonetti:
Keep filling up until page size is fulfilled

### DIFF
--- a/applications/crossbar/src/crossbar_doc.erl
+++ b/applications/crossbar/src/crossbar_doc.erl
@@ -409,20 +409,21 @@ load_view(#load_view_params{dbs = []
     end;
 load_view(#load_view_params{page_size = PageSize
                            ,context = Context
-                           ,should_paginate = true
+                           ,should_paginate = 'true'
                            })
-  when is_integer(PageSize), PageSize =< 0 ->
+  when is_integer(PageSize)
+       andalso PageSize =< 0 ->
     lager:debug("page_size exhausted: ~p", [PageSize]),
-    case success =:= cb_context:resp_status(Context) of
-        false -> Context;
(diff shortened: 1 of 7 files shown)


applications/crossbar/src/crossbar_doc.erl
```diff
@@ -409,20 +409,21 @@ load_view(#load_view_params{dbs = []
...
+        'false' -> Context;
+        'true' -> handle_datamgr_success(cb_context:doc(Context), Context)
     end;
 load_view(#load_view_params{view = View
                            ,view_options = Options
                            ,context = Context
                            ,start_key = StartKey
                            ,page_size = PageSize
-                           ,dbs = [Db|Dbs]
+                           ,dbs = [Db|_]=Dbs
```
fenollp: I don't think that's right (wrt L472).
Have you tried this with the ledgers API? I think this will return duplicates.

applications/crossbar/src/crossbar_doc.erl
```diff
@@ -891,28 +892,35 @@ update_pagination_envelope_params(Context, StartKey, PageSize, NextStartKey, 'tr
...
 -spec handle_datamgr_pagination_success(kz_json:objects(), api_pos_integer(), ne_binary(), load_view_params()) ->
                                                cb_context:context().
+%% If v1, just append results and try next database
 handle_datamgr_pagination_success(JObjs
                                  ,_PageSize
                                  ,?VERSION_1
                                  ,#load_view_params{context = Context
                                                    ,filter_fun = FilterFun
                                                    ,direction = Direction
+                                                   ,dbs=[_Db|Dbs]
```
fenollp: Hm I guess not then. Kinda weird though.

PR #3783, opened 2017-06-01
[master] HELP-30444: check for conflict when creating db

icehess:
CouchDB returns and HTTP error 500 with payload
`{"error":"error","reason":"conflict"}`

This seems and internal error when couchdb tries to create the shard
files, but due to race condition on creating db at same time those files
are already created, so couchdb returns an undocumented error.

The `kazoo_modb:create` falsely thinks that database creation failed and
won't referesh views.

This patches `kz_couch_db` to handle that error.

Also it patches the `kazoo_modb` to retry the operation due db creation
failure.

### DIFF
--- a/core/kazoo_couch/src/kz_couch_db.erl
+++ b/core/kazoo_couch/src/kz_couch_db.erl
@@ -48,9 +48,6 @@ db_create(#server{}=Conn, DbName, Options) ->
         {error, db_exists} ->
             lager:warning("db ~s already exists", [DbName]),
             true;
-        {'error', {bad_response, {500, _Headers, Body}}=Error} ->
-            Reason = kz_json:get_value(<<"reason">>, kz_json:decode(Body)),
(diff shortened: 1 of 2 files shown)


core/kazoo_couch/src/kz_couch_db.erl
```diff
@@ -67,6 +70,14 @@ do_db_create_db(#server{url=ServerUrl, options=Opts}=Server, DbName, Options, Pa
             Error
     end.
 
+-spec check_db_create_error(ne_binary(), any(), any()) -> boolean().
+check_db_create_error(_DbName, _Error, <<"conflict">>) ->
+    lager:warning("db ~s creation failed with HTTP error 500 and conflict reason, assuming db is created", [_DbName]),
+    'true';
```
jamesaimonetti: what about using 'kz_datamgr:db_exists(DbName)` instead of assuming `true`?

PR #3785, opened 2017-06-02
HELP-30462 check channel bridge event in ringback

lazedo:


lazedo:
4.0 https://github.com/2600hz/kazoo/pull/3786

PR #3786, opened 2017-06-02
HELP-30462 check channel bridge event in ringback - 4.0

lazedo:


lazedo:
master https://github.com/2600hz/kazoo/pull/3785

PR #3787, opened 2017-06-02
log channel remove with proper call-id

lazedo:


lazedo:
4.0 https://github.com/2600hz/kazoo/pull/3788

PR #3788, opened 2017-06-02
log channel remove with proper call-id - 4.0

lazedo:


lazedo:
master https://github.com/2600hz/kazoo/pull/3787

PR #3789, opened 2017-06-02
HELP-3044: do not cache false and flush data plan cache in maintenanc…

k-anderson:
…e flush

### DIFF
--- a/core/kazoo_data/src/kazoo_data_maintenance.erl
+++ b/core/kazoo_data/src/kazoo_data_maintenance.erl
@@ -11,7 +11,7 @@
 -include("kz_data.hrl").
 
 -export([flush/0]).
--export([flush_dataplans/0]).
+-export([flush_data_plans/0]).
 -export([flush_docs/0
         ,flush_docs/1
         ,flush_docs/2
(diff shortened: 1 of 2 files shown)


core/kazoo_data/src/kazoo_data_maintenance.erl
```diff
@@ -10,28 +10,42 @@
...
         ,trace_pid/1
         ,stop_trace/1
         ]).
 
 -spec flush() -> 'ok'.
--spec flush(ne_binary()) -> 'ok'.
--spec flush(ne_binary(), ne_binary()) -> 'ok'.
 flush() ->
     _ = kz_datamgr:flush_cache_docs(),
+    _ = kz_cache:flush_local(?KAZOO_DATA_PLAN_CACHE),
```
lazedo: @k-anderson can we call `kzs_plan:flush/0` instead ?

core/kazoo_data/src/kazoo_data_maintenance.erl
```diff
@@ -10,28 +10,42 @@
...
+flush_data_plans() ->
+    _ = kz_cache:flush_local(?KAZOO_DATA_PLAN_CACHE),
+    io:format("flushed all data plans~n").
+
+-spec flush_docs() -> 'ok'.
+-spec flush_docs(ne_binary()) -> 'ok'.
+-spec flush_docs(ne_binary(), ne_binary()) -> 'ok'.
+flush_docs() ->
+    _ = kz_datamgr:flush_cache_docs(),
+    _ = kz_cache:flush_local(?KAZOO_DATA_PLAN_CACHE),
```
lazedo: same here

PR #3792, opened 2017-06-03
KAZOO-5544 use per slot persistence

lazedo:


applications/callflow/src/module/cf_park.erl
```diff
@@ -288,7 +280,7 @@ create_slot(ParkerCallId, PresenceType, Call) ->
 %% was provided
 %% @end
 %%--------------------------------------------------------------------
--spec get_slot_number(kz_json:object(), kapps_call:call()) -> ne_binary().
+-spec get_slot_number(kapps_call:call(), api_binary()) -> ne_binary().
```
jamesaimonetti: `ParkedCalls` isn't a `kapps_call:call()`, its `kz_json:object()`

applications/callflow/src/module/cf_park.erl
```diff
@@ -496,72 +431,31 @@ maybe_get_ringback_id(Call) ->
...
-            JObj;
-        {'error', _R}=E ->
-            lager:info("unable to get parked calls: ~p", [_R]),
-            E
-    end.
+load_parked_calls(JObjs) ->
+    Slots = [load_parked_call(JObj) || JObj <- JObjs],
+    kz_json:from_list([{<<"slots">>, kz_json:from_list(Slots)}]).
+
+load_parked_call(JObj) ->
```
k-anderson: @lazedo specs (call and calls)

PR #3794, opened 2017-06-05
HELP-30386-4oh: fix situation where calculations were off on a Monday

mark2600:


fenollp:
master: https://github.com/2600hz/kazoo/pull/3793

PR #3795, opened 2017-06-06
Fix incorrect binding in bh_object

kalda341:


lazedo:
@kalda341 nice catch, thanks.

PR #3796, opened 2017-06-06
db type mismatch

onnet:
@fenollp (#3782) Hi! 

IMHO we should expect modb according to this line: https://github.com/2600hz/kazoo/blob/master/core/kazoo_transactions/src/kz_transaction.erl#L753

Sorry in advance if I'm wrong :)


fenollp:
I believe you're right! Can't believe nor understand why I'd use these macros instead of the MoDB ones.

Thanks a lot!

PR #3804, opened 2017-06-07
amend auth doc with restrictions

jamesaimonetti:
The new auth doc in crossbar_auth didn't add "restrictions" to the
auth doc (as was done in other auth work in crossbar_util). This
meant token restrictions were never processed during the authorize
phase.

### DIFF
--- a/applications/crossbar/doc/ref/token_restrictions.md
+++ b/applications/crossbar/doc/ref/token_restrictions.md
@@ -11,9 +11,9 @@ Schema for token restrictions
 Key | Description | Type | Default | Required
 --- | ----------- | ---- | ------- | --------
 `restrictions` |   | `object` |   | `false`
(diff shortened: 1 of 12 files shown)


zabizombie:
The restrictions do not seem to be working... tested it on the 4.1 & 4.2 code.

Added GET restrictions using, 
curl -v -X POST \     -H "X-Auth-Token: {AUTH_TOKEN}" \     -d '{"data":{"restrictions":{"cb_user_auth": {"user":{"accounts":[{"rules": {"*": ["GET"]}}]}}}}}' \     http://{SERVER}:8000/v2/accounts/{ACCOUNT_ID}

A user with only GET permissions is able to use PUT method. Please find the attached logs
[user.put.log](https://github.com/2600hz/kazoo/files/1858752/user.put.log)

Please feel free to ask for more information

PR #3809, opened 2017-06-09
Teletype unable to read converted file

tsromana:
Teletype was unable to read converted file because of slow conversion from tiff2pdf, so to slow the read process a timer was added.

### DIFF
--- a/applications/teletype/src/teletype_fax_util.erl
+++ b/applications/teletype/src/teletype_fax_util.erl
@@ -62,14 +62,8 @@ convert(FromFormat0, ToFormat0, Bin) ->
     lager:debug("running conversion command: ~s", [Cmd]),
     Response = case os:cmd(Cmd) of
                    "success" ->
-                       case file:read_file(ToFile) of
-                           {'ok', PDF} ->
-                               lager:debug("convert file ~s to ~s succeeded", [FromFile, ToFile]),
-                               {'ok', PDF};
-                           {'error', _R}=E ->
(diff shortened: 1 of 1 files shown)


fenollp:
`os:cmd` is synchronous so should return only when `tiff2pdf` returns...
Do you have a non default `"tiff_to_pdf_conversion_command"` setting in `system_configs/fax`?

tsromana:
no its the default command, it seems due to slow write to disk it may be causing issues.

lazedo:
@tsromana i don't see how this changes mitigates the problem reported

tsromana:
@lazedo so basically, when the pdf file is tried to be read, it causes no file error (enoent). so i just added a wait period of few seconds to check if the file exist. As @fenollp the `tiff2pdf` is synchronous which means the issue must be related to disk when the file is not being written  fast enough and the file is being read faster by the process, so just to slow it down a bit i added the sleep.

sergey-safarov:
Similar [issue](https://2600hz.atlassian.net/browse/KAZOO-5354) reported by me

tsromana:
@sergey-safarov that's the exact issue i had, with debian 8 virtualized under openstack. I think it is kind of a disk issue as the file could be read after waiting a few ms, using the commit, most of the code does not even gets `enoent` as the delay to call the new fn is enough to make the file readable.

fenollp:
@tsromana are you happy with the alternative solution discussed in the above JIRA ticket?

tsromana:
@fenollp yea no problem :)

PR #3811, opened 2017-06-09
s3 attachments update

lazedo:
. remove credential cache for now, it does not support multiple connections
. support scheme, host, port
. support bucket after host
. support base path


core/kazoo_attachments/src/aws/kz_aws.erl
```diff
@@ -298,16 +299,8 @@ configure(#aws_config{} = Config) ->
 
 -spec get_metadata_credentials(aws_config()) -> {'ok', #metadata_credentials{}} | {'error', any()}.
```
fenollp: How about not creating this function and using `get_credentials_from_metadata/1` directly?
Also, looks like there's no longer a need to `application:set_env(kazoo_attachments, metadata_credentials, ...)`

lazedo: i want to keep the structure to later implement a caching with kazoo_cache, maybe reusing in kazoo_auth and/or kazoo_data and/or own kazoo_cache.

PR #3817, opened 2017-06-09
kazoo-types: move kz_types to a separate app for easier rebar3 building

fenollp:
I'll update the rebar3 branch once that's in


jamesaimonetti:
I don't understand the need for this? How is having the hrl in `/core/kazoo` a problem?

If you are moving it, why not to `kazoo_stdlib` instead of creating another core lib?

I don't recall an initiative to roll out rebar3 in force to the project - has this been documented as desirable? Did I miss a vote? Was this decreed from on high by @k-anderson?

fenollp:
Closing this in favor of https://github.com/2600hz/kazoo/pull/3959

PR #3819, opened 2017-06-12
KAZOO-5392: fix pagination of recordings (#3474) - 4.0

lazedo:


applications/crossbar/src/modules/cb_recordings.erl
```diff
@@ -131,18 +131,81 @@ recording_summary(Context) ->
...
+start_key_builder(Options, Context) ->
+    case crossbar_doc:start_key(Context) of
+        'undefined' ->
+            lager:debug("using default start key"),
+            props:get_value('startkey', Options);
+        CreatedBin ->
+            override_start_key(Options, kz_util:to_integer(CreatedBin))
+    end.
+
+override_start_key(Options, Created) ->
```
jamesaimonetti: spec

applications/crossbar/src/modules/cb_recordings.erl
```diff
@@ -153,19 +216,27 @@ do_load_recording_binary(Context, DocId) ->
...
-                                               );
-                [AName | _] ->
-                    Ctx = crossbar_doc:load_attachment({<<"call_recording">>, DocId}, AName, ?TYPE_CHECK_OPTION(<<"call_recording">>), Context),
-                    set_resp_headers(Ctx, AName)
-            end;
+            do_load_recording_binary_attachment(Context1, DocId);
         _Status -> Context1
     end.
 
+do_load_recording_binary_attachment(Context, DocId) ->
```
jamesaimonetti: spec

applications/crossbar/src/modules/cb_recordings.erl
```diff
@@ -189,11 +260,11 @@ get_disposition(MediaName, Context) ->
 normalize_view_results(JObj, Acc) ->
     [kz_json:public_fields(kz_json:get_value(<<"doc">>, JObj))|Acc].
 
--spec get_view_and_filter(cb_context:context()) -> {ne_binary(), api_binaries(), api_binaries()}.
+-spec get_view_and_filter(cb_context:context()) -> {ne_binary(), api_binaries(), api_objects()}.
```
jamesaimonetti: Should be [binary()] as the third tuple element in the tuple. But since it doesn't vary at all, maybe we just omit returning it and use the `[<<>>]` literal as the `PostFilter`

k-anderson:
@lazedo is there a master version?

lazedo:
@k-anderson no, this is a backport from master, no code changes besides stdlib

PR #3820, opened 2017-06-12
allow set as dialplan

lazedo:


lazedo:
needed for https://github.com/2600hz/kazoo/pull/3824

PR #3821, opened 2017-06-12
alllow forcing dynamic cid - 4.0

lazedo:


lazedo:
master https://github.com/2600hz/kazoo/pull/3824

k-anderson:
@lazedo howdy #3824 requires #3820 but there is no 4.0 version of #3820, is that an issue?

lazedo:
@k-anderson that's ok. https://github.com/2600hz/kazoo/pull/3820 is only for master/4.1

PR #3826, opened 2017-06-13
HELP-27660: migrate cdrs and voicemail to modb

icehess:
Add maintenance command to migrate cdrs and voicemail from AccountDB to MODB.

Useful for cleaning Account DBs purposes in which it would create old modb if they're not present.

### DIFF
--- a/applications/fax/src/fax_maintenance.erl
+++ b/applications/fax/src/fax_maintenance.erl
@@ -24,15 +24,15 @@
 -export([load_smtp_attachment/2]).
 -export([versions_in_use/0]).
 
--define(DEFAULT_MIGRATE_OPTIONS, []).
--define(OVERRIDE_DOCS, ['override_existing_document']).
+-define(DEFAULT_MIGRATE_OPTIONS, [{'allow_old_modb_creation', 'true'}]).
+-define(OVERRIDE_DOCS, ['override_existing_document' | ?DEFAULT_MIGRATE_OPTIONS]).
 -define(DEFAULT_BATCH_SIZE, 100).
 
 -spec migrate() -> 'ok'.
 migrate() ->
     Accounts = kapps_util:get_all_accounts(),
     Total = length(Accounts),
-    lists:foldr(fun(A, C) -> migrate_faxes_fold(A, C, Total,?DEFAULT_MIGRATE_OPTIONS) end, 1, Accounts),
+    lists:foldr(fun(A, C) -> migrate_faxes_fold(A, C, Total, ?DEFAULT_MIGRATE_OPTIONS) end, 1, Accounts),
     migrate_outbound_faxes(),
     'ok'.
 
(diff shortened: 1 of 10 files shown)


applications/fax/src/fax_maintenance.erl
```diff
@@ -408,7 +408,10 @@ migrate_outbound_fax(JObj) ->
     ToDB = kz_util:format_account_modb(AccountMODb, 'encoded'),
     ToId = ?MATCH_MODB_PREFIX(kz_term:to_binary(Year), kz_time:pad_month(Month),FromId),
 
-    case kazoo_modb:move_doc(FromDB, FromId, ToDB, ToId, ['override_existing_document']) of
+    Options = [{'allow_old_modb_creation', 'true'}
```
k-anderson: @icehess why not use ?OVERRIDE__DOCS

core/kazoo_modb/src/kazoo_modb.erl
```diff
@@ -315,24 +327,29 @@ copy_doc(FromDb, FromId, ToDb, ToId, Options, _Reason, Retry) ->
...
                                          'source_not_exists' |
                                          'too_old'|
                                          boolean().
-maybe_create_destination_db(FromDb, ToDb, Options) ->
+maybe_create_destination_db(FromDb, FromId, ToDb, Options) ->
     ShouldCreate = props:get_is_true('create_db', Options, 'true'),
     lager:info("destination modb ~p not found, maybe creating...", [ToDb]),
     case ShouldCreate
-        andalso kz_datamgr:db_exists(FromDb)
+        andalso (kz_datamgr:db_exists(FromDb)
```
fenollp: no need for braces

core/kazoo_modb/src/kazoo_modb_migrate_maintenance.erl
```diff
@@ -0,0 +1,368 @@
...
+                 ,[fun migrate_voicemails/1
+                  ,fun migrate_cdrs/1
+                  ]
+                 ).
+
+%%%===================================================================
+%%% Internal functions
+%%%===================================================================
+
+-spec get_account_id(ne_binary()) -> ne_binary().
```
fenollp: This looks like it's replicating `format_account_id(raw, _)`. Is this an accident?

core/kazoo_modb/src/kazoo_modb_migrate_maintenance.erl
```diff
@@ -0,0 +1,367 @@
...
+-spec next_skip(non_neg_integer(), non_neg_integer(), kz_proplist()) -> kz_proplist().
+next_skip(0, Skip, ViewOptions) -> props:set_value('skip', Skip, ViewOptions);
+next_skip(Remaining, Skip, ViewOptions) ->
+    case Remaining < props:get_value(limit, ViewOptions) of
+        true -> props:set_value('skip', Skip, props:delete(limit, ViewOptions));
+        false -> props:set_value('skip', Skip, ViewOptions)
+    end.
+
+-spec maps_update_with(ne_binary(), fun((any()) -> any()), any(), map()) -> map().
+maps_update_with(Key, UpdateFun, Init, Map) ->
```
fenollp: It's a bit sad to not use the BIF when able. How about using http://erldocs.com/current/erts/erlang.html?i=0&search=erlang:function_exported#function_exported/3 ?

icehess: There is no performance gain, I think update_with implementation is the same as here

icehess: https://github.com/erlang/otp/blob/master/lib/stdlib/src/maps.erl#L169

fenollp: Interesting. I'm ok with this then, but we should keep this kind of things in mind since the `maps` implementation can change in the future (to a C version for example).

PR #3827, opened 2017-06-14
[master] HELP-30620: fix to_csv to join as binary

icehess:
Also replace empty json values with empty binary


k-anderson:
@icehess need a 4.0 version :)

PR #3830, opened 2017-06-15
HELP-30600: send channel_destroy to interested processes

jamesaimonetti:
also add logging to know when the channel_create has been noted in the
channels cache.

### DIFF
--- a/core/kazoo_stdlib/src/kz_json.erl
+++ b/core/kazoo_stdlib/src/kz_json.erl
@@ -324,6 +324,8 @@ merge_left(_K, {'right', ?JSON_WRAPPER(_)=Right}) ->
     {'ok', merge(fun merge_left/2, Right, new())};
 merge_left(_K, {'right', V}) -> {'ok', V};
 
+merge_left(_K, {'both', ?EMPTY_JSON_OBJECT=Left, ?JSON_WRAPPER(_)=_Right}) ->
+    {'ok', Left};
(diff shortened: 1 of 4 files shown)


core/kazoo_stdlib/src/kz_json.erl
```diff
@@ -324,6 +324,8 @@ merge_left(_K, {'right', ?JSON_WRAPPER(_)=Right}) ->
     {'ok', merge(fun merge_left/2, Right, new())};
 merge_left(_K, {'right', V}) -> {'ok', V};
 
+merge_left(_K, {'both', ?EMPTY_JSON_OBJECT=Left, ?JSON_WRAPPER(_)=_Right}) ->
+    {'ok', Left};
```
lazedo: @jamesaimonetti shouldn't this be `{'ok', Right}` ? 
we're merging the `Left` object into the `Right`, the resulting should be `Right` overridden with `Left` values.

core/kazoo_stdlib/src/kz_json.erl
```diff
@@ -342,6 +344,8 @@ merge_right(_K, {'right', ?JSON_WRAPPER(_)=Right}) ->
     {'ok', merge(fun merge_right/2, new(), Right)};
 merge_right(_K, {'right', V}) -> {'ok', V};
 
+merge_right(_K, {'both', ?JSON_WRAPPER(_)=_Left, ?EMPTY_JSON_OBJECT=Right}) ->
+    {'ok', Right};
```
lazedo: @jamesaimonetti see above. i think it should be `{'ok', Left}`

PR #3833, opened 2017-06-15
KAZOO-5532: return empty result for storage instead of error

icehess:


lazedo:
this should be handled by the api client not in the api

PR #3834, opened 2017-06-15
[master] KAZOO-5259: send reload acls/gateways to all connected fs nodes

icehess:


k-anderson:
@icehess as discussed :)

Please revert these changes, add 'federated', then start 2 ecallmgrs in the same zone.  Connect both to a local freeswitch server, then check the remote zone rabbitmq server.  You should find a queue that matches the freeswitch nodes for ecallmgr_fs_node in that zone, and it should have two consumers.  If you find two queues in the remote rabbitmq then we need a different approach.  Let me know!

PR #3836, opened 2017-06-16
Fix register_overwrite_notify

koctep:


lazedo:
@koctep Hi, why do you consider this a fix ? the notification is designed to be sent only on the first registration.

k-anderson:
@lazedo Howdy, the fundamental issue they are working on here is that when they set the flag for overwrite notifications they are not getting them.  It might be related to 4.0 changes, when we routed directly to the endpoint via ecallmgr no later registrar was required so we had both contact strings.  However, in 4.0 we now route to Kamailio then resolve the AOR so perhaps its already flushed when these are sent?  The report also seems to indicate that the message is not being sent from FreeSWITCH but might just be it was on another media server in the cluster and they were unaware that only one will be chosen to initiate this message.  Either way we should look into this and see if we can reproduce it in our lab.

lazedo:
@k-anderson the proposed `fix` is a change in behavior to always send registration notice.

koctep:
@lazedo anyway notification would be sent only if contact would be changed

lazedo:
@koctep check again, that is not what this commit does.
it would be helpful if you actually provide the use case / flow if you want us to take a look. i'm afraid this commit will send `kapi_notifications:publish_register` payloads on registrations and re-registrations which is not intended

koctep:
@lazedo 
Our flow is:
- I'm opening a page with WebRTC client
- then I'm opening this page one more time

This function is preventing not needed notifcations
https://github.com/2600hz/kazoo/blob/master/applications/ecallmgr/src/ecallmgr_fs_notify.erl#L194

k-anderson:
@koctep howdy, do to the age of this PR and since the comments have not been addressed I am going to close it.  However, please feel free to open a new PR if this is still an issue for you, thanks!

PR #3838, opened 2017-06-16
[4oh] KAZOO-5259: send reload acls/gateways to all connected fs nodes

icehess:
backport of #3834


k-anderson:
@icehess ecallmgr_fs_node is started per connected freeswitch node, since each process is bound to AMQP for the events then each runs for the respective node.  

I think the issue is related to zones?

PR #3839, opened 2017-06-16
HELP-30676 fix refer on offnet calls - 4.0

lazedo:


k-anderson:
@lazedo we need more testing before we can accept this to 4.0, I will accept the master so we can run it on sandbox!

The master pr was #3837

PR #3855, opened 2017-06-20
Language docs

jamesaimonetti:


k-anderson:
@jamesaimonetti Hey I talked to JR and the UI is doing the correct thing when setting the language to default but the schemas appear to set it:
https://github.com/2600hz/kazoo/blob/master/applications/crossbar/priv/couchdb/schemas/accounts.json#L68

I think we should remove the default and allow the accounts to use the system_config value unless otherwise defined on the account.  Thoughts?

k-anderson:
@jamesaimonetti oh and need a 4.0 version please :)

PR #3860, opened 2017-06-21
HELP-30703-master: set a more useful label on provision requests for …

mark2600:
…devices

### DIFF
--- a/applications/crossbar/src/modules/provisioner_v5.erl
+++ b/applications/crossbar/src/modules/provisioner_v5.erl
@@ -349,6 +349,29 @@ settings_keys(Assoc, KeyKind, JObj) ->
         LineKey -> kz_json:set_value(<<"account">>, LineKey, Keys)
     end.
 
+-spec get_label(binary(), binary()) -> binary().
+get_label(AccountId, DocId) ->
+    AccountDb = kz_util:format_account_id(AccountId, 'encoded'),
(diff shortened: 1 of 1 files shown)


applications/crossbar/src/modules/provisioner_v5.erl
```diff
@@ -349,14 +349,37 @@ settings_keys(Assoc, KeyKind, JObj) ->
...
+get_label(Doc) ->
+    case {kz_json:get_value(<<"first_name">>, Doc), kz_json:get_value(<<"last_name">>, Doc)} of
+        {'undefined', 'undefined'} ->
+            kz_json:get_value(<<"name">>, Doc);
+
+        {First, 'undefined'} ->
+            First;
+
+        {'undefined', _Last} ->
+            kz_json:get_value(<<"name">>, Doc);
```
k-anderson: @mark2600 this should never happen ;) but if there is no first_name then there will be no 'name'.  Should this not return Last?

Also why the space between clauses?

mark2600: Space because it looks good :)

In reality I think it makes it easier to read, but I don't have a strong opinion either way.

As for just returning last, I was thinking it through and that just seemed like odd behavior.  I suppose that I could be convinced differently.

PR #3871, opened 2017-06-24
Added feature to disable rtcp_mux for sip calls via resources gateway.

sergey-safarov:


sergey-safarov:
Hello @lazedo
Could you look this PR

applications/stepswitch/src/stepswitch_resources.erl
```diff
@@ -1084,6 +1093,11 @@ gateway_from_jobj(JObj, #resrc{is_emergency=IsEmergency
             ,privacy_mode=kz_json:get_value(<<"privacy_mode">>, JObj, PrivacyMode)
             }.
 
+-spec gateway_rtcp_mux(kz_json:object()) -> boolean().
+gateway_rtcp_mux(JObj) ->
+    Default = kapps_config:get_is_true(?SS_CONFIG_CAT, <<"default_rtcp_mux">>, 'true'),
```
sergey-safarov: @lazedo is required to add default value to `applications/stepswitch/src/stepswitch.hrl` file

jamesaimonetti: Does this config setting what's used in #3872 ?

sergey-safarov: Yes, this settings is equal but applied to different objects. One this for carrier gateways other for SIP devices.

applications/stepswitch/src/stepswitch_resources.erl
```diff
@@ -620,28 +622,34 @@ gateway_to_endpoint(DestinationNumber
...
+                 ,{<<"Custom-SIP-Headers">>, SipHeaders}
+                 ,{<<"SIP-Interface">>, SipInterface}
+                 ,{<<"Endpoint-Type">>, EndpointType}
+                 ,{<<"Endpoint-Options">>, EndpointOptions}
+                 ,{<<"Endpoint-Progress-Timeout">>, kz_term:to_binary(ProgressTimeout)}
+                 ,{<<"Custom-Channel-Vars">>, kz_json:from_list(CCVs)}
+                 ,{<<"Outbound-Caller-ID-Number">>, CIDNumber}
+                 ,{<<"Outbound-Caller-ID-Name">>, CIDName}
+                  | maybe_get_t38(Gateway, OffnetJObj)
+                 ]) ++ maybe_rtcp_mux(RTCP_MUX),
```
lazedo: because this is a single value, there's no need to add it to the list of properties being built, just include it upper and it will be remove if not defined.

more, the setting at this level will be removed in the publish and there's no handling at this level in ecallmgr. the `RTCP-MUX` should be included as a `Custom-Channel-Var`

sergey-safarov: @lazedo i cannot add as regular property to `Custom-Channel-Var` [here](https://github.com/sergey-safarov/kazoo/blob/ea272da70b547ed70f08e5abe23539aff4e5d97e/applications/stepswitch/src/stepswitch_resources.erl#L620-L624). Because this feature is enabled in FreeSwitch by default and `<<"RTCP-MUX">>` property with `true` value is useless and with `false` value is been removed by `props:filter_empty`.
I can simple move `++ maybe_rtcp_mux(RTCP_MUX)` to string [624](https://github.com/sergey-safarov/kazoo/blob/ea272da70b547ed70f08e5abe23539aff4e5d97e/applications/stepswitch/src/stepswitch_resources.erl#L624). Is this will be correct?

lazedo: you can remove the `props:filter_empty`. setting `RTCP-MUX` has no effect, it will not be delivered to ecallmgr because the amqp library will remove that header

sergey-safarov:
PR is updated and reopened https://github.com/2600hz/kazoo/pull/3888

PR #3873, opened 2017-06-25
Changed payload, added `request_id` and `data`.

Comsolvia:
Old example didn't work (in 4.0). Added `request_id: '{REQUEST_ID}',` and `data: {}` after discussion with JR^ on IRC.
(Not sure if/how the unsubscribe part works, did the same changes there anyway...)


jamesaimonetti:
@Comsolvia thanks for the update! I've merged this manually but Github doesn't detect it automagically. Closing here

PR #3885, opened 2017-06-27
HELP-30632: fix state picking on number creation

fenollp:


core/kazoo_number_manager/src/knm_number.erl
```diff
@@ -154,42 +154,32 @@ state_for_create(Options) ->
...
 allowed_creation_states(AuthBy) ->
-    %% Note: AuthBy can be ?KNM_DEFAULT_AUTH_BY
-    case knm_phone_number:is_admin(AuthBy) of
+    allowed_creation_states([], AuthBy).
+
+-spec allowed_creation_states(knm_number_options:options(), api_ne_binary()) -> ne_binaries().
+allowed_creation_states(_, undefined) -> [];
+allowed_creation_states(Options, AuthBy) ->
+    case knm_phone_number:is_admin(AuthBy)
+        orelse allow_number_additions(Options, AuthBy)
```
k-anderson: @fenollp please update this so that the user is only provided with the list of states they are allowed to transition to.

PR #3890, opened 2017-06-28
KAZOO-5473: Why limit numbers lookup to Zero?

onnet:
Just a temporary correction while guys thinking: https://groups.google.com/forum/#!topic/2600hz-dev/UaIqtL_OVGY


fenollp:
Good catch! That wasn't caught by our test suite as the limit is always set in the fixtures...
This issue comes from https://github.com/2600hz/kazoo/pull/2605/commits/17bbc475aeab67b22a13c31edef1bca504fa5cc6 and https://github.com/2600hz/kazoo/pull/2605/commits/d539c951a6d681c05dbd98e9358f74e6c93858a3 and it seems the team at Telnyx has not come around to add the sometimes missing `limit` key.

I'll merge this once CI allows me.

PR #3888, opened 2017-06-28
Added feature to configure rtcp_mux for sip calls via resources gateway

sergey-safarov:

PR #3872, opened 2017-06-24
Added feature to disable rtcp_mux for calls to SIP endpoints

sergey-safarov:

PR #3899, opened 2017-06-28
Sergey safarov rtcp mux9

k-anderson:


k-anderson:
Alternative formatting / var placement for #3888

lazedo:
@k-anderson i think you'll need to handle this in ecallmgr (bridge, originate) if you want to move it out of `Custom-Channel-Vars`

sergey-safarov:
@k-anderson 'RTCP-Mux' parameter have default `true` value in FreeSwitch. Please look [FS-10400](https://freeswitch.org/jira/browse/FS-10400).
When `true` is assigned to `RTCP-Mux` [here](https://github.com/2600hz/kazoo/blob/e9036f19dd522f5b97171c5a7e2e37579abf0057/applications/stepswitch/src/stepswitch_resources.erl#L646), then this useless. When assigned `false`, then it will be filtered by [`props:filter_empty`](https://github.com/2600hz/kazoo/blob/e9036f19dd522f5b97171c5a7e2e37579abf0057/applications/stepswitch/src/stepswitch_resources.erl#L627).
Please look similar implementation https://github.com/2600hz/kazoo/pull/3871
Also please look @fenollp comment at this PR https://github.com/2600hz/kazoo/pull/3872

PR #3903, opened 2017-06-29
KAZOO-5595 gproc registration for event/callid

lazedo:


applications/ecallmgr/src/ecallmgr_call_control.erl
```diff
@@ -73,7 +73,7 @@
 
 -define(SERVER, ?MODULE).
 
--define(KEEP_ALIVE, 2 * ?MILLISECONDS_IN_MINUTE). %% after hangup, keep alive for 2 minutes
+-define(KEEP_ALIVE, 2 * ?MILLISECONDS_IN_SECOND). %% after hangup, keep alive for 2 minutes
```
fenollp: comment seems wrong now

core/kazoo_ast/src/fs_prop_usage.erl
```diff
@@ -9,8 +9,8 @@
 -include_lib("kazoo_ast/include/kz_ast.hrl").
 -include_lib("kazoo_stdlib/include/kazoo_json.hrl").
 
--define(DEBUG(_Fmt, _Args), 'ok').
-%% -define(DEBUG(Fmt, Args), io:format([$~, $p, $  | Fmt], [?LINE | Args])).
+%%-define(DEBUG(_Fmt, _Args), 'ok').
+-define(DEBUG(Fmt, Args), io:format([$~, $p, $  | Fmt], [?LINE | Args])).
```
fenollp: Shouldn't this be reverted?
Also could be rewritten as `"~p "++Fmt`: the compiler will output the same code.

lazedo: yep, dev time

PR #3913, opened 2017-06-30
Faster binary reverse

jamesaimonetti:


fenollp:
Nice 100x improvements!
What does this give on inputs ranging from a size of 4 and 15?

PR #3918, opened 2017-06-30
Primary Amqp server fix

tsromana:
make decision as which server is primary based on on its connection pid, where the first added amqp server will have a lower connection pid which does not changes after a connection has been added, and the server are added in accordance to their addition in config file.

### DIFF
--- a/core/kazoo_amqp/include/kz_amqp.hrl
+++ b/core/kazoo_amqp/include/kz_amqp.hrl
@@ -220,6 +220,7 @@
                              ,timestamp=os:timestamp() :: kz_now() | '_'
                              ,zone='local' :: atom() | '$1' | '_'
                              ,manager=self() :: pid() | '_'
+                             ,started = os:timestamp() :: kz_now() | '$2' | '_'
                              ,tags = [] :: list() | '_'
                              ,hidden = 'false' :: boolean() | '$3' | '_'
                              }).
--- a/core/kazoo_amqp/src/kz_amqp_connections.erl
+++ b/core/kazoo_amqp/src/kz_amqp_connections.erl
@@ -201,14 +201,15 @@ primary_broker() ->
                                   ,zone='local'
                                   ,hidden='false'
                                   ,broker='$1'
+                                  ,started='$2'
                                   ,_='_'
                                   },
-    case lists:sort([Broker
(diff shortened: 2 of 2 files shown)


tsromana:
also if you think we should add an id to every amqp server, and filter according to that i will do that way :)

lazedo:
@tsromana hi, what are you trying to fix or decide here and why?

tsromana:
@lazedo so as discussed with @jamesaimonetti on irc, 

kazoo chooses the primary server based on its name, so e.g. when name server1 and server2 it will choose server1 as primary amqp server and when ip address e.g. 192.168.0.10 and 192.168.0.11 it will choose .10 as its primary server disregarding the position they are added in the config file,
which causes issue where kamailio is connected to one server and kazoo to another. 

To fix it this commit tries to make the primary server as the server whose connection  was first added, which translates to the one first mentioned in the config file by checking their connection pid.

jamesaimonetti:
@tsromana if connections are added synchronously (via `gen:call/2` vs `gen:cast/2`) then PID should be fine. It is theoretically possible that a new connection is added after the VM has been running for a long time and the PIDs have cycled around back to the start and so the newly added connection would be "first" now. I wonder if we could select and sort by `[#kz_amqp_connection.started, #kz_amqp_connection.connection]`?

tsromana:
@jamesaimonetti its is using `#kz_amqp_connections.connection` and i guess it does not have `#kz_amqp_connections.started` so should i add started to `#kz_amqp_connections` ? and in the above mentioned case where `pids` are cycled, it means that running process's pid can change overtime ?

tsromana:
@jamesaimonetti updated as requested, tested and its working. but i am just a bit skeptic as if its the right way, as should we do it differently, what you think :)

lazedo:
@tsromana this is not how it works. If you add more than 1 connection to the zone, those added connections should all be part of the same rabbit cluster. There's not a primary or secondary connection awareness in kazoo or a fallback mechanism.

tsromana:
ohk got it meaning from kazoo 4.1 rabbit should be clustered??

lazedo:
It's not a 4.1 thing.

tsromana:
i mean't is it recommended now to use rabbit in cluster as it was always considered to be a bad practice to use with kazoo

PR #3921, opened 2017-07-02
Formatters doc

jamesaimonetti:


k-anderson:
@jamesaimonetti shouldn't we move these to a core library?  If they are in the current location they would be loaded in system_schemas, and I don't think there is any reason to do so...

jamesaimonetti:
@k-anderson well, I debated that. Is there a reason not to load them, aside from storing them? Would someone like to access them via `/v2/schemas` for some reason? We could build a UI for creating formatters that would load `/v2/schemas/kapi.offnet_resource.req` to show you available keys for formatting, for instance.

Anyway, I didn't see a drawback to having them in the db / accessible via the API

applications/crossbar/priv/couchdb/schemas/kapi.dialplan.bridge.json
```diff
@@ -0,0 +1,196 @@
...
+                "peer"
+            ],
+            "type": "string"
+        },
+        "Enable-T38-Passthrough": {
+            "type": "string"
+        },
+        "Endpoints": {
+            "items": {
+                "type": "string"
```
lazedo: should this be `object` or `$ref` to `endpoint` ?

jamesaimonetti: yes, i was going to go through and update some to $ref'd objects. Probably do that now :)

jamesaimonetti:
@k-anderson appears something is up with erlydtl and building a release, not related to this PR.

jamesaimonetti:
@k-anderson @lazedo still a merl / relx issue and unrelated to the PR; want to merge?

PR #3923, opened 2017-07-03
qf-portreq: introduce and use validate_request_data_only/2,3,4

fenollp:
Fixes and issue where, after validating a scheduled port request, the PATCH would forget about the port request JSON and only continue with the req data.

`crossbar_doc:patch_and_validate` was not used here because `cb_port_requests` loads then checks/modifies its documents its own way and I do not want to deal with that mess.

### DIFF
--- a/applications/crossbar/src/cb_context.erl
+++ b/applications/crossbar/src/cb_context.erl
@@ -725,7 +725,8 @@ validate_request_data(SchemaId, Context, OnSuccess, OnFailure) ->
 
 -spec copy_req_data_to_doc(context()) -> context().
 copy_req_data_to_doc(Context) ->
-    set_doc(Context, req_data(Context)).
+    NewDoc = kz_json:merge_jobjs(kz_doc:private_fields(doc(Context)), req_data(Context)),
+    set_doc(Context, NewDoc).
 
 -spec validate_request_data_only(ne_binary(), context()) -> context().
 -spec validate_request_data_only(ne_binary(), context(), after_fun()) -> context().
--- a/applications/crossbar/src/cb_context.erl
+++ b/applications/crossbar/src/cb_context.erl
@@ -722,6 +722,7 @@ validate_request_data(SchemaId, Context, OnSuccess, OnFailure) ->
     OnPassing = fun copy_req_data_to_doc/1,
     do_validate_request_data(SchemaId, Context, OnSuccess, OnFailure, OnPassing).
 
+-spec copy_req_data_to_doc(context()) -> context().
 copy_req_data_to_doc(Context) ->
     set_doc(Context, req_data(Context)).
 
@@ -749,7 +750,7 @@ do_validate_request_data(SchemaId, Context, OnSuccess, OnFailure, OnPassing) ->
 
 -spec do_validate_request_data(ne_binary() | api_object(), context(), on_passing()) -> context().
 do_validate_request_data(undefined, Context, _) ->
-    lager:error("why validate then?"),
+    lager:error("schema id or schema JSON not defined, continuing anyway"),
     passed(Context);
 do_validate_request_data(?NE_BINARY=SchemaId, Context, OnPassing) ->
     Strict = fetch(Context, ensure_valid_schema, ?SHOULD_ENSURE_SCHEMA_IS_VALID),
--- a/applications/crossbar/src/cb_context.erl
+++ b/applications/crossbar/src/cb_context.erl
@@ -116,6 +116,8 @@
 -define(SHOULD_FAIL_ON_INVALID_DATA
        ,kapps_config:get_is_true(?CONFIG_CAT, <<"schema_strict_validation">>, false)).
 
+-type validation_error() :: jesse_error:error_reason().
+-type validation_errors() :: [validation_error()].
 
 -type context() :: #cb_context{}.
 -type setter_fun_1() :: fun((context()) -> context()).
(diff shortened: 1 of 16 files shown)


applications/crossbar/src/cb_context.erl
```diff
@@ -706,88 +714,95 @@ response(#cb_context{resp_error_code=Code
...
-validate_request_data('undefined', Context) ->
+validate_request_data(SchemaId, Context) ->
+    validate_request_data(SchemaId, Context, undefined).
+validate_request_data(SchemaId, Context, OnSuccess) ->
+    validate_request_data(SchemaId, Context, OnSuccess, undefined).
+validate_request_data(SchemaId, Context, OnSuccess, OnFailure) ->
+    OnPassing = fun copy_req_data_to_doc/1,
+    do_validate_request_data(SchemaId, Context, OnSuccess, OnFailure, OnPassing).
+
+copy_req_data_to_doc(Context) ->
```
k-anderson: @fenollp spec

applications/crossbar/src/cb_context.erl
```diff
@@ -706,88 +714,95 @@ response(#cb_context{resp_error_code=Code
...
+        #cb_context{resp_status='success'}=C1 when is_function(OnSuccess, 1) ->
+            OnSuccess(C1);
+        #cb_context{}=C2 when is_function(OnFailure, 1) ->
+            OnFailure(C2);
+        Else -> Else
+    end.
+
+-spec do_validate_request_data(ne_binary() | api_object(), context(), on_passing()) -> context().
+do_validate_request_data(undefined, Context, _) ->
+    lager:error("why validate then?"),
```
k-anderson: @fenollp please explain the error or remove the line ;)

lazedo:
@fenollp why these changes to validation ? it seems to me that you are setting the possible modified req_data with the validated result ? if that's the case, then defaults applied during validation are going to back to req_data which doesn't seem correct to me since those values were not sent by the client.

applications/teletype/src/teletype_port_utils.erl
```diff
@@ -193,13 +187,8 @@ fix_transfer_date(JObj) ->
...
 -spec fix_scheduled_date(kz_json:path()) -> kz_json:object().
 fix_scheduled_date(JObj) ->
-    ScheduledDate = case kz_json:get_integer_value([<<"scheduled_date">>, <<"local">>], JObj) of
-                        %% backward compatibility
-                        undefined -> kz_json:get_integer_value(<<"scheduled_date">>, JObj);
-                        Timestamp -> Timestamp
-                    end,
     kz_json:set_values([{<<"port_scheduled_date">>, kz_json:get_value(<<"scheduled_date">>, JObj)}
-                       ,{<<"scheduled_date">>, ScheduledDate}
+                       ,{<<"scheduled_date">>, kz_json:get_value([<<"scheduled_date">>, <<"local">>], JObj)}
```
icehess: did you test this? it shouldn't be integer?

fenollp: The value in there is a datetime. (it's what `localtime:local_to_*` returns)
Yay for storing non-JSON data in JSON objects amirite

fenollp:
Wait a little I have one more thing to add romorrow

fenollp:
This is ready now

lazedo:
Still feel that there are a lot of unnecessary changes here.
If you really need to do non-standard validation, why not call json_schemas:validate?

fenollp:
don't merge yet!

fenollp:
Ready for real this time!

applications/crossbar/src/crossbar_maintenance.erl
```diff
@@ -458,11 +458,11 @@ validate_user(JObj, Context) ->
                                  ,[{fun cb_context:set_req_data/2, JObj}
                                   ,{fun cb_context:set_req_nouns/2, [{?KZ_ACCOUNTS_DB, []}]}
                                   ,{fun cb_context:set_req_verb/2, ?HTTP_PUT}
-                                  ,{fun cb_context:set_resp_status/2, 'fatal'}
+                                  ,{fun cb_context:set_resp_status/2, success}
```
lazedo: wouldn't this make any request succeed ?

applications/crossbar/src/crossbar_maintenance.erl
```diff
@@ -432,7 +432,6 @@ validate_account(JObj, Context) ->
                                  ,[{fun cb_context:set_req_data/2, JObj}
                                   ,{fun cb_context:set_req_nouns/2, [{?KZ_ACCOUNTS_DB, []}]}
                                   ,{fun cb_context:set_req_verb/2, ?HTTP_PUT}
-                                  ,{fun cb_context:set_resp_status/2, success}
```
lazedo: why the remove of this ? it will use the current resp_status which may be `success`

applications/crossbar/src/crossbar_maintenance.erl
```diff
@@ -458,7 +457,6 @@ validate_user(JObj, Context) ->
...
```
lazedo: same here

applications/crossbar/src/crossbar_maintenance.erl
fenollp: Maybe. Without this validation fails with fatal and no errors though.
I'm trying with the resp_status preset in #cb_context{} now (=error instead of fatal).

I have seen a user validation failure and the reason would be because of missing "first_name" and "last_name" which are however clearly passed in crossbar_maintenance. Lot's of funky things...

applications/crossbar/src/crossbar_maintenance.erl
fenollp: (I removed your other comment cause it's exactly the same as this one)
See my comment about using record preset above.

PR #3924, opened 2017-07-05
use encoded node name in templates

icehess:
Add node to the `system` macro in teletype with encoded version (same for hostname) to allow track the sender server without revealing the name.

### DIFF
--- a/applications/teletype/src/teletype_util.erl
+++ b/applications/teletype/src/teletype_util.erl
@@ -330,7 +330,7 @@ default_content_transfer_encoding(_) -> <<"7BIT">>.
 
 -spec system_params() -> kz_proplist().
 system_params() ->
-    [{<<"hostname">>, kz_term:to_binary(net_adm:localhost())}].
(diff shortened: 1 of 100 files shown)


applications/teletype/src/teletype_util.erl
```diff
@@ -330,7 +330,11 @@ default_content_transfer_encoding(_) -> <<"7BIT">>.
 
 -spec system_params() -> kz_proplist().
 system_params() ->
-    [{<<"hostname">>, kz_term:to_binary(net_adm:localhost())}].
+    [{<<"hostname">>, kz_term:to_binary(net_adm:localhost())}
+    ,{<<"hostname">>, kz_base64url:encode(crypto:hash('md5', kz_term:to_binary(net_adm:localhost())))}
```
k-anderson: @icehess dont think you meant to define this twice ;)

k-anderson:
@icehess if you rebase should pass the ci test

PR #3928, opened 2017-07-05
fix-erlydtl_4oh: use latest tag for ErlyDTL dep

fenollp:


fenollp:
master: https://github.com/2600hz/kazoo/pull/3927

fenollp:
See https://github.com/ninenines/erlang.mk/issues/709

PR #3929, opened 2017-07-05
MWI function fix

tsromana:
`omnip_util:request_probe(Package, Username, Realm)`  was being called with value `user` instead of `username` which contains `username@realm` so the function was changed to `omnip_util:request_probe(Package, User)`

### DIFF
--- a/applications/omnipresence/src/omnip_subscriptions.erl
+++ b/applications/omnipresence/src/omnip_subscriptions.erl
@@ -453,7 +453,7 @@ subscribe(#omnip_subscription{user=_U
                              ,call_id=CallId
                              ,stalker=Stalker
                              ,contact=Contact
-                             ,normalized_user=User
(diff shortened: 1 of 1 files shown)


tsromana:
@lazedo sorry didn't understand what you meant, change the Event message `user` to `username` or change the function `omnip_util:request_probe(Package, Username, Realm)`.

lazedo:
@tsromana the call here should be to `omnip_util:request_probe(Package, Username, Realm)`
if you found that `Username` is actually user@realm then the calling code is passing the wrong param, we should fix the caller not the callee

tsromana:
@lazedo please let me know if anytime needs to be changed :)

lazedo:
@tsromana looks good. please rebase with latest master to pass CI

tsromana:
done!

PR #3930, opened 2017-07-06
Braintree Descriptor crash fix

tsromana:
while creating a new subscription braintree descriptor is undefined which crashes, bookkeeper sync


lazedo:
@tsromana thanks

PR #3931, opened 2017-07-06
Fix crossbar_util:response_400

kalda341:


BorigTheDwarf:
+1

fenollp:
Do you file like this is needed for crossbar modules pre-packaged in Kazoo or more for the ones you might have developed?

kalda341:
@fenollp This is needed wherever resp_error_msg is set in the validation stage. This occurs whenever crossbar_util:response_400 is used, such as in cb_phone_numbers_v2.
This is also used by additional crossbar modules I have developed.

lazedo:
@mark2600 ping
@kalda341 how about fixing in `crossbar_util:response_400` ?

mark2600:
I agree with these changes, we should not clobber an already set error_resp

PR #3934, opened 2017-07-06
circle_4oh: pass circle CI builds again

fenollp:
master: https://github.com/2600hz/kazoo/pull/3933
related issue: https://github.com/ninenines/erlang.mk/issues/709

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ clean-deps:
 	$(if $(wildcard deps/), rm -r deps/)
 
 .erlang.mk:
-	wget 'https://raw.githubusercontent.com/ninenines/erlang.mk/2017.05.18/erlang.mk' -O $(ROOT)/erlang.mk
+	wget 'https://raw.githubusercontent.com/ninenines/erlang.mk/2017.07.06/erlang.mk' -O $(ROOT)/erlang.mk
 
 deps: deps/Makefile
 	$(MAKE) -C deps/ all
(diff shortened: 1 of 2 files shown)


make/deps.mk
```diff
@@ -22,7 +22,6 @@ dep_detergent = git https://github.com/pap/detergent e86dfeded3e4f9f3f9278c6a1ae
 dep_jiffy = hex 0.14.7
 dep_nklib = git https://github.com/NetComposer/nklib
 dep_plists = git https://github.com/essen/plists
-dep_erlydtl = git https://github.com/erlydtl/erlydtl 0.12.0
```
lazedo: why are we removing this ?

Makefile
```diff
@@ -55,7 +55,7 @@ clean-deps:
 	$(if $(wildcard deps/), rm -r deps/)
 
 .erlang.mk:
-	wget 'https://raw.githubusercontent.com/ninenines/erlang.mk/2017.05.18/erlang.mk' -O $(ROOT)/erlang.mk
+	wget 'https://raw.githubusercontent.com/ninenines/erlang.mk/2017.07.06/erlang.mk' -O $(ROOT)/erlang.mk
```
lazedo: if erlang.mk team is deleting tags, maybe we should fork to ensure we use a valid tag over time ?

fenollp: the tag deletion episode was a one-of so we can still rely on that IMO

make/deps.mk
fenollp: cause it appears their master branch has less issues (today's merl issue is due to that)

PR #3939, opened 2017-07-07
HELP-30966: treat an internal FS error as an execution error

k-anderson:


applications/ecallmgr/src/ecallmgr_originate.erl
```diff
@@ -574,9 +574,6 @@ originate_execute(Node, Dialstrings, Timeout) ->
             Media = get('hold_media'),
             _Pid = kz_util:spawn(fun set_music_on_hold/3, [Node, UUID, Media]),
             {'ok', UUID};
-        {'ok', <<"-ERR ", Other/binary>>} ->
-            lager:debug("recv other 'ok': ~s", [Other]),
-            {'error', kz_util:strip_binary(binary:replace(Other, <<"\n">>, <<>>))};
```
lazedo: Why are we removing this?

PR #3942, opened 2017-07-11
MWI sub feature code fix

tsromana:
fix use case when feature code is used for MWI sub

### DIFF
--- a/applications/omnipresence/src/omnip_subscriptions.erl
+++ b/applications/omnipresence/src/omnip_subscriptions.erl
@@ -454,6 +454,7 @@ subscribe(#omnip_subscription{user=_U
                              ,stalker=Stalker
                              ,contact=Contact
                              ,username=Username
+                             ,normalized_from=From
                              ,realm=Realm
                              ,event=Event
                              }=S) ->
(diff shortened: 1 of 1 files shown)


lazedo:
@tsromana thanks. not sure what you're trying to fix, but this seems wrong. what's the use case here ? 
not sure what you mean with `feature code is used for MWI sub`

tsromana:
@lazedo sorry for not being clear, earlier the code was when `dialog` sub and username `*98username`, than send request, but after seeing some devices, it seems that some subscribe using `message-summary` so thats why `maybe_probe(_, {_, <<"*", _/binary>> = Username, From, Realm, _}) ` i have added the wildcard to accept both use cases, than after that the `From` is added because some devices register using `*98@realm` and there is no username inside the `username` section, so that's why if someone subscribe using voicemail number it looks for who subsribed for it, make sense? let me know if its not clear enough :)

lazedo:
@tsromana if the device is subscribing to `mwi` with *98@realm then its not well configured. you should configure mwi to subscribe to user@realm where user can be either the device user or a vm box.
the existing `dialog` handling in probe is to allow devices the use of blf keys to subscribe to mwi.
a side car configured to handle multiple voicemail boxes is an example if that.

tsromana:
@lazedo got it, sorry for the issue, still trying to learn how everything works, thanks for the help :)

PR #3945, opened 2017-07-11
Fix service plan not loading when no overrides present

powerpbx:
Service plans do not load when there are no overrides present.  I believe this will fix that.


lazedo:
@powerpbx thanks for the pr. it seems that the issue will be better handled at https://github.com/2600hz/kazoo/blob/master/core/kazoo_stdlib/src/kz_json.erl#L296
i'll let @jamesaimonetti figure out.

jamesaimonetti:
@lazedo, @powerpbx and I talked briefly on IRC about this. My thought was:

Undoing the change where merging the empty object onto a non-empty object removes the non-empty object properties (this would make it more like merge_recursive/2 in that case). Tests would need to be refactored to support this assumption. This was more invasive and had wider reaching consequences.

For now, ignoring empty objects as a no-op gets things going they way they should for service plan overrides. So I'm fine with this PR to unblock the issue with empty overrides, with the thought that we reevaluate `kz_json:merge/2` and empty object handling.

core/kazoo_services/src/kz_service_plans.erl
```diff
@@ -236,8 +238,9 @@ get_plan(PlanId, ResellerId, Services, ServicePlans) ->
 maybe_fetch_vendor_plan(PlanId, VendorId, VendorId, Overrides) ->
     case kz_service_plan:fetch(PlanId, VendorId) of
         'undefined' -> 'undefined';
-        ServicePlan ->
-            kzd_service_plan:merge_overrides(ServicePlan, Overrides)
+        ServicePlan when Overrides /= ?EMPTY_JSON_OBJECT ->
```
jamesaimonetti: My only comment is the project favors using `=/=` and `=:=` for equality tests.

fenollp:
Sorry to jump on this a bit late: to me it would be cleaner to have 

```erlang
OverridesAreEmpty = kz_json:is_empty(Overrides),
...
+        ServicePlan when OverridesAreEmpty ->
```

and not include the json header file altogether.

powerpbx:
What is the status of this? I tested my suggested fix and know it works.  I don't have my dev environment set up anymore so I am not in a position to test other peoples fixes.  It doesn't matter to me how it's done as long as it gets fixed.

What is the etiquette here when other people suggest something different?

lazedo:
@powerpbx request was made to have the changes suggested by @fenollp commited instead. it does not rely on macros defined elsewhere. if the json library changes, then this would have to change also.

powerpbx:
Do I need to do anything?

fenollp:
This needs rebase for circleci, sorry...

powerpbx:
I don't know what that means.

fenollp:
I can also do this patch myself, will probably backport it too. Is this fine with you?

powerpbx:
Hi,

Whatever you guys decide is fine with me.  Circleci is showing passed now.

lazedo:
@powerpbx thanks for spotting and for your effort on fixing this. we merged @fenollp PR's  for master and 4.0 because of rebase issues with your branch.

PR #3951, opened 2017-07-13
update device schema for media.webrtc on master

mark2600:


applications/crossbar/priv/couchdb/schemas/devices.json
```diff
@@ -241,6 +241,11 @@
                         }
                     },
                     "type": "object"
+                },
+                "webrtc": {
+                    "description": "If true forces a webrtc compatible SDP on the INVITE.",
+                    "required": false,
```
lazedo: `"required": false,` is for `draft-03`, please remove

PR #3953, opened 2017-07-14
Fix create number with auth_by system

kalda341:
Creating a number authorised by system currently crashes with the following stacktrace:

```
02:23:28.322 warning kzs_util.86 unknown type for database account%2Fsy%2Fst%2Fem
02:23:28.322 error kz_util.108 stacktrace:
02:23:28.323 error kz_util.120 st: kzs_util:db_classification/1 at (87)
02:23:28.323 error kz_util.120 st: kzs_plan:get_dataplan/1 at (70)
02:23:28.323 error kz_util.120 st: kzs_plan:plan/2 at (47)
02:23:28.323 error kz_util.120 st: kz_datamgr:open_doc/3 at (631)
02:23:28.323 error kz_util.120 st: kzs_cache:open_cache_doc/3 at (52)
02:23:28.323 error kz_util.120 st: knm_number:allow_number_additions/2 at (253)
02:23:28.323 error kz_util.120 st: knm_number:allowed_creation_states/2 at (171)
02:23:28.326 error kazoo_bindings.692 excepted: error: {badmatch,{error,not_found}}
02:23:28.326 error kz_util.108 stacktrace:
02:23:28.326 error kz_util.120 st: knm_number:allow_number_additions/2 at (253)
02:23:28.326 error kz_util.120 st: knm_number:allowed_creation_states/2 at (171)
02:23:28.326 error kz_util.120 st: knm_number:state_for_create/1 at (157)
02:23:28.326 error kz_util.120 st: knm_numbers:create/2 at (307)
02:23:28.326 error kz_util.120 st: knm_number:create/2 at (143)
02:23:28.326 error kz_util.120 st: cb_purchase_numbers:create_number_kazoo/2 at (291)
02:23:28.326 error kz_util.120 st: lists:foreach/2 at (1338)
```

This fixes that issue.

### DIFF
--- a/core/kazoo_number_manager/src/knm_number.erl
+++ b/core/kazoo_number_manager/src/knm_number.erl
@@ -249,6 +249,8 @@ ensure_can_create(Num, Options) ->
 -endif.
 
 -spec allow_number_additions(knm_number_options:options(), ne_binary()) -> boolean().
+allow_number_additions(_Options, ?KNM_DEFAULT_AUTH_BY) ->
+    'true';
 allow_number_additions(_Options, _AccountId) ->
     {'ok', JObj} = ?LOAD_ACCOUNT(_Options, _AccountId),
     kz_account:allow_number_additions(JObj).


fenollp:
Hi there, I am not sure this is the right way to fix this...
Are you calling `knm_number:create(Num, [{auth_by,?KNM_DEFAULT_AUTH_BY}|...])`?
Shouldn't you be calling this with `cb_context:auth_account_id/1` (from cb_purchase_numbers) instead?
Also, `knm_numbers` has a similar API and allows for bulk number operations.

You could call with `auth_by` set to the `MasterAccountId`, or any account you know has `wnm_allow_additions` set to true, too.

What do you think?

kalda341:
Yes I could probably use the master account id, but that's beside the point.
It currently breaks when you do something that seems reasonable. There is code to handle KNM_DEFAULT_AUTH_BY elsewhere, so why not here?

fenollp:
This case is not handled there because this function fetches a value that is account-specific. Passing it the equivalent of `sudo` sounds strange to me. Should `allowed_creation_states/2` thus return the full list?
Hm I guess it already does, given that `?KNM_DEFAULT_AUTH_BY` is considered an admin.
I'll talk about this to the team and get back to you :)

kalda341:
Regarding knm_numbers vs the loop, knm_numbers was not working correctly when porting this code from 3.22 (and I seem to remember the master account as auth_id also not working) and I had trouble debugging the number manager code,  so I ended up using the loop. I should probably refactor this to use knm_numbers now.

Do you view kazoo_number_manager as being stable at the moment? I've had to modify cb_purchase_numbers several times recently to work with kazoo_number_manager changes and would like to avoid having to do it again. 

The reason we have a custom API for purchasing numbers is we do it differently - rather than having numbers stockpiled we actually purchase them when a user wants them, and then add them to the account.

kalda341:
Thanks :)

fenollp:
KNM is pretty stable now :)
You should use `knm_numbers` over `knm_number` all the time (in 4.1 at least) because is does many things in bulk (DB reads & writes, but at some point more things too).
Buying numbers from carriers is not optimized for bulk requests though. If you want you can take a look into this [here is where you should start](https://github.com/2600hz/kazoo/blob/master/core/kazoo_number_manager/src/carriers/knm_carriers.erl#L237).
Note: the some carriers may not have a bulk order API, others have asynchronous APIs...

icehess:
This LGTM, and @fenollp have no issue with it too.

Thanks @kalda341.

PR #3957, opened 2017-07-14
Rework creating initial accounts

jamesaimonetti:


applications/crossbar/src/crossbar_maintenance.erl
```diff
@@ -435,13 +497,13 @@ validate_account(JObj, Context) ->
                                   ,{fun cb_context:set_resp_status/2, 'fatal'}
                                   ])
               ],
-    Context1 = crossbar_bindings:fold(<<"v1_resource.validate.accounts">>, Payload),
+    Context1 = crossbar_bindings:fold(<<"v2_resource.validate.accounts">>, Payload),
```
fenollp: There's no cb_accounts_v2 though, right? This does not actually change anything or am I missing something?

applications/crossbar/src/crossbar_maintenance.erl
```diff
@@ -413,20 +404,91 @@ create_account(AccountName, Realm, Username, Password) ->
...
     io:format("updated master account id in system_config.~s~n", [?KZ_SYSTEM_CONFIG_ACCOUNT]).
 
+-spec prechecks(cb_context:context()) -> {'ok', cb_context:context()}.
+prechecks(Context) ->
+    Funs = [fun db_accounts_exists/0
+           ,fun db_system_config_exists/0
+           ,fun db_system_schemas_exists/0
+           ],
+    'true' = lists:all(fun(F) -> F() end, Funs),
+    {'ok', Context}.
```
fenollp: So this is the only thing that PR adds, right? It is fixing the race condition that CircleCI found. Was there something else that changed in this PR?

applications/crossbar/src/crossbar_maintenance.erl
jamesaimonetti: There is no diff in /v1/accounts and /v2/accounts, this just makes sure if there was a split, that we would use /v2

circle.yml
```diff
@@ -39,7 +40,12 @@ test:
     - . ~/.kerl/$OTP_VERSION/activate && make apis
     - make docs
     - make validate-schemas
-    - make validate-swagger
+    - |
```
fenollp: Can you move this to a dedicated PR so that we can merge it ASAP?

core/kazoo_apps/src/kapps_notify_publisher.erl
```diff
@@ -97,7 +97,7 @@ maybe_handle_error(NotifyType, Req, Error) ->
 %% @doc save pay load to db to retry later
 -spec handle_error(ne_binary(), api_terms(), any()) -> 'ok'.
 handle_error(NotifyType, Req, Error) ->
-    lager:error("attempt for publishing notifcation ~s was unsuccessful: ~p", [NotifyType, Error]),
+    lager:warning("attempt for publishing notifcation ~s was unsuccessful: ~p", [NotifyType, Error]),
```
fenollp: Is hiding this from CI really the right solution?

rel/ci-sys.config
```diff
@@ -1,7 +1,7 @@
 [
  {lager, [
           {handlers, [
-                      {lager_console_backend, [debug,
+                      {lager_console_backend, [info,
```
fenollp: If we're doing this there should also be a debug backend going to file and have debug.log + error.log uploaded as CI artefacts so that they can be downloaded fully.

scripts/check-release-startup.sh
```diff
@@ -15,9 +15,9 @@ sup() {
 script() {
     sup crossbar_maintenance create_account 'compte_maitre' 'royaume' 'superduperuser' 'pwd!'
     sleep 3
-    sup kazoo_perf_maintenance json_metrics | python -m json.tool
+#    sup kazoo_perf_maintenance json_metrics | python -m json.tool
     sleep 1
-    sup kazoo_perf_maintenance graphite_metrics 'compte_maitre' 'clu1' 'royaume'
+#    sup kazoo_perf_maintenance graphite_metrics 'compte_maitre' 'clu1' 'royaume'
```
fenollp: Not sure why these would be disabled?

jamesaimonetti: Log clutter for no benefit afaict. Was going to ask tomorrow about why they're here.

circle.yml
jamesaimonetti: https://github.com/2600hz/kazoo/pull/3961

core/kazoo_apps/src/kapps_notify_publisher.erl
jamesaimonetti: According to @icehess these logs are expected for CircleCI. Will defer to him with proposal for this

icehess: I don't expect it to work here, there is no smtp server.

But what triggered `service_added` notification? From log it appears it triggered after `sup kapps_maintenance migrate`.

Also even without a smtp server, teletype should respond with a message says timeout to connect to smtp or something like that. This is weird yes, though it could be a race condition with initializing kazoo, creating the very first account and migration.

scripts/check-release-startup.sh
fenollp: Revert this. We don't remove tests because "they generate too much output".

PR #3958, opened 2017-07-15
KAZOO-5572: 'doodle' application is enabled by default

sergey-safarov:


lazedo:
@sergey-safarov hi, this is not something we're ready to accept. `doodle` is not `yet` a member of default apps

PR #3964, opened 2017-07-18
test-notifs: tests for teletype

fenollp:
This tests the generation of teletype templates from a few notifications. To make this easier a behaviour was introduced.
In the future it would be appreciable to test the whole 32 templates as well as their preview.
Since this PR is already pretty large I only went through 6 notifications (and not their preview, this should however be easier to test).


applications/teletype/src/teletype_util.erl
```diff
@@ -376,17 +388,34 @@ find_account_params(AccountId) ->
...
+fetch_account_for_params(?AN_ACCOUNT_ID) -> {ok, fixture("an_account.json")};
+fetch_account_for_params(?A_MASTER_ACCOUNT_ID) -> {ok, fixture("a_master_account.json")};
+fetch_account_for_params(?MATCH_ACCOUNT_RAW(_)) -> {error, testing_too_hard}.
+-else.
+fetch_account_for_params(AccountId) -> kz_account:fetch(AccountId).
+-endif.
+
+-ifdef(TEST).
+-spec fixture(nonempty_string()) -> kz_json:object().
+fixture(JSONFileName) ->
```
icehess: why not using `read_preview_doc/1`?

applications/teletype/test/teletype_render_tests.erl
```diff
@@ -0,0 +1,99 @@
...
+    %% Above is only when adding new tests
+    [?_assertEqual(T0, Line)
+     || {T0, Line} <- lists:zip(t0(TemplateId, CT), lines(iolist_to_binary(Rendered)))
+    ].
+
+t0(TemplateId, CT) ->
+    Ext = ct_to_ext(CT),
+    Path = filename:join([code:lib_dir(?APP), "test", <<TemplateId/binary,".",Ext/binary>>]),
+    ?LOG_DEBUG("reading t0 template ~s", [Path]),
+    {ok, Bin} = file:read_file(Path),
```
icehess: if `Path` is not exists call `overwrite_t0`

applications/teletype/test/teletype_render_tests.erl
```diff
@@ -0,0 +1,99 @@
...
+
+render(TemplateId, CT, Macros) ->
+    TmpModule = teletype_templates:renderer_name(TemplateId, CT),
+    {ok, Template} = fetch_template(TemplateId, CT),
+    {ok, Rendered} = kz_template:render(Template, TmpModule, Macros),
+    %% Below is only when adding new tests
+    %% overwrite_t0(TemplateId, CT, Rendered),
+    %% Above is only when adding new tests
+    [?_assertEqual(T0, Line)
+     || {T0, Line} <- lists:zip(t0(TemplateId, CT), lines(iolist_to_binary(Rendered)))
```
icehess: Even though I like testing rendered with a pre-rendered version, but it makes changing the template files and keeping the them in sync hard (changing a simple style or text in all templates needs changing 140 files)

fenollp: You realize that you just have to uncomment the line above to update your test data?

applications/teletype/test/teletype_render_tests.erl
fenollp: I'd rather not: uncomment the line to add new tests.

applications/teletype/src/teletype_util.erl
fenollp: As the description of this PR says: this is not testing preview documents.

icehess: the name of func says preview but it's actually a fixture

applications/teletype/test/teletype_render_tests.erl
icehess: another reason that I'm not liking it. But anyway I approved this one.

Btw it's good to check and print error or warning from erlydtl.

applications/teletype/src/teletype_util.erl
fenollp: `read_preview_doc/1` reads things in priv/preview_data. This `fixture` function is a test-mode only function that fetches files from test/. Why should I use read_preview_doc?

applications/teletype/test/teletype_render_tests.erl
fenollp: Well if there are errors/warnings to be printed from erlydtl I would love to have tests that check for them! Maybe you have some code that fetches these?

PR #3967, opened 2017-07-19
Allow HTTP partial response

kalda341:
This allows seeking of streamed media in HTML5, and will show the correct media length in Safari.


jamesaimonetti:
@kalda341 This looks good! Can you add a doc to `crossbar/doc` about doing range queries and include a curl example (or a sample request from Safari)? Would be good to show how to leverage this new functionality!

fenollp:
I'm not sure I get this. Range requests are already supported by the HTTP protocol through headers.
Why this?

kalda341:
@jamesaimonetti I've added documentation.

PR #3974, opened 2017-07-20
Update default conversion command

extremerotary:
Update default pdf conversion command to read success or failure based on the exit code of the script


fenollp:
Hi there, couple notes:
1. I agree with the stderr redirection
1. The error code checking is however not changing anything and should be reverted to `&& echo ...` (the syntax is wrong, that's why it fails CI)

fenollp:
Sorry the log redirection should probably be added to the other commands too. Otherwise LGTM :)

fenollp:
Thanks!

PR #3977, opened 2017-07-20
HELP-31124: fix check type and add numbers as reserved

k-anderson:


fenollp:
master: https://github.com/2600hz/kazoo/pull/3978

PR #3979, opened 2017-07-20
set default reverse_proxy, include req_id in auth log

icehess:
* set `reverse_proxies` default value `127.0.0.1`
* include crossbar request id in logging log
* return `client_ip` from logging attempts view
* add a system parameter to skip reverse proxy addresses check in crossbar


core/kazoo_modb/priv/couchdb/views/auth.json
```diff
@@ -62,6 +65,7 @@
                 "    'status': doc.status,",
                 "    'message': doc.message,",
                 "    'timestamp': doc.timestamp",
+                "    'client_ip': doc.client_ip"
```
fenollp: needs more commas

k-anderson:
@icehess can you add a system_config param to skip the maybe_allow_proxy_req and accept any x-forward -for please :)

icehess:
@k-anderson added

PR #3980, opened 2017-07-21
Allow get groups by user in crossbar

kalda341:


lazedo:
@kalda341 thanks

lazedo:
@kalda341 thanks

PR #3985, opened 2017-07-22
looks like a typo

onnet:


jamesaimonetti:
@fenollp can you check whether the code was meant to use the old `Context` and not the new `Context1` here?

onnet:
@fenollp @jamesaimonetti Hi again guys!
Here is my thoughts: not using Context1 you are loosing db_doc and unable to track numbers assignment. Over here for example: https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/modules/cb_connectivity.erl#L157-L158
Regards,

PR #3986, opened 2017-07-22
Added option to enable fail_on_single_reject

tsromana:
When forwarding call or during call failover, stepswitch creates bridge request with multiple endpoints which causes freeswitch to call again and again on the same device even if the call is rejected, this will allow to reject the call if one of the user's device cancel's the call. This does not effect group call.

Do not know if this is the right way to handle this issue but for the time being, it seems to work fine.

TODO: still need to added json docs for the kapps_config

### DIFF
--- a/core/kazoo_endpoint/src/kz_endpoint.erl
+++ b/core/kazoo_endpoint/src/kz_endpoint.erl
@@ -1468,6 +1468,7 @@ generate_ccvs(Endpoint, Call, CallFwd) ->
               ,fun maybe_set_owner_id/1
               ,fun maybe_set_account_id/1
               ,fun maybe_set_call_forward/1
+              ,fun maybe_set_call_forward_fail/1
               ,fun maybe_set_confirm_properties/1
               ,fun maybe_enable_fax/1
               ,fun maybe_enforce_security/1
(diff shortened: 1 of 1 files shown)


core/kazoo_endpoint/src/kz_endpoint.erl
```diff
@@ -1562,6 +1563,22 @@ bowout_settings('false') ->
...
     ].
 
+
+-spec maybe_set_call_forward_fail(ccv_acc()) -> ccv_acc().
+maybe_set_call_forward_fail({_Endpoint, _Call, 'undefined', _CCVs}=Acc) ->
+    Acc;
+maybe_set_call_forward_fail({Endpoint, Call, CallFwd, CCVs}=Acc) ->
+    case kapps_config:get_is_true(?CONFIG_CAT, <<"fail_on_single_call_forward_reject">>, 'false') of
+        'true' ->
+            Fail = kapps_config:get_binary(?CONFIG_CAT, <<"fail_on_reject_val">>, <<"NORMAL_CLEARING">>),
```
jamesaimonetti: I would consider using a config name like `fail_on_reject_cause` or something more specific.

fenollp: And please use `get_ne_binary/3`.

tsromana: @jamesaimonetti @fenollp will do so :) also do you know about behaviour of `fail_on_single_reject` apparently if i send `NORMAL_CLEARING` only than also it disconnects the calls, even if the disconnect reason is `NO_ANSWER` or `USER_BUSY`. i have tried `^^:NORMAL_CLEARNING` that also seems to do the same, or normal termination is a part of `NORMAL_CLEARNING`, do you guys know anything about it. its not a big issue just a bit of annoyance.

icehess: @tsromana maybe you have to set `continue_on_fail` too? setting `fail_on_single_reject` would cause the bridge attempt to failed, and if the `fail_on_single_reject` has not been set the call is terminated.

tsromana: @icehess thanks for the help, i have tried continue_on_fail too. after some digging it seems that, when fail_on_single_reject is set to something it will set that value with three more causes which are `NORMAL_CLEARNING` `USER_BUSY` and `NO_ANSWER`

tsromana:
closing this will reopen with all the work done :)

PR #3988, opened 2017-07-24
add comment to cccp pin/cid

onnet:
If user has several pins, it would be convenient to add notes to them..

### DIFF
--- a/applications/crossbar/doc/cccps.md
+++ b/applications/crossbar/doc/cccps.md
@@ -12,6 +12,7 @@ Key | Description | Type | Default | Required
 --- | ----------- | ---- | ------- | --------
 `active` | Show's whether CID/PIN active | `boolean()` | `false` | `false`
 `cid` | CID to authorize | `string()` |   | `false`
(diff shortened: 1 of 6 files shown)


jamesaimonetti:
Hi @onnet! Typically, fields emitted in a view are also fields that exist in the schema. So adding `doc.comments` to the view while not adding it to the schema, to ensure it is of a valid form, is unusual.

Would you mind adding a `comments` property (and sub-property for a comment) to the schema for cccp?

onnet:
@jamesaimonetti Hi, sure, will do!

onnet:
@jamesaimonetti Done.

fenollp:
Does this have anything to do with https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/modules/cb_comments.erl ?

onnet:
@fenollp Sounds similar, but no, it is just one more field for cccp docs crossbar_listing view :)

jamesaimonetti:
Merged! Thanks @onnet

PR #3989, opened 2017-07-24
PISTON-459: if all resources have Bypass-Media, still set bypass_media=true

danielfinke:
Previous version of this function would only enable bypass_media if there was a single endpoint and Bypass-Media was enabled on it. New version enables bypass_media as long as _all_ endpoints have Bypass-Media set to true.

### DIFF
--- a/applications/ecallmgr/src/ecallmgr_fs_bridge.erl
+++ b/applications/ecallmgr/src/ecallmgr_fs_bridge.erl
@@ -158,15 +158,27 @@ handle_bypass_media(DP, _Node, _UUID, #channel{profile=ChannelProfile}, JObj) ->
     end.
 
 -spec maybe_bypass_endpoint_media(kz_json:objects(), ne_binary(), ne_binary(), kz_proplist()) -> kz_proplist().
-maybe_bypass_endpoint_media([Endpoint], BridgeProfile, ChannelProfile, DP) ->
-    EndpointProfile = kz_json:get_value(<<"SIP-Interface">>, Endpoint, BridgeProfile),
-    case kz_json:is_true(<<"Bypass-Media">>, Endpoint)
-        andalso EndpointProfile =:= ChannelProfile of
+maybe_bypass_endpoint_media(Endpoints, BridgeProfile, ChannelProfile, DP) ->
+    BypassOnAll = lists:foldl(fun(EP, 'none') ->
+                                      case bypass_endpoint_media_enabled(EP, BridgeProfile, ChannelProfile) of
+                                          'true' -> 'true';
+                                          'false' -> 'none'
+                                      end;
+                                 (EP, BypassOnAll1) ->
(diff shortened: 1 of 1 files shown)


applications/ecallmgr/src/ecallmgr_fs_bridge.erl
```diff
@@ -158,15 +158,27 @@ handle_bypass_media(DP, _Node, _UUID, #channel{profile=ChannelProfile}, JObj) ->
     end.
 
 -spec maybe_bypass_endpoint_media(kz_json:objects(), ne_binary(), ne_binary(), kz_proplist()) -> kz_proplist().
-maybe_bypass_endpoint_media([Endpoint], BridgeProfile, ChannelProfile, DP) ->
-    EndpointProfile = kz_json:get_value(<<"SIP-Interface">>, Endpoint, BridgeProfile),
-    case kz_json:is_true(<<"Bypass-Media">>, Endpoint)
-        andalso EndpointProfile =:= ChannelProfile of
+maybe_bypass_endpoint_media(Endpoints, BridgeProfile, ChannelProfile, DP) ->
+    BypassOnAll = lists:foldl(fun(EP, 'none') ->
```
jamesaimonetti: `lists:all/2` feels more appropriate to use instead of the fold.

```erlang
ShouldBypass = lists:all(fun(Endpoint) -> is_bypass_endpoint_media_enabled(Endpoint, BridgeProfile, ChannelProfile) end, Endpoints),
case ShouldBypass of
    'true' -> [{...} | DP];
    'false' -> DP
end.
```
Something like that looks a little more concise (but does it work? :) )

applications/ecallmgr/src/ecallmgr_fs_bridge.erl
```diff
@@ -158,15 +158,27 @@ handle_bypass_media(DP, _Node, _UUID, #channel{profile=ChannelProfile}, JObj) ->
...
-    end;
-maybe_bypass_endpoint_media(_, _, _, DP) ->
-    DP.
+        'false' -> DP;
+        'none' -> DP
+    end.
+
+-spec bypass_endpoint_media_enabled(kz_json:object(), ne_binary(), ne_binary()) -> boolean().
+bypass_endpoint_media_enabled(Endpoint, BridgeProfile, ChannelProfile) ->
+    EndpointProfile = kz_json:get_value(<<"SIP-Interface">>, Endpoint, BridgeProfile),
```
fenollp: If this is json, `=:=` cannot be used.
If this is an `ne_binary()`, please use `get_ne_binary_value`.

danielfinke:
@jamesaimonetti @fenollp Changes addressed, thanks for the suggestions 👍

jamesaimonetti:
I like it but will defer to @k-anderson or @lazedo for the final call.

PR #3990, opened 2017-07-24
Add previously failing test and update jesse dep location

jamesaimonetti:
With the 2600Hz branch of jesse, the supplied test would fail. Pointing to the 1.5 rc2 in the for-GET repo causes the test to pass successfully.

### DIFF
--- a/core/kazoo_schemas/src/kz_json_schema.erl
+++ b/core/kazoo_schemas/src/kz_json_schema.erl
@@ -896,4 +896,4 @@ fix_path(Path) ->
 -spec fix_el(kz_json:key() | non_neg_integer()) -> kz_json:key() | non_neg_integer().
 fix_el(I) when is_integer(I) -> I+1;
 fix_el(El) -> El.
-    
+
--- a/core/kazoo_schemas/src/kz_json_schema.erl
+++ b/core/kazoo_schemas/src/kz_json_schema.erl
@@ -33,7 +33,7 @@
 -define(DEFAULT_OPTIONS, [{'schema_loader_fun', fun load/1}
                          ,{'allowed_errors', 'infinity'}
                          ,{'extra_validator', fun kz_json_schema_extensions:extra_validator/2}
-                         ,{'setter_fun', fun kz_json:set_value/3}
+                         ,{'setter_fun', fun set_value/3}
                          ,{'validator_options', ['use_defaults'
                                                 ,'apply_defaults_to_empty_objects'
                                                 ]}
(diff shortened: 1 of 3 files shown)


jamesaimonetti:
@onnet can you try this PR locally? Make sure you `rm -r deps/jesse && make deps` to pull in the new version of jesse.

fenollp:
@jamesaimonetti These are the tests you are looking for core/kazoo_documents/test/kzd_callflow_test.erl

onnet:
@jamesaimonetti Sure, will do

lazedo:
I think defaults are not merged yet upstream. What's failing on our repo that is not in upstream?

jamesaimonetti:
@lazedo i'm rebasing our 1.5-rc5 against upstream's 1.5.0-rc2 (which works) and will tag 1.5-rc6 in our repo. Will update this PR once that's done (assuming it works).

onnet:
@jamesaimonetti Added patch to my 4.1 and problem disappeared. Thanks!

jamesaimonetti:
@onnet great! But I need to patch jesse to support our custom extensions so I wouldn't roll this out just yet :)

jamesaimonetti:
@lazedo I think the issue is that we used to have separate steps for validating the object and setting defaults. If you look at how [we iterate on arrays](https://github.com/2600hz/kazoo/blob/master/core/kazoo_schemas/src/kz_json_schema.erl#L128) vs [how jesse iterates](https://github.com/2600hz/jesse/blob/1.5-rc5/src/jesse_validator_draft4.erl#L550), you'll see we use the Erlang 1-indexed lists and jesse uses JSON 0-indexed arrays.

So the bug is that our jesse's set_default/3 is building the JSON path using the 0-indexed number but trying to call `kz_json:set_value/3` which uses it as the default. We can see in this test case, where we add a DID to object 3 (JSON index 2, Erlang index 3) that jesse is calling kz_json:set_value with `setting [<<"servers">>,2,<<"DIDs">>,<<"+78121111111">>,<<"force_outbound">>] to false` as the path (logging I added to jesse_validator_draft4). So it finds object 2 (server BBB1) and sets the default (which creates the DID on the BBB1 object.

Thoughts on using `kz_json_schema:set_value/3` to increment indexes before calling `kz_json:set_value/3`?

lazedo:
Great finding. I'm ok with your solution to fix the path.

lazedo:
@onnet can you test this again ? Tks

onnet:
Hi @lazedo, retested, still works OK. Regards,

PR #3991, opened 2017-07-24
Allow fetch full docs in crossbar

kalda341:


lazedo:
Why do you need this?
I don't think this is acceptable as is.

jamesaimonetti:
I would add, this may be fine for summaries with docs in the 10s but will easily break all the things when you get into 100s or 1000s of docs.

If I were to accept this, I would at least require a system config flag to allow the use of this request flag (defaulting to false - not allowed - of course). Then admins can choose whether to allow their clients to make this type of request.

kalda341:
@lazedo Often the summary does not contain enough information to be useful, but n+1 (where n is the number of docs to be loaded) requests is too inefficient. This was the solution I came up with.

@jamesaimonetti I agree that it could be slow for loading a very large number of documents, but that's what pagination is for! Adding the system config option is a good idea.

lazedo:
@kalda341 how often? Can you provide examples? And isn't this leaking private fields?
I would prefer to have a way to load alternative views

kalda341:
@lazedo No private fields remain in the response JSON. 
For example, a fetch all of groups contains no information about what is in the group.

BorigTheDwarf:
This would be extremely useful for me. I am currently writing an application which needs to fetch a lot of information at once about users and groups within Kazoo. Currently I have to do individual requests for each user and three different requests per group. Performance is pretty poor :(

fenollp:
Time for a GraphQL crossbar module @icehess ! :)

grahamsnz:
@lazedo @jamesaimonetti This would be very helpful for us when retrieving CDR records (and other record types). Very often, when retrieving the shortened records from the default list view, we need specific fields that are missing from the list view. The result is an initial list query and then often thousands of individual queries. 

Using a flag for full documents from the list view would help us reduce the number of API queries needed to access the same data. My feeling is that with pagination as a safety, this flag for full documents would likely reduce the overall load to retrieve the same information from Crossbar. 

If this specific flag is problematic, perhaps we can use a different or more RESTful query format such as `/v2/{account_id}/{document_type}/all` to show all fields for a specific document type?

lazedo:
I still prefer to provide a way to load alternative views or have PRs to change existing ones if needed.

BorigTheDwarf:
@Lazedo, could you explain why this is preferable?

On 27 Jul 2017 4:32 a.m., "lazedo" <notifications@github.com> wrote:

I still prefer to provide a way to load alternative views or have PRs to
change existing ones if needed.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/2600hz/kazoo/pull/3991#issuecomment-318109279>, or mute
the thread

jamesaimonetti:
@fenollp should be a new app but yes! @icehess make that happen (name it well) ;)

@grahamsnz I get the utility but I also know the potential for accidental or intentional abuse. Step one is hiding it behind a config flag so system administrators can opt-in (hopefully knowing the risks it poses). Like @lazedo I wonder if there is a field or three that are commonly wanted, and missing from the summary view, that could be added to the summary view in place of this.

Another thought is introducing a filter to the collection endpoints that allows you to request a list of IDs. For instance, you could hit `/v2/accounts/{ACCOUNT_ID}/cdrs?filter_id=XYZ&filter_id=ABC&...` or `filter_ids=[XYZ, ABC,...]`

kalda341:
Perhaps I should clarify exactly what we want this for.
We are working on a web app which will load all the data from several endpoints on startup, and keep this data up to date using  blackhole object. Making individual requests is a pain, and is slower than we would like.
We need the full data and not just a summary, so that we can update these documents without having to perform a get immediately before the post. A patch is not sufficient as we need to be able to delete fields.

I like all these suggestions a lot and GraphQL would be **AWESOME**, but they don't completely fit our use case.

jamesaimonetti:
@kalda341 Thanks for the use case. I think I didn't explain my filter idea. It basically mirrors how Couch's `all_docs` query works where you can specify a list of "keys" that map to the IDs you want to fetch. So not quite a view but allows you to fetch bigger chunks of data.

Essentially, `/v2/accounts/{ACCOUNT_ID}/cdrs?keys=[ABC, XYZ,...]` would turn into `kz_datamgr:all_docs(<<"account_modb">>, [{'keys', Keys},include_docs])` with the results being the full doc of public fields (as if you have executed GETs for ABC, XYZ, etc). This will also do type checks so you can't include IDs of users in a request to /cdrs and get results.

lazedo:
You can delete fields/ objects with patch.
Set the field to null ( or nul can't remember)
"media: null" will delete the media field/object

kalda341:
@lazedo That's so helpful! Thanks!

kalda341:
@jamesaimonetti So there's nothing that we can change about our approach to make you happy to merge it?
If that's the case, your filter suggestion is the next best thing. Should I write it and submit a pull request, or would you rather do it yourself? If I am going to develop it (which I am happy to do), can you suggest a preferred approach?

Thanks!

jamesaimonetti:
@kalda341 realize we never circled back here. Would be great if you amended the PR to allow for the keys querystring option.

icehess:
You can also consider using  the new `crossbar_view` module for adding this. The new module is supporting chunked response, so it's more memory efficient for the crossbar node.

PR #3992, opened 2017-07-25
no need to filter as from_list already does it

sergey-safarov:
Related to https://github.com/2600hz/kazoo-configs-freeswitch/pull/3


lazedo:
Is there a reason for this?

sergey-safarov:
Props filter changes done only to allow pass `rtcp_mux` to freeswitch.
Now such changes is no more required.

lazedo:
What if we want to pass false?

sergey-safarov:
Then 'false' value is filtered at `props:filter` and `rtcp_mux` will be disabled by FS global var.

lazedo:
'False' is not filtered by filter_undefined, I would leave this as is or even change it to from_list since it already does a filter_undefined

PR #3993, opened 2017-07-25
Added feature to enable rtcp_mux for calls to SIP endpoints

sergey-safarov:


sergey-safarov:
Related to https://github.com/2600hz/kazoo-configs-freeswitch/pull/3

lazedo:
Hi, I think we should stay with freeswitch defaults. What's the reason for inverting the logic here?

sergey-safarov:
Kazoo designed to define all CCV as default false.
Now rtcp_mux implemented like other vars.

About default FreeSwitch value.
We can define `true` value via kapps_config call as default, then nothing changed from FreeSwitch view.

sergey-safarov:
@lazedo also please look Fs profile config. RTCP is disabled at sip-profile setting. And rtcp_mux value is not important.

lazedo:
But why would we want to change the way freeswitch works by default?

sergey-safarov:
@lazedo is will be acceptable if I set default rtcp_mux to `true` at kazoo level.
Then FreeSwitch behavior will same as current.

lazedo:
Can you answer the question? Why would we want to change default behavior?

sergey-safarov:
Because kazoo pass 'media -> rtcp_mux' of endpoint json is not possible to pass in current implementation.
You not ready to merge PR with workarround (bad code) in https://github.com/2600hz/kazoo/pull/3872.

lazedo:
So the change would be to enable passing false from kazoo_endpoint. It seems to me that changing the default behavior because of this is not acceptable

lazedo:
Maybe someone from the team can help you with #3872. I can try to help you but only on next week

core/kazoo_endpoint/src/kz_endpoint.erl
```diff
@@ -1552,6 +1554,11 @@ maybe_set_call_forward({Endpoint, Call, CallFwd, CCVs}) ->
                        )
     }.
 
+-spec maybe_set_rtcp_mux(ccv_acc()) -> ccv_acc().
+maybe_set_rtcp_mux({Endpoint, Call, CallFwd, CCVs}) ->
+    RTCP_MUX = kz_json:get_binary_value([<<"media">>, <<"rtcp_mux">>], Endpoint, ?DEFAULT_RTCP_MUX),
```
fenollp: Shouldn't this be a call to `is_true`?

sergey-safarov: I will update after 1 hour

lazedo:
@sergey-safarov we still don't see why we would change the defaults in freeswitch and this PR  depends on it. maybe close this in favour of #3872 (and fix the merge code) ?

PR #3994, opened 2017-07-25
Crossbar auth: support for refresh tokens to acquire extended JWT

danielfinke:
_This code isn't ready for merge. Looking for suggestions first._
Since the switch to JWTs in Kazoo 4, we've had a lot of frustrated users about session timeouts even while users are active. Instead of simply extending the timer, we wanted a way to retain short idle timeouts.
This PR adds in "refresh tokens" - tokens that are long duration, but one-time use. When a user authenticates, a refresh token is generated and included in the response data. This token can be swapped for a brand new JWT at any time before expiry. Once it is used, it is invalid for re-use. At any time, an authenticated user can revoke one or all of their active refresh tokens.
Please give your suggestions on how this should be improved. There are 2 areas in particular that still need to be addressed:

- DELETEing expired refresh tokens from db
- compacting the token_auth db

### DIFF
--- a/applications/crossbar/src/crossbar_auth.erl
+++ b/applications/crossbar/src/crossbar_auth.erl
@@ -9,6 +9,7 @@
 
 -export([create_auth_token/2
         ,validate_auth_token/1, validate_auth_token/2
+        ,validate_refresh_token/1
         ,authorize_auth_token/1
         ,log_success_auth/4, log_success_auth/5, log_success_auth/6
         ,log_failed_auth/4, log_failed_auth/5, log_failed_auth/6
@@ -20,6 +21,9 @@
 -define(TOKEN_AUTH_EXPIRY(Method, AuthConfig)
        ,kz_json:get_integer_value(method_config_path(Method, <<"token_auth_expiry">>), AuthConfig, ?DEFAULT_AUTH_EXPIRY)
        ).
+-define(TOKEN_AUTH_REFRESH_EXPIRY, kapps_config:get_integer(?APP_NAME, <<"token_auth_refresh_expiry", ?SECONDS_IN_DAY)).
+
+-define(ACCOUNT_AUTH_CONFIG_ID, <<"kazoo_auth_configs">>).
 
 -define(SHOULD_LOG_FAILED,
         kapps_config:get_is_true(?AUTH_CONFIG_CAT, <<"log_failed_attempts">>, 'false')
@@ -90,7 +94,8 @@ create_auth_token(Context, Method, JObj) ->
                       ,{<<"owner_id">>, OwnerId}
                       ]),
             RespObj = kz_json:set_values(Props, kz_json:delete_key(<<"Claims">>, JObj)),
-            Resp = crossbar_util:response_auth(RespObj, AccountId, OwnerId),
+            AuthRefreshToken = maybe_create_auth_refresh_token(AccountId, OwnerId),
+            Resp = crossbar_util:response_auth(RespObj, AccountId, OwnerId, AuthRefreshToken),
 
             lager:debug("created new local auth token: ~s", [kz_json:encode(Resp)]),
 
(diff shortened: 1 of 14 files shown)


applications/crossbar/src/crossbar_auth.erl
```diff
@@ -160,13 +183,41 @@ validate_auth_token(Token, Options) ->
...
+    case kz_datamgr:open_cache_doc(?KZ_TOKEN_DB, Token) of
+        {'ok', TokenJObj} -> validate_refresh_token_expiry(TokenJObj);
+        E -> E
+    end.
+
+-spec validate_refresh_token_expiry(kz_json:object()) ->
+                                           {'ok', kz_json:object()} |
+                                           {'error', ne_binary()}.
+validate_refresh_token_expiry(TokenJObj) ->
+    case erlang:system_time('seconds') > kz_json:get_integer_value(<<"exp">>, TokenJObj) of
```
jamesaimonetti: Look at `kz_time:elapsed_s/2` as an alternative (or just use `kz_time:now_s()` again).

applications/crossbar/src/modules/cb_token_auth.erl
```diff
@@ -42,30 +49,66 @@ init() ->
...
+
+resource_exists(?REFRESH_TOKENS_PATH_TOKEN, _RefreshToken) -> 'true'.
+
 -spec validate(cb_context:context()) -> cb_context:context().
+-spec validate(cb_context:context(), path_token()) -> cb_context:context().
+-spec validate(cb_context:context(), path_token(), path_token()) -> cb_context:context().
 validate(Context) ->
     cb_context:put_reqid(Context),
-    validate(Context, cb_context:req_verb(Context)).
+    validate_1(Context, cb_context:req_verb(Context)).
```
jamesaimonetti: This naming goes against how most crossbar modules name the specific validators. I would name this `validate_tokens/2`, and `validate_3` should probably be `validate_refresh_token`.

applications/crossbar/src/modules/cb_token_auth.erl
```diff
@@ -114,21 +230,28 @@ authorize(Context) ->
...
                           {'true' | 'halt', cb_context:context()}.
 authenticate(Context) ->
     cb_context:put_reqid(Context),
-    authenticate(Context, cb_context:auth_account_id(Context), cb_context:auth_token_type(Context)).
+    authenticate(Context, cb_context:auth_account_id(Context), cb_context:req_nouns(Context), cb_context:auth_token_type(Context)).
 
-authenticate(_Context, ?NE_BINARY = _AccountId, 'x-auth-token') -> 'true';
-authenticate(Context, 'undefined', 'x-auth-token') ->
+authenticate(Context, _, [{<<"token_auth">>, [?REFRESH_PATH_TOKEN]}], _) ->
+    case is_rate_limited(Context) of
```
jamesaimonetti: The clauses seem backwards? Or the naming of this function is backwards? :)

lazedo:
First impression, Refresh tokens should be used as part of implementing oauth server inside kazoo. I don't see this as a desired  approach.

danielfinke:
@lazedo doesn't sound like a bad idea to me. I'm looking at the kazoo_auth and kazoo_oauth core applications. They appear to both be auth clients, kazoo_oauth being an older version of the same code. Can you verify this?
And that I am right in seeing that there is no server side implementation yet?

lazedo:
Kazoo_oauth is being deprecated in favour of kazoo_auth. The work to implement oauth server in kazoo is still at early stage and not in the scope of this PR which tries to address a different use case.
Yu can have jwt tiken wíth larger expiry times and renew them when needed as it is aleady done in monster-ui.

danielfinke:
@lazedo cool, thanks.

I agree, it's a bigger project than the scope of this PR. I am interested in exploring the implementation of OAuth 2 server. It looks like the appropriate model for the problem I'm trying to solve is the [Resource Owner Password Credentials Grant](https://tools.ietf.org/html/draft-ietf-oauth-v2-31#section-4.3) grant mode of the OAuth 2 spec. Other reading (https://aaronparecki.com/oauth-2-simplified/#others) backs this up (it's what is used by the client apps created by the service itself, i.e. Twitter app against Twitter OAuth 2). This grant mode is permitted to return refresh tokens. So we'd use this for first-party authentication in crossbar.

Any chance you could direct me to where this refresh is done in Monster today to solve my problem in the short-term?

k-anderson:
@danielfinke Howdy, I must admit I don't like this approach.  Primarily, the token_auth database was a scaling issue that was resolved by moving to the JWT.  I am worried that writing the referesh tokens to the db is re-introducing that original issue.  Additionally, having another mechanism for authentication (session renewal) feels awkward and potentially insecure.  That is, unless the refresh preforms the same checks against external systems (such as SSO integrations) then it bypasses external business logic.  If it does the same checks then its duplicating the existing authentication mechanisms.

The use case is valid, and the current approach by Monster UI is not ideal (sorry I am not sure where it is implemented, you could ping JR or Joris about that).  If I had to come up with a suggestion (that I have not solicited the team for input on yet, so take it lightly) I think I would recommend a way to renew a JWT with the JWT.  That is to say, after the application gets the initial JWT it would be informed when that token expires.  Prior to the expiration the application could issue the JWT to an API that would re-execute the JWT creation using the provided information (and same code path as initial creation) then possibly issue a new token and expiration.  Rinse and repeat.

As much as I hate to turn away work, in this case I am going to reject the PR.  Normally we would leave it open as we had no place to continue the design discussion but we have fancy new forums now and I request that we move this discussion to there (forums.2600hz.com/forums/)!  I also hope that you don't think we don't appreciate the work done but this is not how we would like to solve this valid issue.

danielfinke:
@lazedo @k-anderson happy to move the discussion. Created a forum post [here](https://forums.2600hz.com/forums/topic/9033-oauth-2-server-refresh-tokens-jwt/). Would appreciate some more chat if you guys are up for it!

PR #3996, opened 2017-07-26
Ensure that when creating the first account, if it fails, so will CircleCI

jamesaimonetti:
More reports coming in of first account creation failing. Better logging of errors from Crossbar helps; this PR will make sure Circle lets us know if it fails there too. Hopefully then we can analyze the logs to see what's going on and how to address this intermittent problem

### DIFF
--- a/core/sup/src/sup.erl
+++ b/core/sup/src/sup.erl
@@ -61,11 +61,11 @@ main(CommandLineArgs, Loops) ->
                 {'badrpc', {'EXIT',{'undef', _}}} ->
                     print_invalid_cli_args(CommandLineArgs);
                 {badrpc, {'EXIT', {timeout_value,[{Module,Function,_,_}|_]}}} ->
-                    stderr("Command failed: timeout", []),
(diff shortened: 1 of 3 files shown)


core/sup/src/sup.erl
```diff
@@ -172,9 +169,10 @@ long_or_short_name() ->
         'false' -> 'shortnames'
     end.
 
--spec print_invalid_cli_args() -> no_return().
-print_invalid_cli_args() ->
-    stderr("Invalid command or wrong number of arguments, please try again", []),
+-spec print_invalid_cli_args(string()) -> no_return().
+print_invalid_cli_args(Args) ->
+    stderr("Invalid command or wrong number of arguments, please try again:", []),
+    stderr("~p", [Args]),
```
fenollp: Not sure what printing arguments brings to the table?

jamesaimonetti: Relic from debugging things going wrong, will remove

jamesaimonetti:
@fenollp ready

PR #3997, opened 2017-07-26
Create admin user

jamesaimonetti:
With changes to how PATCH worked, the _rev from the account creation step would be set on the admin user doc. This clears out the 'doc' portion of the context so the user's doc doesn't pick up account-related pvt fields.


jamesaimonetti:
@fenollp may as well merge this too

PR #3998, opened 2017-07-26
PISTON-460: truncate DTMF after terminators in early DTMF before cf_collect_dtmf

danielfinke:
While cf_collect_dtmf was already checking whether the max number of DTMF digits were entered early via kapps_call:get_dtmf_collection/1, it wasn't checking if the digits contained terminators.

We were getting odd cases where despite '#' being a terminator, it was appearing in the collected DTMF after this module. The culprit was the passing the length check without honouring the terminator.

After this PR, if a terminator is found in early DTMF, it is ignored, and subsequent digits are truncated.

### DIFF
--- a/applications/callflow/src/module/cf_collect_dtmf.erl
+++ b/applications/callflow/src/module/cf_collect_dtmf.erl
@@ -36,8 +36,9 @@ handle(Data, Call) ->
             'undefined' -> <<>>;
             <<_/binary>> = D -> D
         end,
+    AlreadyCollected1 = truncate_after_terminator(AlreadyCollected, terminators(Data)),
 
-    maybe_collect_more_digits(Data, kapps_call:set_dtmf_collection('undefined', Call), AlreadyCollected).
+    maybe_collect_more_digits(Data, kapps_call:set_dtmf_collection('undefined', Call), AlreadyCollected1).
 
 -spec maybe_collect_more_digits(kz_json:object(), kapps_call:call(), binary()) -> 'ok'.
 maybe_collect_more_digits(Data, Call, AlreadyCollected) ->
@@ -77,6 +78,10 @@ collect_more_digits(Data, Call, AlreadyCollected, MaxDigits) ->
         {'error', _E} ->
             lager:debug("failed to collect DTMF: ~p", [_E])
     end.
+
+-spec truncate_after_terminator(binary(), ne_binaries()) -> binary().
+truncate_after_terminator(AlreadyCollected, Terminators) ->
+    hd(binary:split(AlreadyCollected, Terminators, ['global'])).
 
 -spec collection_name(kz_json:object()) -> ne_binary().
 collection_name(Data) ->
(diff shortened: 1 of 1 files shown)


applications/callflow/src/module/cf_collect_dtmf.erl
```diff
@@ -78,6 +79,10 @@ collect_more_digits(Data, Call, AlreadyCollected, MaxDigits) ->
             lager:debug("failed to collect DTMF: ~p", [_E])
     end.
 
+-spec truncate_after_terminator(binary(), ne_binaries()) -> binary().
+truncate_after_terminator(AlreadyCollected, Terminators) ->
+    hd(binary:split(AlreadyCollected, Terminators, ['global'])).
```
jamesaimonetti: I don't think you need global here, as I think you just want the first sequence of DTMF that doesn't contain a terminator. For instance:

```erlang
(pdx@pdx.2600hz.com)2> binary:split(<<"1234#456#789">>, [<<"#">>, <<"*">>]).                                                                                                                                       
[<<"1234">>,<<"456#789">>]                                                                                                                                                                                         
(pdx@pdx.2600hz.com)3> binary:split(<<"1234#456#789">>, [<<"#">>, <<"*">>], [global]).                                                                                                                             
[<<"1234">>,<<"456">>,<<"789">>] 
```
`global` requires traversing the whole binary looking for terminators; omitting it ends execution once a terminator is found.

danielfinke:
@jamesaimonetti I'm not sure if a unit test fits here, it's more of an integration test thing. It's also very intermittent so it may be hard to test for. It's particularly bad for us as heavy users of pivot with a client. They do this:
- cf_play
- cf_pivot
Pivot returns
- cf_collect_dtmf
Because there is a bit of a delay between module execution (this can happen anywhere, but pivot is particularly sensitive due to the network request), cf_exe can accumulate a digit or two prior to cf_collect_dtmf.
An example of the exact circumstance:
- cf_play ("please enter dob in format MMYY")
- cf_pivot starts in background
- user presses 1234
- cf_collect_dtmf (max digits = 4)
- cf_play ("invalid dob" starts playing)
- cf_pivot starts in the background
- for some reason this client decides to "end" their input with '#'
- invalid dob gets cut off
- cf_collect_dtmf (max digits = 4)
The bug manifests in that last step - that pound doesn't work as a terminator, it is set as the first DTMF of the collection

jamesaimonetti:
@danielfinke ah, got it. Yeah, then just remove the global and it should be good to go

danielfinke:
@jamesaimonetti alright, change me

fenollp:
I do think that EUnit tests for `truncate_after_terminator` would be a good idea just to show what kind of inputs and outputs it has.
Tests can be embedded in the module with `-ifdef(TEST). ... -include_lib("eunit/... my_test_() -> ... -endif.`

jamesaimonetti:
@danielfinke thanks, merged.

danielfinke:
@fenollp added unit test here: https://github.com/2600hz/kazoo/pull/4010

PR #4000, opened 2017-07-27
Low hanging type fixes

jamesaimonetti:
Not sure why I missed them but fixes calls to `crossbar_doc:delete/2` in other modules.

### DIFF
--- a/applications/crossbar/src/modules/cb_security.erl
+++ b/applications/crossbar/src/modules/cb_security.erl
@@ -70,19 +70,23 @@ init() ->
 %% allowed to access the resource, or false if not.
 %% @end
 %%--------------------------------------------------------------------
--spec authorize(cb_context:context()) -> boolean().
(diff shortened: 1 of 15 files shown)


applications/crossbar/src/modules/cb_security.erl
```diff
@@ -70,19 +70,23 @@ init() ->
...
 %%--------------------------------------------------------------------
--spec authorize(cb_context:context()) -> boolean().
+-spec authorize(cb_context:context()) ->
+                       boolean() |
+                       {'halt', cb_context:context()}.
 authorize(Context) ->
     authorize_list_available_module(Context, cb_context:req_nouns(Context), cb_context:req_verb(Context)).
 
--spec authorize(cb_context:context(), path_token()) -> boolean().
+-spec authorize(cb_context:context(), path_token()) -> 'true'.
```
fenollp: This is not how specs work

PR #4002, opened 2017-07-27
Update build-autocomplete.escript

powerpbx:
Add kz_datamgr module to build sup command list escript.

I find this script useful for building a list of sup commands.  Are there any other useful modules that should be added?

### DIFF
--- a/core/sup/priv/build-autocomplete.escript
+++ b/core/sup/priv/build-autocomplete.escript
@@ -16,6 +16,7 @@
                           ,"kapps_account_config"
                           ,"kapps_config"
                           ,"kapps_controller"
+                          "kz_datamgr"
                           ]).
 
 %% API
--- a/core/sup/priv/build-autocomplete.escript
+++ b/core/sup/priv/build-autocomplete.escript
@@ -16,7 +16,7 @@
                           ,"kapps_account_config"
                           ,"kapps_config"
                           ,"kapps_controller"
-                          "kz_datamgr"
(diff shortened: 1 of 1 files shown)


k-anderson:
Howdy,

Why kz_datamgr?  Most of those functions will not work properly via SUP, you should be using kazoo_data_maintenance...

powerpbx:
Why not?  Unless all functions can be done elsewhere.

https://forums.2600hz.com/forums/topic/9037-adding-custom-sip-headers/#comment-48945

k-anderson:
@powerpbx that is the intention, for kz_datamgr there is kazoo_data_maintenance.  On that ticket I did find that the kazoo_endpoints library had a local cache but no maintenance module so it is my intent to add one shortly.  My concern with adding this is commands will show up in auto-complete that will always fail, and that will confuse people.  That and it doesn't block anyone who knows what they are doing from running them.  If you have some that are useful we could add them as wrappers to the maintenance module, for example (simple) in the maintenance module we would have:

open_document(Database, DocumentId) ->
    {'ok', JObj} = kz_datamgr:open_doc(Database, DocumentId),
    io:format("~s~n", [kz_json:encode(JObj)]),
    'no_return'.

This is explicitly allowing json to be consumed from the maintenance module, which is likely more useful to an external app than Erlang terms anyway ;)

fenollp:
I opened https://github.com/2600hz/kazoo/pull/4035 
Do you think it covers your needs?

PR #4005, opened 2017-07-27
KAZOO-5624: return an empty array for user channels if the user has n…

k-anderson:
…o devices assigned


fenollp:
4.0: https://github.com/2600hz/kazoo/pull/4006

PR #4010, opened 2017-07-28
PISTON-460: add unit test for truncate_after_terminator/2

danielfinke:
Add unit test as requested by @fenollp in https://github.com/2600hz/kazoo/pull/3998.


fenollp:
Thanks!

PR #4011, opened 2017-07-28
PISTON-461: Callflow module for flushing early DTMF collection

danielfinke:
When a callflow looks like:
- cf_play
- cf_play
- cf_collect_dtmf

If a caller presses a digit during the first cf_play, it cancels playback and gets added to the kapps_call dtmf_collection. Then the next media file starts playing.
There isn't a way to allow a caller to just skip the first cf_play without accumulating that DTMF. The solution is this:

- cf_play
- cf_flush_dtmf
- cf_play
- cf_collect_dtmf

### DIFF
--- a/applications/callflow/src/module/cf_flush_dtmf.erl
+++ b/applications/callflow/src/module/cf_flush_dtmf.erl
@@ -1,0 +1,35 @@
+%%%-------------------------------------------------------------------
+%%% @copyright (C) 2017 Voxter Communications
+%%% @doc
+%%% Flush DTMF collection on a call
+%%% "data":{
+%%%   "collection_name":"your_name_here" // the collection to flush
+%%% }
+%%% @end
+%%% @contributors
+%%%-------------------------------------------------------------------
+-module(cf_flush_dtmf).
+
+-behaviour(gen_cf_action).
+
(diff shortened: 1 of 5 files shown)


applications/callflow/src/module/cf_flush_dtmf.erl
```diff
@@ -0,0 +1,35 @@
...
+%% @end
+%%--------------------------------------------------------------------
+-spec handle(kz_json:object(), kapps_call:call()) -> 'ok'.
+handle(Data, Call) ->
+    Collection = collection_name(Data),
+    cf_exe:continue(kapps_call:set_dtmf_collection('undefined', Collection, Call)).
+
+-spec collection_name(kz_json:object()) -> ne_binary().
+collection_name(Data) ->
+    case kz_json:get_value(<<"collection_name">>, Data) of
```
fenollp: How about `kz_json:get_ne_binary_value(<<"collection_name">>, Data, <<"default">>)`?

fenollp:
Looks like a `make apis` should fix circleci.

icehess:
@danielfinke why not add a boolean parameter to cf_play to flush any dtmf after it played?

PR #4015, opened 2017-08-01
Fix fax attachments and some others changes

icehess:
* Fix an issue when sending a fax error template with no fax doc or attachment resulted in crash
* Increase timeout for notify resender to 10s to allow templates like voicemail to fetch attachments
* ~~Remove obsolete `register` template~~
* Return and log failure when data is invalid instead of crash so the publisher won't wait to timeout
* See if any templates binding was successful, if not return failure so the publisher won't wait to timeout

### DIFF
--- a/applications/teletype/src/teletype_bindings.erl
+++ b/applications/teletype/src/teletype_bindings.erl
@@ -36,6 +36,7 @@ flush_mod(Module) ->
 
 -spec notification(kz_json:object()) -> 'ok'.
 notification(JObj) ->
+    kz_util:put_callid(JObj),
     {EventCategory, EventName} = kz_util:get_event_type(JObj),
     RoutingKey = ?ROUTING_KEY(EventCategory, EventName),
     lager:debug("dispatching notification ~s", [RoutingKey]),
--- a/applications/teletype/src/teletype_bindings.erl
+++ b/applications/teletype/src/teletype_bindings.erl
@@ -43,15 +43,12 @@ notification(JObj) ->
     Res = kazoo_bindings:map(RoutingKey, JObj, []),
     case kazoo_bindings:succeeded(Res, fun filter_out_failed/1) of
         [] ->
-            lager:debug("notification ~s did not result in successes", [RoutingKey]),
-            teletype_util:send_update(JObj, <<"failed">>, <<"not_succeeded">>);
+            FailureMsg = build_failure_message(Res),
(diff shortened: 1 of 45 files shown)


applications/teletype/src/teletype_bindings.erl
```diff
@@ -36,11 +36,22 @@ flush_mod(Module) ->
...
     {EventCategory, EventName} = kz_util:get_event_type(JObj),
     RoutingKey = ?ROUTING_KEY(EventCategory, EventName),
     lager:debug("dispatching notification ~s", [RoutingKey]),
-    _ = kazoo_bindings:map(RoutingKey, JObj, []),
-    'ok'.
+    Res = kazoo_bindings:map(RoutingKey, JObj, []),
+    case kazoo_bindings:succeeded(Res, fun filter_out_failed/1) of
+        [] ->
+            lager:debug("notification ~s did not result in successes", [RoutingKey]),
+            teletype_util:send_update(JObj, <<"failed">>, <<"not_succeeded">>);
```
fenollp: Is there no way to get more info than failed/not_succeeded? Or is that intended?

icehess: hehe just saw this, I updated it, take a look.

applications/teletype/src/teletype_fax_util.erl
```diff
@@ -22,129 +20,194 @@
...
-            ,[<<"fax">>, <<"tx_result">>, <<"timezone">>]
-            ,[<<"fax">>, <<"rx_result">>, <<"timezone">>]
-            ],
-    find_timezone(kz_json:get_first_defined(Paths, DataJObj), FaxBoxJObj).
+-spec add_attachments(kz_json:object(), kz_proplist(), boolean()) -> {kz_proplist(), attachments()}.
+add_attachments(DataJObj, Macros, ShouldTerminate) ->
+    IsPreview = teletype_util:is_preview(DataJObj),
+    FaxDoc = props:get_value(<<"fax">>, Macros),
+    case kz_json:is_json_object(FaxDoc)
+        andalso kz_json:is_empty(FaxDoc)
```
lazedo: @icehess it seems that if there is a nonempty FaxDoc it will not try to get the attachment ?

icehess: yes

PR #4018, opened 2017-08-03
Added feature to configure rtcp_mux for calls to SIP endpoints

sergey-safarov:


core/kazoo_endpoint/src/kz_endpoint.erl
```diff
@@ -1552,6 +1554,11 @@ maybe_set_call_forward({Endpoint, Call, CallFwd, CCVs}) ->
                        )
     }.
 
+-spec maybe_set_rtcp_mux(ccv_acc()) -> ccv_acc().
+maybe_set_rtcp_mux({Endpoint, Call, CallFwd, CCVs}) ->
+    RTCP_MUX = kz_json:is_true([<<"media">>, <<"rtcp_mux">>], Endpoint, ?DEFAULT_RTCP_MUX),
+    {Endpoint, Call, CallFwd, kz_json:set_value(<<"RTCP-MUX">>, RTCP_MUX, CCVs)}.
```
lazedo: @sergey-safarov as commented before, there's no need in sending `rtcp-mux` if its not defined as that will unnecessary increase the payload size that we send over amqp, specially in the case of users with lots of devices or large ring-groups.
lets remove the system default and set a case here as commented earlier.

sergey-safarov:
PR is tested. rtcp_mux may be set to true/false value at device document and at system_config level.
Value true/false is passed to FreeSwitch correctly

if rtcp_mux not configured, then 'undefined' falue is filtered

core/kazoo_endpoint/src/kz_endpoint.erl
sergey-safarov: @lazedo PR updated
CCVs is not updated when rtcp_mux is not defined.
http://prntscr.com/g3ypbq

sergey-safarov: @lazedo i reviewed PR according last recommendation in https://github.com/2600hz/kazoo/pull/3872
But leaved default value because case got 'undefined' if default value is not present in system_config

lazedo:
@sergey-safarov not acceptable as is, if a value is set at system level then we will be sending rtcp-mux for every endpoint. please follow  previous comment.

sergey-safarov:
@lazedo PR is updated.
As admin i want set settings at global level. Global off/on
May do this at freeswitch config?

core/kazoo_endpoint/src/kz_endpoint.erl
```diff
@@ -1552,6 +1553,13 @@ maybe_set_call_forward({Endpoint, Call, CallFwd, CCVs}) ->
                        )
     }.
 
+-spec maybe_rtcp_mux(ccv_acc()) -> ccv_acc().
+maybe_rtcp_mux({Endpoint, Call, CallFwd, CCVs} = Acc) ->
+    case kz_json:is_true([<<"media">>, <<"rtcp_mux">>], Endpoint) of
```
lazedo: `kz_json:is_true/2` will never return `undefined`
you need to use `kz_json:get_value/2` and then use `kz_term:to_boolean(RTCP_MUX)`
you can also add a `kz_json:get_boolean_value/2/3` like  `kz_json:get_atom_value/2/3` and others and use here without the need to use `kz_term`. `kz_json:is_true` and `kz_json:is_false` always return `true/false`

sergey-safarov:
Luis (@lazedo), added kz_json:get_boolean_value/2/3

core/kazoo_endpoint/src/kz_endpoint.erl
```diff
@@ -1552,6 +1553,13 @@ maybe_set_call_forward({Endpoint, Call, CallFwd, CCVs}) ->
                        )
     }.
 
+-spec maybe_rtcp_mux(ccv_acc()) -> ccv_acc().
+maybe_rtcp_mux({Endpoint, Call, CallFwd, CCVs} = Acc) ->
+    case kz_json:get_boolean_value([<<"media">>, <<"rtcp_mux">>], Endpoint, 'undefined') of
```
lazedo: @sergey-safarov last one, can you change this call from `kz_json:get_boolean_value/3` to `kz_json:get_boolean_value/2` ?

sergey-safarov: Thanks you Luis @lazedo for the detailed explanation of why so and not otherwise
Also I added comments to FreeSwitch config to allow globaly disable/enable rtcp_mux without overloading rabbitmq

https://github.com/2600hz/kazoo-configs-freeswitch/pull/3

lazedo:
@sergey-safarov  thanks for all your work here.

PR #4019, opened 2017-08-03
re-work cb_security to use hierarchy config

icehess:


applications/crossbar/src/modules/cb_security.erl
```diff
@@ -341,91 +277,84 @@ update(Id, Context) ->
...
-handle_singularity_update(Context, ?HTTP_POST, Id, 'undefined', _) -> not_found(Context, Id).
-
--spec check_mod_config_exists(cb_context:context(), ne_binary()) -> cb_context:context().
-check_mod_config_exists(Context, Id) ->
-    case kz_json:get_value(module_config_path(Id), cb_context:fetch(Context, 'db_doc')) of
-        'undefined' -> not_found(Context, Id);
-        _ModConfig -> Context
+-spec has_configuration_id(kz_json:object()) -> boolean().
+has_configuration_id(JObj) ->
+    case kz_json:get_ne_binary_value([<<"multi_factor">>, <<"configuration_id">>], JObj) of
```
fenollp: Also known as `get_ne.... =/= undefined`

core/kazoo_ast/src/kapps_config_usage.erl
```diff
@@ -138,6 +142,24 @@ config_to_schema(Source, F='get_global', [_Account, Cat, K, Default], Schemas) -
...
+    end;
+config_to_schema(Source, F='get_with_strategy', [Strategy, Account, Cat, K], Schemas) ->
+    config_to_schema(Source, F, [Strategy, Account, Cat, K, 'undefined'], Schemas);
+config_to_schema(Source, F='get_with_strategy', [?BINARY_MATCH([?BINARY_STRING("hierarchy_merge")]), _Account, Cat, K, Default], Schemas) ->
+    Document = category_to_document(Cat),
+    case key_to_key_path(K) of
+        'undefined' -> Schemas;
+        Key -> config_key_to_schema(Source, F, Document, Key, Default, Schemas)
+    end;
+config_to_schema(_, 'get_with_strategy', _Args, Schemas) ->
```
fenollp: Shouldn't there be a catchall clause on the `/5` version of `get_with_strategy` that uses the `Default` value too?

icehess: maybe for `global` but for `reseller` I'm not sure that we want to set its Default value in system_config.

icehess: added guards to func header for global

PR #4021, opened 2017-08-04
created maintenance function to drop the rev of service documents

k-anderson:


icehess:
@k-anderson probably you need to add the `KZ_SERVICES_DB_TMP` to `kzs_util:db_classification/1` to avoid logging unnecessary stacktrace.

PR #4022, opened 2017-08-04
created maintenance function to drop the rev of service documents

k-anderson:


fenollp:
master: https://github.com/2600hz/kazoo/pull/4021

PR #4031, opened 2017-08-04
fixing fixing fixing and fixing fax again and forever

icehess:


applications/teletype/src/teletype_util.erl
```diff
@@ -721,7 +721,11 @@ find_address(DataJObj, TemplateMetaJObj, _ConfigCat, Key, 'undefined') ->
     {Key, check_address_value(kz_json:find(Key, [DataJObj, TemplateMetaJObj]))};
 find_address(DataJObj, TemplateMetaJObj, _ConfigCat, Key, ?EMAIL_SPECIFIED) ->
     lager:debug("checking template for '~s' email addresses", [Key]),
-    Emails = kz_json:find_first_defined([[Key, <<"email_addresses">>], Key], [DataJObj, TemplateMetaJObj]),
+    Email0 = kz_json:get_first_defined([[Key, <<"email_addresses">>], Key], TemplateMetaJObj),
+    case kz_term :is_empty(Email0) of
```
jamesaimonetti: Personally, I dislike how vars can escape the case clause like this. Why not

```erlang
Emails = case kz_term:is_empty(Email0) of
    'false' -> Email0;
    'true' -> kz_json:get_first_defined(...)
end,
{Key, check_address_value(Emails)};
```

icehess: fixed

PR #4036, opened 2017-08-04
Dialyzer adjustments

jamesaimonetti:
cherry picked #4032 commits:

* get attachment properly for voicemail

* fix types

* fix getting account id if json object is used

* fix type/spec of proplist

* fix setting custom sip headers

* unreachable clause (or request_return() is wrong)

* unreachable

* return is JObj, call expects JObjs

* fix spec

* fix return types

* io:format removed

### DIFF
--- a/core/kazoo_apps/src/kapps_notify_publisher.erl
+++ b/core/kazoo_apps/src/kapps_notify_publisher.erl
@@ -70,10 +70,9 @@ cast(Req, PublishFun) ->
 %% @private
 %% @doc handle amqp worker responses
 -spec handle_resp(api_ne_binary(), api_terms(), kz_amqp_worker:request_return()) -> 'ok'.
-handle_resp(_NotifyType, _Req, 'ok') -> 'ok';
 handle_resp(NotifyType, Req, {'ok', _}=Resp) -> check_for_failure(NotifyType, Req, Resp);
 handle_resp(NotifyType, Req, {'error', Error}) -> maybe_handle_error(NotifyType, Req, error_to_failure_reason(Error));
-handle_resp(NotifyType, Req, {'returned', _, Resp}) -> check_for_failure(NotifyType, Req, {'returned', Resp});
+handle_resp(NotifyType, Req, {'returned', _, Resp}) -> check_for_failure(NotifyType, Req, {'returned', [Resp]});
 handle_resp(NotifyType, Req, {'timeout', _}=Resp) -> check_for_failure(NotifyType, Req, Resp).
 
 %% @private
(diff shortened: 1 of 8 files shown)


core/kazoo_apps/src/kapps_notify_publisher.erl
```diff
@@ -167,20 +167,25 @@ maybe_ignore_failure(_) -> 'false'.
...
+    Get([<<"account_id">>
+        ,[<<"account">>, <<"_id">>]
+        ,<<"pvt_account_id">>
+        ,<<"_id">>, <<"id">>
+        ,<<"Account-ID">>
+        ,[<<"details">>, <<"account_id">>]
+        ,[<<"Details">>, <<"Account-ID">>]
+        ,[<<"details">>, <<"custom_channel_vars">>, <<"account_id">>]
+        ,[<<"Details">>, <<"Custom-Channel-Vars">>, <<"Account-ID">>]
+        ]
```
fenollp: Why not put the list of paths in a `/0` function?
This way they can be const-referenced and `erlang:apply` does not need to be called.

jamesaimonetti: I don't follow. This change just makes sure that if Req is a json object, the appropriate getter is used.

fenollp: Yes but with your changes `find_account_id/2` has to do more than it needs to at runtime.
I am suggesting hand-unrolling the 2 possible cases explicitly so that the code generated by the compiler does not need to call `erlang:apply/3` at all.
I'm saying let's put this list of paths in a `/0` function and write

```erlang
find_account_id(Req) when is_list(Req) ->
    props:get_first_defined(account_id_paths(), Req);
find_account_id(Req) ->
    kz_json:get_first_defined(account_id_paths(), Req).
```

jamesaimonetti: This is just a backport so I'd rather a ticket be filed to make the change.

jamesaimonetti: Out of curiosity, I tried a horse comparison:
```
kapps_notify_publisher:find_account_id in 0.585478s
kapps_notify_publisher:fai in 0.512621s
```
where find_account_id is the existing and fai implements your suggestion. Repeat factor was 1000000.  5.85478e-7 and 5.12621e-7 per invocation...so I would say this is a meaningless change anyway.

fenollp: I'm glad you took the time to implement it, write benchmarks for it then post the results here though

PR #4043, opened 2017-08-08
add kazoo_globals to app files

jamesaimonetti:


fenollp:
What is this solving?
Do you know that `kazoo_apps` already depend on `kazoo_globals`?
https://github.com/2600hz/kazoo/blob/32441633ea64c467c7ba70e73464642634495351/core/kazoo_apps/src/kazoo_apps.app.src#L23

jamesaimonetti:
@fenollp there's a report on the forum of fax not starting. fax_monitor is failing to start, no useful logging that I could see in the forum post's logs. Seeing fax_monitor is named using kz_globals, thought it would be useful to explicitly name kazoo_globals as a dependent app, since fax uses it directly.

fenollp:
So you have no idea if this solves anything then?
As long as the fax kapp is started using either kapps_controller or application:ensure_all_start

jamesaimonetti:
I think it's important to have the .app files be more explicit about what dependent apps are used. These apps use `kazoo_globals` directly and that should be reflected in the .app file.

PR #4048, opened 2017-08-09
Support Erlang/OTP 20.0 (well, 19 for now)

fenollp:


core/kazoo_bindings/src/kazoo_bindings.erl
```diff
@@ -665,15 +665,6 @@ fold_bind_results([#kz_responder{module=M
         {'error', _E}=E ->
             lager:debug("error: ~p", [_E]),
             E;
-        {'EXIT', {'undef', [{_M, _F, _A, _}|_]}} ->
-            ST = erlang:get_stacktrace(),
-            log_undefined(M, F, length(Payload), ST),
```
icehess: move these log stack trace thing to a function and call it here instead

core/kazoo_bindings/src/kazoo_bindings.erl
```diff
@@ -665,15 +665,6 @@ fold_bind_results([#kz_responder{module=M
...
-            fold_bind_results(Responders, Payload, Route, RespondersLen, ReRunResponders);
-        {'EXIT', _E} ->
-            ST = erlang:get_stacktrace(),
-            lager:error("~s:~s/~p died unexpectedly: ~p", [M, F, length(Payload), _E]),
-            kz_util:log_stacktrace(ST),
```
icehess: same as above

jamesaimonetti:
I would want @danielfinke to weigh in on ACDc changes. Daniel, can you run the code in this PR locally and confirm it continues to operate as you all expect?

danielfinke:
@jamesaimonetti Thanks, I'm going to give this a go today

danielfinke:
I have a PR that fixes numerous things for gen_statem. I think you may want to reference that for this PR too? (I'm thinking for the presence of the callback_mode/1 required function for the gen_statem behaviour)

core/kazoo_bindings/src/kazoo_bindings.erl
fenollp: From [this](https://gist.github.com/fenollp/a362f1dfa94f2033700ba20d092c8588) it looks like the stacktrace is `[]` (between `try` and `catch`). So `log_undefined` would never log anything.

Still want me to do this? Would be dead code to me.

core/kazoo_bindings/src/kazoo_bindings.erl
fenollp: ditto

fenollp:
## Note

Merging this PR as of now would drop support for OTP18 and not add support for OTP20 yet.
We'd only be able to build/use Kazoo on OTP19.

PR #4052, opened 2017-08-09
[4.1] get merge result of account, it's parents and system config

icehess:
back port of #4008


core/kazoo_documents/src/kz_account.erl
```diff
@@ -466,6 +467,33 @@ parent_account_id(JObj) ->
 %% @doc
 %% @end
 %%--------------------------------------------------------------------
+-spec get_parent_account(ne_binary()) -> {'ok', doc()} | {'error', any()}.
+get_parent_account(AccountId) ->
+    case get_parent_account_id(AccountId) of
+        'undefined' -> {'ok', 'not_found'};
```
fenollp: this doesn't look right...

icehess: what should I return when can't find parent id then?

PR #4060, opened 2017-08-13
Update registrar and kazoo_auth Makefiles for solaris platform

noahmehl:
Needed to add the CC and LDFLAGS for the solaris platform to be able to build.

PR #4063, opened 2017-08-14
Update registrar and kazoo_auth Makefiles for solaris platform

noahmehl:


fenollp:
4.1: https://github.com/2600hz/kazoo/pull/4060

Thanks @noahmehl !

PR #4068, opened 2017-08-14
[master] HELP-31652: check allowed and authorized apps in one fold

icehess:


lazedo:
@icehess it seems some tests are failing for `kazoo_services` after this merge. can you check ?

lazedo:
@icehess nm. it seems ok now.

PR #4070, opened 2017-08-15
KAZOO-5639_4one: move MDN check on write instead of read, to allow reads

fenollp:
master: https://github.com/2600hz/kazoo/pull/4069


fenollp:
4.0 is coming up next

PR #4076, opened 2017-08-16
[master] KAZOO-5641: return provider name and various fixes and clean up

icehess:
Some changes as request by UI team for `security` and `multi_factor` APIs.

* move and update list providers to list by type and returns more info (name, type and etc.)
* add `name` field to multi factor configuration, rename `provider_type` to `pvt_provider_type`
* add API to list/create/change system multi factor provider
* return `name` and `provider_name` from security api if a module has multi factor configured
* remove default values from `account_module_config` schema so if an account only want to configure a parameter and inherit rest of the configuration.
* rename `auth_provider` schema to `multi_factor_provider` to avoid confusion.
* for inherited auth config, instead of using the same account id, use it's parent so the UI knows what is inherited what is not (Only if the account is NOT a reseller though, if reseller read system_config)
* use `merge_recursive` in `kapps_account_config`
* when setting multi factor for an auth module, check if the account is decedents of the authorized account or only if the `include_subaccounts` is true check if the authenticated account has the account id in its tree.
* Duo config in `system_auth` is disabled now
* MFA default provider is Duo now
* change account's mfa configuration document type from `auth_provider` to `provider`
* fixing various typos


core/kazoo_auth/priv/couchdb/providers/google.json
```diff
@@ -25,6 +25,7 @@
     "public_key_jwt_location": "header",
     "public_key_lookup_field": "kid",
     "public_key_method": "lookup",
+    "pvt_provider_type": "sso",
```
lazedo: should be `oauth`

core/kazoo_auth/priv/couchdb/providers/office365.json
```diff
@@ -25,6 +25,7 @@
     "public_key_jwt_location": "header",
     "public_key_lookup_field": "kid",
     "public_key_method": "lookup",
+    "pvt_provider_type": "sso",
```
lazedo: `oauth`

core/kazoo_auth/priv/couchdb/providers/salesforce.json
```diff
@@ -25,6 +25,7 @@
     "public_key_jwt_location": "header",
     "public_key_lookup_field": "kid",
     "public_key_method": "lookup",
+    "pvt_provider_type": "sso",
```
lazedo: `oauth`

applications/crossbar/src/crossbar_auth.erl
```diff
@@ -101,7 +101,7 @@ create_auth_token(Context, Method, JObj) ->
 
             lager:debug("created new local auth token: ~s", [kz_json:encode(Resp)]),
 
-            log_success_auth(Method, <<"jwt_auth_token">>, <<"authentiaction resulted in token creation">>, Context, AccountId, AuthConfig),
+            log_success_auth(Method, <<"jwt_auth_token">>, <<"authentication resulted in token creation">>, Context, AccountId, AuthConfig),
```
fenollp: I found some more matches of this same typo in security.md

applications/crossbar/src/crossbar_auth.erl
```diff
@@ -224,7 +224,7 @@ is_auth_module_enabled(Method, Config) ->
 %%--------------------------------------------------------------------
 %% @private
 %% @doc
-%% Checks if authenticator module is configured to do mutli factor auth
+%% Checks if authenticator module is configured to do multi factor auth
```
fenollp: Same here: more matches hiding.

applications/crossbar/src/modules/cb_multi_factor.erl
```diff
@@ -284,14 +283,15 @@ merge_summary(_Context, Available, _) ->
...
+    cb_context:set_doc(Context, kz_doc:set_type(Doc, <<"provider">>));
 on_successful_validation(Id, Context) ->
-    crossbar_doc:load_merge(Id, Context, ?TYPE_CHECK_OPTION(<<"auth_provider">>)).
+    crossbar_doc:load_merge(Id, Context, ?TYPE_CHECK_OPTION(<<"provider">>)).
 
 %%--------------------------------------------------------------------
 %% @private
 %% @doc
-%% Normalizes the resuts of a view
+%% Normalizes the results of a view
```
fenollp: 37 more matches :D 
we can probably forget about these: they are useless comments anyway.

applications/crossbar/src/modules/cb_security.erl
```diff
@@ -326,14 +326,29 @@ has_account_id_or_db(JObj) ->
...
 check_account_hierarchy(AuthModule, JObj, Context) ->
     AuthAccountId = cb_context:auth_account_id(Context),
     IsSysAdmin = cb_context:is_superduper_admin(AuthAccountId),
     AccountId = kz_json:get_ne_binary_value([<<"multi_factor">>, <<"account_id">>], JObj),
+    NotIncludeSubAccounts = kz_json:is_false([<<"multi_factor">>, <<"include_subaccounts">>], JObj),
 
     case IsSysAdmin
-        orelse kz_util:is_in_account_hierarchy(AccountId, AuthAccountId, 'true')
+        orelse kz_util:is_in_account_hierarchy(AuthAccountId, AccountId, 'true')
+        orelse (NotIncludeSubAccounts
```
fenollp: Braces are useless here.
And `NotIncludeSubAccounts` should probably be replaced by its contents (ie. not introducing the var) but that's just my taste.
And the second call to `is_in_account_hierarchy/3` can be `is_in_account_hierarchy/2` as the first one already catches the `AccountId = AuthAccountId` case, for clarity.

icehess:
It's ready!

jamesaimonetti:
The clause for [`kz_mfa_auth:authenticate/1`](https://github.com/2600hz/kazoo/blob/64ad6e8447bc9b75688255014c04d555a959552f/applications/crossbar/src/crossbar_auth.erl#L141) is `{'ok', 'authenticated'}` but the [header](https://github.com/2600hz/kazoo/blob/master/core/kazoo_auth/src/kazoo_auth.hrl#L63) says it should be the bare atom `'authenticated'`. Can you update the header to be the OK tuple?

applications/crossbar/src/modules/cb_auth.erl
```diff
@@ -274,7 +274,10 @@ validate_path(Context, ?LINKS_PATH, ?HTTP_GET) ->
 
 %% validating /auth/providers
 validate_path(Context, ?PROVIDERS_PATH, ?HTTP_GET) ->
-    crossbar_doc:load_view(?PROVIDERS_VIEW, [], Context, fun normalize_view/2);
+    ViewOptions = [{'startkey', [<<"oauth">>]}
+                  ,{'endkey', [<<"oauth">>, kz_json:new()]}
```
lazedo: @icehess why are we limiting the scope to `oauth` here ?

icehess:
me merge please :-(

PR #4077, opened 2017-08-16
[4.1] KAZOO-5641: return provider name and various fixes and clean up

icehess:
backport of #4076 

Some changes as request by UI team for `security` and `multi_factor` APIs.

* move and update list providers to list by type and returns more info (name, type and etc.)
* add `name` field to multi factor configuration, rename `provider_type` to `pvt_provider_type`
* add API to list/create/change system multi factor provider
* return `name` and `provider_name` from security api if a module has multi factor configured
* remove default values from `account_module_config` schema so if an account only want to configure a parameter and inherit rest of the configuration.
* rename `auth_provider` schema to `multi_factor_provider` to avoid confusion.
* for inherited auth config, instead of using the same account id, use it's parent so the UI knows what is inherited what is not (Only if the account is NOT a reseller though, if reseller read system_config)
* use `merge_recursive` in `kapps_account_config`
* when setting multi factor for an auth module, check if the account is decedents of the authorized account or only if the `include_subaccounts` is true check if the authenticated account has the account id in its tree.
* Duo config in `system_auth` is disabled now
* MFA default provider is Duo now
* change account's mfa configuration document type from `auth_provider` to `provider`
* fixing various typos


fenollp:
master: https://github.com/2600hz/kazoo/pull/4076

icehess:
It's ready!

PR #4082, opened 2017-08-18
add erlang 20 for test

getong:


fenollp:
Hi there. 
If you want to work on getting Kazoo to work on OTP20, look here https://github.com/2600hz/kazoo/pull/4048 and use what you want.

This PR is bringing nothing and I am not sure why you opened it.

PR #4087, opened 2017-08-22
KAZOO-5618: Service plans planner API

icehess:


applications/crossbar/src/modules/cb_service_planner.erl
```diff
@@ -0,0 +1,227 @@
...
+%% @private
+%% @doc
+%%
+%% @end
+%%--------------------------------------------------------------------
+-spec on_successful_validation(api_binary(), cb_context:context()) -> cb_context:context().
+on_successful_validation('undefined', Context) ->
+    cb_context:set_doc(Context, kz_doc:set_type(cb_context:doc(Context), kzd_service_plan:type()));
+on_successful_validation(Id, Context) ->
+    crossbar_doc:load_merge(Id, Context, ?TYPE_CHECK_OPTION(kzd_service_plan:type())).
```
fenollp: This shouldn't be needed after the last fixes to patch_and_validate.

PR #4089, opened 2017-08-23
Kz json fix

jamesaimonetti:


core/kazoo_stdlib/test/kz_json_test.erl
```diff
@@ -758,6 +758,13 @@ merge_vs_merge_recursive_test_() ->
     ,?_assert(kz_json:are_equal(M3, ?REQUEST))
     ].
 
+set_value_test_() ->
+    JObj = kz_json:from_list([{<<189>>,[]}]),
+    Key = [<<189>>,<<0>>],
```
fenollp: Shouldn't the offsets into arrays be integers instead of binaries (`<<0>>`) ?
And does the key really need to be outside the ascii range or could it be `<<"a">>` instead of `<<189>>`?

PR #4098, opened 2017-08-24
[4.1] url encode the attachment name when put

icehess:


core/kazoo_data/src/kzs_attachments.erl
```diff
@@ -119,7 +120,8 @@ put_attachment(#{att_handler := {Handler, Params}}=Map
             handle_put_attachment(Map, Att, DbName, DocId, AName, Contents, Options, Props);
         {'error', _} = E -> E
     end;
-put_attachment(#{server := {App, Conn}}, DbName, DocId, AName, Contents, Options) ->
+put_attachment(#{server := {App, Conn}}, DbName, DocId, AttName, Contents, Options) ->
+    AName = kz_http_util:urlencode(AttName),
```
lazedo: @icehess not sure if this should be done here or in kazoo_couch. need to take a better look at code

lazedo:
master https://github.com/2600hz/kazoo/pull/4097

core/kazoo_data/src/kzs_attachments.erl
fenollp: I'd say in kazoo_couch: another backend may want to encode this its own way.

PR #4099, opened 2017-08-24
bad-bad-json: should fail when reading from not JSON data

fenollp:


core/kazoo_stdlib/src/kz_json.erl
```diff
@@ -876,9 +876,14 @@ get_value1([K|Ks], JObjs, Default) when is_list(JObjs) ->
...
-get_value1([K|Ks], ?JSON_WRAPPER(Props)=_JObj, Default) ->
+get_value1([K|Ks], ?JSON_WRAPPER(Props), Default) ->
     get_value1(Ks, props:get_value(K, Props, Default), Default);
-get_value1(_, _, Default) -> Default.
+get_value1(_, undefined, Default) ->
+    Default;
+get_value1(_, ?JSON_WRAPPER(_), Default) ->
+    Default;
+get_value1(_K, _V, _D) ->
+    erlang:error(badarg).
```
lazedo: @jamesaimonetti apparently, this breaks  `get_first_defined`

jamesaimonetti:
@kalda341 yeah, we're working through some of the issues in #4219; should have it patched up soon. I think more data cleansing is needed. I would even entertain a "task" that runs through all docs in all databases and validates them against the schema, reporting those that are "bad" so they can be fixed, either manually or programatically.

jamesaimonetti:
@kalda341 this has also shown some poor test coverage so I think, while a bit painful now, kazoo will ultimately be improved long term.

kalda341:
@jamesaimonetti A "task" would be fantastic. I suspect we have a lot of docs that don't validate against the  schemas on our system.

PR #4100, opened 2017-08-25
[master] flush registrar and fs_xml when enable/disable an account or a device

icehess:
4.1: #4101 
4.0: #4102


lazedo:
https://2600hz.atlassian.net/browse/KAZOO-5634

PR #4105, opened 2017-08-25
fix displaying Path in ecallmgr_maintenance registrar_summary

danielfinke:


applications/ecallmgr/src/ecallmgr_registrar.erl
```diff
@@ -1298,6 +1295,18 @@ breakup_contact(Contact) when is_binary(Contact) ->
 breakup_contact(Contact) ->
     breakup_contact(kz_term:to_binary(Contact)).
 
+-spec proxy_path(api_binary(), api_binary(), api_integer()) -> binary().
+proxy_path('undefined', 'undefined', 'undefined') -> <<>>;
+proxy_path('undefined', 'undefined', Port) -> proxy_path('undefined', <<>>, Port);
+proxy_path('undefined', IP, 'undefined') -> IP;
+proxy_path('undefined', IP, Port) -> <<IP/binary, ":", (kz_term:to_binary(Port))/binary>>;
+proxy_path(Proxy, _, Port) ->
+    Proxy1 = binary:replace(Proxy, [<<"sip:">>], <<>>),
```
fenollp: The list is superfluous.

applications/ecallmgr/src/ecallmgr_registrar.erl
```diff
@@ -1298,6 +1295,18 @@ breakup_contact(Contact) when is_binary(Contact) ->
 breakup_contact(Contact) ->
     breakup_contact(kz_term:to_binary(Contact)).
 
+-spec proxy_path(api_binary(), api_binary(), api_integer()) -> binary().
+proxy_path('undefined', 'undefined', 'undefined') -> <<>>;
+proxy_path('undefined', 'undefined', Port) -> proxy_path('undefined', <<>>, Port);
+proxy_path('undefined', IP, 'undefined') -> IP;
+proxy_path('undefined', IP, Port) -> <<IP/binary, ":", (kz_term:to_binary(Port))/binary>>;
```
fenollp: Here `IP` can be `<<>>`. Is this okay?

danielfinke: I'm not sure previous versions of Kazoo would have ever had an empty IP, but I'm not sure it shouldn't be permitted. We are parsing the output of this command by splitting on the colon, so it makes more sense for us to keep it, ensuring the parsed IP is always an IP or blank, and not the port.

PR #4107, opened 2017-08-28
std-rebar3: compile Kazoo with rebar3 (still works with make)

fenollp:


fenollp:
Merrrrge ittttt

fenollp:
Not sure why that perf maintenance test times out...

icehess:
@fenollp thanks for breaking it

icehess:
somehow `plist` is not in good term with rebar3+eunit I think. It execute the list but is not fuse them together and it timeout

fenollp:
You're always welcome :)

I don't see an issue with `plist` on my part. What makes you say that?
I can reproduce the perf_maintenance timeouts when running `rebar3 eunit` but not when running `rebar3 eunit -m kazoo_perf_maintenance`.
If I disable these tests I can reproduce the old KNM failures somehow due to knm_search. I tried for weeks to fix these to no avail though...

Tried with rebar3 3.4.3 and 3.4.4 on 19.3. Trying 3.4.4 on 20.1 now for the lulz

fenollp:
Same results with 3.4.4 / 20.1

jamesaimonetti:
If someone wants to pick up this work for 5.x so be it but for now, closing until that time.

PR #4108, opened 2017-08-29
PISTON-470: ACDc position/wait time announcements

danielfinke:
Add support for periodic position and estimated wait time announcements while waiting for call to be answered in ACDc queue.
Either can be enabled independently, time interval and media files are customizable.

Relevant media files at: https://github.com/2600hz/kazoo-sounds/pull/6

### DIFF
--- a/applications/acdc/src/acdc_announcements_sup.erl
+++ b/applications/acdc/src/acdc_announcements_sup.erl
@@ -49,12 +49,8 @@ start_link() ->
 maybe_start_announcements(Manager, Call, Props) ->
     Enabled = props:get_is_true(<<"position_announcements_enabled">>, Props, 'false')
         orelse props:get_is_true(<<"wait_time_announcements_enabled">>, Props, 'false'),
-    case Enabled of
-        'true' ->
-            supervisor:start_child(?SERVER, [Manager, Call, Props]);
-        'false' ->
-            'false'
-    end.
+    Enabled
+        andalso supervisor:start_child(?SERVER, [Manager, Call, Props]).
 
 %%--------------------------------------------------------------------
 %% @doc
--- a/applications/acdc/src/acdc_announcements_sup.erl
+++ b/applications/acdc/src/acdc_announcements_sup.erl
@@ -1,0 +1,94 @@
+%%%-------------------------------------------------------------------
+%%% @copyright (C) 2017, Voxter Communications
+%%% @doc
+%%%
+%%% @end
+%%% @contributors
+%%%   Daniel Finke
+%%%-------------------------------------------------------------------
+-module(acdc_announcements_sup).
+
+-behaviour(supervisor).
+
+%% API
+-export([start_link/0]).
+-export([maybe_start_announcements/3
+        ,stop_announcements/1
+        ]).
+
+%% Supervisor callbacks
+-export([init/1]).
+
+-include("acdc.hrl").
+
+-define(SERVER, ?MODULE).
+-define(CHILDREN, [?WORKER('acdc_announcements')]).
+
+%%%===================================================================
+%%% API functions
+%%%===================================================================
+
+%%--------------------------------------------------------------------
+%% @doc
+%% Starts the supervisor
+%%
+%% @end
+%%--------------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


kalda341:
This looks AMAZING!

noahmehl:
Wow, this is awesome!!

applications/acdc/src/acdc_queue_manager.erl
```diff
@@ -947,4 +958,47 @@ update_properties(QueueJObj, State) ->
...
+
+-spec cancel_position_announcements(kapps_call:call() | 'false', map()) ->
+                                           map().
+cancel_position_announcements('false', Pids) -> Pids;
+cancel_position_announcements(Call, Pids) ->
+    CallId = kapps_call:call_id(Call),
+    case maps:is_key(CallId, Pids) of
+        'true' ->
+            lager:debug("cancelling announcements for ~s", [CallId]),
+            {Pid, Pids1} = maps:take(CallId, Pids),
```
mark2600: If using maps:take here you don't really need to check maps:is_key beforehand (you can move the take into that case statement and handle either the return or error return).

Also as you may be aware, maps:take will NOT work on Erlang 18, and afaik Kazoo has not moved to 19 yet.

applications/acdc/src/acdc_queue_manager.erl
```diff
@@ -947,4 +958,47 @@ update_properties(QueueJObj, State) ->
...
+                      ],
+            kapps_call_command:send_command(Command, Call),
+            Pids1;
+        'false' ->
+            lager:debug("did not have the announcements for call ~s", [CallId]),
+            Pids
+    end.
+
+-spec maybe_remove_queue_member(api_binary(), mgr_state()) -> mgr_state().
+maybe_remove_queue_member(CallId, #state{current_member_calls=CurrentCalls
```
mark2600: A small nitpick, but "maybe_*" implies that it may or may not do the action, this will always remove the member.

applications/acdc/src/acdc_announcements_sup.erl
```diff
@@ -0,0 +1,94 @@
...
+%% Start an announcements child if either position or wait time
+%% announcements are enabled
+%%
+%% @end
+%%--------------------------------------------------------------------
+-spec maybe_start_announcements(pid(), kapps_call:call(), kz_proplist()) -> supervisor:startchild_ret() | 'false'.
+maybe_start_announcements(Manager, Call, Props) ->
+    Enabled = props:get_is_true(<<"position_announcements_enabled">>, Props, 'false')
+        orelse props:get_is_true(<<"wait_time_announcements_enabled">>, Props, 'false'),
+    case Enabled of
```
jamesaimonetti: You can cheat and do:
```erlang
Enabled
    andalso supervisor:start_child(...)
```
`andalso` will not evaluate the right-hand side if the left is 'false' (and returns false).

jamesaimonetti:
It would be wise to add these announcement configs to the [queue schema](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/priv/couchdb/schemas/queues.json) so people know how to set these values.

applications/acdc/src/acdc_queue_manager.erl
fenollp: We should move soon.

danielfinke:
OK guys, I have addressed your comments

applications/acdc/src/acdc_queue_manager.erl
kalda341: "Kazoo 4 targets Erlang 19+" - from https://github.com/2600hz/kazoo/blob/master/doc/installation.md

danielfinke: Either way, I've moved to using get/2 and remove/2 which are compatible with 18 through 20 :)

fenollp: @kalda341 That was merged soon after my comment ;)

lazedo:
@danielfinke thank you for this great contribution

danielfinke:
No problem!

PR #4111, opened 2017-08-31
event is metaflow.bind

N4COM:
defined in kapi_metaflow:
define(METAFLOW_BIND_ROUTING_KEY(AccountId, CallId), <<"metaflow.bind.", (amqp_util:encode(AccountId))/binary, ".", (amqp_util:encode(CallId))/binary>>).

### DIFF
--- a/applications/konami/src/konami_listener.erl
+++ b/applications/konami/src/konami_listener.erl
@@ -31,7 +31,7 @@
 -type state() :: #state{}.
 
 %% By convention, we put the options here in macros, but not required.
--define(BINDINGS, [{'metaflow', [{'restrict_to', ['bindings']}]}
+-define(BINDINGS, [{'metaflow', [{'restrict_to', ['bind']}]}
                   ,{'route', []}
                   ]).
 -define(RESPONDERS, [{{?MODULE, 'handle_metaflow'}


jamesaimonetti:
Hi @N4COM 
I don't think your proposed change fixes the perceived problem. The `restrict_to` list is used by the `kapi_metaflows` module to bind the AMQP queue to the broker [here](https://github.com/2600hz/kazoo/blob/master/core/kazoo_amqp/src/api/kapi_metaflow.erl#L187). So if this change is merged, `bind_q/3` will use the unknown restriction clause instead and not bind the queue to anything.

The [routing key](https://github.com/2600hz/kazoo/blob/master/core/kazoo_amqp/src/api/kapi_metaflow.erl#L104) used is defined here as you note. Payloads are published to AMQP using [this function](https://github.com/2600hz/kazoo/blob/master/core/kazoo_amqp/src/api/kapi_metaflow.erl#L287) which uses that routing key.

Perhaps you want to debug your issue in [kz_endpoint's metaflow code](https://github.com/2600hz/kazoo/blob/master/core/kazoo_endpoint/src/kz_endpoint.erl#L840)?

If you find anything in your exploration, please open a new PR with those findings and code changes! Thanks!

PR #4112, opened 2017-08-31
Allow more than one rate doc per prefix

onnet:
Since options based rate selection exists, we should be able to create several docs for particular prefix..

### DIFF
--- a/core/kazoo_documents/src/kzd_rate.erl
+++ b/core/kazoo_documents/src/kzd_rate.erl
@@ -57,6 +57,7 @@ ensure_id(Rate, 'undefined') ->
     ID = list_to_binary([iso_country_code(Rate, <<"XX">>)
                         ,<<"-">>
                         ,prefix(Rate)
+                        ,rate_suffix(Rate)
                         ]),
     kz_doc:set_id(Rate, ID);
 ensure_id(Rate, ID) ->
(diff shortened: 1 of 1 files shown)


jamesaimonetti:
Hi @onnet I'm not sure on whether a random suffix is appropriate or some other solution to allow for this. Thanks for bringing it up; I'll try to get some thought towards this soon.

jamesaimonetti:
@onnet my concern is with updates to ratedecks could change the ID used. So if I have a rate with ID `XX-1222` and I upload a new ratedeck that updates the parameters of the rate, now the ID is changed to `XX-1222-abc`, my scripts that previously expected `XX-1222` will have the old rate data.

What about another optional field that can be added to a rate doc? If present, append it to the ID. Maybe `rate_suffix`, or `rate_id_suffix`, or???

Thoughts?

onnet:
@jamesaimonetti Hi! The shorter, the better :) On the other side, it seems we can have just one rate per carrier. If its true, we can use carrier field as a suffix..

lazedo:
@onnet doesn't this defeat the purpose of having multiple ratedecks and moving the options based selection forward?

onnet:
@lazedo In my particular usecase an account can have seats in different cities with at least one local carrier per city (which means different rates for different seats). IMHO only one ratedeck could be applied to an account. If this statement is true, I need options based selection within one ratedeck. Even it is not, it looks to be cool option anyway. :)

onnet:
@lazedo And multiple ratedecks purpose isn't defeated anyway. Just because it is needed by resellers tree thing, different rates for different accounts within one reseller, etc ... :)

onnet:
@jamesaimonetti @lazedo  this works for me
Do you guys have some more concerns regarding this approach? 
Docs looks like this:
{
   "_id": "XX-1-spb",
   "_rev": "1-ca1e77a1ddd7f3b8785c26462780a3f5",
   "carrier": "spb_beeline",
   "description": "США",
   "direction": "outbound",
   "prefix": "1",
   "rate_cost": "3.97",
   "rate_name": "SŠA",
   "rate_suffix": "spb",
   "ratedeck_id": "plan1",
   "pvt_type": "rate",
   "routes": [
       "^\\+?1.+$"
   ],
   "weight": 1
}

lazedo:
@onnet actually, i would prefer not to mess with the `_id` right now. i would even think about letting it use the default generated `uuid` and work with views. that way you could compose the key and filtering in many other ways. but this is not a clear idea yet. if @jamesaimonetti is ok and it doesn't break anything (trie), i'm ok.

jamesaimonetti:
The trie stores the prefix and a list of doc ids. When matched, it will load all docs from the stored IDs. So ID is opaque to the trie implementation.

jamesaimonetti:
@onnet thanks! merged

PR #4130, opened 2017-09-07
remove user codec defaults to avoid merge issues

k-anderson:


bazemoreb:
Hey guys - we are looking at an issue where the user codec seems to override the device codec and the B leg of the call is always done as PCMU following the user document.  Is this commit meant to help that?

PR #4154, opened 2017-09-08
Shutdown faster 4.1

lazedo:


lazedo:
@k-anderson lets change the reason that we handle. this doesn't work as expected. it only brings benefits in CI but in runtime is a mess.

lazedo:
we need to solve in another way

PR #4155, opened 2017-09-08
PISTON-463: add pusher.apple spec

danielfinke:
Just a couple docs/specs as follow-up to https://github.com/2600hz/kazoo/pull/4033


applications/crossbar/priv/couchdb/schemas/system_config.pusher.json
```diff
@@ -12,10 +12,19 @@
...
+                    "type": "string"
+                },
                 "certificate": {
                     "description": "PEM-encoded key and certificate",
                     "type": "string"
                 },
+                "extra_headers": {
+                    "default": {},
+                    "description": "Additional headers for INVITEs sent to push devices in the UA",
+                    "type": "object"
```
lazedo: @danielfinke maybe we can describe here the headers that define kamailio behaviour. thoughts ? need help ?

danielfinke: @lazedo I can document them here if you think that makes sense, but is there any other place where I should document supported headers on endpoints or originate reqs or anything else

lazedo: @danielfinke i was thinking on adding them as properties and allow additionalProperties (expand the definition of extra_headers

danielfinke:
@lazedo OK I have added it as a property in the extra_headers. Default behaviour for other properties is to permit them

lazedo:
@danielfinke tks. i added `push` and `invite` to the enumeration, but wasn't able to edit swagger file. can you please add those to swagger and push ?

lazedo:
@danielfinke if you're ok with it, we'll merge https://github.com/2600hz/kazoo/pull/4182 to spare you extra work

danielfinke:
@lazedo yeah no problem, works for me

PR #4156, opened 2017-09-08
kz_device_test additional coverage

claytonfinney:


core/kazoo_documents/test/kz_device_test.erl
```diff
@@ -28,7 +50,7 @@ no_legacy_sip_headers_test_() ->
                             ,{<<"out">>, OutCSH}
                             ]
                            ),
-    TestDevice = kz_json:set_value([<<"sip">>, <<"custom_sip_headers">>], CSH, Device),
+	TestDevice = kz_device:set_custom_sip_headers(Device, CSH),
```
jamesaimonetti: formatting

core/kazoo_documents/src/kz_device.erl
```diff
@@ -16,6 +16,7 @@
         ,sip_ip/1, sip_ip/2, set_sip_ip/2
         ,sip_invite_format/1, sip_invite_format/2, set_sip_invite_format/2
         ,sip_route/1, sip_route/2, set_sip_route/2
+		,set_custom_sip_headers/2
```
jamesaimonetti: formatting is borked.

core/kazoo_documents/test/kz_device_test.erl
```diff
@@ -179,6 +200,38 @@ outbound_dynamic_flags_test_() ->
      }
     ].
 
+device_param_setting_test_() ->
+    {'ok', Device} = kz_device:fetch(?MASTER_ACCOUNT, ?DEVICE_1_ID),
+
+    Device1 = kz_device:set_sip_username(Device, <<"rick_and_morty">>),
```
jamesaimonetti: It seems like you could create a list of `[{Value, Setter, Getter}]` and then use a LC to create the tests. Something like:
```erlang
Setup = [{<<"rick_and_morty">>, fun kz_device:set_sip_username/2, fun kz_device:sip_username/1}
         ,...
        ],
[?_assertEqual(Value, Get(Set(Device, Value))) || {Value, Set, Get} <- Setup].
```

PR #4158, opened 2017-09-08
[master] log successful login for cb_user_auth only

icehess:
4.1: #4159


lazedo:
@icehess please rebase for circleci

PR #4164, opened 2017-09-11
add role counter for app/node

lazedo:


core/kazoo_globals/src/kz_nodes.erl
```diff
@@ -1071,3 +1072,119 @@ node_encoded() ->
             Encoded;
         {'ok', Encoded} -> Encoded
     end.
+
+-spec whapp_role_count(text(), text()) -> integer().
```
mark2600: Any reason we are not moving these to kapp* functions ?

lazedo: since the others weren't i just followed the pattern

PR #4177, opened 2017-09-12
cleanup the parking slot if the retrieve fails

k-anderson:


applications/callflow/src/module/cf_park.erl
```diff
@@ -139,8 +139,8 @@ retrieve(SlotNumber, Call) ->
                     {'ok', 'retrieved'};
                                                 %                    kapps_call_command:wait_for_hangup();
                 {'error', _E}=E ->
-                    update_presence(<<"terminated">>, Slot),
```
lazedo: @k-anderson can we add this back ?

PR #4186, opened 2017-09-14
[master] KAZOO-5625: fix cdrs pagination

icehess:
* calculating page_size correctly
* loop over databases until requested page_size is exhausted or database is exhausted.
* reverse the order of databases to query

### DIFF
--- a/applications/crossbar/src/modules/cb_cdrs.erl
+++ b/applications/crossbar/src/modules/cb_cdrs.erl
@@ -270,7 +270,7 @@ validate(Context, _, _) ->
 %% @end
 %%--------------------------------------------------------------------
 -spec load_cdr_summary(cb_context:context(), req_nouns()) -> cb_context:context().
-load_cdr_summary(Context, [_, {?KZ_ACCOUNTS_DB, [_]} | _]) ->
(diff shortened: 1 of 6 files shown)


applications/crossbar/src/modules/cb_cdrs.erl
```diff
@@ -270,7 +270,7 @@ validate(Context, _, _) ->
 %% @end
 %%--------------------------------------------------------------------
 -spec load_cdr_summary(cb_context:context(), req_nouns()) -> cb_context:context().
-load_cdr_summary(Context, [_, {?KZ_ACCOUNTS_DB, [_]} | _]) ->
+load_cdr_summary(Context, [_, {?KZ_ACCOUNTS_DB, _} | _]) ->
```
jamesaimonetti: When would this not be a list of one? `/v2/accounts/{ACCOUNT_ID}/cdrs` would give you `[{<<"cdrs">>, []}, {<<"accounts">>, [<<"{ACCOUNT_ID}">>]}]`. What request is coming in with more parameters to the `accounts` endpoint?

icehess: There is no such a request, so basically `[_] =:= _` and also `[_]` is 3 characters to type compare to `_`.

PR #4188, opened 2017-09-15
fix endpoint call recording merging "any"

lazedo:


lazedo:
4.1 https://github.com/2600hz/kazoo/pull/4189

PR #4191, opened 2017-09-18
kazoo attachments fixes & enhancements

lazedo:
- moves common functions to kz_att_util
- allow composition of url attachment for all drivers
- add google storage
- add azure storage


jamesaimonetti:
Can you create some basic docs to go along with these backends? I can flesh them out with more text, sample code to set them up, etc, but it would be great to have the minimal settings documented at least.

lazedo:
@jamesaimonetti lets create a pr for that ? if you layout the docs the way you want, i can fill them with the basic info, and then you complete ?

PR #4193, opened 2017-09-18
[4.1] fix pass recovery template when previewing

icehess:
if `Email` is not present in payload and it's previewing use `to` email
addresses

master: #4192


jamesaimonetti:
@icehess added the schema changes and merged

PR #4195, opened 2017-09-20
HELP-32352 fix endpoint inbound sip headers

lazedo:
inbound sip headers are not set when there's no authorizing endpoint


lazedo:
4.1 https://github.com/2600hz/kazoo/pull/4204

PR #4196, opened 2017-09-20
encode start_keys in request and response

icehess:
We usually normalizing `start_key` and `next_start_key` to be a timestamp, and for every db request we construct the `startkey` and `endkey` based on that, but in some situation like CDRs this is not helping. View's key in cdrs is a list and to have accurate view result we need at least two element in view's key list.

This PR is base64 encoding (url safe) erlang term binary representation of `start_key` and `next_start_key`, so for the next request we can decode it to erlang term again use that as the `startkey` as is.  

For example if `next_start_key` is `[63660927599, <<"abc123">>]` will encoded to `"g2wAAAACbgUAb6p80g5tAAAABmFiYzEyM2o"` and the client can set it as `start_key` value, and in crossbar we can decode it as erlang term again to use as `startkey` for CouchDB query.

Also this PR is returning `descending` result of MODBs for some api like recordings. (by adding a helper function `cb_modules_util:make_modb_view_descending/1`)

Also fixed KAZOO-5678 - Ledgers pagination doesn't return all results if range across 2 months

### DIFF
--- a/applications/crossbar/src/api_util.erl
+++ b/applications/crossbar/src/api_util.erl
@@ -1356,11 +1356,11 @@ do_create_resp_envelope(Context) ->
 -spec encode_start_keys(kz_json:object()) -> kz_json:object().
 encode_start_keys(Resp) ->
     lists:foldl(fun(Key, JObj) ->
-                       case kz_json:get_value(Key, JObj) of
-                           'undefined' -> JObj;
-                           Value ->
-                               kz_json:set_value(Key, encode_start_key(Value), JObj)
-                       end
+                        case kz_json:get_value(Key, JObj) of
+                            'undefined' -> JObj;
+                            Value ->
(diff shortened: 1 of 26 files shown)


applications/crossbar/src/api_util.erl
```diff
@@ -1349,10 +1351,30 @@ do_create_resp_envelope(Context) ->
...
+                                kz_json:set_value(Key, encode_start_key(Value), JObj)
+                        end
+                end
+               ,Resp
+               ,[<<"start_key">>, <<"next_start_key">>]
+               ).
+
+-spec encode_start_key(kz_json:path()) -> ne_binary().
+encode_start_key(StartKey) ->
+    kz_binary:hexencode(erlang:term_to_binary(StartKey)).
```
lazedo: why `kz_binary:hexencode` ? did you consider `kz_base64:encode` ? or `base64:encode` ?

PR #4198, opened 2017-09-20
add switch_user api to allow super user masquerade other account's users

icehess:


applications/crossbar/doc/user_authentication.md
```diff
@@ -172,3 +172,34 @@ curl -v -X POST \
     "status": "error"
 }
 ```
+
+#### Switch User
+
+You can switch to another user in another account if you're already is logged in as an admin in your master account.
+
+> PUT /v2/accounts/{ACCOUNT_ID}/users/{USER_ID}/user_auth/switch_user
```
jamesaimonetti: I don't think you need the `switch_user` path token. Toggle `user_auth`'s PUT handling based on `req_nouns` being `[{user_auth, []}]` vs `[{user_auth, []}, {users, [user_id]}, {accounts, [account_id]}]`

PR #4202, opened 2017-09-20
if the slot is considered available, clear it - 4.1

lazedo:


lazedo:
master https://github.com/2600hz/kazoo/pull/4201

PR #4204, opened 2017-09-21
HELP-32352 fix endpoint inbound sip headers - 4.1

lazedo:
inbound sip headers are not set when there's no authorizing endpoint


lazedo:
master https://github.com/2600hz/kazoo/pull/4195

PR #4208, opened 2017-09-22
Prod 154 master

k-anderson:
This allows you to add the following to a resource gateway:

```
           "invite_parameters": {
               "dynamic": [
                   "custom_channel_vars.pass-through",
                   {
                       "tag": "id",
                       "key": "custom_channel_vars.account_id"
                   }
               ],
               "static": [
                   "npid"
               ]
           }
```

Assuming the inbound call has pass-through (with value "pass-through=0288") and account_id (with value "XXXX") CCVs it will result in an INVITE request such as:

```
INVITE sip:+14158867900@10.26.0.88;npid;id=XXXX;pass-through=0288 SIP/2.0
```


jamesaimonetti:
@k-anderson can you add the comment above with usage example to the [resources](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/doc/resources.md) doc? Or maybe somewhere in stepswitch?

PR #4209, opened 2017-09-25
HELP-32303: correctly load rulesets for cf_temporal_route

mark2600:


mark2600:
Note, per our discussion in standup today I have verified that using an explicit rule_set = {RSID} will route to the "rule_set" callflow child, per the old behavior.

Using a ruleset as part of a temporal route callflow created in Advanced Callflows will correctly route to the ID of the appropriate child callflow (that of the matching ruleset).

PR #4228, opened 2017-09-29
4one: log crashes in handle_event for gen_listener

mark2600:


core/kazoo_amqp/src/gen_listener.erl
```diff
@@ -861,8 +861,14 @@ callback_handle_event(JObj
...
-        'true' -> catch Module:handle_event(JObj, BasicDeliver, ModuleState);
-        'false' -> catch Module:handle_event(JObj, ModuleState)
+        'true' ->
+            try Module:handle_event(JObj, BasicDeliver, ModuleState)
+            catch E:O -> lager:error("CRASH in handle_event ~p:~p", [E, O])
+            end;
+        'false' ->
+            try Module:handle_event(JObj, ModuleState)
+            catch E:O -> lager:error("CRASH in handle_event ~p:~p", [E, O])
+            end
```
lazedo: @mark2600 this commit creates a lot of crashes because it now returns `ok` when it catches the exception, but `ok` is not handled by caller `callback_handle_event/3`. you should move any logging to `callback_handle_event/3`  `{ 'EXIT´, _Why}` line 851

mark2600: That will not achieve what I was going for here, but I will fix the crashes.

mark2600: https://github.com/2600hz/kazoo/pull/4240 addresses this

PR #4231, opened 2017-09-29
knm_quantites: restore old behaviour and ignore knm_local in phone_nu…

k-anderson:
…mbers (also fix default for prepend apply_to)


k-anderson:
4.1 version #4230

PR #4236, opened 2017-10-04
Invalid log format

tsromana:
lager invalid format fix, added .idea folder for intellij ide into gitignore, please let me know if idea should be removed


lazedo:
@tsromana thanks!

PR #4237, opened 2017-10-04
change s3 doc host to vhost

tsromana:
change s3 doc


lazedo:
@tsromana awesome, tks.

PR #4247, opened 2017-10-04
Time is on your side

jamesaimonetti:
Looking at `kz_time:current_tstamp/0` vs `kz_time:now_s/0` performance:

```
core/kazoo_stdlib/$ make perf
...
kz_time_tests:now_s in 0.117612s
kz_time_tests:current_tstamp_s in 0.552993s
kz_time_tests:mono_now_s in 0.106005s
kz_time_tests:erlang_timestamp in 0.102483s
kz_time_tests:os_timestamp in 0.053582s
```

Also improves test coverage of the module (looking at the cover report)


jamesaimonetti:
@icehess this PR fixes that so `now_s` is in gregorian too. Use `current_unix_timestamp/0` for unix timestamps

jamesaimonetti:
@icehess also, apart from duo, I didn't see any usages that cared about unix vs gregorian - mostly just looking at diffs between two timestamps.

icehess:
Ah, I didn't notice you changed `now_s` to gregorian. 

You right about that unix vs. gregorian usage mostly. One usage I think is `pvt_created`, which using to figure out the right MODBs some times.

Also if we're planning to change from gregorian to unix we should announce it so everybody (UI team for example) knows they have to change their code if they expecting gregorian.

PR #4253, opened 2017-10-06
does more dialyzing

jamesaimonetti:
- nicer errors for emacs *compilation* buffers
- dialyze in batches (little faster)


jamesaimonetti:
Included in #4256 so ignoring here

PR #4258, opened 2017-10-10
Rate direction fixes

jamesaimonetti:
When specifying a direction on a rate doc, it would default to both directions (since binary() is not list()). Add conversion code in from_map/1 to ensure direction is a list of one when specified in the CSV.


jamesaimonetti:
backport to 4.1 at #4259

PR #4263, opened 2017-10-12
Better and more generic load view for crossbar

icehess:
Separated function for loading view result from MODBs, a time-range limited query or regular query. 

Also provides loading view and sending the result in chunks!


applications/crossbar/src/crossbar_filter.erl
```diff
@@ -1,42 +1,183 @@
...
+%%--------------------------------------------------------------------
+-spec build(cb_context:context(), boolean()) -> filter_fun().
+build(_, 'false') -> fun true/1;
+build(Context, 'true') ->
+    fun(Doc) ->
+            filter_doc_by_querystring(Doc, cb_context:query_string(Context))
+    end.
+
+-spec true(any()) -> 'true'.
+true(_) -> 'true'.
```
jamesaimonetti: what about `kz_term:always_true/1`? :)

applications/crossbar/src/crossbar_view.erl
```diff
@@ -0,0 +1,1165 @@
...
+%%
+%% Note: Keep the original passed JOBjs ORDER in your mapper!
+%% @end
+%%--------------------------------------------------------------------
+-spec apply_chunked_mapper(chunked_mapper_fun(), load_params(), kz_json:objects(), ne_binary()) ->
+                                  chunked_mapper_ret().
+apply_chunked_mapper('undefined', #{context := Context
+                                   ,chunk_response_type := 'json'
+                                   }, JObjs, _) ->
+    {JObjs, Context};
```
jamesaimonetti: This clause would crash line 795 above I think.

jamesaimonetti: I think you either need to update `chunked_mapper_ret()` to include the cowboy_req record or remove the pattern match at line 795

applications/crossbar/src/modules/cb_notifications.erl
```diff
@@ -245,7 +245,7 @@ validate(Context) ->
     validate_notifications(maybe_update_db(Context), ReqVerb).
 
 validate(Context, ?SMTP_LOG) ->
-    crossbar_modb_view:load(Context, ?CB_LIST_SMTP_LOG, [{mapper, fun normalize_view_result/1}]);
+    crossba_view:load_modb(Context, ?CB_LIST_SMTP_LOG, [{mapper, crossba_view:map_value_fun()}]);
```
jamesaimonetti: `s/crossba_view/crossbar_view/`
Try `make xref` to check for these

applications/crossbar/src/modules/cb_cdrs.erl
```diff
@@ -255,465 +231,172 @@ validate(Context, CDRId) ->
...
-    StartKey = maybe_get_start_key_from_req(Context, [CreatedTo]),
-    {'ok'
+get_view_options([{<<"cdrs">>, [?PATH_INTERACTION]}, {?KZ_ACCOUNTS_DB, _}|_]) ->
+    {?CB_INTERACTION_LIST
     ,props:filter_undefined(
-       [{'startkey', StartKey}
-       ,{'endkey', [CreatedFrom, kz_json:new()]}
-       ,{'limit', pagination_page_size(Context)}
+       [{'start_key_map', []}
+       ,{'end_key_map', crossbar_view:suffix_key_fun([kz_json:new()])}
```
jamesaimonetti: `[kz_json:new()]` is not covered by the `crossbar_view:keymap()` type.

applications/crossbar/src/modules/cb_cdrs.erl
```diff
@@ -255,465 +231,172 @@ validate(Context, CDRId) ->
...
        ,{'reduce', 'true'}
-       ,'descending'
         | maybe_add_stale_to_options(?STALE_CDR)
        ])
-    }.
+    };
+get_view_options(_) ->
+    {'undefined', []}.
 
 -spec maybe_add_stale_to_options(crossbar_doc:view_options()) -> crossbar_doc:view_options().
```
jamesaimonetti: Can you fix this spec? Should be `boolean()` instead of `crossbar_doc:view_options()`. Let's make the second clause match `false` too instead of `_`.

applications/crossbar/src/crossbar_view.erl
icehess: I forgot to change `chunked_mapper_ret()`. fixed now.

applications/crossbar/src/modules/cb_cdrs.erl
icehess: added to `crossbar_view:keymap()`

jamesaimonetti: Not quite :) Empty json object was added but this is passing in a list of json objects. Should be `kz_json:objects()` instead of `?EMPTY_JSON_OBJECT`.

jamesaimonetti:
Also, it would be great to see some data on how this performs with large databases. Ask Vinix for one of their MODBs with 250,000+ CDRs so we can compare memory/latency before and after this change.

applications/crossbar/src/modules/cb_allotments.erl
```diff
@@ -106,16 +107,21 @@ post(Context) ->
...
 
 
--spec load_consumed(cb_context:context()) -> cb_context:context().
-load_consumed(Context) ->
-    Allotments = kz_json:get_json_value(?PVT_ALLOTMENTS
-                                       ,cb_context:doc(crossbar_doc:load(?PVT_TYPE, Context, ?TYPE_CHECK_OPTION(?PVT_TYPE)))
-                                       ,kz_json:new()),
+-spec load_consumed(cb_context:context(), crossbar_status()) -> cb_context:context().
+load_consumed(Context, 'success') ->
+    Allotments = kz_json:get_json_value(?PVT_ALLOTMENTS, cb_context:doc(cb_context:doc(Context))),
```
jamesaimonetti: This can't be right. Take the doc of a doc?

applications/crossbar/src/modules/cb_port_requests.erl
```diff
@@ -614,89 +604,130 @@ update(Context, Id) ->
...
+              ,{'databases', [?KZ_PORT_REQUESTS_DB]}
+              ,'include_docs'
+               | Opts
+              ],
+    case IsRanged of
+        'true' -> crossbar_view:load_range(Context1, View, Options);
+        'false' -> crossbar_view:load(Context1, View, Options)
+    end.
+
+-spec view_key_options(cb_context:context(), ne_binary(), boolean()) -> {boolean(), cb_context:context()}.
```
jamesaimonetti: Does not return cb_context:context()

PR #4265, opened 2017-10-13
device lookup by it's cid number

onnet:
in continuation to the conversation https://forums.2600hz.com/forums/topic/9281-dynamic-extension-lookup-callflow/

### DIFF
--- a/applications/callflow/doc/route_to_cid.md
+++ b/applications/callflow/doc/route_to_cid.md
@@ -1,0 +1,58 @@
+## Route To Cid
+
+### About Route To Cid
+
+This module could be used to route calls without creating static callflows for each and every endpoint. 
+This could be especially usefull for Asterisk based newbies to eliminate shock of absence of explicit binding between devices and extensions  :)  
+
+
+```
+{
+   "_id": "34396ff84198c03df70b6f5073f04a38",
+   "flow": {
+       "children": {
+       },
+       "data": {
+           "cid_types": [
+               "internal"
+           ],
+           "endpoint_types": [
+               "device",
+               "useer"
+           ],
+           "timeout": 60,
+           "can_call_self": true,
+           "delay": 0
+       },
+       "module": "route_to_cid"
+   },
+   "numbers": [
+   ],
+   "patterns": [
+       "^([0-9]{3,4})$"
+   ],
+   "metadata": {
+   },
+   "pvt_type": "callflow"
+}
+```
+
+#### Schema
+
+Endpoints lookup by cid number
+
+
+
+Key | Description | Type | Default | Required
+--- | ----------- | ---- | ------- | --------
+`can_call_self` | Toggle whether devices of the same owner can call each other | `boolean()` |   | `false`
+`can_text_self` | Toggle whether devices of the same owner can text each other | `boolean()` |   | `false`
+`cid_types` | CID types to perform seacrh: internal, external, custom | `array()` | `[]` | `false`
+`delay` | How long to delay ringing the device, in seconds | `integer()` | `0` | `false`
+`endpoint_types` | Endpoint types to perform seacrh: user, device | `array()` | `[]` | `false`
+`static_invite` | Override the SIP Username | `string()` |   | `false`
+`suppress_clid` | Suppress sending caller ID | `boolean()` |   | `false`
+`timeout` | Time, in seconds, to wait for device to bridge | `integer()` | `0` | `false`
+
+
+
--- a/applications/callflow/doc/route_to_cid.md
+++ b/applications/callflow/doc/route_to_cid.md
@@ -47,7 +47,8 @@ Key | Description | Type | Default | Required
 --- | ----------- | ---- | ------- | --------
 `can_call_self` | Toggle whether devices of the same owner can call each other | `boolean()` |   | `false`
 `can_text_self` | Toggle whether devices of the same owner can text each other | `boolean()` |   | `false`
-`cid_types` | CID types to perform seacrh: internal, external, custom | `array()` | `[]` | `false`
+`cid_types.[]` |   | `string()` |   | `false`
+`cid_types` | CID types to perform seacrh: internal, external, custom | `array(string())` | `["internal"]` | `false`
 `delay` | How long to delay ringing the device, in seconds | `integer()` | `0` | `false`
 `endpoint_types` | Endpoint types to perform seacrh: user, device | `array()` | `[]` | `false`
 `static_invite` | Override the SIP Username | `string()` |   | `false`
--- a/applications/callflow/src/module/cf_route_to_cid.erl
+++ b/applications/callflow/src/module/cf_route_to_cid.erl
@@ -1,0 +1,72 @@
+%%%-------------------------------------------------------------------
+%%% @copyright (C) 2011-2017, 2600Hz INC
+%%% @doc
+%%%
+%%% @end
+%%% @contributors
+%%%   Kirill Sysoev
+%%%-------------------------------------------------------------------
+-module(cf_route_to_cid).
+
+-behaviour(gen_cf_action).
+
+-include("callflow.hrl").
+
+-export([handle/2]).
+
+%%--------------------------------------------------------------------
+%% @public
+%% @doc
+%% This module attempts to lookup endpoints by it's cid number.
+%% Returns continue if fails to connect or stop when successful.
+%% @end
+%%--------------------------------------------------------------------
+-spec handle(kz_json:object(), kapps_call:call()) -> 'ok'.
+handle(Data, Call) ->
+    case endpoints_lookup(Data, Call) of
+        [] ->
+            lager:debug("~p not found", [kapps_call:callee_id_number(Call)]),
+            cf_exe:continue(Call);
+        EndpointJObjs ->
+            Vals = [{<<"strategy">>,<<"simultaneous">>}
+                   ,{<<"endpoints">>, EndpointJObjs}
+                   ,{<<"timeout">>,kz_json:get_integer_value(<<"timeout">>, Data, ?DEFAULT_TIMEOUT_S)}
+                   ],
+            cf_group:handle(kz_json:set_values(Vals, Data), Call)
+    end.
+
+-spec endpoints_lookup(kz_json:object(), kapps_call:call()) -> kz_proplist().
+endpoints_lookup(Data, Call) ->
+    lists:foldl(fun(CIDType, Acc) -> cid_type_based_lookup(CIDType, Data, Call) ++ Acc end
+               ,[]
+               ,kz_json:get_value(<<"cid_types">>, Data, [])
+               ).
+
+-spec cid_type_based_lookup(ne_binary(), kz_json:object(), kapps_call:call()) -> kz_proplist().
+cid_type_based_lookup(CIDType, Data, Call) ->
+    CIDNum = kapps_call:callee_id_number(Call),
+    ViewOptions = [{'key', [CIDNum, CIDType]}],
+    AccountDb = kapps_call:account_db(Call),
+    EndpointTypes = kz_json:get_value(<<"endpoint_types">>, Data, []),
+    case kz_datamgr:get_results(AccountDb, <<"attributes/endpoints_lookup">>, ViewOptions) of
+        {'ok', JObjs} ->
+            [endpoint_format(kz_json:get_ne_binary_value(<<"id">>, JObj)
+                            ,kz_json:get_ne_binary_value(<<"value">>, JObj)
+                            ,Data
+                            )
+             || JObj <- JObjs
+                    ,kz_json:get_ne_binary_value(<<"id">>, JObj) =/= 'undefined'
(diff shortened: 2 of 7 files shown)


applications/callflow/src/module/cf_route_to_cid.erl
```diff
@@ -0,0 +1,91 @@
...
+%% @public
+%% @doc
+%% This module attempts to lookup device by it's internal cid number.
+%% Returns continue if fails to connect or stop when successful.
+%% @end
+%%--------------------------------------------------------------------
+-spec handle(kz_json:object(), kapps_call:call()) -> 'ok'.
+handle(Data, Call) ->
+    case find_device_by_cid_number(Call) of
+        {'ok', EndpointId} ->
```
jamesaimonetti: One immediate thought, without looking too deeply at the code, is the create a `Data` object with the EndpointId and call into `cf_device:handle/2`. So this module's only job is to resolve CallerID Number into an Endpoint ID.
Also, it could instead call into `cf_group:handle/2` and set "endpoints" to `["{ENDPOINT_ID}"]` which would handle users or devices.

onnet: @jamesaimonetti Thanks for a hint. Decided to go cf_group:handle/2 way..

applications/callflow/doc/route_to_cid.md
```diff
@@ -0,0 +1,25 @@
+## Route To Cid
+
+### About Route To Cid
+
+#### Schema
+
+Skeleton JSON schema
```
jamesaimonetti: :) Probably add a "description" to the main JSON schema

applications/callflow/src/module/cf_route_to_cid.erl
```diff
@@ -0,0 +1,101 @@
...
+-spec handle(kz_json:object(), kapps_call:call()) -> 'ok'.
+handle(Data, Call) ->
+    case endpoints_lookup(Data, Call) of
+        [] ->
+            lager:debug("~p not found", [kapps_call:callee_id_number(Call)]),
+            cf_exe:continue(Call);
+        EndpointJObjs ->
+            Vals = [{<<"strategy">>,<<"simultaneous">>}
+                   ,{<<"endpoints">>, EndpointJObjs}
+                   ,{<<"timeout">>,kz_json:get_integer_value(<<"timeout">>, Data, ?DEFAULT_TIMEOUT_S)}],
```
jamesaimonetti: Drop the ending list bracket to a new line:
```erlang
[{<<"foo">>, <<"bar">>}
,{<<"biz">>, <<"bam">>}
],
```

applications/callflow/src/module/cf_route_to_cid.erl
```diff
@@ -0,0 +1,101 @@
...
+        ++ maybe_user_external_cid_lookup(Data, Call).
+
+maybe_device_external_cid_lookup(Data, Call) ->
+    case kz_json:is_true(<<"device_external_cid_lookup">>, Data, 'false') of
+        'false' -> [];
+        'true' ->
+            type_based_cid_lookup(<<"device">>
+                                 ,<<"devices/listing_by_external_cidnum">>
+                                 ,Data
+                                 ,Call)
```
jamesaimonetti: Drop the end-paren to a newline

applications/callflow/src/module/cf_route_to_cid.erl
```diff
@@ -0,0 +1,101 @@
...
+    end.
+
+maybe_device_internal_cid_lookup(Data, Call) ->
+    case kz_json:is_true(<<"device_internal_cid_lookup">>, Data, 'false') of
+        'false' -> [];
+        'true' ->
+            type_based_cid_lookup(<<"device">>
+                                 ,<<"devices/listing_by_internal_cidnum">>
+                                 ,Data
+                                 ,Call)
```
jamesaimonetti: Drop paren

applications/callflow/src/module/cf_route_to_cid.erl
```diff
@@ -0,0 +1,101 @@
...
+    end.
+
+maybe_user_external_cid_lookup(Data, Call) ->
+    case kz_json:is_true(<<"user_external_cid_lookup">>, Data, 'false') of
+        'false' -> [];
+        'true' ->
+            type_based_cid_lookup(<<"user">>
+                                 ,<<"users/listing_by_external_cidnum">>
+                                 ,Data
+                                 ,Call)
```
jamesaimonetti: drop paren

applications/callflow/src/module/cf_route_to_cid.erl
```diff
@@ -0,0 +1,101 @@
...
+    end.
+
+maybe_user_internal_cid_lookup(Data, Call) ->
+    case kz_json:is_true(<<"user_internal_cid_lookup">>, Data, 'false') of
+        'false' -> [];
+        'true' ->
+            type_based_cid_lookup(<<"user">>
+                                 ,<<"users/listing_by_internal_cidnum">>
+                                 ,Data
+                                 ,Call)
```
jamesaimonetti: drop paren

applications/callflow/src/module/cf_route_to_cid.erl
```diff
@@ -0,0 +1,101 @@
...
+            lager:debug("~p not found", [kapps_call:callee_id_number(Call)]),
+            cf_exe:continue(Call);
+        EndpointJObjs ->
+            Vals = [{<<"strategy">>,<<"simultaneous">>}
+                   ,{<<"endpoints">>, EndpointJObjs}
+                   ,{<<"timeout">>,kz_json:get_integer_value(<<"timeout">>, Data, ?DEFAULT_TIMEOUT_S)}],
+            cf_group:handle(kz_json:set_values(Vals, Data), Call)
+    end.
+
+endpoints_lookup(Data, Call) ->
```
jamesaimonetti: specs on all functions from here on are needed

applications/callflow/src/module/cf_route_to_cid.erl
```diff
@@ -0,0 +1,101 @@
...
+
+type_based_cid_lookup(EndpointType, View, Data, Call) ->
+    CIDNum = kapps_call:callee_id_number(Call),
+    ViewOptions = [{'key', CIDNum}],
+    AccountDb = kapps_call:account_db(Call),
+    case kz_datamgr:get_results(AccountDb, View, ViewOptions) of
+        {'ok', JObjs} ->
+            [endpoint_format(kz_json:get_ne_binary_value(<<"id">>, JObj), EndpointType, Data)
+             || JObj <- JObjs
+                    ,kz_json:get_ne_binary_value(<<"id">>, JObj) =/= 'undefined'];
```
jamesaimonetti: Drop end list bracket

applications/callflow/src/module/cf_route_to_cid.erl
```diff
@@ -0,0 +1,101 @@
...
+        _E ->
+            []
+    end.
+
+endpoint_format(EndpointId, EndpointType, Data) ->
+    Values = [{<<"endpoint_type">>, EndpointType}
+             ,{<<"id">>, EndpointId}
+             ,{<<"delay">>,0}
+             ,{<<"timeout">>,kz_json:get_integer_value(<<"timeout">>, Data, ?DEFAULT_TIMEOUT_S)}],
```
jamesaimonetti: Drop end list bracket

applications/crossbar/priv/couchdb/account/devices.json
```diff
@@ -20,6 +20,24 @@
                 "}"
             ]
         },
+        "listing_by_external_cidnum": {
+            "map": [
+                "function (doc) {",
+                "  if (doc.pvt_type != 'device' || doc.pvt_deleted) return;",
+                "  if(doc.caller_id.external.number)",
+                "    emit(doc.caller_id.external.number, {'id': doc._id})",
```
jamesaimonetti: You get `_id` in the view results by default so no need to emit it. You get "key", "value", and "id" (and "doc" if you include docs) in each view result.

applications/crossbar/priv/couchdb/account/devices.json
```diff
@@ -20,6 +20,24 @@
...
+                "}"
+            ]
+        },
+        "listing_by_internal_cidnum": {
+            "map": [
+                "function (doc) {",
+                "  if (doc.pvt_type != 'device' || doc.pvt_deleted) return;",
+                "  if(doc.caller_id.internal.number)",
```
jamesaimonetti: Also, we do support "custom" caller IDs. I wonder if instead of having a view for internal and a view for external, if you just have one view that emits a composite key like:
```js
function(doc) {
  if (typeof doc.caller_id === "undefined"); return;

  for(var type in doc.caller_id) {
    emit([doc[type].number, type], doc.pvt_type);
  }
}
```
Untested, but the idea is you would get rows like:
```erlang
{["+1234567890", "external"], "user"}
,{["1005", "internal"], "device"}
 ,...
```

Then the query in the callflow becomes:
```erlang
{key, [kapps_call:callee_id_number(Call)
      | case kz_json:get_ne_binary_value(<<"caller_id_type">>, Data) of 'undefined' -> []; Type -> [Type] end %% this should be a function
      ]
}
```

Again, all untested but it feels like it could clean up a lot of duplicate code and allow people to create callflows triggering on custom caller IDs depending on how the call entered the system.

Then you can iterate over the results to create the endpoints (since you have the types of the document in the "value" you can filter there based on Data too).

jamesaimonetti: To add, the view can go in "attributes.json" design doc.

onnet:
@jamesaimonetti 
Hi James, your comments have been addressed.

applications/callflow/src/module/cf_route_to_cid.erl
```diff
@@ -0,0 +1,72 @@
...
+                   ,{<<"timeout">>,kz_json:get_integer_value(<<"timeout">>, Data, ?DEFAULT_TIMEOUT_S)}
+                   ],
+            cf_group:handle(kz_json:set_values(Vals, Data), Call)
+    end.
+
+-spec endpoints_lookup(kz_json:object(), kapps_call:call()) -> kz_proplist().
+endpoints_lookup(Data, Call) ->
+    lists:foldl(fun(CIDType, Acc) -> cid_type_based_lookup(CIDType, Data, Call) ++ Acc end
+               ,[]
+               ,kz_json:get_value(<<"cid_types">>, Data, [])
```
jamesaimonetti: Do you want this to create 0 endpoints if no "cid_types" restriction exists in `Data`? Perhaps it should default to `[<<"device">>]` or `[<<"device">>, <<"user">>]`?

applications/callflow/src/module/cf_route_to_cid.erl
```diff
@@ -0,0 +1,72 @@
...
+    end.
+
+-spec endpoint_format(ne_binary(), ne_binary(), kz_json:object()) -> kz_json:object().
+endpoint_format(EndpointId, EndpointType, Data) ->
+    Values = [{<<"endpoint_type">>, EndpointType}
+             ,{<<"id">>, EndpointId}
+             ,{<<"delay">>,0}
+             ,{<<"timeout">>,kz_json:get_integer_value(<<"timeout">>, Data, ?DEFAULT_TIMEOUT_S)}
+             ],
+    kz_json:set_values(Values, kz_json:new()).
```
jamesaimonetti: I would just do a `kz_json:from_list(Values)` instead.

applications/crossbar/priv/couchdb/schemas/callflows.route_to_cid.json
```diff
@@ -0,0 +1,44 @@
...
+    "properties": {
+        "can_call_self": {
+            "description": "Toggle whether devices of the same owner can call each other",
+            "type": "boolean"
+        },
+        "can_text_self": {
+            "description": "Toggle whether devices of the same owner can text each other",
+            "type": "boolean"
+        },
+        "cid_types": {
```
jamesaimonetti: let's add the items schema:

```json
"cid_types":{
    "type":"array",
    "items":{
        "type":"string",
    }
}
```

applications/callflow/src/module/cf_route_to_cid.erl
jamesaimonetti: Sorry, should be `["internal"]` or `["internal", "external"]` - not sure what happened :)

applications/crossbar/priv/couchdb/schemas/callflows.route_to_cid.json
jamesaimonetti: Another thought - set `minItems` to `1` to make sure no empty arrays (and remove `default:[]`)

onnet:
@jamesaimonetti  we've got 'minItems 1' as a guard and probably do not need default over here: kz_json:get_value(<<"cid_types">>, Data, [<<"internal">>])

... but anyway I feel myself more confident with defaults.. What would be your decision, remove or not to remove? :)

jamesaimonetti:
@onnet I would prefer no default so it requires the API client to send a properly-configured callflow action.

PR #4269, opened 2017-10-16
Added bh_ping

kalda341:


jamesaimonetti:
Hi @kalda341, can you add a quick doc to blackhole/doc/ping.md that describes the basic usage? Maybe just a quick sample bit of Javascript (or pseudo-code) that shows how to issue the ping and branch on pong/error? Then add it to the appropriate place in `kazoo/doc/mkdocs/mkdocs.yml` so `make docs` succeeds.

I plan to write an AST parser for blackhole modules to auto-gen bindings, commands, etc, but until then I'd like now PRs to add even basic usage examples for each module. Thanks!

jamesaimonetti:
@kalda341 perfect! Thanks, merged manually.

PR #4273, opened 2017-10-17
Cps dialyzer

jamesaimonetti:
Some basic dialyzing to add to your branch @lazedo if you like.


jamesaimonetti:
Ignore circle, it dialyzes against master and not the cps branch. I wonder if we can detect that in the circle environment though...hmm

jamesaimonetti:
OK, `CIRCLE_COMPARE_URL` is only populated properly on the initial PR open. If you push commits after that, the compare URL will be from the last commit of the PR to the new HEAD. So that won't work...

lazedo:
@jamesaimonetti BIG THANK YOU! i cherry picked the commits in the branch

PR #4276, opened 2017-10-18
fix s3 region parameter - 4.1

lazedo:


lazedo:
master https://github.com/2600hz/kazoo/pull/4275

PR #4287, opened 2017-10-20
Change missed_call message + fix teletype_binding

icehess:


applications/teletype/src/teletype_bindings.erl
```diff
@@ -70,16 +70,16 @@ notification(JObj) ->
 maybe_send_update(JObj, RoutingKey, #{'completed' := _Completed}=Map) ->
     %% for now we just only care about at least one success
     print_result(RoutingKey, Map),
-    Metadata = kz_json:from_list(maps:to_list(Map)),
+    Metadata = kz_json:from_list_recursive(maps:to_list(Map)),
```
jamesaimonetti: Doesn't `kz_json` have a `from_map/1` now? It is recursive too; maybe test to make sure it works as expected and use it instead?

icehess: yes, it does have `from_map`, but first it does not convert keys to binary (which i need here) and second it's only converting nested maps, so if you have a key in the maps with a proplist value, it does not convert proplist to jobj.

PR #4288, opened 2017-10-22
core/kazoo: Added "audio/x-wave" mime support

sergey-safarov:
x-wave mime is used when voicemail in wav format
Fixed file extension generation for "x-wave" mime


jamesaimonetti:
@sergey-safarov thanks! merged manually

PR #4291, opened 2017-10-26
KAZOO-5737: add conference relate to metaflows

jamesaimonetti:


core/kazoo/src/kz_util.erl
```diff
@@ -143,26 +143,26 @@ log_stacktrace_mfa(M, F, Args, Info) ->
...
+                  | 'notice'
+                  | 'info'
+                  | 'debug'
+                  | ne_binary().
 
 -spec change_console_log_level(log_level()) -> 'ok'.
 change_console_log_level(L) when is_atom(L) ->
-    lager:info("updated console_log to level ~s", [L]),
-    lager:set_loglevel('lager_console_backend', L);
+    lager:set_loglevel('lager_file_backend', "log/console.log", L),
```
lazedo: only thing i don't agree. you are forcing the location of the console. when running in the foreground (docker images or shell for example) we want to configure where console output goes aswell as its level.

PR #4300, opened 2017-10-31
PISTON-494: wrap as JSON the values in cf_webhook's props to publish

danielfinke:
Right now the JSON payload to kapi_notifications:publish_webhook/1 is recursively converted to proplist, but kz_api does not recursively convert back before the publish, leading to "invalid_ejson". The "Hook" and "Data" props are at fault.
PR just leaves the types as kz_json:object() so the conversion is not required.


jamesaimonetti:
Merged in 30ea4baca9

PR #4302, opened 2017-10-31
KAZOO-5749: do not defer if progress is not possible

k-anderson:


core/kazoo_amqp/src/api/kapi_route.erl
```diff
@@ -105,8 +105,7 @@ is_actionable_resp(Prop) when is_list(Prop) ->
         <<"sms">> -> 'true';
         <<"plan">> -> 'true';
         <<"application">> -> 'true';
-        <<"error">> ->
-            kz_term:is_true(props:get_value(<<"Defer-Response">>, Prop));
+        <<"error">> -> 'true';
```
k-anderson: @jamesaimonetti This was originally added by you (iirc).  What other types of error responses was it guarding against?  I looked but seems like an error reply to a route_req is actionable...

PR #4305, opened 2017-11-01
fix presence update

lazedo:
requires https://github.com/2600hz/kazoo-configs-kamailio/pull/35


lazedo:
4.1 https://github.com/2600hz/kazoo/pull/4306

PR #4310, opened 2017-11-02
[DON'T MERGE] Add /harrys API

harenson:
**This PR should NOT be accepted/merged**

From inside `applications/crossbar` folder run the following command for running `cb_harrys_SUITE` common test suite:

```shell
ct_run -suite cb_harrys_SUITE -include src/ -pa ../../core/kazoo_stdlib/ -pa ebin/ -pa ../../deps/jiffy/ebin -pa ../../deps/jiffy/priv -logdir ./testlogs
```

### DIFF
--- a/applications/crossbar/src/modules/cb_harrys.erl
+++ b/applications/crossbar/src/modules/cb_harrys.erl
@@ -1,0 +1,195 @@
+%%% GET|PUT     /:api_version/accounts/:account_id/harrys
+%%% POST|DELETE /:api_version/accounts/:account_id/harrys/:harry_id
+-module(cb_harrys).
+
+-export([init/0
+        ,allowed_methods/0, allowed_methods/1
+        ,resource_exists/0, resource_exists/1
+        ,validate/1, validate/2
+        ,put/1
+        ,post/2
+        ,delete/2
+        ]).
+
+-include("crossbar.hrl").
+
+-define(CB_LIST, <<"harrys/crossbar_listing">>).
+
+%%%===================================================================
+%%% API
+%%%===================================================================
+
+%%--------------------------------------------------------------------
+%% @public
+%% @doc
+%% Initializes the bindings this module will respond to.
+%% @end
+%%--------------------------------------------------------------------
+-spec init() -> 'ok'.
+init() ->
+    _ = crossbar_bindings:bind(<<"*.allowed_methods.harrys">>, ?MODULE, 'allowed_methods'),
+    _ = crossbar_bindings:bind(<<"*.resource_exists.harrys">>, ?MODULE, 'resource_exists'),
+    _ = crossbar_bindings:bind(<<"*.validate.harrys">>, ?MODULE, 'validate'),
+    _ = crossbar_bindings:bind(<<"*.execute.get.harrys">>, ?MODULE, 'get'),
+    _ = crossbar_bindings:bind(<<"*.execute.put.harrys">>, ?MODULE, 'put'),
+    _ = crossbar_bindings:bind(<<"*.execute.post.harrys">>, ?MODULE, 'post'),
+    _ = crossbar_bindings:bind(<<"*.execute.delete.harrys">>, ?MODULE, 'delete').
+
+%%--------------------------------------------------------------------
+%% @public
+%% @doc
+%% Given the path tokens related to this module, what HTTP methods are
+%% going to be responded to.
+%% @end
+%%--------------------------------------------------------------------
+-spec allowed_methods() -> http_methods().
+allowed_methods() ->
+    [?HTTP_GET, ?HTTP_PUT].
+
+-spec allowed_methods(PathToken::path_token()) -> http_methods().
+allowed_methods(_PathToken) ->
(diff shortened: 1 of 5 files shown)


applications/crossbar/src/modules/cb_harrys.erl
```diff
@@ -0,0 +1,195 @@
...
+%% Does the path point to a valid resource
+%% So /harrys => []
+%%    /harrys/foo => [<<"foo">>]
+%%    /harrys/foo/bar => [<<"foo">>, <<"bar">>]
+%% @end
+%%--------------------------------------------------------------------
+-spec resource_exists() -> 'true'.
+resource_exists() -> 'true'.
+
+-spec resource_exists(PathToken::path_token()) -> 'true'.
```
jamesaimonetti: We typically won't name variables in a spec unless they are returned unchanged.

applications/crossbar/test/cb_harrys_SUITE.erl
```diff
@@ -0,0 +1,187 @@
...
+%% Private (helpers)
+%% =============================================================================
+-spec login_user() ->
+  ok.
+login_user() ->
+  URL = "http://192.168.99.100:8000/v2/user_auth",
+  Creds = <<"1141aeae55223480eca2d9e50e3ecb3b">>,
+  ID = <<"harenson">>,
+  % JSON={"data": {"credentials": "value here", "account_name": "value here"}}.
+  ReqBody = jiffy:encode({[{<<"data">>, {[{<<"credentials">>, Creds},
```
jamesaimonetti: We don't hard-code the internal representation of JSON. Use kz_json:from_list/1 or kz_json:set_value(s) to create objects.

applications/crossbar/test/cb_harrys_SUITE.erl
```diff
@@ -0,0 +1,187 @@
...
+                                          {<<"account_name">>, ID}]} }]}),
+  Request = {URL, [], "application/json", ReqBody},
+  {ok, {{_Vsn, 201, _HTTPText}, _RespHeaders, RespBody}} =
+    httpc:request(put, Request, [], []),
```
jamesaimonetti: Take a look at the core libraries. In this case, `kz_http` will be useful.

applications/crossbar/test/cb_harrys_SUITE.erl
```diff
@@ -0,0 +1,187 @@
...
+  DecodedResp = jiffy:decode(RespBody),
```
jamesaimonetti: kz_json:decode/1 instead

applications/crossbar/test/cb_harrys_SUITE.erl
```diff
@@ -0,0 +1,187 @@
...
+  Data = get_value(<<"data">>, DecodedResp),
+  AccID = get_value(<<"account_id">>, Data),
+  Token = get_value(<<"auth_token">>, DecodedResp),
+  ok = application:set_env(cb_harrys_tests, account_id, binary_to_list(AccID)),
```
jamesaimonetti: We ticky all atoms, despite protests from a select few :)

applications/crossbar/test/cb_harrys_SUITE.erl
```diff
@@ -0,0 +1,187 @@
...
+  AccID.
+
+-spec auth_token() ->
+  string().
+auth_token() ->
+  {ok, AuthToken} = application:get_env(cb_harrys_tests, auth_token),
+  AuthToken.
+
+-spec build_harry(FName::binary(), LName::binary(), Email::binary()) ->
+  jiffy:json_value().
```
jamesaimonetti: kz_json has a lot of exported types to use. Also, look in kz_types.hrl for lots of types we use. For instance, if the arg cannot be an empty binary (it must be non-empty), we have defined `ne_binary()` to reflect that. A weird convention if a term can be `'undefined' | type()` is to call it `api_type()`. For instance, `'undefined' | ne_binary()` is typed as `api_ne_binary()`.

applications/crossbar/test/cb_harrys_SUITE.erl
```diff
@@ -0,0 +1,187 @@
...
+  {ok, {{_Vsn, HTTPCode, _HTTPText}, _RespHeaders, RespBody}} =
+    httpc:request(Method, Request, [], []),
+  DecodedResp = jiffy:decode(RespBody),
+  ct:log("Got ~p response with:~nRespBody: ~p~n", [HTTPCode, DecodedResp]),
+  {HTTPCode, DecodedResp}.
+
+-spec get_value(Key::binary(), Data::jiffy:json_value()) ->
+  term().
+get_value(Key, {Data}) ->
+  proplists:get_value(Key, Data).
```
jamesaimonetti: Again, unpacking the internal structure is verboten. I would write this as:

```erlang
get_value(Key, JObj) ->
    kz_json:get_value(Key, JObj).
```

And you'll probably just want to change your `get_value/2` calls to call `kz_json:get_value/2` directly. Also, kz_json has many typed versions of get_value. If you know the value must be a ne_binary, use `get_ne_binary_value`; if an object, `get_json_value`, etc.

PR #4312, opened 2017-11-02
add custom CCVs to conference dialout

jamesaimonetti:


jamesaimonetti:
Add to #4314 instead so I can track all the client work.

PR #4314, opened 2017-11-02
KAZOO-5751: more client fixes for pivot/conferences

jamesaimonetti:


core/kazoo_documents/src/kzd_conference.erl
```diff
@@ -0,0 +1,31 @@
+-module(kzd_conference).
```
lazedo: @jamesaimonetti can you check the `cps` branch ? there's a module there's a `kzd_conference` module

jamesaimonetti: @lazedo i guess we'll race on getting @k-anderson to merge one or the other first! :)

jamesaimonetti:
@lazedo since my kzd_conference is much smaller, want to merge this and rebase cps with it ? :)

lazedo:
@jamesaimonetti not sure, this kzd_conference is document oriented and the one in cps is event oriented. i'm ok with renaming the one in cps to something more suitable. thoughts ?

jamesaimonetti:
@lazedo we have kz_call_event, could yours be kz_conference_event?

lazedo:
@jamesaimonetti sold!

PR #4320, opened 2017-11-06
[master] KAZOO-5736: properly handling retention for voicemail operations

icehess:
4.1: #4321

### DIFF
--- a/core/kazoo_modb/priv/couchdb/views/mailbox_messages.json
+++ b/core/kazoo_modb/priv/couchdb/views/mailbox_messages.json
@@ -3,14 +3,69 @@
     "language": "javascript",
     "views": {
         "count_per_folder": {
-            "map": "function(doc) { if (doc.pvt_type != 'mailbox_message' || doc.pvt_deleted) return; emit([doc.source_id, doc.metadata.folder], 1); }",
-            "reduce": "_count"
+            "map": [
+                "function(doc) {",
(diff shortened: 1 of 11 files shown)


core/kazoo_modb/priv/couchdb/views/mailbox_messages.json
```diff
@@ -2,15 +2,70 @@
...
+            ]
+        },
         "count_per_folder": {
-            "map": "function(doc) { if (doc.pvt_type != 'mailbox_message' || doc.pvt_deleted) return; emit([doc.source_id, doc.metadata.folder], 1); }",
-            "reduce": "_count"
+            "map": [
+                "function(doc) {",
+                "  if (doc.pvt_type != 'mailbox_message' || doc.pvt_deleted)",
+                "    return;",
+                "  emit([doc.metadata.folder, doc.metadata.timestamp, doc.source_id]);",
```
jamesaimonetti: does `emit/1` use `null` as the value by default?

jamesaimonetti:
@icehess i think you need to rebase and all will be well with the PR

core/kazoo_modb/priv/couchdb/views/mailbox_messages.json
icehess: yes, though i'm not sure about bigcouch :-)

icehess: well I see karl added `missing_metadata` without specifying value, so it is working in bigcouch too.

PR #4329, opened 2017-11-09
cb_registrations loses endpoints

onnet:
Got two registrations from FreePBX host.
One registration is a device.
Another one a trunkstore registration. 
Contact appears to be the same and dict:store/3 overwrites one another.


lazedo:
@onnet if the contact is the same shouldn't that be fixed in freepbx ?

onnet:
Hi @lazedo,
 
My reasons are:
- despite contact fields are the same, these two different registrations exist and work fine and cb_registrations should just reflect real situation to not to confuse user/admin;
- shouldn't Kazoo be foolproof itself?
- cb_registrations/count gives the real amount of registrations, but list of registrations, provided by cb_registrations has less members, which feels like an inconsistency.

Anyway, I certainly do not insist on this. Just let me know and I'll close this pr.

Regards,
Kirill

lazedo:
@onnet thanks

PR #4336, opened 2017-11-13
Fix bad catch in sup.erl

efine:
In `sup:print_result/2`, the code

```erlang
try stdout("~s", [String])
catch
    erro:badarg -> stdout("Result: ~p", [String])
end
```

is not catching `badarg` errors because `erro:` is not a valid exception class. This causes an exception to be propagated when it should not be.

This was fixed by changing the offending line to

```erlang
    error:badarg -> stdout("Result: ~p", [String])
```


jamesaimonetti:
@efine merged!

PR #4338, opened 2017-11-13
HELP-33189-4two: fix return value for GET on port_in numbers

mark2600:


k-anderson:
This will be pulled into 4.2 when master is merged into it next

PR #4349, opened 2017-11-15
looks like a typo, but not sure

onnet:
It seems "t38_passthrough" should be "t38_passthru" or fax should be just enabled "fax_enable_t38"

Sorry if I'm wrong, but just offending to the eye..


jamesaimonetti:
https://freeswitch.org/confluence/display/FREESWITCH/mod_spandsp certainly agrees with you!

PR #4352, opened 2017-11-17
t38_passthru fax_enable_t38 requirement

onnet:
https://freeswitch.org/confluence/display/FREESWITCH/T.38 says: 
As of FS v1.6, t38_passthru=true can *NOT* be used if fax_enable_t38 is not also true.


lazedo:
@onnet thanks for the PR and sorry for late reply. this looks ok, i just need some time to test this.

noahmehl:
We're definitely a +1 for this, as it's broken for us.  Is there anything we can do to help with testing?

lazedo:
@onnet haven't has a chance to test all cases but as the code seems good to me, i'm merging it and testing it before 4.2 becomes stable

noahmehl:
@lazedo thank you so much :)

noahmehl:
@lazedo one other question, is it reasonable/possible to backport this to 4.1?

lazedo:
@noahmehl as mentioned above, since all the tests were not made, we won't backport it until we're sure it doesn't break any other use case.

arekfryz:
@lazedo I will be testing it on 4.1 tonight. Do you have special test cases? T38 is broken on 4.1

PR #4353, opened 2017-11-17
KAZOO-5744: fixes for webhook filtering and other purposes

jamesaimonetti:
Fix setting outbound call id for conference dial

Fix application vars showing up on bridge strings

### DIFF
--- a/applications/callflow/src/callflow.hrl
+++ b/applications/callflow/src/callflow.hrl
@@ -27,6 +27,8 @@
 -define(APP_VERSION, <<"4.0.0">> ).
 -define(CF_CONFIG_CAT, ?APP_NAME).
 
+-define(DEFAULT_CHILD_KEY, <<"_">>).
+
 -define(RECORDED_NAME_KEY, [<<"media">>, <<"name">>]).
 -define(CF_RECORDING_ID_KEY, <<"Recording-ID">>).
 
--- a/applications/callflow/src/cf_exe.erl
+++ b/applications/callflow/src/cf_exe.erl
@@ -689,8 +689,8 @@ launch_cf_module(#state{call=Call
                        ,flow=Flow
                        ,cf_module_pid=OldPidRef
                        }=State) ->
-    Module = <<"cf_", (kz_json:get_value(<<"module">>, Flow))/binary>>,
-    Data = kz_json:get_value(<<"data">>, Flow, kz_json:new()),
+    Module = <<"cf_", (kz_json:get_ne_binary_value(<<"module">>, Flow))/binary>>,
+    Data = kz_json:get_json_value(<<"data">>, Flow, kz_json:new()),
 
     {PidRef, Action} =
         case maybe_start_cf_module(Module, Data, Call) of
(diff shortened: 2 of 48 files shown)


applications/ecallmgr/src/ecallmgr.hrl
```diff
@@ -256,9 +256,9 @@
                               ,{<<"Media-Encryption-Enforce-Security">>,<<"sdp_secure_savp_only">>}
                               ,{<<"Media-Files-Separator">>, <<"playback_delimiter">>}
                               ,{<<"Media-Webrtc">>, <<"media_webrtc">>}
-                              ,{<<"Origination-Call-ID">>, <<"sip_origination_call_id">>}
+                              ,{<<"Origination-Call-ID">>, <<"origination_uuid">>}
```
lazedo: why this change ?

jamesaimonetti: @lazedo I can't find any mention of `sip_origination_call_id` in FreeSWITCH wiki nor did I see the asked-for call-id being set, but setting `origination_uuid` worked and is found in [Channel Variables](https://freeswitch.org/confluence/display/FREESWITCH/Channel+Variables)

lazedo: @jamesaimonetti its set by us so that we can bind to a know call-id from the originating process (fax worker uses it) for calls that have 2+ destinations (2+ /= uuids) and the first one fails we can still get events from the other legs.
ecallmgr_call_event sets `{<<"Origination-Call-ID">>, kzd_freeswitch:origination_call_id(Props)}` from that value and `kapi_call:publish_event` prefers it when setting the routing key.
i propose we leave `Origination-Call-ID` as-is and change fax_worker to send that instead of `Outbound-Call-ID` and have stepswitch forward `Origination-Call-ID` and  `Outbound-Call-ID` (with value = `Origination-Call-ID).

this can also be useful to track quickcall

applications/stepswitch/src/stepswitch_originate.erl
```diff
@@ -271,34 +271,35 @@ build_originate(#state{endpoints=Endpoints
...
       ,{<<"Ignore-Early-Media">>, kz_json:get_value(<<"Ignore-Early-Media">>, OffnetReq)}
+      ,{<<"Loopback-Bowout">>, kz_json:get_value(<<"Loopback-Bowout">>, OffnetReq)}
       ,{<<"Media">>, kz_json:get_value(<<"Media">>, OffnetReq)}
-      ,{<<"Hold-Media">>, kz_json:get_value(<<"Hold-Media">>, OffnetReq)}
-      ,{<<"Presence-ID">>, kz_json:get_value(<<"Presence-ID">>, OffnetReq)}
-      ,{<<"Outbound-Callee-ID-Number">>, kz_json:get_value(<<"Outbound-Callee-ID-Number">>, OffnetReq)}
+      ,{<<"Msg-ID">>, MsgId}
+      ,{<<"Originate-Immediate">>, kz_json:get_value(<<"Originate-Immediate">>, OffnetReq)}
+      ,{<<"Origination-Call-ID">>, kz_json:get_value(<<"Origination-Call-ID">>, OffnetReq)}
+      ,{<<"Outbound-Call-ID">>, kz_json:get_value(<<"Outbound-Call-ID">>, OffnetReq)}
```
lazedo: can we default `Outbound-Call-ID` to `Origination-Call-ID` ? we still want the originate to use the provided call_id. we can also do this in ecallmgr_originate.

PR #4358, opened 2017-11-20
 [Fix #kazoo-5636] Split kz_account_crawler tasks into task modules

harenson:
- Trigger this new tasks checks along doodle checks
- Fix dialyzer warnings

### DIFF
--- a/applications/doodle/src/doodle_app.erl
+++ b/applications/doodle/src/doodle_app.erl
@@ -10,6 +10,7 @@
 -behaviour(application).
 
 -include("doodle.hrl").
+-define(ACCOUNT_CRAWLER_BINDING, <<"tasks.account_crawler">>).
 
 -export([start/2, stop/1]).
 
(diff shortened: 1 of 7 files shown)


applications/doodle/src/doodle_app.erl
```diff
@@ -29,6 +29,10 @@ start(_Type, _Args) ->
         {'error', Err} -> lager:error("default sms is undefined and cannot read default from file: ~p", [Err]);
         JObj -> kapps_config:set(?CONFIG_CAT, <<"reschedule">>, JObj)
     end,
+    lager:debug("Start listening for tasks.account_crawler trigger"),
+    _ = kazoo_bindings:bind(<<"tasks.account_crawler">>,
+                            'doodle_maintenance',
+                            'start_check_sms_by_account'),
```
lazedo: @harenson you also need to `unbind` when app stops

applications/tasks/src/kz_account_crawler.erl
```diff
@@ -210,239 +201,13 @@ check_then_process_account(AccountId, {'ok', AccountJObj}) ->
...
-    kapps_notify_publisher:cast(Req, fun kapi_notifications:publish_low_balance/1),
-
-    update_account_low_balance_sent(AccountJObj).
-
--spec update_account_low_balance_sent(kz_account:doc()) -> 'ok'.
-update_account_low_balance_sent(AccountJObj0) ->
-    AccountJObj1 = kz_account:set_low_balance_sent(AccountJObj0),
-    AccountJObj2 = kz_account:set_low_balance_tstamp(AccountJObj1),
-    _ = kz_util:account_update(AccountJObj2),
+    _ = tasks_bindings:map(<<"tasks.account_crawler">>, [AccountId, AccountJObj]),
```
lazedo: since the modules binding to account crawler shouldn't depend on the execution of others, you can add `pmap` to `tasks_bindings` and use it here. check kazoo_bindings `pmap`

PR #4359, opened 2017-11-21
PISTON-497, PISTON-498: allow custom Event SIP NOTIFY to be sent via CrossBar

danielfinke:
https://forums.2600hz.com/forums/topic/9334-can-i-send-a-notify-using-crossbar/


danielfinke:
@lazedo OK, added the logging. Also fixed CircleCI

jamesaimonetti:
My feedback is I think the API should be more [like metaflows on /channels](https://docs.2600hz.com/dev/applications/crossbar/doc/channels/#put-a-feature-metaflow-on-a-channel) where you PUT `/devices/{DEVICE_ID}` `-d {"action":"notify", "data":{"event":"check-sync"}}`, if that makes sense. We should not be adding verbs to the URL.

danielfinke:
@jamesaimonetti that sort of seems weird to me, as it results in us overloading the behaviour of the API endpoint. The result would be a somewhat confusing API specification for that endpoint in which for example some fields are required or not depending on certain circumstances. Unlike /channels, there is an actual object model (in the database) associated with the endpoint. My stance on these matters is always that services, functions, etc should also be single-purpose and concise

jamesaimonetti:
@danielfinke I hear that but I consider HTTP Verb+Endpoint  to be the identifier of functionality (not sure how to phrase that better). So `PUT + /devices/{DEVICE_ID}` is different from `GET + /devices/{DEVICE_ID}` in my mind. See also [conference dial](https://docs.2600hz.com/dev/applications/crossbar/doc/conference/#dialing-an-endpoint) where we PUT an action onto a conference (existing or dynamic).

danielfinke:
@jamesaimonetti I see what you're trying to say, given PUT /device and PUT /device/<device_id> are different. What I think makes it most odd is that other methods on /device/<device_id> all deal with the model; GET fetches it, POST replaces it, PATCH modifies it, DELETE deletes it. So it seems somewhat inconsistent that the PUT operation be used for something unrelated, despite its consistency with other existing APIs (for which I feel the same).
If your (and your team's preference) is to go with PUT /devices/<device_id>, ultimately I'll make the change as I value compatibility between your and our implementations the most. Just want my thoughts shared on that

jamesaimonetti:
@danielfinke I'll bring it up on standup tomorrow morning and see what the consensus is (@lazedo and @k-anderson can weigh in here as well); I haven't discussed with them yet.

lazedo:
@danielfinke i think @jamesaimonetti  comment is correct and we're trying to pursuit the implementation of `actions` on a `thing` with the `PUT` verb instead of adding verbs to the url.

a `PUT` on a `collection` (devices, accounts, users, ...) creates the `thing` , the `PUT` on a `../things/thing` takes an action on the `thing`.
instead of having POST `../accounts/xxx/enable` we would have PUT `../accounts/xxx` {action: enable}
this also allows for extra parameters needed for the action to be sent in the data payload which supports complex types (json, arrays) hardly handled in url parameters

jamesaimonetti:
@danielfinke discussed and, despite @k-anderson agreeing with you in theory, in principal, for consistency, the consensus is to move this to `PUT /devices/{DEVICE_ID}` with `action=notify` on the root request JSON object, in line with how `/channels` is working. If it helps, you are PUTting a ephemeral feature on the config of a device :) It just happens to be added/removed without affecting the doc itself!

danielfinke:
@lazedo @jamesaimonetti works for me guys. :) I'll make an update

danielfinke:
@lazedo @jamesaimonetti OK, this feature has been moved into PUT /devices/<device_id>

jamesaimonetti:
@lazedo i'm fine with this; any outstanding comments for you before approval?

jamesaimonetti:
@danielfinke thanks for your patience! merged

PR #4361, opened 2017-11-21
KAZOO-5718: Set user info when sending port comment notification

icehess:


Makefile
```diff
@@ -248,7 +249,6 @@ validate-schemas:
 	@$(ROOT)/scripts/validate-schemas.sh $(ROOT)/applications/crossbar/priv/couchdb/schemas
 
 CHANGED := $(shell git --no-pager diff --name-only HEAD origin/master -- applications core scripts)
-TO_FMT := $(shell git --no-pager diff --name-only HEAD origin/master -- "*.erl" "*.hrl" "*.escript")
```
jamesaimonetti: Why remove this? By default, format the project Erlang files, unless overridden explicitly in the `make fmt` invocation.

icehess: I re-added `TO_FMT`.

PR #4368, opened 2017-11-30
Check spelling of forward in config

jamesaimonetti:


applications/crossbar/priv/couchdb/schemas/system_config.callflow.json
```diff
@@ -150,9 +150,13 @@
                     "description": "callflow save after notify",
                     "type": "boolean"
                 },
-                "vm_message_foraward_type": {
+                "vm_message_forward_type": {
                     "default": "only_forward",
-                    "description": "Enable or disable the ability to prepend a message when forwarding a voicemail message",
+                    "description": "callflow voicemail vm_message_forward_type",
```
icehess: Why changing the description?

jamesaimonetti: Oops, I didn't realize the old one had a "proper" description :)

jamesaimonetti:
My guess, re migration, is that no one has used this setting yet; otherwise they would have noticed the spelling. :) But probably better safe than sorry; I'll add migration.

jamesaimonetti:
@icehess have your migration!

PR #4371, opened 2017-11-30
KAZOO-5527: do not print annoying message if there is no errors

icehess:


core/kazoo_number_manager/src/kazoo_number_manager_maintenance.erl
```diff
@@ -137,13 +137,22 @@ convert_carrier_module_numbers(Nums, Target) ->
...
             Routines = [{fun knm_phone_number:set_module_name/2, Target}],
             #{ok := Ns, ko := KOs} = knm_numbers:update(Nums, Routines),
-            io:format("updated carrier module to ~s for ~p:\n", [Target, length(Ns)]),
-            F = fun (N) -> io:format("\t~s\n", [knm_phone_number:number(knm_number:phone_number(N))]) end,
-            lists:foreach(F, Ns),
-            io:format("updating carrier module failed for ~p:\n", [maps:size(KOs)]),
-            G = fun (Num, R) -> io:format("\t~s: ~p\n", [Num, R]) end,
-            _ = maps:map(G, KOs),
-            ok
+            case length(Ns) > 0 of
```
jamesaimonetti: Typically if you are testing for empty vs non-empty lists, use pattern matching instead of calling length:

```erlang
case Ns of
    [] -> 'ok';
    _ -> ...
end
```

jamesaimonetti: Better yet, call a function and have the clauses do the matching

PR #4372, opened 2017-12-01
Update conferences.md

extremerotary:
Specify that some schema elements are under development and not available yet.
Add additional information about adding a random filter at the end of a participant action to ensure it's viewed as a new request to crossbar.

### DIFF
--- a/applications/crossbar/doc/conference.md
+++ b/applications/crossbar/doc/conference.md
@@ -43,6 +43,12 @@ Key | Description | Type | Default | Required
 `play_name` | Do we need to announce new conference members? | `boolean()` | `false` | `false`
 `play_welcome` | Whether to play the welcome prompt | `boolean()` |   | `false`
 `profile` | The XML profile name used to configure the conference | `string()` |   | `false`
+
+
+
+#### Development not complete on the below ####
+Key | Description | Type | Default | Required
+--- | ----------- | ---- | ------- | --------
 `require_moderator` | does the conference require a moderator | `boolean()` |   | `false`
 `wait_for_moderator` | should members wait for a moderator before joining the conference | `boolean()` |   | `false`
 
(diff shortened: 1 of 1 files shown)


applications/crossbar/doc/conference.md
```diff
@@ -43,6 +43,12 @@ Key | Description | Type | Default | Required
 `play_name` | Do we need to announce new conference members? | `boolean()` | `false` | `false`
 `play_welcome` | Whether to play the welcome prompt | `boolean()` |   | `false`
 `profile` | The XML profile name used to configure the conference | `string()` |   | `false`
+
+
+
+#### Development not complete on the below ####
```
jamesaimonetti: This is not valid to do, as the scripts checking these schema tables are inserting the now missing fields from the main schema. This is what CircleCI is complaining about - the scripts added the rows back to the table above and now there are unstaged changes.

Instead, let's leave the rows in the table and add a `Note: ` below about the fields' status

applications/crossbar/doc/conference.md
```diff
@@ -247,6 +253,15 @@ curl -v -X PUT \
     http://{SERVER}:8000/v2/accounts/{ACCOUNT_ID}/conferences/{CONFERENCE_ID}/participants/{PARTICIPANT_ID}
 ```
 
+When performing actions on participants, it's best practice to add a random string to the end of the call to ensure the request is viewed as 'unique'. For example:
```
jamesaimonetti: I don't understand this best practice? At best, you're cache-busting the browser but there's nothing advantageous in doing this from an API perspective. What's the thought here?

extremerotary: Hey James,
Many times I'll get a HTTP/1.1 304 Not Modified for API calls to conferences for some reason in 4.0. I now add the random filter to make the request unique. Check out support ticket 29528

extremerotary:
Hey James,
I'm a newbie at this. I've updated the changes as a second commit. How do i resubmit the pull request? Do I open a new one? It's a little different than sourcetree that i'm used to.

k-anderson:
@jamesaimonetti Howdy, when you are back can you complete this review :)

jamesaimonetti:
Hi @extremerotary sorry for the delay. I've merged this into master. Thanks!

PR #4378, opened 2017-12-04
PISTON-506: keep Account-ID in webhooks-parking hook data

danielfinke:


applications/webhooks/src/modules/webhooks_parking.erl
```diff
@@ -81,8 +81,9 @@ maybe_send_event(AccountId, JObj) ->
 %% @doc
 %% @end
 %%--------------------------------------------------------------------
--spec format(kz_json:object()) -> kz_json:object().
-format(JObj) ->
+-spec format(kz_json:object(), ne_binary()) -> kz_json:object().
+format(JObj, AccountId) ->
+    JObj1 = kz_json:set_value(<<"Account-ID">>, AccountId, JObj),
```
lazedo: you can get the same without changing the function arity
AccountId = kz_json:get_value([<<"Custom-Channel-Vars">>, <<"Account-ID">>], JObj),

danielfinke:
@lazedo OK, kept format's function arity as the original

danielfinke:
@lazedo anything else required here?

jamesaimonetti:
Merged

PR #4385, opened 2017-12-07
[Fix #kazoo-5748] Fix send worker update check

harenson:
Also:
-  Mark job as partial when listing or dumping numbers and some of those numbers cannot be obtained
- Fix typo within kazoo/Makefile

### DIFF
--- a/applications/tasks/src/modules/kt_numbers.erl
+++ b/applications/tasks/src/modules/kt_numbers.erl
@@ -47,13 +47,15 @@
 -define(MOD_CAT, <<(?CONFIG_CAT)/binary, ".numbers">>).
 -define(IMPORT_DEFAULTS_TO_CARRIER
         %% Defaults to knm_carriers:default_carrier()'s default value
-       ,kapps_config:get_binary(?MOD_CAT, <<"import_defaults_to_carrier">>, ?CARRIER_LOCAL)).
+       ,kapps_config:get_binary(?MOD_CAT, <<"import_defaults_to_carrier">>, ?CARRIER_LOCAL)
+       ).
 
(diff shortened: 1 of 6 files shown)


icehess:
@harenson you need to rebase

applications/tasks/src/modules/kt_numbers.erl
```diff
@@ -768,9 +768,9 @@ list_assigned_to(AuthBy, [{Next,NumberDb}|Rest]) ->
     ViewOptions = [{limit,?DB_DUMP_BULK_SIZE} | view_for_list_assigned(Next)],
     case kz_datamgr:get_result_keys(NumberDb, <<"numbers/assigned_to">>, ViewOptions) of
         {ok, []} -> {ok, Rest};
-        {error, _R} ->
-            lager:error("could not get ~p's numbers in ~s: ~p", [ViewOptions, NumberDb, _R]),
-            {ok, Rest};
+        {error, R} ->
```
icehess: We usually prefix variable with `_` if we only are using them for log.

core/kazoo_number_manager/src/knm_phone_number.erl
```diff
@@ -1611,7 +1611,7 @@ sanitize_public_fields(JObj) ->
                    (knm_phone_number()) -> boolean().
 is_authorized(T) when is_map(T) -> is_authorized_collection(T);
 is_authorized(#knm_phone_number{auth_by = ?KNM_DEFAULT_AUTH_BY}) ->
-    lager:info("bypassing auth"),
+    lager:debug("bypassing auth"),
```
icehess: Why decreasing the log level here?

PR #4386, opened 2017-12-07
PISTON-492: find a valid to_user for presence event so ecallmgr_fs_presence doesn't crash

danielfinke:
https://forums.2600hz.com/forums/topic/9453-blf-when-using-fast-pickup

[log-snippet.txt](https://github.com/2600hz/kazoo/files/1540326/log-snippet.txt)

[culprit code](https://github.com/2600hz/kazoo/blob/master/applications/ecallmgr/src/ecallmgr_fs_presence.erl#L236)

Attached is a snippet that shows the presence props when picking up a KFP. Propose having a backup of "Other-Leg-Destination-Number" to prevent the crash. It matches the "Caller-Destination-Number" of the presence props for the other channel's dialog (parked channel).

### DIFF
--- a/applications/ecallmgr/src/ecallmgr_fs_presence.erl
+++ b/applications/ecallmgr/src/ecallmgr_fs_presence.erl
@@ -234,6 +234,7 @@ to_user(Props) ->
 
 to_user(<<"initiator">>, Props) ->
     props:get_first_defined([<<"Caller-Destination-Number">>
+                            ,<<"Other-Leg-Destination-Number">>
                             ,<<"variable_sip_to_user">>
                             ], Props);
 to_user(<<"recipient">>, Props) ->


lazedo:
@danielfinke this doesn't seem right.
the `initiator` shouldn't get the value from `<<"Other-Leg-Destination-Number">>`

also, you're not adding a backup, you're putting it before `variable_sip_to_user`.

but..., have you tried setting verbose-events in switch.conf.xml ? or maybe you are sending an originate that needs variables set ?

danielfinke:
@lazedo may I ask why the initiator shouldn't get the value from Other-Leg-Destination-Number if there is no alternative?
Also I am not sending an originate, the INVITE is coming from the phone, as a result of pressing a BLF key with a KFP cookie

lazedo:
`if there is no alternative?` what do you mean ? did you set `verbose-channels-events` as recommended ? variable_sip_to should be set in that case

danielfinke:
@lazedo I think I misunderstood the behaviour of `verbose-channel-events`. So, if enabling that flag in the FS config is required for proper behaviour, should I/we make a change in [https://github.com/2600hz/kazoo-configs-freeswitch](https://github.com/2600hz/kazoo-configs-freeswitch) to default it on?

lazedo:
@danielfinke we created a pr for freeswitch to avoid that. https://freeswitch.org/stash/projects/FS/repos/freeswitch/pull-requests/1428/overview

danielfinke:
@lazedo looks like I can't see the PR. Is it from your repo here: [https://freeswitch.org/stash/users/lazedo/repos/freeswitch/browse](https://freeswitch.org/stash/users/lazedo/repos/freeswitch/browse)?
Which branch?

jamesaimonetti:
@danielfinke @lazedo would this discussion be better suited to a forum post at this point (and close this PR)?

lazedo:
yes, this should be closed

PR #4391, opened 2017-12-11
just an attempt to make crossbar reply with xml

onnet:


jamesaimonetti:
@onnet which endpoints actually support XML responses? 406 error when trying with `/v2/accounts/{ACCOUNT_ID}/devices`.

onnet:
Hi @jamesaimonetti 
none in kazoo repo yet, but probably will appear in the future if such a possibility will be provided. 
As I wrote in forums, I need to implement Yandex.Money payment gateway crossbar module for Kazoo based billing. 
Yandex.Money expects reply in XML to its POST requests. So, I'm setting content types provided to XML and with this patch can reply with XML to Yandex.Money.

Here is an example - yandex.Money crossbar module (just an alpha - Yandex understands responses, transactions saved to Kazoo modb):
https://bitbucket.org/kirillsysoev/onbill/src/ddfcf53549e48b8c3a93866fa5a8a4f580b79ccb/src/crossbar/cb_yandex_money.erl?at=master&fileviewer=file-view-default#cb_yandex_money.erl-89:90

Regards,

jamesaimonetti:
@onnet ok. hmm. I'm not sure what to do at the moment since I can't test this PR without adding more code to an endpoint (or better, making a generic to_xml function like we have for to_csv). Want to add that? Then people can receive XML for any endpoint if they choose...

onnet:
@jamesaimonetti got you. Can't imagine why would I need devices in XML, but will check if I can do that. :))

applications/crossbar/src/api_resource.erl
```diff
@@ -796,6 +797,29 @@ to_json(Req, Context, Accept) ->
...
+to_xml(Req, Context) ->
+    lager:debug("run: to_xml"),
+    [{Mod, _Params}|_] = cb_context:req_nouns(Context),
+    Verb = cb_context:req_verb(Context),
+    Event = api_util:create_event_name(Context, [<<"to_xml">>
+                                                ,kz_term:to_lower_binary(Verb)
+                                                ,Mod
+                                                ]),
+    {Req1, Context1} = crossbar_bindings:fold(Event, {Req, Context}),
+    case cb_context:fetch(Context1, 'is_chunked') of
```
icehess: This won't work anymore. In 4.2+ we added a generic chunked response to crossbar. You can look at `to_json/2` or `to_csv/2` as example.

Basically you have to use `crossbar_view` to get view results,. If the request wants chunked response `crossbar_view` stores the necessary values in `Context` and then in `api_resource:to_xml/2`, if `is_chunk` is set, it should call `to_chunk/3` which then starts getting results from view and send response to client in chunk.

You simply have to write appropriate function clause for `send_chunk_response/3` and `finish_chunked_response/1` to generate XML from JSON.

jamesaimonetti:
@onnet I just meant we should have a generic json->XML similar to how we have a json->csv if CSV is requested. In addition to @icehess's comments about using crossbar_view.

jamesaimonetti:
Hi @onnet, had a chance to look at a generic XML responder and using crossbar_view?

onnet:
@jamesaimonetti Hi! Not yet. Postponed it maybe till January since I don't play with 4.2+ yet

PR #4392, opened 2017-12-12
fix typos

katrinleinweber:
see diffs :-)


jamesaimonetti:
@katrinleinweber thanks for the fixes! I added them to `scripts/misspellings.txt` for future reference as well.

katrinleinweber:
Neat! I used [something](https://github.com/katrinleinweber/PhD-thesis/blob/doctor/makefile#L97) [similar](https://github.com/katrinleinweber/PhD-thesis/blob/doctor/regular-expressions.md) once :-)

PR #4393, opened 2017-12-12
Fix Content-Length header for all crossbar endpoints

kalda341:
The Content-Length header that is manually set in many Crossbar endpoints is incorrect when compression is used (most of the time).
This leaves many clients waiting for more data that never arrives, and causes many issues.
When we do not set the Content-Length header, Cowboy correctly computes the length, and sets the header for us.

If possible it would be fantastic to get this fix in 4.1 also. I'm happy to submit another pull request when this is approved.


jamesaimonetti:
@kalda341 thanks. Can you fix the compilation errors so it builds cleanly?

jamesaimonetti:
@kalda341 thanks, i fixed a couple spelling issues (unrelated) and merged

PR #4394, opened 2017-12-12
KAZOO-5695: increase key press read timeout in call forward

icehess:


core/kazoo_endpoint/src/kz_endpoint.erl
```diff
@@ -1653,7 +1653,8 @@ maybe_set_confirm_properties({Endpoint, Call, CallFwd, CCVs}=Acc) ->
         'true' ->
             lager:info("call forwarding configured to require key press"),
             Confirm = [{<<"Confirm-Key">>, <<"1">>}
-                      ,{<<"Confirm-Cancel-Timeout">>, <<"2">>}
+                      ,{<<"Confirm-Cancel-Timeout">>, 'true'}
+                      ,{<<"Confirm-Read-Timeout">>, kz_term:to_binary(7 * ?MILLISECONDS_IN_SECOND)}
```
lazedo: @icehess is freeswitch interpreting the `group_confirm_read_timeout` value as milliseconds ?

icehess: yes, it's in milliseconds, default value is 5000.

applications/ecallmgr/src/ecallmgr_fs_xml.erl
```diff
@@ -546,7 +546,6 @@ get_channel_vars({<<"Timeout">>, V}, Vars) ->
     case kz_term:to_integer(V) of
         TO when TO > 0 ->
             [<<"call_timeout=", (kz_term:to_binary(TO))/binary>>
-            ,<<"originate_timeout=", (kz_term:to_binary(TO))/binary>>
```
lazedo: @icehess why do you want to remove this ? isn't this going to affect the originate command timeout (not the ringing time) ?

icehess: hmm, I'm trying to cancel `leg_timeout` in group confirm.

PR #4396, opened 2017-12-13
HELP-33420 keep a record of local recording requests - 4.1

lazedo:
shared queue should be enough here, but we may get duplicated events


lazedo:
master https://github.com/2600hz/kazoo/pull/4397

lazedo:
DO NOT MERGE

PR #4397, opened 2017-12-13
HELP-33420 Keep a record of local recording requests

lazedo:
shared queue should be enough here, but we may get duplicated events


lazedo:
4.1 https://github.com/2600hz/kazoo/pull/4396

lazedo:
DO NOT MERGE

PR #4398, opened 2017-12-14
HELP-33420 fix recording events duplication

lazedo:


lazedo:
alternative implementation, needs testing

k-anderson:
@lazedo this did not resolve the issue, I don't think the issue was/only duplication in a single ecallmgr.  Multiple ecallmgrs receive the stop event so even if it isn't duplicated internally it is by having multiple services.

Another sidenote, looking at the logs there is an extra '.' in the file name:

kz_http:233 (<0.32244.1415>) 23.65ms: put https://127.0.0.1:2000/call_recording_REDACTED..mp3?REDACTED=REDACTED

PR #4399, opened 2017-12-15
PISTON-514: do not use ANI for outbound get_sip_to as it contains caller details

danielfinke:
https://forums.2600hz.com/forums/topic/9465-41-incorrect-to-in-cdr-outbound-call/


jamesaimonetti:
@danielfinke thanks, merged.

PR #4400, opened 2017-12-16
Logging improvements 4.1

mark2600:


core/kazoo/src/kz_util.erl
```diff
@@ -640,10 +640,10 @@ try_load_module(Name) ->
 -spec put_callid(kz_json:object() | kz_proplist() | ne_binary() | atom()) ->
                         api_binary().
 put_callid(?NE_BINARY = CallId) ->
-    lager:md([{'callid', CallId}]),
+    lager:md([{'callid', CallId}]++lager:md()),
```
lazedo: found this yesterday, we're in sync :)

rel/sys.config
```diff
@@ -7,15 +7,15 @@
         ,{lager_file_backend, [
             {file, "log/error.log"}, {level, error}, {size, 10485760}, {date, "$D0"}, {count, 5}
             ,{formatter, lager_default_formatter}
-            ,{formatter_config, [time," [",severity,"] ", "|", {callid, <<"0000000000">>}, "|", module, ":", line, " (",pid, ") ", message, "\n"]}
+            ,{formatter_config, [time," [",severity,"] ", "|", {log_id, {callid, <<"0000000000">>}}, "|", module, ":", line, " (",pid, ") ", message, "\n"]}
```
lazedo: should it be `msg_id`instead of `log_id` ?

lazedo: and why not both ? callid and msg_id ?

PR #4404, opened 2017-12-19
set start_key in build_options funs to avoid crash in get_result

icehess:


applications/crossbar/src/crossbar_view.erl
```diff
@@ -234,6 +235,7 @@ build_load_range_params(Context, View, Options) ->
                                      ,start_time => StartTime
                                      ,view_options => build_view_query(Options, Direction, StartKey, EndKey, HasQSFilter)
                                      ,has_qs_filter => HasQSFilter
+                                     ,start_key => StartKey
```
lazedo: @icehess  not related but, can you fix the duplicated `has_qs_filter` assignment ?

PR #4407, opened 2017-12-20
Fix distinctive ring on ring groups

kalda341:


jamesaimonetti:
Hi @kalda341, I've merged this in but made one change - instead of using custom_channel_vars for storage, I used the kvs storage in the kapps_call record (kvs_store/kvs_fetch). Using CCVs will put the Override-Ringtone on the channel vars which seems noisy and unnecessary.

Can you backport to 4.1 - we think this constitutes a bug patch.

Thanks!

PR #4408, opened 2017-12-20
Fix Content-Length header for all crossbar endpoints (port #4393 to 4.1)

kalda341:
Fix compilation error

fix spelling


jamesaimonetti:
@kalda341 thanks for the backport!

PR #4410, opened 2017-12-20
don't set the default callid, it provides no value

mark2600:


k-anderson:
I reverted the prior change for now, we can continue the discussion on the standup tomorrow if you feel strongly about this ...

PR #4411, opened 2017-12-20
Fix logging again master

mark2600:


k-anderson:
I reverted the commit for now, we can continue the discussion on the standup

PR #4413, opened 2017-12-21
Fix distinctive ring on ring groups (backport #4407 to 4.1)

kalda341:


jamesaimonetti:
@kalda341 Thanks! merged

PR #4414, opened 2017-12-21
Resolve application dependent applications

jamesaimonetti:
Since `core` is a mess as far as inter-app dependencies, I've broken out the app dependencies of `applications` so we can at least be confident we're getting the right dependent applications in support of the app. This also identified some inter-app dependencies between apps in `applications` which should not be the case.

I've moved cccp-related crossbar endpoints to the cccp app (@onnet take note) as well as cb_access_lists to the frontier app as they are tightly coupled to each other. Since the crossbar modules required the related app to be on the same node already, the change should not be noticeable.

There are a couple more inter-app dependencies to resolve but this is a good step forward.


onnet:
@jamesaimonetti thanks for info!

PR #4417, opened 2017-12-22
KAZOO-5787: adding metadata to storage URL

jamesaimonetti:


lazedo:
@jamesaimonetti the comment above can be done in another PR, kz_endpoint_recording needs the same `Ext`  fix. and if you agree, we could move both recording from kazoo_call & kazoo_endpoint to a kazoo_recording core library as it seems we have duplicated code that could be moved to a util module.
thoughts ?

PR #4419, opened 2017-12-25
Added account_name to webhook meta data

Prots:


jamesaimonetti:
@Prots Hi! I don't think `account_name` is worth adding to all event payloads. What I'd like to see is `format_event` call two functions - `base_event_fields` and `specific_event_fields`. The base fields being what's there now and `specific_event_fields` taking the object type and adding fields as appropriate.

Does that make sense? Is that something you'd like to add to this PR?

PR #4424, opened 2017-12-27
Add Web-hook for receiving notification events

icehess:


applications/crossbar/src/modules/cb_webhooks.erl
```diff
@@ -86,12 +86,20 @@ maybe_revise_schema(MasterDb, SchemaJObj) ->
         {'error', _E} ->
             lager:warning("failed to find registered webhooks: ~p", [_E]);
         {'ok', Hooks} ->
-            revise_schema(SchemaJObj, [kz_json:get_value(<<"key">>, Hook) || Hook <- Hooks])
+            ToRemovHooks = [<<"callflow">>
```
jamesaimonetti: Not a big deal, but the variable should be `ToRemoveHooks` (missing the second e)

applications/webhooks/src/webhooks_util.erl
```diff
@@ -628,22 +727,22 @@ available_events() ->
 
 -spec fetch_available_events() -> ne_binaries().
 fetch_available_events() ->
-    {'ok', MasterAccountDb} = kapps_util:get_master_account_db(),
-    View = <<"webhooks/webhook_meta_listing">>,
-    case kz_datamgr:get_all_results(MasterAccountDb, View) of
+    MasterAccountDb = get_master_account_db(),
```
jamesaimonetti: I don't see the advantage of this change. If the master account can't be fetched why should this work proceed?

core/kazoo_amqp/src/api/kz_api.erl
```diff
@@ -63,6 +63,21 @@
         ,validate_message/4
         ]).
 
+%% #kapi_definition accessors
```
jamesaimonetti: I would think a kapi_definition module would be a good place to put this stuff (and the record definition).

core/kazoo_amqp/src/api/kapi_notifications.erl
```diff
@@ -12,486 +12,274 @@
...
+
+
+%% Notify Update
+notify_update_definition() ->
+    #kapi_definition{name = <<"notify_update">>
+                    ,friendly_name = <<"Notify Status Update">>
+                    ,description = <<"This event is triggered by notification consumer, e.g. teletype application, to send status of notification being process to publisher">>
+                    ,build_fun = fun notify_update/1
+                    ,validate_fun = fun notify_update_v/1
+                    ,publish_fun = fun publish_notify_update/2
```
jamesaimonetti: Spec'd as being an arity-1 function. Update spec to account for arity-2 functions, or ?

applications/webhooks/src/webhooks_util.erl
icehess: It won't proceed, `kz_datamgr:get_results/2` would return `{error,invalid_db_name}` because master account db is `undefined`

jamesaimonetti: But it will return `[]` instead of crashing, right?

core/kazoo_amqp/src/api/kapi_notifications.erl
icehess: a quick look at all API publish functions, we have so many functions with different arity (range from 0 to 6) with different args order. So I changed it to just have_`fun()` now.

applications/webhooks/src/modules/webhooks_notifications.erl
```diff
@@ -0,0 +1,149 @@
...
+-spec init(kapi_definition:apis(), kz_json:objects()) -> 'ok'.
+init([], Acc) ->
+    _ = webhooks_util:init_metadata(?ID, ?METADATA(Acc)),
+    'ok';
+init([EventDefinition|Rest], Acc) ->
+    Description = kz_json:from_list([{<<"friendly_name">>, kapi_definition:friendly_name(EventDefinition)}
+                                    ,{<<"description">>, kapi_definition:description(EventDefinition)}
+                                    ]
+                                   ),
+    init(Rest, [{kapi_definition:name(EventDefinition), Description} | Acc]).
```
jamesaimonetti: The spec says the second arg is `kz_json:objects()` but this looks like `[{ne_binary(), kz_json:object()}]`

icehess: fixed

PR #4425, opened 2017-12-27
Inherit voicemail keys (and fix them at the same time)

kalda341:


jamesaimonetti:
@kalda341 this looks good with the exception of the schemas. You've only included the base generated schemas and obviously there are many sub-keys defined and the existing schema itself is insufficient (no type, for instance). Can you add them manually (faster than adjusting the kazoo_ast stuff to figure it out, probably) so people have a clearer picture of what's stored in those docs?

kalda341:
@jamesaimonetti Done! Will update the 4.1 PR if you're happy with what I've done.

jamesaimonetti:
@kalda341 so close! Can you actually put the changes in the system_config/account_config docs? The `make apis` will then pull those in and update the swagger.json file

kalda341:
@jamesaimonetti I put the schema in a separate file which is referenced by the system_config/account_config json files to avoid duplicating all the fields. I've tested invalid API calls and this appears to work as intended.

It has automatically updated the swagger.json file, and the references appear to be correct, but I am happy to move this into both the system_config and account_config json files if you prefer. I just hate the duplication.
Or perhaps I'm misunderstanding you?

jamesaimonetti:
@kalda341 nope, you're right. I saw the swagger update and didn't see the "$ref" updates in account/system config json. My bad.

jamesaimonetti:
merged!

PR #4430, opened 2017-12-28
migrate system_config doc to another named doc

jamesaimonetti:


core/kazoo_config/test/kapps_config_migrate_test.erl
```diff
@@ -0,0 +1,390 @@
+-module(kapps_config_migrate_test).
+
+-include_lib("eunit/include/eunit.hrl").
+
+-define(WHAPPS_CONTROLLER
+       ,<<"{
```
icehess: why not storing it as a fixture file and load it?

jamesaimonetti: Is that on 4.1?

jamesaimonetti:
@k-anderson or @lazedo should we delete the whapps_controller (if found) once migrated? That's my only outstanding question for this PR

core/kazoo_config/test/kapps_config_migrate_test.erl
icehess: I didn't mean FixtureDB, just save it as a JSON file in `test` directory and read/decode it in your test function.

PR #4435, opened 2017-12-29
Fix incorrect account realm used when an inbound offnet call is blind…

kalda341:
… transferred to an offnet destination

### DIFF
--- a/applications/stepswitch/src/stepswitch_util.erl
+++ b/applications/stepswitch/src/stepswitch_util.erl
@@ -301,7 +301,10 @@ endpoint_format_from(Endpoint, Number, OffnetReq, CCVs) ->
 get_endpoint_format_from(OffnetReq, CCVs) ->
     DefaultRealm = default_realm(OffnetReq),
     case kz_json:is_true(<<"From-Account-Realm">>, CCVs) of
-        'true' -> DefaultRealm;
+        'true' ->
+            %% DefaultRealm isn't the account's realm for a blind transfer of an inbound offnet call
+            {'ok', AccountJObj} = kz_account:fetch(kz_json:get_ne_binary_value(<<"Account-ID">>, OffnetReq)),
+            kz_account:realm(AccountJObj);
         'false' -> kz_json:get_value(<<"From-URI-Realm">>, CCVs, DefaultRealm)
     end.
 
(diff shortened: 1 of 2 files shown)


applications/stepswitch/src/stepswitch_util.erl
```diff
@@ -301,7 +301,10 @@ endpoint_format_from(Endpoint, Number, OffnetReq, CCVs) ->
 get_endpoint_format_from(OffnetReq, CCVs) ->
     DefaultRealm = default_realm(OffnetReq),
     case kz_json:is_true(<<"From-Account-Realm">>, CCVs) of
-        'true' -> DefaultRealm;
+        'true' ->
+            %% DefaultRealm isn't the account's realm for a blind transfer of an inbound offnet call
+            {'ok', AccountJObj} = kz_account:fetch(kz_json:get_ne_binary_value(<<"Account-ID">>, OffnetReq)),
```
k-anderson: @kalda341 Howdy, rather than doing a DB lookup would you be able to use:

kapi_offnet_resource:account_realm(OffnetReq)?

Also, what is the call scenario that this is correcting?  We want to make sure that there isn't a more fundamental issue where the From realm was not properly set earlier in the call :)

kalda341: This won't work as the account realm in the offnet request is incorrect.
I suspect there is a more fundamental issue that has been around for a long time (it was also an issue in 3.22).

Our carrier requires the realm to be set to the account's realm, which we achieve by setting "from_account_realm" to true on the gateway. This however doesn't work in this particular scenario.
This hack sets the realm used as the from on the INVITE when an inbound offnet call is blind transferred to an offnet destination to the account realm.

k-anderson: @kalda341 Got it, ya we believe this is the real issue:
https://github.com/2600hz/kazoo/blob/4.1/applications/callflow/src/module/cf_resources.erl#L83

That should not be from_realm but should actually be set to the accounts realm.  If I am not mistaken the accounts realm should be present in the kapps_call or as a CCV.  We think that would be a more appropriate correction, in addition to removing the need for the DB lookup.

Could you try changing it there?

kalda341:
@jamesaimonetti Cool - that seems to work. My fix felt like a hack, but I wasn't sure what the proper fix was.
I'll update the 4.1 PR if you're happy with this.

lazedo:
we're happy @kalda341 . tks

PR #4436, opened 2017-12-29
Fix incorrect account realm used when an inbound offnet call is blind transferred to an offnet destination (backport #4435 to 4.1)

kalda341:


k-anderson:
See comments on 4435

kalda341:
@k-anderson I have updated to match #4435

PR #4437, opened 2017-12-29
handle fetch configuration for new mod_kazoo

lazedo:


applications/ecallmgr/src/ecallmgr_fs_config.erl
```diff
@@ -119,6 +119,16 @@ handle_info({'fetch', 'configuration', <<"configuration">>, <<"name">>, Conf, ID
     lager:debug("fetch configuration request from ~s: ~s", [Node, ID]),
     _ = kz_util:spawn(fun handle_config_req/4, [Node, ID, Conf, Data]),
     {'noreply', State};
+handle_info({'fetch', 'configuration', <<"configuration">>, <<"name">>, Conf, ID, [_UUID | Data]}, #state{node=Node}=State)
+  when is_binary(_UUID) ->
```
jamesaimonetti: @lazedo typically using `_Var` is used when only logging the value. Could you do `[?NE_BINARY | Data]` instead?

PR #4438, opened 2017-12-30
To Cowboy 2.2.0!

jamesaimonetti:


lazedo:
@jamesaimonetti this breaks (for me) the existing `php-sdk`. can you try to run `php-sdk` tests ?

lazedo:
if we change the SDK.php headers option to include `content-type` (currently `content_type`) it works.
but, the problem is on crossbar side as this previously was working.  it returns `HTTP/1.1 415 Unsupported Media Type` because the default CTA is now is `[]`

jamesaimonetti:
@lazedo i explained elsewhere, but cowboy now [requires the content-type header](https://github.com/ninenines/cowboy/blob/master/doc/src/manual/cowboy_rest.asciidoc#content_types_accepted) to exist on the request when calling content_types_accepted.
I've added a patch to add the content-type header into the request headers if its missing on the cowboy_req:req().

PR #4439, opened 2018-01-02
Make ordering

jamesaimonetti:
Orders the core/application folders for compiling. 

Changes to use `JOBS` in top level Makefile instead of setting/overriding `MAKEFLAGS` as the `-j` was problematic. 

Quieter compiling too.


efine:
I'm going to try this commit from my baggage-claim branch and then I'll review, ok?

efine:
Sadly, the build still breaks with the default setting of `JOBS`, so whatever dependency changes were done didn't fix that issue.

PR #4442, opened 2018-01-04
KAZOO-5785: allow specifying device JSON in conference dial

jamesaimonetti:


core/kazoo_ast/src/cb_api_endpoints.erl
```diff
@@ -253,13 +253,12 @@ process_schema(Filename, Definitions) ->
...
-                    false -> KV;
-                    true -> {Path, maybe_fix_ref(V)}
+                    'false' -> KV;
+                    'true' -> {Path, maybe_fix_ref(V)}
                 end
                 || {Path, V}=KV <- kz_json:to_proplist(kz_json:flatten(JObj0)),
                    not lists:member(<<"patternProperties">>, Path),
-                   not lists:member(<<"kazoo-validation">>, Path),
-                   not lists:member(<<"oneOf">>, Path)
+                   not lists:member(<<"kazoo-validation">>, Path)
```
lazedo: since you're changing this, can you add `not lists:member(<<"kazoo-", _/binary>>, Path)` instead ?

jamesaimonetti: hehe, that's not exactly valid but I get what you're asking :)

PR #4443, opened 2018-01-04
Change Erlang version to 19.3.6.5

efine:
This can be set up in the build environment as follows:

1. Add `KERL_BUILD_BACKEND=git` to `.kerlrc`.
2. Build and install the indicated Erlang version in the build
   environment prior to building `kazoo-core`.


k-anderson:
Requires https://github.com/2600hz/baggage-handler/pull/1 and build system agents to be rebuilt before this can be accepted.

PR #4448, opened 2018-01-04
remove defaults from voicemail key schema

jamesaimonetti:
on further discussion, it was deemed good to not have defaults because
accounts overriding individual keys would get the defaults from the
schema added on validation. if the global values were to be changed,
these accounts would no longer use the global values of keys they
didn't override.

Without defaults, the account config doc will reflect only the keys
overridden by the account and all other keys will be pulled from the
reseller/global configs.


kalda341:
Makes sense!

PR #4455, opened 2018-01-06
spawn schema update

lazedo:


lazedo:
DO NOT MERGE

applications/crossbar/src/crossbar_maintenance.erl
```diff
@@ -1185,9 +1185,17 @@ set_app_screenshots(AppId, PathToScreenshotsFolder) ->
              ],
     update_screenshots(AppId, MA, SShots).
 
+-spec update_schemas() -> 'ok'.
+update_schemas() ->
+    kz_datamgr:suppress_change_notice(),
```
k-anderson: @lazedo you should not suppress change notices here, what about running systems the cache would not be flushed...

lazedo: @k-anderson i missed a line to published a db change, will make update

PR #4456, opened 2018-01-07
remove abs on rollup update

onnet:
Looks like there is no need in abs/1 over here, otherwise there will be always credit type of transaction..


k-anderson:
@onnet Howdy, in 4.3 we have refactor the rollups, and this no longer applies. :)

PR #4460, opened 2018-01-08
HELP-34176: from_iso_week fails some years due to incorrect first ISO…

k-anderson:
… 8601 week calculation

PR #4463, opened 2018-01-08
Removed special calendar arithmetics for years starting on Monday

fcastellano:
If I understand correctly, iso_week_to_gregorian_date({Y,W}) should
return the starting day (Monday) of the ISO week; but when the year Y
starts on Monday, it computes the starting day of the _next_ week.

The change could be terribly broken if its purpose is not what I
guessed.

Sometimes you can get unexpected results, e.g.: iso_week_to_gregorian_date({2015,1}) = {2014,12,29}.
But probably is better than iso_week_to_gregorian_date({2018,1}) = {2018,1,8}

### DIFF
--- a/core/kazoo_stdlib/src/kz_date.erl
+++ b/core/kazoo_stdlib/src/kz_date.erl
@@ -52,12 +52,7 @@ from_gregorian_seconds(Seconds, <<_/binary>>=TZ) when is_integer(Seconds) ->
 from_iso_week({Year, Week}) ->
     Jan1 = calendar:date_to_gregorian_days(Year, 1, 1),
     Offset = 4 - calendar:day_of_the_week(Year, 1, 4),
-    Days =
-        case Offset =:= 0 of
-            'true' -> Jan1 + ( Week * 7 );
-            'false' ->
-                Jan1 + Offset + ( ( Week - 1 ) * 7 )
-        end,
+    Days = Jan1 + Offset + ( ( Week - 1 ) * 7 ),
     calendar:gregorian_days_to_date(Days).
 
 %%--------------------------------------------------------------------


lazedo:
@fcastellano thanks for your contribution. did you check this https://github.com/2600hz/kazoo/pull/4460

sergey-safarov:
Tested on my test host, PR resolves current issue.

lazedo:
@fcastellano thanks for looking into this. please check latest master.

fcastellano:
Thank you @lazedo. I'm checking it. I 'm trying to figure out why this PR tests failed, but the changes in HELP-34176-master seem to solve the issue we got. Actually, our case test would have been:

`,?_assertEqual({2018,1,8}, cf_temporal_route:next_rule_date(#rule{cycle = <<"weekly">>, interval=1, wdays=[<<"monday">>,<<"tuesday">>,<<"wednesday">>,<<"thursday">>,<<"friday">>], start_date={1983,4,11}}, {2018,1,7}))
`

k-anderson:
@fcastellano Howdy and thank you! So the one test failure that you have I also found when I worked on this issue.  Using outlook to verify (by setting up a recurring meeting with the same parameters) I found that the test is incorrect.  The value provided is correct and the expected does not match the events outlook created.  I think the test must have been created to match the code and it was influenced by this bug.

At this time my original PR has been accepted primarily because it was the first and already had confirmation from multiple sources that it corrected the issue.  The reason I wrote it that way was because I was re-implementing an existing known working algorithm to make sure this won't be an issue in the future.

I really appreciate your contribution here, specially in a timely manor for this but I am going to close this PR.  If you would like to have a discussion around using your approach I am happy to do so, like I say this might be a cleaner valid approach but the other is proven.

fcastellano:
Dear @lazedo , I finished my review. As I understand, my PR failed a test, but probably it was a test fault (at least, I see @k-anderson changed it). Indeed I don't know if, when one has a weekly temporal rule with a start_day on Sunday, the _active_ week starts from the previous Monday, or the next Monday.

Anyway, as I understand, the master change is actually equivalent to mine. The only point one can consider is that this function could be probably dropped:
```
-spec weekday_distance(1..7, 1..7) -> 1..7.
weekday_distance(D0, D1) when D0 =< 7, D1 =< 7 ->
    case D0 - D1 of
        Days when Days =< 7 -> Days;
        Days -> Days + 7
    end.
```

fcastellano:
Hi @k-anderson , I missed your kind comment, before my final note. I totally agree. Thank you for your hard and great work (not just on this issue, obviously).

PR #4474, opened 2018-01-10
Fix proxy for recordings

jamesaimonetti:
Also addresses some issues when dialyzing from an app (vs the project vs CI).

Refactors the proxy handler to have more useful logging as well (around errors mostly).


jamesaimonetti:
Incorporates changes #4473 as well.

PR #4479, opened 2018-01-12
make easy to add attachment to fixturedb

icehess:
Allow to call fixturedb functions utility without starting without need
to start kazoo_data and config.

Also enhance add attachment to fixturedb by creating sym linking att
file to the media_files/* if you're using files from media_files

(I need to update the document, maybe tomorrow :)

### DIFF
--- a/core/kazoo_fixturedb/doc/README.md
+++ b/core/kazoo_fixturedb/doc/README.md
@@ -1,18 +1,18 @@
 # Kazoo FixtureDB
 
-What can you say about a data application? That it loves to be distributed and fault tolerant? That it listens to data streams and not understand a bit? yeah, that.
+What can you say about a data application? That it loves to be distributed and fault tolerant? That it listens to data streams and not understand a bit?
 
 ## Overview
 
-General speaking, FixtureDB removes dependency of Kazoo Data Manager for AMQP and Kazoo Cache and removes the burden of isolating source codes with `-ifdef(TEST)` when it comes to reading from database, which in parts result in a tiny little bit more coverage (if any).
+General speaking, FixtureDB removes dependency of Kazoo Data Manager to CouchDB and RabbitMQ for running EUnit tests and removes the burden of isolating source codes with `-ifdef(TEST)` when it comes to reading from database, which in part result in a tiny little bit more coverage (if any) by running through all Kazoo Data Manager code.
 
-FixtureDB acts as central repository for all fixtures which we simply use during tests.
+FixtureDB acts as a central repository for all fixtures which we simply use during tests.
 
-When writing your EUnit test you can use [_fixture_ test representation](http://erlang.org/doc/apps/eunit/chapter.html#Fixtures) to setup a dummy connection to FixtureDB, and cleanup the connection later.
+When writing your EUnit test you can use [_fixture_ test representation](http://erlang.org/doc/apps/eunit/chapter.html#Fixtures) to setup a your test environment by starting a dummy connection to FixtureDB, and cleanup the connection later.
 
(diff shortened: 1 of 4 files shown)


core/kazoo_fixturedb/doc/README.md
```diff
@@ -88,17 +68,50 @@ core/kazoo_fixturedb/priv/dbs/
...
+
+### Database
+
+Each database has a directory of their own inside `core/kazoo_fixturedb/priv/dbs/`.
+
+> **NOTE:** Database name in URL encoded, e.g. `account/xxxx` becomes `account%2Fac%2Fco%2Funtxxxx`.
+
+In each database there are two directories: `docs` for storing regular document inside database and `views` for storing view results.
+
+For make it easy to map each attachment or complex view to their correspond files in `docs` and `views` directories there are two CSV index files the root of the database directory.
```
jamesaimonetti: "To make it easy..." probably makes more sense here

jamesaimonetti: And "corresponding"

jamesaimonetti: "at the root"

core/kazoo_fixturedb/doc/README.md
```diff
@@ -113,7 +126,9 @@ Delete document(s) is result in a bulk save in CouchBeam, so to mimic this behav
        {<<"accepted">>,true}]}]}
 ```
 
-If you changed an already exists JSON file manually with a text editor and need to update `pvt_document_hash` you can use `kz_fixturedb_util:update_pvt_doc_hash(Path)`. This is necessary for example in Teletype notification templates in `system_config` database since during initializing a template teletype checks if document hash is changed or not so it does not updates template which is customized directly in database.
+#### Utility Functions
+
+If you changed an already existence JSON file manually with a text editor and need to update `pvt_document_hash` you can use `kz_fixturedb_util:update_pvt_doc_hash(Path)`. This is necessary for example in Teletype notification templates in `system_config` database since during initializing a template, teletype checks if document hash is changed or not so it does not update template which is customized directly in database.
```
jamesaimonetti: "If you change an existing JSON file manually and need to update..."

core/kazoo_fixturedb/doc/README.md
```diff
@@ -136,50 +151,58 @@ doc_id, attachment_name, attachment_file_name
 201710-vm_message0000000000000000000001, vm-new_message.wav, 5cef38cd0260632cf119c6fcf0c44f58.att
 ```
 
-When you're adding an attachment use `kz_fixturedb_util:get_att_path/3,4` to get the file path and save it there then use `kz_fixturedb_util:add_att_path_to_index/3,4` to add the attachment file name to the `attachment-index.csv` file.
-
-Putting an attachment is just returning the updated document JObj back. Deleting an attachment is returning tuple `{'ok', JObj}` whic JObj just have `id` and `rev`. Both put and delete will return error tuple if the JSON document file does not exists in database.
+Putting an attachment is just returning the updated document JObj back. Deleting an attachment is returning tuple `{'ok', JObj}` which the JObj just have the `id` and `rev`. Both put and delete will return error tuple if the JSON document file does not exists in database.
```
jamesaimonetti: Delete should include `"_deleted":true` too I believe. Not sure if code checks for this but to avoid surprises later.

core/kazoo_fixturedb/doc/README.md
```diff
@@ -188,6 +211,8 @@ If `include_docs` is set, you don't need to include docs in your view result, `k
 
 As a rule write your view result in the `ascending` order. Again `kz_fixturedb_view` is reversing the the result if `descending` is set.
 
+This only works if you're query does not have `reduce`, `group` or `group_level` view options.
```
jamesaimonetti: s/you're/your/

core/kazoo_fixturedb/doc/README.md
icehess: nope: http://docs.couchdb.org/en/2.1.1/api/document/attachments.html#delete--db-docid-attname

jamesaimonetti: TIL

PR #4495, opened 2018-01-18
ensure number db format is sane

jamesaimonetti:


jamesaimonetti:
@kalda341 can you apply this and test? I think this catches the issues being seen with refresh/migrate.

lazedo:
@jamesaimonetti do we need a master version just in case?

jamesaimonetti:
i'll check, will add if needed

kalda341:
@jamesaimonetti Working for me!
Thanks heaps.

PR #4498, opened 2018-01-18
HELP-34437: for boxes setup via APIs with is_setup and no pin explici…

k-anderson:
…tly reject access to main_menu

### DIFF
--- a/applications/callflow/src/module/cf_voicemail.erl
+++ b/applications/callflow/src/module/cf_voicemail.erl
@@ -237,19 +237,19 @@ check_mailbox(#mailbox{exists='false'
             %% can't find mailbox, set Loop to max to play abort prompts in above func clause
             check_mailbox(Box, Call, MaxLoginAttempts + 1)
     end;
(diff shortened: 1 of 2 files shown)


applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -237,19 +237,19 @@ check_mailbox(#mailbox{exists='false'
...
     main_menu(Box, Call);
-check_mailbox(#mailbox{pin = <<>>}=Box, 'true', Call, _) ->
-    %% If this is the owner of the mailbox calling in and it doesn't require a pin then jump
-    %% right to the main menu
-    lager:info("caller is the owner of this mailbox, and it has no pin"),
-    main_menu(Box, Call);
-check_mailbox(#mailbox{pin = <<>>, require_pin='true'}, 'false', Call, _) ->
-    %% If the caller is not the owner or the mailbox requires a pin to access it but has none set
-    %% then terminate this call.
+check_mailbox(#mailbox{pin = <<>>}, _, Call, _) ->
```
icehess: should you check for `require_pin` here?

PR #4499, opened 2018-01-18
Compile on hrl changes

jamesaimonetti:


jamesaimonetti:
WTH, when I SSH onto the circle server and run `JOBS="4" make` is succeeds...

PR #4508, opened 2018-01-23
Handle case where teletype template exists on account but not system

kalda341:
The request was crashing before, with no useful information as to why.


icehess:
Hi @kalda341 

I was wondering what is use case for this PR? Are you expecting to add a new template to one account only not system? Or one of the the kazoo default templates is missing?

kalda341:
@icehess This happened as a result of renaming/deleting teletype templates on the system. It's not a case you'd want to be in, but I figured probably a case worth handling, as I couldn't make any notifications API requests.

PR #4510, opened 2018-01-23
remove unused filters object

k-anderson:
Just a maintenance PR, removing the unused filters from the view.


k-anderson:
Don't merge this yet, I would like to do some more testing.  Just didn't want to loose it :)

k-anderson:
this is ready :)

PR #4515, opened 2018-01-25
PISTON-523: handle Lang when using account db, not just system_media

danielfinke:
Prompts overridden by cb_media right now cannot be found by kz_media_file (kz_media_file:get_uri/2 returns <<>>).
Fixed by this PR. Tested various other calls to get_uri and they all still work.

### DIFF
--- a/core/kazoo_media/src/kz_media_file.erl
+++ b/core/kazoo_media/src/kz_media_file.erl
@@ -24,7 +24,7 @@ get_uri(#media_store_path{}=Store, JObj) ->
     maybe_proxy(JObj, Store);
 get_uri(Media, JObj) when is_binary(Media) ->
     kz_util:put_callid(JObj),
-    Paths = [Path
+    Paths = [kz_http_util:urldecode(Path)
              || Path <- binary:split(Media, <<"/">>, ['global', 'trim']),
                 not kz_term:is_empty(Path)
             ],
--- a/core/kazoo_media/src/kz_media_map.erl
+++ b/core/kazoo_media/src/kz_media_map.erl
@@ -392,7 +392,7 @@ maybe_add_prompt(AccountId, JObj, PromptId) ->
     lager:debug("adding language ~s for prompt ~s to map for ~s", [Lang, PromptId, AccountId]),
     Languages = kz_json:set_value(Lang
                                  ,kz_media_util:prompt_path(kz_doc:account_id(JObj, ?KZ_MEDIA_DB)
-                                                           ,kz_doc:id(JObj)
+                                                           ,kz_http_util:urlencode(kz_doc:id(JObj))
                                                            )
                                  ,Langs
                                  ),


jamesaimonetti:
@danielfinke interesting, we found a case where, if the attachment upload failed in cb_media, it continued on its way without reporting the error. Could that be related to this issue?

jamesaimonetti:
See [here](https://github.com/2600hz/kazoo/pull/4511/commits/d3c3c7dc3f04f74e94ad70319213ed13f21e325f#diff-560c8b8a140609ad9e7fd8859287a2d9R959)

danielfinke:
@jamesaimonetti I think that is a different issue. This will affect media docs whose ids contain a slash

jamesaimonetti:
gotcha, thanks. taking a look

lazedo:
@danielfinke can you give an example of what you're finding ? `This will affect media docs whose ids contain a slash`

danielfinke:
@lazedo upload a custom prompt at the account level using the cb_media API. It won't be found by kz_media_file:get_uri without this change

lazedo:
@danielfinke attach the relevant crossbar logs and payload sent

lazedo:
@danielfinke makes sense, needs attention and testing

danielfinke:
The prompt creation payload:
PUT /v1/accounts/63463bc16b7af744f88c05567219c839/media
[510388243e4114ce4aa7ec7a9d0dc6f3.log](https://github.com/2600hz/kazoo/files/1669411/510388243e4114ce4aa7ec7a9d0dc6f3.log)
```
{"data":{"prompt_id":"agent-already_logged_in","language":"en-us","description":"C:\\fakepath\\test.wav","name":"en-us/agent-already_logged_in"},"verb":"PUT"}
```
POST /v1/accounts/63463bc16b7af744f88c05567219c839/media/en-us%2Fagent-already_logged_in/raw
[42d761bed269826ae72d0b06a9b7cf8f.log](https://github.com/2600hz/kazoo/files/1669409/42d761bed269826ae72d0b06a9b7cf8f.log)
```
<base64 audio data>
```

[53e560db062133da.log](https://github.com/2600hz/kazoo/files/1669410/53e560db062133da.log)
Call attempt to log in the agent is supposed to play the already logged in message. In particular note the line:
```
Jan 26 22:34:01 kazoo41-1 2600hz[12860]: |53e560db062133da|kz_media_file:70 (<0.3608.0>) fetching attachment url for '<<"account%2F63%2F46%2F3bc16b7af744f88c05567219c839">>' , '<<"en-us">>', '<<"agent-already_logged_in">>'
```
which shows that the parsing is broken. Those arguments should be Db, Id, Attachment.

lazedo:
@danielfinke thanks, please provide the full log for the play. with the existing code it should work (get first attachment)

danielfinke:
[kazoo_media.log](https://github.com/2600hz/kazoo/files/1669472/kazoo_media.log)
@lazedo it's not going to work with the existing code. There will be 3 elements in the list passed to kz_media_file:find_attachment/1.

lazedo:
@danielfinke i don't think the problem is here. can you provide full log of the previous ? i think its failing a match or not calling in the proper way, maybe because of map addition (1st time).
```
playback(<<"prompt://", PromptPath/binary>>, Options) ->
    lager:debug("looking up prompt path ~s", [PromptPath]),
    case binary:split(PromptPath, <<"/">>, ['global']) of
        [AccountId, PromptId, Language] ->
            Media = kz_media_map:prompt_path(AccountId, PromptId, Language),
            playback(Media, Options);
        _Path ->
            lager:warning("invalid prompt path: ~p", [_Path]),
            {'error', 'invalid_media_name'}
    end;
playback(<<_/binary>> = Media, JObj) ->
    lager:debug("lookup media url for ~s", [Media]),
    kz_media_file:get_uri(Media, JObj);
```
from your log above, it matched `playback(<<_/binary>> = Media, JObj) ->` and i think it should have matched `playback(<<"prompt://", PromptPath/binary>>, Options) ->`

danielfinke:
@lazedo it does go through the last clause as part of the last part of the "prompt://" clause. Look at the second line in the last log I attached. It's logging "looking up prompt path..." line 35

lazedo:
@danielfinke that's why i asked for the full log.
you can also try to add some log here
```
        [AccountId, PromptId, Language] ->
            Media = kz_media_map:prompt_path(AccountId, PromptId, Language),
            playback(Media, Options);
```
but the full log with evidence would help

danielfinke:
@lazedo the whole call's log is here: https://github.com/2600hz/kazoo/pull/4515#issuecomment-360924097 in my earlier comment (53e560db062133da.log)

lazedo:
@danielfinke can you try these changes instead ? worked for me but needs further tests.  the problem is that we are creating a two part path without urlencoding the `id` of the prompt which contains the `/`
`kz_in media_map`
```
maybe_add_prompt(AccountId, JObj, PromptId) ->
    lager:debug("add prompt ~s to ~s (~s)", [PromptId, AccountId, kz_doc:id(JObj)]),
    Lang = kz_term:to_lower_binary(
             kz_json:get_value(<<"language">>, JObj, kz_media_util:prompt_language(AccountId))
            ),

    #media_map{languages=Langs}=Map = get_map(AccountId, PromptId),

    lager:debug("adding language ~s for prompt ~s to map for ~s", [Lang, PromptId, AccountId]),
    Languages = kz_json:set_value(Lang
                                 ,kz_media_util:prompt_path(kz_doc:account_id(JObj, ?KZ_MEDIA_DB)
                                         ====>             ,kz_http_util:urlencode(kz_doc:id(JObj))
                                                           )
                                 ,Langs
                                 ),
```
`kz_media_file`
```
get_uri(#media_store_path{}=Store, JObj) ->
    maybe_proxy(JObj, Store);
get_uri(Media, JObj) when is_binary(Media) ->
    kz_util:put_callid(JObj),
    Paths = [kz_http_util:urldecode(Path)      <=====
             || Path <- binary:split(Media, <<"/">>, ['global', 'trim']),
                not kz_term:is_empty(Path)
            ],
    get_uri(Paths, JObj);
```

danielfinke:
@lazedo though that does work for playing the media, it will require more changes in kz_media_map (for removal - more changes to what I added in https://github.com/2600hz/kazoo/pull/4527). I'm afraid it will add double encoding here, most likely with prompts retained from 3.22.

Is there an issue with the change that I proposed? It's minimal and no functionality has been removed. Particular removing 'first' was fine as it is implicit anyway, when not using kz_media_file:find_attachment/4.

danielfinke:
@lazedo after a second look, it looks like deletion is actually fine when my other patch #4527 is applied. I'll run with your change and revisit this in the future if I discover an issue. I'll update the PR in a minute

lazedo:
@danielfinke your tries to fix an error. we should fix where the error is.

danielfinke:
@lazedo updated

lazedo:
@danielfinke i mean, even the `KZ_MEDIA_DB` shouldn't be there. the `kz_media_map` should cache the attachment name and revision, and kz_media_map:prompt_path should return the `#media_store_path`.
but that's for another take

danielfinke:
@lazedo I would agree, standardizing on the record return value is a good idea

lazedo:
@danielfinke thanks.

PR #4516, opened 2018-01-25
PISTON-522: add end_wrapup kapi to cancel agent wrapup early

danielfinke:


applications/acdc/src/acdc_agent_fsm.erl
```diff
@@ -267,6 +268,14 @@ pause(ServerRef, Timeout) ->
 resume(ServerRef) ->
     gen_statem:cast(ServerRef, {'resume'}).
 
+%%--------------------------------------------------------------------
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec end_wrapup(kz_types:server_ref()) -> 'ok'.
+end_wrapup(FSM) ->
+    gen_fsm:send_all_state_event(FSM, {'end_wrapup'}).
```
lazedo: use `gen_statem:cast/2`

PR #4519, opened 2018-01-25
[4.1] HELP-34519: don't do after-notify action if notify was not successful

icehess:
Check for any completed in the notification response and if it wasn't successful don't do nothing.

This for the case when the user don't have any email address set or doesn't want to receive notification or the template is disabled in account but `delete_after_notify` or `save_after_notify` is set on voicemail box, so kazoo_voicemail would do that action regardless of the reality that it wasn't successful.

master: https://github.com/2600hz/kazoo/pull/4520

### DIFF
--- a/applications/tasks/src/kz_notify_resend.erl
+++ b/applications/tasks/src/kz_notify_resend.erl
@@ -33,7 +33,7 @@
 -define(MOD_CONFIG_CAT, <<(?CONFIG_CAT)/binary, ".notify_resend">>).
 -define(DEFAULT_TIMEOUT, 10 * ?MILLISECONDS_IN_SECOND).
 
-%% notify resned crawler settungs
+%% notify resend crawler settings
 -define(NOTIFY_RESEND_ENABLED,
         kapps_config:get_is_true(?MOD_CONFIG_CAT, <<"notify_resend_enabled">>, 'true')).
 -define(TIME_BETWEEN_CYCLE,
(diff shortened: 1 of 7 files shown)


elenamarinaatooma:
I applied the patch for core/kazoo_voicemail. The patch solves the following problem: if the requested voicemail notification fails  action_after_notify is not considered. However you need to make an additional change to fix the original issue: application/teletype has to set notification status to 'failed' if voicemail notification is disabled:
1. applications/teletype/src/teletype_util.erl:
notification_disabled(DataJObj, TemplateId) ->
     AccountId = kapi_notifications:account_id(DataJObj),
     lager:debug("notification ~s handling not configured for account ~s", [TemplateId, AccountId]),
     send_update(DataJObj, <<"failed">>, <<"notification disabled">>).
2.  applications/teletype/src/templates/teletype_voicemail_to_email.erl:
maybe_process_req(DataJObj, false) ->
     Msg = io_lib:format("request or box ~s has no emails or owner doesn't want emails"
                        ,[kz_json:get_value(<<"voicemail_box">>, DataJObj)]
                        ),
     lager:debug(Msg),
     teletype_util:send_update(DataJObj, <<"failed">>, kz_term:to_binary(Msg));
Could you please give it a try?

icehess:
@elenamarinaatooma So I added new `status` to `notify_update` to let the sender knows what is happening. For voicemail case, it set the action to `nothing` if the message is disabled or ignored or failed.

elenamarinaatooma:
Hi Team,
Thanks very much for the help! The fix resolved the problem.
Regards,
Elena

PR #4527, opened 2018-01-26
PISTON-524: extract_prompt_id/1 to support "/" as per Kazoo 4 prompt id format

danielfinke:
Account-level overridden prompts have the format lang/prompt_id. Support this in extract_prompt_id/1 in kz_media_map. Then deletions of those prompts will be properly removed from the map

### DIFF
--- a/core/kazoo_media/src/kz_media_map.erl
+++ b/core/kazoo_media/src/kz_media_map.erl
@@ -123,13 +123,20 @@ handle_media_doc_change(JObj, _Change) ->
     {'ok', Doc} = kz_datamgr:open_doc(Db, kz_json:get_value(<<"ID">>, JObj)),
     gen_listener:cast(?MODULE, {'add_mapping', Db, Doc}).
 
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%% Parse the kz_media_map prompt ID from media doc ID.
+%% Keeping old format separator "%2F" for backwards compatibility,
+%% i.e. deleting prompts carried forward from Kazoo < 4.0
+%% @end
+%%--------------------------------------------------------------------
 -spec extract_prompt_id(kz_term:ne_binary()) -> kz_term:ne_binary().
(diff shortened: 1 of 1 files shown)


core/kazoo_media/src/kz_media_map.erl
```diff
@@ -123,13 +123,20 @@ handle_media_doc_change(JObj, _Change) ->
...
-    PromptId;
-extract_prompt_id(<<_Lang:2/binary, "%2F", PromptId/binary>>) ->
-    PromptId;
-extract_prompt_id(MediaId) ->
-    MediaId.
+extract_prompt_id(<<_Lang:5/binary, "/", PromptId/binary>>) -> PromptId;
+extract_prompt_id(<<_Lang:2/binary, "/", PromptId/binary>>) -> PromptId;
+extract_prompt_id(<<_Lang:5/binary, "%2F", PromptId/binary>>) -> PromptId;
+extract_prompt_id(<<_Lang:2/binary, "%2F", PromptId/binary>>) -> PromptId;
+extract_prompt_id(MediaId) -> MediaId.
```
lazedo: @danielfinke why are adding this ? risk of going back to double encoding. we need example examples before accepting this. i don't like the risk.

lazedo:
this seems somehow related to #4515

core/kazoo_media/src/kz_media_map.erl
danielfinke: @lazedo this function is only called by the ?DOC_DELETED clause of handle_media_doc_change/2. Will only contain an ID of whatever the deleted doc was. I'm keeping the old clauses ("%2F") for accounts that were brought forward from Kazoo 3.22 and below. The stuff I added ("/") fixes the function for new prompt overrides, since those docs' ids contain "/". kz_media_map shouldn't be getting a prompt_id that contains the lang, because that's not how it's stored in the ets table.

lazedo:
@danielfinke thanks

PR #4528, opened 2018-01-26
Master pr disable voicemail envelope playback

bradfordben:
Added feature to disable the playback of the voicemail envelope ("received on ....") when listening to voicemail

### DIFF
--- a/applications/callflow/src/module/cf_voicemail.erl
+++ b/applications/callflow/src/module/cf_voicemail.erl
@@ -792,7 +792,7 @@ message_count_prompts(New, Saved) ->
 %% @end
 %%--------------------------------------------------------------------
 -spec message_prompt(kz_json:objects(), binary(), non_neg_integer(), mailbox()) ->
-                          kapps_call_command:audio_macro_prompts().
+                            kapps_call_command:audio_macro_prompts().
 message_prompt([H|_]=Messages, Message, Count, #mailbox{timezone=Timezone, skip_envelope='false'}) ->
     [{'prompt', <<"vm-message_number">>}
     ,{'say', kz_term:to_binary(Count - length(Messages) + 1), <<"number">>}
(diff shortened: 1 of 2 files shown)


applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -803,13 +829,7 @@ play_messages([H|T]=Messages, PrevMessages, Count, #mailbox{timezone=Timezone
...
     Message = kvm_message:media_url(AccountId, H),
     lager:info("playing mailbox message ~p (~s)", [Count, Message]),
-    Prompt = [{'prompt', <<"vm-message_number">>}
-             ,{'say', kz_term:to_binary(Count - length(Messages) + 1), <<"number">>}
-             ,{'play', Message}
-             ,{'prompt', <<"vm-received">>}
-             ,{'say',  get_unix_epoch(kz_json:get_integer_value(<<"timestamp">>, H), Timezone), <<"current_date_time">>}
-             ,{'prompt', <<"vm-message_menu">>}
-             ],
+    Prompt = message_prompt(Messages, Message, Count, Box),
```
lazedo: you need to remove `timezone=Timezone` since `Timezone` is not used within the function
```
 play_messages([H|T]=Messages, PrevMessages, Count, #mailbox{timezone=Timezone
                                                            }=Box, Call) ->
```

bradfordben: My mistake i missed that. I have removed it now. Thanks

core/kazoo_documents/src/kzd_voicemail_box.erl
```diff
@@ -124,6 +126,13 @@ skip_greeting(Box) ->
 skip_greeting(Box, Default) ->
     kz_json:is_true(?KEY_SKIP_GREETING, Box, Default).
 
+-spec skip_envelope(doc()) -> boolean().
+-spec skip_envelope(doc(), Default) -> boolean() | Default.
```
icehess: Please do not group `spec` line together. Move this line above its own function counterpart line.

jamesaimonetti: I think `scripts/evil_specs_remover.escript -k` will do the updates needed

icehess:
@bradfordben Thank you for your contribution!

PR #4537, opened 2018-01-31
subscribe to all accounts' calls

onnet:
Being a cluster's superadmin, I would like to monitor all the calls passing though my cluster.
Since we've already got cb_channels able to provide all cluster's current calls, the only thing we need for live monitoring is ability to subscribe to call events of all the cluster over  Websockets


lazedo:
@onnet thanks

PR #4540, opened 2018-02-02
Implement withholding of caller id for offnet outbound calls via property on user, device and account doc.

bpar476:
Caller id privacy settings can be controlled through the User, Device and Account docs. Configuration works similarly to how it does in cf_privacy; common code has been refactored into cf_util.

### DIFF
--- a/applications/callflow/src/module/cf_resources.erl
+++ b/applications/callflow/src/module/cf_resources.erl
@@ -357,7 +357,7 @@ get_privacy_prefs(Call) ->
 get_privacy_prefs_from_endpoint(Call) ->
     lager:debug("Call is outbound, checking caller_id_outbound_privacy value"),
     {'ok', Endpoint} = kz_endpoint:get(Call),
-    case kz_json:get_value(<<"caller_id_outbound_privacy">>, Endpoint) of
+    case kz_json:get_value([<<"caller_id_options">>, <<"outbound_privacy">>], Endpoint) of
         'undefined' ->
             [];
         %% can't call kapps_call_command:privacy/2 with Mode = <<"none">>
--- a/applications/callflow/src/module/cf_resources.erl
+++ b/applications/callflow/src/module/cf_resources.erl
@@ -21,6 +21,9 @@
 %%% @end
 %%% @contributors
 %%%   Karl Anderson
+%%%
+%%%   Account, user, device level privacy - Sponsored by Raffel Internet B.V.
+%%%       implemented by Voyager Internet Ltd.
 %%%-------------------------------------------------------------------
 -module(cf_resources).
 -behaviour(gen_cf_action).
@@ -127,7 +130,7 @@ get_channel_vars(Call) ->
 -spec add_privacy_flags(kapps_call:call(), kz_term:proplist()) -> kz_term:proplist().
 add_privacy_flags(Call, Acc) ->
     CCVs = kapps_call:custom_channel_vars(Call),
-    kz_privacy:flags(CCVs) ++ Acc.
+    props:set_values(get_privacy_prefs(Call), kz_privacy:flags(CCVs)) ++ Acc.
 
 -spec maybe_require_ignore_early_media(kapps_call:call(), kz_term:proplist()) -> kz_term:proplist().
 maybe_require_ignore_early_media(Call, Acc) ->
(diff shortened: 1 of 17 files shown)


applications/callflow/src/module/cf_resources.erl
```diff
@@ -341,3 +344,26 @@ handle_channel_destroy(<<"loopback", _/binary>>, _JObj) ->
...
+    case kapps_call:inception(Call) of
+        'undefined' -> get_privacy_prefs_from_endpoint(Call);
+        _Else -> []
+    end.
+
+-spec get_privacy_prefs_from_endpoint(kapps_call:call()) -> kz_term:proplist().
+get_privacy_prefs_from_endpoint(Call) ->
+    lager:debug("Call is outbound, checking caller_id_outbound_privacy value"),
+    {'ok', Endpoint} = kz_endpoint:get(Call),
+    case kz_json:get_value(<<"caller_id_outbound_privacy">>, Endpoint) of
```
k-anderson: @bpar476 Howdy!  This looks great with one exception.  We prefer a hierarchical data structure to 'group' like parameters and have an existing object called 'caller_id_options' that we believe is a more appropriate location for this.  More generally we are also avoiding adding non-object root level properties to documents for a number of reasons :)

As such, we request that you move this to caller_id_options.outbound_privacy, let us know your thoughts.  Thanks!

bpar476: Will get right onto it and comment again when I have updated the pull request :smile:

bpar476:
Updated and tested!

applications/callflow/src/module/cf_resources.erl
```diff
@@ -341,3 +344,26 @@ handle_channel_destroy(<<"loopback", _/binary>>, _JObj) ->
...
+    lager:debug("Call is outbound, checking caller_id_outbound_privacy value"),
+    {'ok', Endpoint} = kz_endpoint:get(Call),
+    case kz_json:get_value([<<"caller_id_options">>, <<"outbound_privacy">>], Endpoint) of
+        'undefined' ->
+            [];
+        %% can't call kapps_call_command:privacy/2 with Mode = <<"none">>
+        <<"none">>=NoneMode ->
+            cf_util:ccvs_by_privacy_mode(NoneMode);
+        Mode ->
+            kapps_call_command:privacy(Mode, Call),
```
lazedo: why is this being called here ?

applications/callflow/src/module/cf_resources.erl
```diff
@@ -341,3 +344,26 @@ handle_channel_destroy(<<"loopback", _/binary>>, _JObj) ->
...
+        _Else -> []
+    end.
+
+-spec get_privacy_prefs_from_endpoint(kapps_call:call()) -> kz_term:proplist().
+get_privacy_prefs_from_endpoint(Call) ->
+    lager:debug("Call is outbound, checking caller_id_outbound_privacy value"),
+    {'ok', Endpoint} = kz_endpoint:get(Call),
+    case kz_json:get_value([<<"caller_id_options">>, <<"outbound_privacy">>], Endpoint) of
+        'undefined' ->
+            [];
```
lazedo: what happens if cf_privacy is called but there's not caller_id_options for privacy ?

lazedo: or what happens if cf_privacy is called and there's calker_id_options for privacy ?

applications/callflow/src/module/cf_resources.erl
bpar476: I assume you mean the call to kapps_call_command:privacy. This is mimicking the behaviour from the cf_privacy module and the caller ID attributes won't be hidden unless it's called.

applications/callflow/src/module/cf_resources.erl
bpar476: This doesn't interact with cf_privacy. If there are no called_id_options for outbound_privacy then this module will not add any ccvs to the call.

applications/callflow/src/module/cf_resources.erl
lazedo: i don't think this should be called here, we are creating properties for the amqp payload. we may need to handle this on stepswitch and/or ecallmgr_fs_bridge

bpar476: Okay, valid point. Seeing as it's already merged, how would you like to proceed? Should I change it and create a new pull request?

applications/callflow/src/module/cf_resources.erl
lazedo: what if there are caller_id_options for outbound_privacy and cf_privacy was called before ? cf_privacy settings should override these

bpar476: when cf_privacy and outbound_privacy are both set, the resulting privacy will be the union of the two. If there are no outbound_privacy options, cf_privacy will have full control. If outbound_privacy is name only and cf_privacy is number only, both name and number will be hidden.

lazedo: my point exactly, this is not correct

bpar476: From my understanding you think that if one privacy option is set with cf_privacy, and the other is set with caller_id_options, then only the cf_privacy option should be used? I think it makes more sense that the user's decisions to hide caller ID should be respected and cf_privacy can be used to partially override some of the caller_id_options on a case-by-case basis.

lazedo: `caller_id_options` should be treated as defaults overridden by cf_privacy or set as an option in cf_privacy.
if `both` are set into caller_id_options and i want to send display name and/or number, how would cf_privacy help with that ?
this needs to be worked

bpar476: As far as I know, there is no way for cf_privacy to specify that you want to display everything, only hide which would mean you could not user cf_privacy to show your name and number if they are set to be hidden with the caller_id_options. However I can change it so that if cf_privacy and caller_id_options.outbound_privacy are both set, and are different to each other, then cf_privacy will be preferred completely instead of a merge.

lazedo: this PR introduced a change in behaviour, cf_privacy was mandatory and no longer is.  this will certainly cause issues and unnecessary support tickets. cf_privacy must be changed to either `merge` or `overwrite`, and cf_resources should respect cf_privacy if set or add caller_id_options if not set.
the default behaviour should be preserved, ie, default in cf_privacy should be `overwrite`

bpar476:
Requested changes implemented for master in: #4590

PR #4552, opened 2018-02-07
PISTON-534: fix results limit logic

danielfinke:


lazedo:
@danielfinke keep them coming :)

PR #4554, opened 2018-02-08
[master] HELP-34868: enforce media_source when recording in cf_voicemail

icehess:
Set `media_source` to `recording` for greetings, name, and temp greeting in voicemail callflow module.

Also enforcing `media_source` in schema for Media API and only handling TTS based on it.

4.1: #4560

### DIFF
--- a/applications/crossbar/src/modules/cb_media.erl
+++ b/applications/crossbar/src/modules/cb_media.erl
@@ -386,7 +386,7 @@ post_media_doc(Context, MediaId, 'undefined') ->
 post_media_doc(Context, _MediaId, _AccountId) ->
     case is_tts(cb_context:doc(Context)) of
         'true' -> create_update_tts(Context, <<"update">>);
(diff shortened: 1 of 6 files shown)


applications/crossbar/src/modules/cb_media.erl
```diff
@@ -402,68 +383,59 @@ post(Context, MediaId) ->
...
-                            'success' -> crossbar_doc:save(remove_tts_keys(C));
-                            _Status -> Context
-                        end
-                end
-               ],
-    lists:foldl(fun(F, C) -> F(C) end, Context, Routines).
+post_media_doc(Context, _MediaId, _AccountId) ->
+    case is_tts(cb_context:doc(Context)) of
+        'true' -> create_update_tts(Context, <<"update">>);
+        'false' -> crossbar_doc:save(crossbar_doc:save(remove_tts_keys(Context)))
```
lazedo: two saves ?

lazedo:
needs rebase after #4561

icehess:
@lazedo It rebased.

PR #4557, opened 2018-02-08
PISTON-535: allow agents/status queries with no limit

danielfinke:
Conflicts:
	applications/acdc/src/acdc_agent_stats.erl
	applications/acdc/src/acdc_stats.erl
	applications/acdc/src/acdc_stats_util.erl


lazedo:
@danielfinke thanks

PR #4558, opened 2018-02-08
Master pr ablity to set cdr csv file name in crossbar request

bradfordben:
We would like to be able to set the file name of the returned csv file.
The reason for this is to allow us to update the way kazoo-ui exports CDR's to CSV files to use the CSV returned by crossbar as opposed to building the csv file in JS


jamesaimonetti:
Merged. Thanks for the work!

PR #4559, opened 2018-02-08
Handle a json object as a default value in kz_ast_util

kalda341:


core/kazoo_ast/src/kz_ast_util.erl
```diff
@@ -482,9 +482,12 @@ cell_wrap('undefined') -> <<" ">>;
 cell_wrap([]) -> <<"`[]`">>;
 cell_wrap(L) when is_list(L) -> [<<"`[\"">>, kz_binary:join(L, <<"\", \"">>), <<"\"]`">>];
 cell_wrap(<<>>) -> <<"\"\"">>;
-cell_wrap(?EMPTY_JSON_OBJECT) -> <<"`{}`">>;
+cell_wrap(B) when is_binary(B) -> [<<"`">>, B, <<"`">>];
 cell_wrap(Type) ->
-    [<<"`">>, kz_term:to_binary(Type), <<"`">>].
+    case kz_json:is_json_term(Type) of
```
jamesaimonetti: Out of curiosity, what's triggering this change to be necessary?

kalda341: We have set `properties.gateways.custom_sip_headers.default` in the resources schema to be:

    {
        "X-Account-ID": "{account_id}",
        "X-Billing-Number": "{caller_id_number}"
    }

This crashes kz_ast_util while trying to build the table.

kalda341:
@jamesaimonetti I know this is the wrong place to ask, but it's driving me crazy at the moment, is vaguely related, and I don't know where else to ask.

I've added documentation to the first Fetch section in a markdown file in `applications/crossbar/doc/` (not the ref directory!) and it's removed every time I run `make apis`. Any ideas?

jamesaimonetti:
Merged.

jamesaimonetti:
@kalda341 nope, can you share the diff of when you make the change (before its removed by `make apis`)?

kalda341:
```
diff --git a/applications/crossbar/doc/hero_preferences.md b/applications/crossbar/doc/hero_preferences.md
index be776fc..36bebb7 100644
--- a/applications/crossbar/doc/hero_preferences.md
+++ b/applications/crossbar/doc/hero_preferences.md
@@ -14,10 +14,7 @@ Key | Description | Type | Default | Required
 
 
 
-#### Fetch Hero user preferences for the currently logged in user
-
-When the user has no preferences, crossbar will return the default preferences, but it wont save them against the user.
-When the user has created custom preferences via the POST, they will be in the response.
+#### Fetch
 
 > GET /v2/accounts/{ACCOUNT_ID}/hero_preferences
 

```

jamesaimonetti:
@kalda341 check the regex in `scripts/generate-doc-schemas.sh`. Maybe it is only capturing the first word after `####` instead of the whole line? Since you have a schema table above, that's were I'd look first.

kalda341:
@jamesaimonetti I just noticed it's fixed in master. Cheers!

PR #4561, opened 2018-02-09
use defaults from jesse

lazedo:
jesse was fixed for using the applied defaults after properties/items andalso json_path fix for integer as part of the path to access items in arrays ([foo, 0, bar] ) 
added tests for array access/validation

### DIFF
--- a/scripts/check-whitespace.sh
+++ b/scripts/check-whitespace.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+echo "Files with trailing whitespace:"
+# From /
+
+ws="$(echo -ne ' \t\v')"
+
+count=0
+for file in $(git grep -IE "[$ws]$" -- $* | cut -d: -f1 | sort -u); do
+    echo "$file"
+    sed -i 's/\s*$//g' "$file"
+    ((count++))
+done
+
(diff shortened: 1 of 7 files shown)


scripts/check-whitespace.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+echo "Files with trailing whitespace:"
+# From /
+
+ws="$(echo -ne ' \t\v')"
+
+count=0
+for file in $(git grep -IE "[$ws]$" -- $* | cut -d: -f1 | sort -u); do
```
jamesaimonetti: I would change this to be override-able:
```bash
FILES ?= $(git grep...)
for file in $FILES; do
```
Then we could add this to `make code_checks` with `FILES=$(CHANGED)`

lazedo: i think we already do that in circle (check for whitespaces). i committed this because it happens a lot to me and one of these days i'm going to loose the script :)
feel free to modify it

icehess:
@lazedo need backported to 4.1? for #4560

lazedo:
@jamesaimonetti good point

PR #4563, opened 2018-02-09
mwi presence update

lazedo:


lazedo:
to fully work needs https://github.com/2600hz/kazoo-configs-kamailio/pull/50

PR #4564, opened 2018-02-10
Edocify

icehess:
* Adding script `scripts/edocify.escript` which fix most issues when generating EDoc from source files.
* Manually editing comments in source files to escape inline code or code block in the comments.
* Change `include` directory header file inclusion in some header file
* Remove `&` from comments
* Move `spec`s to correct functions.
* Other various fixes to make EDoc passes.

UPDATE:

* add missing `end` tags to comment block started with `doc` tag
* remove any empty comment line between `doc` tag line and first comment line or next tag line to avoid EDoc to interpret that empty lines as a single dot `.`. Yes as single dot `.`!
* add an empty comment block between comment separator and function header to avoid inclusion of separator in comment

UPDATE 2:
* Adding `.editorconfig` (www.editorconfig.org)

UPDATE3 3:
* state-of-edoc script
* update edocify script to bump copyright year or fix/format copyright tag


applications/blackhole/src/blackhole_tracking.erl
```diff
@@ -1,10 +1,8 @@
 %%%-------------------------------------------------------------------
 %%% @copyright (C) 2018, 2600Hz inc
 %%% @doc
-%%%
+%%% @author Peter DEfebvre
```
jamesaimonetti: Probably fix this `E` to `e` :)

applications/crossbar/src/modules/cb_alerts.erl
```diff
@@ -1,12 +1,11 @@
 %%%-------------------------------------------------------------------
 %%% @copyright (C) 2011-2018, 2600Hz INC
 %%% @doc
-%%%
 %%% Listing of all expected v1 callbacks
 %%%
+%%%
+%%% @author PEter Defebvre
```
jamesaimonetti: This one too!

lazedo:
@icehess we should wait for the modules that process/produce the output to be ready. there could be some tweaks needed (regex, whatever), and this would would be a one time bulk update

icehess:
@lazedo Are you sure? That could make this PR pretty big. I have another branch which I'm adding some documentation to important modules. Also I was planning to add an overview for each application.

fenollp:
Why edoc over erldocs? ;)

icehess:
@fenollp because .... ;) And it runs edoc too anyway, and edoc is not working for multi app project unless you fold over application list in correct dependency order, it was not working for kazoo because evil specs (specs grouped together), unescaped binary in comments and other thing this PR is fixing.

Oh and edoc ... too :)

icehess:
@fenollp but feel free to run other_erldocs after this PR merged :)

fenollp:
yep I need to fix the issues in other_erldocs!
Erlang needs Golang's golint and go vet...

PR #4573, opened 2018-02-13
KAZOO-5832: fixed different device activation_charges for dirty and synchronized services

sergey-safarov:


k-anderson:
@sergey-safarov Howdy, in 4.3 we have refactored kazoo-services and this no longer applies :)

PR #4577, opened 2018-02-15
Fixed Teletype typos

kalda341:


applications/teletype/priv/templates/new_user.text
```diff
@@ -3,7 +3,7 @@
 
 Welcome {{user.first_name}},
 
-Your user profile is ready to use, here are logindetails to sign in.
+Your user profile is ready to use, here are login details to sign in.
```
k-anderson: @kalda341 Howdy, thanks for this, looks good with one exception.  This is actually part of the render test.  If you change it here need to update the fixture too :)

https://github.com/2600hz/kazoo/blob/master/applications/teletype/test/rendered-templates/new_user.text#L6

kalda341: @k-anderson Passing all tests now

PR #4579, opened 2018-02-15
Fix /dev/null being replaced by log rotation

kalda341:


lazedo:
@kalda341 why do you want to change this ? `data_lager_event` needs to be registered

kalda341:
@lazedo when run with scripts/dev-start-apps.sh /dev/null is rotated, meaning it is replaced with an empty file.

lazedo:
@kalda341 cannot reproduce this. what's your env ? docker ?
i've seen a problem with docker permissions, but the fix was
``` 
{extra_sinks, [{data_lager_event,[]}
                      ]}
````

kalda341:
This occurs on CentOS 7 when `scripts/dev-start-apps.sh` is used to start Kazoo (as root).
After a while /dev/null is replaced with an empty file.

Your docker fix looks like it would solve the problem. Is there any reason for the file backend to `/dev/null`?

lazedo:
i believe that initially we thought that `{handlers, [{lager_file_backend, ..` needed to be configured but didn't want to actually log anything if tracing was not up. can you try the suggested fix to see if it works for you ?
i see it more like a log rotation issue in lager

kalda341:
@lazedo Your docker fix works. I've updated the PR.

lazedo:
@kalda341 thanks

PR #4581, opened 2018-02-15
Konami call record "Stop" action fix

arjanmvos:
Rewrite of handle function for "Stop" Action to fix Action not stopping recording where using different media name than "CallID.Format" as it invokes a hardcoded "CallID.Format" stop.

Used same functions as in Callflow app and launch function kapps_call:stop_recording() to invoke Stop command through kapps_call module instead of calling kapps_call_command:record_call directly. This to be able to stop the recording by name as saved in the custom_channel_variable for the media file name instead of invoking a stop on hardcoded CallID.Format.


jamesaimonetti:
Thanks!

PR #4583, opened 2018-02-15
knm_simwood_update

onnet:
Otherwise it returns empty list


lazedo:
@onnet can you please rebase? there's a ci isssue with a not changed file

onnet:
@lazedo rebased. Thanks for hint!

PR #4585, opened 2018-02-15
Account user device privacy 4.1

bpar476:
This is a 4.1 port of #4585, allowing caller_id privacy to be set on the device, user or account document


k-anderson:
Additionally, new features should not be backported.  This would be introduced with 4.2 by policy.

It is sometimes necessary to bring a feature to a specific version but we try really hard not to, if you have a compelling argument happy to hear it!

bpar476:
Fair enough. We will just let it arrive through 4.2. Cheers.

PR #4590, opened 2018-02-19
Allow for cf_privacy to overwrite or merge with caller_id_options.outbound_privacy

bpar476:
Disambiguates the behaviour of caller_id_options.outbound_privacy with cf_privacy as described in #4540 
"cf_privacy must be changed to either merge or overwrite, and cf_resources should respect cf_privacy if set or add caller_id_options if not set.
the default behaviour should be preserved, ie, default in cf_privacy should be overwrite"


lazedo:
@bpar476 looks good, and thank you for taking care of this.

bpar476:
Not a problem! :smile:

PR #4594, opened 2018-02-20
HELP=32969: add temp_hold_music to the mix

jamesaimonetti:
https://freeswitch.org/confluence/display/FREESWITCH/temp_hold_music

Suggests this is the way to set hold music while transferring a caller


jamesaimonetti:
4.1 PR at #4595

PR #4604, opened 2018-02-23
Using API filters on /cdrs

jamesaimonetti:


jamesaimonetti:
I'll open a 4.1 if review passes muster

jamesaimonetti:
For reference, the API query used to test:

```bash
curl -v  -H "X-Auth-Token: $AUTH_TOKEN" "http://{SERVER}:8000/v2/accounts/$ACCOUNT_ID/cdrs?filter_call_id=781b548c-1753-11e8-bbbc-97bb7cab3610" | python -mjson.tool
```

Obviously use whichever field you need to filter on.

applications/crossbar/src/modules/cb_cdrs.erl
```diff
@@ -360,9 +360,14 @@ load_chunked_cdr_ids(Context, RespType, Db, Ids) ->
     of
         'false' -> Context;
         {'ok', Results} ->
-            JObjs = [kz_json:get_value(<<"doc">>, Result)
+            HasFilters = crossbar_filter:is_defined(Context),
```
icehess: You don't need to call it directly here, `crossbar_filter:by_doc/2` is calling it.

applications/crossbar/src/modules/cb_cdrs.erl
```diff
@@ -360,9 +360,14 @@ load_chunked_cdr_ids(Context, RespType, Db, Ids) ->
...
+
+            JObjs = [kz_json:get_json_value(<<"doc">>, Result)
                      || Result <- Results,
-                        crossbar_filter:by_doc(kz_json:get_value(<<"doc">>, Result), Context)
+                        %% if there are no filters, include doc
+                        %% otherwise run filters against doc for inclusion
+                        (not HasFilters)
```
icehess: no need for parenthesis.

applications/crossbar/src/modules/cb_cdrs.erl
jamesaimonetti: Well, the idea is that we don't need to compute is_defined for every doc. So we can short-circuit calling by_doc because we already know if filters are defined (which is the first thing by_doc does).

icehess: `is_defined` is doing the _short-circuit_ itself: https://github.com/2600hz/kazoo/blob/edocify/applications/crossbar/src/crossbar_filter.erl#L63

jamesaimonetti: By why call the function X times (where X could be large depending on pagination settings) when calling it once outside the loop and then doing the boolean test is available?

PR #4606, opened 2018-02-23
KAZOO-5784: Improve cb_storage:maybe_check_storage_settings/2 tests using FixtureDB

harenson:


applications/crossbar/test/cb_storage_test.erl
```diff
@@ -20,10 +64,113 @@ maybe_check_storage_settings_test_() ->
...
+            %% why meck can't call the original mfa properly?
+            %% meck_code_gen:get_current_call/1 returns undefined and it cause bad_match
+            %% in meck:passthrough/1
+            (kz_datamgr:db_classification(Db) =:= 'modb'
+             andalso is_db_under_test(Db, kazoo_modb:get_modb(kz_util:format_account_id(Db))))
+                orelse erlang:apply('kz_fixturedb_db_meck_original', db_exists, [Server, Db])
+    end.
+
+is_db_under_test(ThisMonth, ThisMonth) ->                                             true;
+is_db_under_test(<<"account%2Fac%2Fco%2Funt0000000000000000000000001-201710">>, _) -> true;
```
icehess: This test is not using this db so remove this line.

applications/crossbar/test/cb_storage_test.erl
```diff
@@ -3,13 +3,57 @@
...
+%% =======================================================================================
+%% Setup
+%% =======================================================================================
+setup() ->
+    ?debugMsg(":: Setting up Kazoo FixtureDB"),
+    %% Required for s3 attachment handler
+    _ = application:ensure_all_started(erlcloud),
+    Pid = kz_fixturedb_util:start_me(),
+    meck:new(kz_datamgr, [unstick, passthrough]),
+    meck:expect(kz_datamgr, get_results, check_kz_datamgr_history()),
```
icehess: You don't need to make `kz_datamgr:get_results`, remove it.

applications/crossbar/test/cb_storage_test.erl
```diff
@@ -20,10 +64,113 @@ maybe_check_storage_settings_test_() ->
...
+            (kz_datamgr:db_classification(Db) =:= 'modb'
+             andalso is_db_under_test(Db, kazoo_modb:get_modb(kz_util:format_account_id(Db))))
+                orelse erlang:apply('kz_fixturedb_db_meck_original', db_exists, [Server, Db])
+    end.
+
+is_db_under_test(ThisMonth, ThisMonth) ->                                             true;
+is_db_under_test(<<"account%2Fac%2Fco%2Funt0000000000000000000000001-201710">>, _) -> true;
+is_db_under_test(_, _) ->                                                             false.
+
+-define(GET_LEGACY_CALL, {kz_datamgr, get_results, [?ACCOUNT_DB_UNDER_TEST, ?LEGACY_VIEW, [{limit, 2000}, descending]]}).
```
icehess: This is for kazoo_voicemail migration test, remove it, also remove `check_kz_datamgr_history` since you don't need to meck `get_result` function.

applications/crossbar/test/cb_storage_test.erl
```diff
@@ -3,13 +3,57 @@
...
+%% =======================================================================================
+setup() ->
+    ?debugMsg(":: Setting up Kazoo FixtureDB"),
+    %% Required for s3 attachment handler
+    _ = application:ensure_all_started(erlcloud),
+    Pid = kz_fixturedb_util:start_me(),
+    meck:new(kz_datamgr, [unstick, passthrough]),
+    meck:expect(kz_datamgr, get_results, check_kz_datamgr_history()),
+    %% Fix `kz_att_util:format_url/3' issue
+    meck:expect(kz_datamgr, open_cache_doc, fun open_cache_doc/2),
```
icehess: I'm think instead of mecking `kz_datamgr:open_cache_doc/2` meck `kz_fixturedb_doc:open_doc`.

applications/crossbar/test/cb_storage_test.erl
```diff
@@ -20,10 +64,113 @@ maybe_check_storage_settings_test_() ->
...
+                _ ->
+                    meck:passthrough([Db, View, Options])
+            end;
+       (Db, View, Options) ->
+            meck:passthrough([Db, View, Options])
+    end.
+
+-spec open_cache_doc(kz_term:ne_binary(), kz_term:ne_binary()) ->
+                            {ok, kz_json:object()}.
+open_cache_doc(_DbName, _DocId) ->
```
icehess: After mecking `kz_fixturedb_doc:open_doc` check the `DbName` to see it is the current month (`kazoo_modb:get_modb(kz_util:format_account_id(Db)))`) and if it is return success.

Since this is test is only interested in attachment of a document in current month db, by checking db for this month only, `open_doc` will work for other db as normal.

icehess:
That's all. Thanks for writing test by the way :1st_place_medal:

applications/crossbar/test/cb_storage_test.erl
```diff
@@ -3,18 +3,70 @@
...
+%% =======================================================================================
+%% Setup
+%% =======================================================================================
+
+%%--------------------------------------------------------------------
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+setup() ->
+    ?debugMsg(":: Setting up Kazoo FixtureDB"),
```
icehess: no need for logging, since `kz_fixturedb_util:start_me/0`  is doing this itself.

applications/crossbar/test/cb_storage_test.erl
```diff
@@ -3,18 +3,70 @@
...
+    meck:new(kz_fixturedb_doc, [unstick, passthrough]),
+    meck:expect(kz_fixturedb_db, db_exists, this_month_db_exists()),
+    meck:expect(kz_fixturedb_doc, open_doc, fun open_doc/4),
+    Pid.
+
+cleanup(Pid) ->
+    _ = application:stop(erlcloud),
+    kz_fixturedb_util:stop_me(Pid),
+    meck:unload(),
+    ?debugMsg(":: Stopped Kazoo FixtureDB").
```
icehess: same as above.

applications/crossbar/test/cb_storage_test.erl
```diff
@@ -25,15 +77,95 @@ maybe_check_storage_settings_test_() ->
...
+
+%% this is required for voice mail messages which their timestamp is missing from metadata and
+%% their private media is missing or both private media and mailbox doesn't have create/modified
+this_month_db_exists() ->
+    fun(Server, Db) ->
+            %% why meck can't call the original mfa properly?
+            %% meck_code_gen:get_current_call/1 returns undefined and it cause bad_match
+            %% in meck:passthrough/1
+            (kz_datamgr:db_classification(Db) =:= 'modb'
+             andalso is_db_under_test(Db, kazoo_modb:get_modb(kz_util:format_account_id(Db))))
```
icehess: use your `?ACCOUNT_DB_CURRENT_MONTH` here.

applications/crossbar/test/cb_storage_test.erl
```diff
@@ -3,18 +3,70 @@
...
+    meck:expect(kz_fixturedb_doc, open_doc, fun open_doc/4),
+    Pid.
+
+cleanup(Pid) ->
+    _ = application:stop(erlcloud),
+    kz_fixturedb_util:stop_me(Pid),
+    meck:unload(),
+    ?debugMsg(":: Stopped Kazoo FixtureDB").
+
+%% =======================================================================================
```
icehess: It is the test file and I don't want nagging, but anyway I call this line comment separator. Generally after edocification, separator lines should be 80 characters in total.

This line (and 2 below) is used to separator code section. So usually I recommend using three `%` at beginning followed by 77 `=` (no space between `%` and `=`).

For comments in the module beginning, use `%%%` and 77 `-` for separator.
For function comments, use `%%` and 78 `-` for separator.

In all case do not add space between `%` and `-` or `=` character at line beginning.

This makes life so easy if later we want to use `grep`, `sed` and other bulk changing commands later.

Sorry I had to write explanation for commenting format before, but I didn't. I'll try to write it soon. Also `edocify` script is not a very good written code, I'll try to expand it to print more info of why it fails and where.

PR #4608, opened 2018-02-25
Filter rates by caller_id number

onnet:
Use case example:
Could be useful when inbound toll_free rating should be done CID based. 
Like calls from landline and cell phone to local toll_free number should have different rates


jamesaimonetti:
@onnet Thanks!

PR #4612, opened 2018-02-27
PROD-154: fix missing default resource when using dynamic flags

k-anderson:
4.1 backport #4611

### DIFF
--- a/applications/stepswitch/src/stepswitch_resources.erl
+++ b/applications/stepswitch/src/stepswitch_resources.erl
@@ -29,6 +29,7 @@
         ,get_resrc_gateways/1
         ,get_resrc_is_emergency/1
         ,get_resrc_require_flags/1
+        ,get_resrc_ignore_flags/1
         ,get_resrc_global/1
         ,get_resrc_format_from_uri/1
         ,get_resrc_from_uri_realm/1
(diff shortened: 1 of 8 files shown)


applications/stepswitch/src/stepswitch_resources.erl
```diff
@@ -431,6 +435,9 @@ resource_has_flags(Flags, Resource) ->
     lists:all(HasFlag, Flags).
 
 -spec resource_has_flag(kz_term:ne_binary(), resource()) -> boolean().
+resource_has_flag(_, #resrc{ignore_flags='true', id=_Id}) ->
+    lager:debug("resource ~s is used regardless of flags", [_Id]),
```
jamesaimonetti: Be honest, was `regardless` auto-corrected for you from `irregardless` ;)

core/kazoo_documents/src/kzd_resources.erl.src
```diff
@@ -154,6 +155,18 @@ grace_period(Doc, Default) ->
 set_grace_period(Doc, GracePeriod) ->
     kz_json:set_value([<<"grace_period">>], GracePeriod, Doc).
 
+-spec ignore_flags(doc()) -> kz_term:api_boolean().
```
jamesaimonetti: These should be included in kzd_resources.erl

k-anderson: done

PR #4613, opened 2018-02-27
OpenTelecom: Fix blackhole amqp creds

remcovanvugt:
This is a very small fix to avoid emitting AMQP credentials in blackhole events. It simply deletes the amqp_broker key from the event payload.

### DIFF
--- a/applications/blackhole/src/blackhole_data_emitter.erl
+++ b/applications/blackhole/src/blackhole_data_emitter.erl
@@ -24,7 +24,7 @@ event(Binding, RK, Name, Data) ->
           ,{<<"data">>, kz_json:delete_key(<<"amqp_broker">>, Data)}
           ],
     SessionPid ! {'send_data', kz_json:from_list(Msg)},
-   'ok'.
+    'ok'.
 
 -spec reply(pid(), kz_term:ne_binary(), kz_term:ne_binary(), kz_json:object()) -> 'ok'.
 reply(SessionPid, RequestId, Status, Data) ->
--- a/applications/blackhole/src/blackhole_data_emitter.erl
+++ b/applications/blackhole/src/blackhole_data_emitter.erl
@@ -21,10 +21,10 @@ event(Binding, RK, Name, Data) ->
           ,{<<"subscription_key">>, SubscriptionKey}
           ,{<<"name">>, Name}
           ,{<<"routing_key">>, RK}
-          ,{<<"data">>, Data}
+          ,{<<"data">>, kz_json:delete_key(<<"amqp_broker">>, Data)}
           ],
     SessionPid ! {'send_data', kz_json:from_list(Msg)},
-    'ok'.
+   'ok'.
 
 -spec reply(pid(), kz_term:ne_binary(), kz_term:ne_binary(), kz_json:object()) -> 'ok'.
 reply(SessionPid, RequestId, Status, Data) ->


applications/blackhole/src/blackhole_data_emitter.erl
```diff
@@ -21,7 +21,7 @@ event(Binding, RK, Name, Data) ->
           ,{<<"subscription_key">>, SubscriptionKey}
           ,{<<"name">>, Name}
           ,{<<"routing_key">>, RK}
-          ,{<<"data">>, Data}
+          ,{<<"data">>, kz_json:delete_key(<<"amqp_broker">>, Data)}
```
lazedo: @remcovanvugt why is `amqp_broker` here in the first place ?

remcovanvugt: @lazedo I was surprised to see it there, I found this on a 4.1-53 clean install on which only a data migration from 3.22 was performed. It's not there on 3.22. 

It happens on a CHANNEL_CREATE event, I haven't check with the other events. The auth token does not belong to superduper admin. The rest of the emitted data looks fine, only channel related data..

lazedo: @remcovanvugt we can't recreate this, maybe something particular to your setup or the data you had in 3.22

lazedo: this is federated related as we talked in the community call.

k-anderson:
@remcovanvugt Howdy, so as we were discussing on todays call we would prefer to do this as a generic method that is used any time an AMQP payload is exposed externally.  This will give us a common place for future maintenance, ensure that there is no deviations between sanitation functions and use the defined headers so future extensions to the payloads are sanitized by default.

The thought would be to create something like:

kz_api.erl
```
-export([public_fields\1]).

-spec public_fields(kz_json:object()) -> kz_json:object().
public_fields(JObj) ->
    Routines = [fun remove_optional_defaults/1
               ,fun kz_json:normalize_jobj/1
               ],
    lists:foldl(fun(F, J) -> F(J) end, JObj, Routines.

-spec public_fields(kz_json:object()) -> kz_json:object().
remove_option_defaults(JObj) ->
    Keys = [Key || Key <- ?OPTIONAL_DEFAULT_HEADERS
            ,Key /= ?KEY_API_ACCOUNT_ID
            ,Key /= ?KEY_API_CALL_ID
            ],
    kz_json:delete_keys(Keys, JObj).
```

Then add AMQP-Broker and AMQP-Broker-Zone to the OPTIONAL_DEFAULT_HEADERS in kz_api.hrl.


Please note that I typed this in the comments so this is completely untested and should be taken as pseudo code :)
Also, I added kz_json:normalize_jobj to the routines so that all uses get a normalized and sanitized version of the payload, need to make sure that does not introduce a non-backward compatible payload in blackhole.


Thanks!

k-anderson:
@remcovanvugt after further discussion lets not include kz_json:normalize_jobj as kz_doc:public_fields and kz_json:public_fields do not normalize as well so by convention it would be unexpected.

k-anderson:
@remcovanvugt Howdy, since this is vulnerability I didn't want to leave it too much longer so I have opened #4623 (and backports) with the changes I mentioned.  Let us know if you have any comments on those but I hope to get them into builds shortly.  Thanks!

PR #4618, opened 2018-02-28
fix json key when converting from map

lazedo:


core/kazoo_stdlib/test/kz_json_test.erl
```diff
@@ -833,3 +833,28 @@ horse_merge_recursive() ->
                 ).
 
 -endif.
+
+from_map_test_() ->
+    M1 = #{options => [{opt1,<<"my-opt1">>},{opt2,<<"my-opt2">>}]},
+    Expected1 = {[{options,{[{opt1,<<"my-opt1">>},{opt2,<<"my-opt2">>}]}}]},
```
jamesaimonetti: Can we use the macros/functions to bulid JObjs

PR #4621, opened 2018-02-28
Ignore non-filter keys when checking if filters are defined

jamesaimonetti:
The 'time-only' filter check wasn't differentiating between filter keys and non-filter keys on the query string. So you'd get `paginate` being compared with `created_to` and `created_from`, not matching, and returning `false` when `true` was the appropriate return (as seen when omitting `paginate=false` from the query string).

### DIFF
--- a/applications/crossbar/src/crossbar_view.erl
+++ b/applications/crossbar/src/crossbar_view.erl
@@ -188,6 +188,7 @@ build_load_params(Context, View, Options) ->
     try build_general_load_params(Context, View, Options) of
         #{direction := Direction}=LoadMap ->
             HasQSFilter = crossbar_filter:is_defined(Context),
+
             UserMapper = props:get_value('mapper', Options),
 
             {StartKey, EndKey} = start_end_keys(Context, Options, Direction),
@@ -223,7 +224,6 @@ build_load_range_params(Context, View, Options) ->
 
             HasQSFilter = crossbar_filter:is_defined(Context)
                 andalso not crossbar_filter:is_only_time_filter(Context, TimeFilterKey),
-            lager:debug("has qs filter: ~s", [HasQSFilter]),
 
             case time_range(Context, Options, TimeFilterKey) of
                 {StartTime, EndTime} ->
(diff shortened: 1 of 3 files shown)


applications/crossbar/src/modules/cb_cdrs.erl
```diff
@@ -113,6 +113,8 @@ to_json({Req, Context}) ->
 to_csv({Req, Context}) ->
     {Req, to_response(Context, <<"csv">>, cb_context:req_nouns(Context))}.
 
+-spec to_response(cb_context:context(), kz_term:ne_binary(), req_nouns()) ->
+                         {cb_cowboy_payload(), cb_context:context()}.
```
icehess: It returns `cb_context:context()`

applications/crossbar/src/crossbar_view.erl
```diff
@@ -223,7 +224,6 @@ build_load_range_params(Context, View, Options) ->
 
             HasQSFilter = crossbar_filter:is_defined(Context)
                 andalso not crossbar_filter:is_only_time_filter(Context, TimeFilterKey),
-            lager:debug("has qs filter: ~s", [HasQSFilter]),
```
icehess: Why removing this? I only prints one time.

applications/crossbar/src/modules/cb_cdrs.erl
jamesaimonetti: So it does. Come on Dialyzer! You had one job!

jamesaimonetti: Actually, if I told dialyzer about cb_context, it would have known. And, in fact, including it in the mix gets an error reported. I'm sorry Dialyzer...

applications/crossbar/src/crossbar_view.erl
jamesaimonetti: I was just deleting all the "has qs filter" logs since I had added a bunch throughout. I'll re-add this one

PR #4623, opened 2018-03-02
KAZOO-5848: filter optional AMQP headers prior to exposing externally

k-anderson:


k-anderson:
backport to 4.1 #4624
backport to 4.2 #4625
replaces #4613

PR #4636, opened 2018-03-05
add sbc_discovery

lazedo:


applications/ecallmgr/src/ecallmgr_discovery.erl
```diff
@@ -0,0 +1,177 @@
...
+%%% gen_server callbacks
+%%%=============================================================================
+
+%%------------------------------------------------------------------------------
+%% @doc Initializes the server
+%%
+%% @end
+%%------------------------------------------------------------------------------
+-spec init(list()) -> {'ok', state(), pos_integer()}.
+init([]) -> {'ok', #{startup => kz_time:now_s()}, ?MILLISECONDS_IN_SECOND}.
```
jamesaimonetti: Is there a reason not to just track the `startup` directly? Do you have plans for more state to be tracked here?

lazedo: not sure i understand your comment, but as always, feel free to improve it.
i need to keep the initial `startup` time to be used in `next_timeout(kz_time:elapsed_s(Startup))` so that the timer will increase (more often to less often checks)

jamesaimonetti: Sorry, let me restate (heh). State is a map here. But looking through the code, I only ever see the map unpacked to get the `startup` attribute. So why not just return `{'ok', kz_time:now_s(), ?MILLISECONDS_IN_SECOND}` and use it directly in the other functions?

lazedo: ah, right. can do that change, thanks

PR #4637, opened 2018-03-05
config less storage

lazedo:


core/kazoo_apps/src/kapps_maintenance.erl
```diff
@@ -980,36 +980,33 @@ handle_module_rename_doc(JObj) ->
...
     end.
 
-maybe_new({ok, Doc}) -> Doc;
-maybe_new(_) -> kz_json:new().
-
-get_config_document(Id) ->
-    kz_doc:public_fields(maybe_new(kapps_config:get_category(Id))).
+-spec validate_system_configs() -> [{kz_term:ne_binary(), _}].
+validate_system_configs() ->
+    Results = [ {Config, validate_system_config(Config)} || Config <- kapps_config_doc:list_configs() ],
```
jamesaimonetti: What about:

```erlang
[ {Config, Status}
  || Config <- kapps_config_doc:list_configs(),
     Status <- [validate_system_config(Config)],
     [] =/= Status
 ].
```

Save a second list interation.

core/kazoo_config/src/kapps_config.erl
```diff
@@ -503,12 +504,14 @@ get_zone_value(Category, _Node, Keys, Default, JObj) ->
...
-        'undefined' when Default /= 'undefined' ->
-            lager:debug("setting default for ~s ~p: ~p", [Category, Keys, Default]),
-            _ = set_default(Category, Keys, Default),
-            Default;
+        'undefined' -> Default;
+        Else -> Else
+    end;
+get_default_value(_Category, Keys, Default, JObj) ->
+    case kz_json:get_value([?KEY_DEFAULT | Keys], JObj) of
+        'undefined' when Default /= 'undefined' -> Default;
```
jamesaimonetti: Should use `=/=`

jamesaimonetti:
@lazedo let me know what you think of these changes. It returns the schema error JObjs that crossbar is used to manipulating (I know we're just logging it for now but eventually we could make it prettier at the least).

PR #4640, opened 2018-03-06
Master pr added utc time offset to cdr request

bradfordben:
Added the ability to request the 'datetime' field to be returned in local time zone. 

This is useful when requesting CDR's as a CSV. At the moment in kazoo ui the time zone conversion is done in kazoo ui to display a list of the cdrs in the browser as local time, but when you request the CSV the times retuned do not match the browsers list as they are returned in the server time zone. With this change the 'seconds from utc' can be provided and the datetime returned will be set accordingly.

### DIFF
--- a/applications/crossbar/src/modules/cb_cdrs.erl
+++ b/applications/crossbar/src/modules/cb_cdrs.erl
@@ -23,6 +23,10 @@
         ]).
 
 -export([load_chunked_cdr_ids/3]).
+-ifdef(TEST).
+    -export([handle_utc_time_offset/2]).
+-endif.
+
 
 -include("crossbar.hrl").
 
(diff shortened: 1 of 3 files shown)


applications/crossbar/src/modules/cb_cdrs.erl
```diff
@@ -403,49 +408,66 @@ csv_rows(Context) ->
...
+handle_utc_time_offset(Timestamp, Atom) when is_atom(Atom) -> Timestamp;
+handle_utc_time_offset(Timestamp, UTCSecondsOffset) ->
+    try offset_timestamp_from_UTC(Timestamp, kz_term:to_number(UTCSecondsOffset))
+    catch error:badarg -> Timestamp
+    end.
+
+-spec offset_timestamp_from_UTC(pos_integer(), integer()) -> pos_integer().
+offset_timestamp_from_UTC(Timestamp, UTCSecondsOffset) ->
+    lager:debug("Adjusting CDR timestamp with UTC Time Offset: ~p", [UTCSecondsOffset]),
+    LocalDateTime = calendar:gregorian_seconds_to_datetime(Timestamp),
```
jamesaimonetti: There is a requirement that servers within a cluster are all using UTC so timestamps from FreeSWITCH or Kamailio (and Kazoo nodes) should be UTC already. So the conversion should be direct from UTC to local TZ.

applications/crossbar/src/modules/cb_cdrs.erl
```diff
@@ -403,49 +408,66 @@ csv_rows(Context) ->
...
+col_recording_url(JObj, _Timestamp, _Context) -> kz_json:get_value([<<"recording_url">>], JObj, <<>>).
+col_media_recordings(JObj, _Timestamp, _Context) -> format_recordings(JObj).
+col_media_server(JObj, _Timestamp, _Context) -> kz_json:get_value(<<"media_server">>, JObj, <<>>).
+col_call_priority(JObj, _Timestamp, _Context) -> kz_json:get_value([?KEY_CCV, <<"call_priority">>], JObj, <<>>).
+
+col_reseller_cost(JObj, _Timestamp, _Context) -> kz_term:to_binary(reseller_cost(JObj)).
+col_reseller_call_type(JObj, _Timestamp, _Context) -> kz_json:get_value([?KEY_CCV, <<"reseller_billing">>], JObj, <<>>).
+
+-spec handle_utc_time_offset(pos_integer(), integer() | 'undefined') -> pos_integer().
+handle_utc_time_offset(Timestamp, Atom) when is_atom(Atom) -> Timestamp;
```
jamesaimonetti: The guard is too general; we know it will be `integer()` or `'undefined'` so let's match `'undefined'` here explicitly. If its ever not `integer() | 'undefined'` we want a crash because bad data is bad.

applications/crossbar/src/modules/cb_cdrs.erl
```diff
@@ -403,49 +408,66 @@ csv_rows(Context) ->
...
+handle_utc_time_offset(Timestamp, UTCSecondsOffset) ->
+    try offset_timestamp_from_UTC(Timestamp, kz_term:to_number(UTCSecondsOffset))
+    catch error:badarg -> Timestamp
```
jamesaimonetti: I wouldn't bother; if the client sends a non-integer value for the offset, their request should fail and they should be scolded! :) If the request succeeds but the data is "wrong" because the client expected a different value, then crossbar gets blamed. So I would just call `offset_timestamp_from_UTC/2` directly.

applications/crossbar/src/modules/cb_cdrs.erl
```diff
@@ -403,49 +408,66 @@ csv_rows(Context) ->
...
+    end.
+
+-spec offset_timestamp_from_UTC(pos_integer(), integer()) -> pos_integer().
+offset_timestamp_from_UTC(Timestamp, UTCSecondsOffset) ->
+    lager:debug("Adjusting CDR timestamp with UTC Time Offset: ~p", [UTCSecondsOffset]),
+    LocalDateTime = calendar:gregorian_seconds_to_datetime(Timestamp),
+    UTCDateTimeList = calendar:local_time_to_universal_time_dst(LocalDateTime),
+    UTCTimestamp = calendar:datetime_to_gregorian_seconds(lists:last(UTCDateTimeList)),
+    UTCTimestamp+UTCSecondsOffset.
```
jamesaimonetti: Spaces between args to `+/2`

applications/crossbar/test/cb_cdrs_test.erl
```diff
@@ -0,0 +1,44 @@
...
+    UTCTimestamp = calendar:datetime_to_gregorian_seconds(lists:last(UTCDateTimeList)),
+
+    [{"Verify the Atom 'undefined' returns unaltered timestamp"
+     ,?_assertEqual(LocalTimestamp, cb_cdrs:handle_utc_time_offset(LocalTimestamp, 'undefined'))
+     }
+    ,{"Verify the Atom 'true' returns unaltered timestamp"
+     ,?_assertEqual(LocalTimestamp, cb_cdrs:handle_utc_time_offset(LocalTimestamp, 'true'))
+     }
+    ,{"Verify the binary <<\"abc\">> returns unaltered timestamp as its NAN"
+     ,?_assertEqual(LocalTimestamp, cb_cdrs:handle_utc_time_offset(LocalTimestamp, <<"bla">>))
```
jamesaimonetti: This test should assert a crash occurs

applications/crossbar/test/cb_cdrs_test.erl
```diff
@@ -0,0 +1,44 @@
...
+%% time zones and offsets
+handle_utc_time_offset_test_() ->
+    %% Simulate a time stamp from a CDR
+    %% (generated in freeswitch depending on system TZ)
+    LocalTimestamp = 63687252084,
+
+    %% Depending on the systems timezone, Calculate the UTC time from the LocalTimestamp
+    LocalDateTime = calendar:gregorian_seconds_to_datetime(LocalTimestamp),
+    UTCDateTimeList = calendar:local_time_to_universal_time_dst(LocalDateTime),
+    UTCTimestamp = calendar:datetime_to_gregorian_seconds(lists:last(UTCDateTimeList)),
```
jamesaimonetti: Also, this unit test is not useful at all. It is literally copy/paste of the implementation. Instead, I would like to see something like:

```erlang
Tests = [{63687252084, 63687237684, -14400}],
[?_assertEqual(Local, handle_utc_time_offset(UTC, Offset)) || {UTC, Local, Offset} <- Tests].
```

jamesaimonetti:
@bradfordben just checking in to see if there were any questions or further discussion you wanted to have here.

bradfordben:
@jamesaimonetti Thanks for checking in, I just have not had a chance to get back to this yet. I will soon :) Thanks for your feedback.

bradfordben:
@jamesaimonetti Sorry for the delay on getting back to this. I have changes some things around as i feel its better this way. 

I added input data verification to allow a 404 response if the utc_offset is not a number and also assume the Time stamp is in UTC as you suggested. I found that if i crashed on the string to number conversion, crossbar would not return any response so i felt the data verification was worth doing. 

The test cases are probably not needed now as the function is so simple now but ive included them for now, I can remove them if you would prefer.

bradfordben:
@jamesaimonetti Was great to meet you at Kazoo Con and put a face to the name, I was wondering if you could let me know what is failing tests on your side?

bradfordben:
Thanks.

PR #4644, opened 2018-03-06
Faster module check

jamesaimonetti:


core/kazoo_stdlib/src/kz_module.erl
```diff
@@ -0,0 +1,47 @@
...
+is_exported(Module, Function, Arity)
+  when is_atom(Module),
+       is_atom(Function),
+       is_integer(Arity),
+       Arity >= 0 ->
+    case erlang:module_loaded(Module) of
+        'false' -> _ = code:ensure_loaded(Module), 'ok';
+        'true' -> 'ok'
+    end,
+    erlang:function_exported(Module, Function, Arity).
```
lazedo: why not use
```
   case ensure_loaded(Module) of
        'false' -> _ = code:ensure_loaded(Module), 'ok';
        _ -> erlang:function_exported(Module, Function, Arity)
    end.
```

core/kazoo_stdlib/src/kz_module.erl
```diff
@@ -0,0 +1,47 @@
...
+        ]).
+
+%% @doc lifted from erlang ML
+%% http://erlang.2086793.n4.nabble.com/What-is-the-fastest-way-to-check-if-a-function-is-defined-in-a-module-tp4723096p4723108.html
+-spec is_exported(module(), atom(), arity()) -> boolean().
+is_exported(Module, Function, Arity)
+  when is_atom(Module),
+       is_atom(Function),
+       is_integer(Arity),
+       Arity >= 0 ->
```
lazedo: maybe do `Module` convert here and accept `ne_binary`.
` Module = kz_term:to_atom(Mod, 'true'),`

applications/crossbar/src/api_util.erl
```diff
@@ -611,27 +611,19 @@ parse_path_tokens(Context, [Mod|T], Events) ->
...
+    Modules = [<<"cb_", Elem/binary>>
+              ,<<"cb_", Elem/binary, "_", ApiVersion/binary>>
+              ],
+
+    is_cb_module(Modules).
+
+-spec is_cb_module(kz_term:ne_binaries()) -> boolean().
+is_cb_module([]) -> 'false';
+is_cb_module([Module|Modules]) ->
+    'true' =:= (catch erlang:function_exported(kz_term:to_atom(Module), 'init', 0))
```
lazedo: look in `kz_module` comment. if suggestion accepted, then this would be 
```
kz_module:is_exported(Module, 'init', 0)
orelse is_cb_module(Modules).
```

applications/crossbar/src/modules/cb_sup.erl
```diff
@@ -205,9 +205,9 @@ does_resource_exist(ModuleBin, FunctionBin, Args) ->
     try {module_name(ModuleBin), kz_term:to_atom(FunctionBin)} of
         {Module, Function} ->
```
lazedo: i think there's no need to call  `kz_module:is_exported` when `Module` is `false`
if taking changes suggested in kz_module, this `fun` would call `kz_module:is_exported(ModuleBin, Function, Arity)`

core/kazoo_stdlib/src/kz_module.erl
jamesaimonetti: I would rather this be strict since it is new functionality to use.

applications/crossbar/src/api_util.erl
```diff
@@ -622,7 +622,7 @@ is_cb_module(Context, Elem) ->
 -spec is_cb_module(kz_term:ne_binaries()) -> boolean().
 is_cb_module([]) -> 'false';
 is_cb_module([Module|Modules]) ->
-    kz_module:is_exported(Module, 'init', 0)
+    'true' =:= (catch kz_module:is_exported(Module, 'init', 0))
```
lazedo: @jamesaimonetti it shouldn't crash, what am i missing ?

core/kazoo_stdlib/src/kz_module.erl
```diff
@@ -17,12 +17,15 @@ is_exported(Module, Function, Arity)
...
-        'true' -> 'ok'
-    end,
-    erlang:function_exported(Module, Function, Arity).
-
+    case ensure_loaded(Module) of
+        'false' -> 'false';
+        Module -> erlang:function_exported(Module, Function, Arity)
+    end;
+is_exported(Module, Function, Arity) ->
+    is_exported(kz_term:to_atom(Module)
```
lazedo: should we call kz_term:to_atom(Module, 'true') ?

jamesaimonetti: @lazedo I didn't want to since this is used in crossbar to check path tokens, so every token on the path will be prefixed with cb_ and tested here. I'd rather the `to_atom/1` fail and have the caller handle that

applications/crossbar/src/api_util.erl
lazedo: i see this in mkbusy test, was that the reason ? why are we trying to convert `cb_d247712c7b03b2114335aa13b95b0849` ?
```
erlang:binary_to_existing_atom(<<"cb_d247712c7b03b2114335aa13b95b0849">>, utf8)
{'EXIT',{{case_clause,[api_resource,<0.2725.0>,13,<0.3106.0>,badarg,[{erlang,binary_to_existing_atom,[<<"cb_d247712c7b03b2114335aa13b95b0849">>,utf8],[]},{kz_term,to_atom,1,[{file,"src/kz_term.erl"},{line,332}]},{kz_module,is_exported,3,[{file,"src/kz_module.erl"},{line,25}]},{api_util,is_cb_module,1,[{file,"src/api_util.erl"},{line,625}]},{api_util,'-parse_path_tokens/3-fun-4-',2,[{file,"src/api_util.erl"},{line,608}]},{lists,splitwith,3,[{file,"lists.erl"},{line,1413}]},{api_util,parse_path_tokens,...},...]]},...}}
```
if this call (with `cb_SOME_ID`)  can be avoided, then maybe the conversion in kz_module can use `kz_term:to_atom/2`
or... do the try catch in kz_module

core/kazoo_stdlib/src/kz_module.erl
lazedo: got it, and i makes sense (saw that in the mkbusy tests). maybe catch the exception here ? that would make the call to `kz_module:is_module` safe.

PR #4646, opened 2018-03-07
Navi - bringing custom push notifications to kazoo

bpar476:
Navi is a new Kazoo application that allows for custom push notifications to be sent to arbitrary users and devices that are interested in events within Kazoo. Navi can push notifications to iOS and Android devices using FCM and APNS.

Navi has a kapi module which can be used to push notifications, or navi_listener can be modified to respond to already existing kazoo events.

For more detail, refer to the Navi README.


lazedo:
@bpar476 what does this differ from the `pusher` app ?

kalda341:
@lazedo Pusher is entirely focused on push notifications on failover of calls to a device. Navi is meant to be far more general, and designed to handle user focused notifications for other alerts such as when a voicemail is received.

Rather than rewriting Pusher to fulfil our needs we decided it was better to create a new, more generic push notification application and leave Pusher for calling. In the future we intend to change Pusher to use Navi to relay push notifications.

Navi also supports FCM over the nearly discontinued GCM.

lazedo:
@kalda341 that assumption on pusher is wrong, it is mainly used for calls but not restricted to calls.  FCM is not a reason at all, its a dependency with a module. the rest is crossbar and vm. but more detailed discussion will come for sure. i don't see a reason to have 1+ app to handle push notifications and i don't see a good move for pusher to depend on another app

applications/crossbar/doc/push_notification_subscriptions.md
```diff
@@ -0,0 +1,64 @@
+### Push Notification Subscriptions
+
+#### About Push Notification Subscriptions
+
```
jamesaimonetti: Can you add some text here so folks looking at this doc get an idea of what the API is useful for?

applications/crossbar/priv/couchdb/schemas/system_config.navi.json
```diff
@@ -0,0 +1,85 @@
...
+                    {
+                        "properties": {
+                            "notification_type": {
+                                "enum": [
+                                    "apns"
+                                ]
+                            }
+                        },
+                        "required": [
+                            "certificate",
```
jamesaimonetti: Seems strange to require keys but not define their properties? What's happening here?

applications/crossbar/src/modules/cb_push_notification_subscriptions.erl
```diff
@@ -0,0 +1,215 @@
...
+post(Context, _, _) ->
+    crossbar_doc:save(Context).
+
+%%------------------------------------------------------------------------------
+%% @doc If the HTTP verb is DELETE, execute the actual action, usually a db delete
+%% @end
+%%------------------------------------------------------------------------------
+-spec delete(cb_context:context(), path_token(), path_token()) -> cb_context:context().
+delete(Context, _, _) ->
+    Id = kz_json:get_value(<<"id">>, cb_context:doc(Context)),
```
jamesaimonetti: Why was `crossbar_doc:delete(Context).` not used?

applications/navi/doc/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Architectural Overview
+Navi functionality is implemented across several modules:
+* cb\_push\_notification\_subscriptions
+* navi\_listener
+* navi\_module\_sup
+* navi\_notification\_server\_sup
+* nv\_apns
+* nv\_fcm
+* kapi\_navi
```
jamesaimonetti: AMQP APIs that are shared among apps should go in `core/kazoo_amqp/api` and be named generically, like `kapi_websockets` instead of `kapi_blackhole` even though `blackhole` does the websocket interactions.

If the intent is for other kazoo applications (like crossbar) to use the `kapi_navi` payloads, `kapi_navi` should move to core and be renamed to `kapi_push_notifications` or the like.

applications/navi/doc/README.md
```diff
@@ -0,0 +1,88 @@
...
+### nv\_apns
+This module contains the logic for sending push notifications to iOS apps. You will not need to interact with this module if everything is set up properly unless you want to add
+custom data to your notifications.
+
+### nv\_fcm
+This module contains the logic for sending push notifications to Android apps. You will not need to interact with this module if everything is set up properly unless you want to add
+custom data to your notifications.
+
+### kapi\_navi
+This module allows for applications other than navi to push notifications. It exposes a service which can make requests through amqp to Navi. It allows for pushing to a specific device and to pushing to all devices owned by a specific user.
```
jamesaimonetti: Ok yes, this should be moved to core and renamed. If it stays here and other applications use it, those applications will then depend on having navi installed and running on the same VM as the application. This could conflict with how folks choose to deploy Kazoo. No Kazoo application should depend on another being present in the same VM - all comms should be done via AMQP.

make/deps.mk
```diff
@@ -50,14 +54,18 @@ endif
 
 dep_amqp_client = hex 3.7.3
 dep_apns = git https://github.com/inaka/apns4erl.git 2.2.0
+dep_base64url = git https://github.com/dvv/base64url v1.0 ## inaka apns dependency
 dep_certifi = hex 0.3.0
+dep_chatterbox = git https://github.com/joedevivo/chatterbox v0.5.0 ## inaka apns dependency
```
jamesaimonetti: Do these not get pulled in as part of building apns?

bpar476:
I've tried to address each of the requests for changes with the latest update to the branch.

applications/crossbar/src/modules/cb_push_notification_subscriptions.erl
bpar476: I first tried using `crossbar_doc:delete` but that didn't work, the request would return successfully and say the document has been deleted but really it was still in the database without `pvt_deleted` being set and could be retrieved with subsequent gets.

applications/crossbar/priv/couchdb/schemas/system_config.navi.json
bpar476: The keys had descriptions further along in the schema, they are only simple strings. To improve the readability of the schema I have re-templated it so that the descriptions for the required keys for either branch of the anyOf are in that subschema.

applications/crossbar/doc/push_notification_subscriptions.md
bpar476: Added a brief description of the purpose that this crossbar endpoint serves.

applications/navi/doc/README.md
bpar476: I have renamed the kapi module kapi_push_notifications and moved it into kazoo_amqp/api, changed the event name to push and the event categories to device, and user. I've also updated navi_listener so that it will pick up the new events.

applications/crossbar/src/modules/cb_push_notification_subscriptions.erl
```diff
@@ -0,0 +1,215 @@
...
+normalize_view_results(JObj, Acc) ->
+    [kz_json:get_json_value(<<"value">>, JObj)|Acc].
+
+-spec subscription_does_exist(kz_term:ne_binary(), kz_term:ne_binary(), cb_context:context()) -> {'not_exists' | 'error', cb_context:context()} | {'exists', kz_term:api_object()}.
+subscription_does_exist(App, MobileDeviceId, Context) ->
+    ViewContext = crossbar_doc:load_view(?BY_APP_BY_DEVICE, [{'key', [cb_context:auth_user_id(Context), MobileDeviceId, App]}], Context, fun normalize_view_results/2),
+    case {cb_context:has_errors(ViewContext), cb_context:doc(ViewContext)} of
+        {'true', _} -> {'error', ViewContext};
+        {'false', []} -> {'not_exists', crossbar_util:response_bad_identifier(MobileDeviceId, Context)};
+        {'false', [Item|[]]=Doc} ->
```
jamesaimonetti: `[Item | []]` is the same as `[Item]`

applications/crossbar/src/modules/cb_push_notification_subscriptions.erl
```diff
@@ -0,0 +1,215 @@
...
+%% @doc Normalizes the results of a view
+%% @end
+%%------------------------------------------------------------------------------
+-spec normalize_view_results(kz_json:object(), kz_json:objects()) -> kz_json:objects().
+normalize_view_results(JObj, Acc) ->
+    [kz_json:get_json_value(<<"value">>, JObj)|Acc].
+
+-spec subscription_does_exist(kz_term:ne_binary(), kz_term:ne_binary(), cb_context:context()) -> {'not_exists' | 'error', cb_context:context()} | {'exists', kz_term:api_object()}.
+subscription_does_exist(App, MobileDeviceId, Context) ->
+    ViewContext = crossbar_doc:load_view(?BY_APP_BY_DEVICE, [{'key', [cb_context:auth_user_id(Context), MobileDeviceId, App]}], Context, fun normalize_view_results/2),
```
jamesaimonetti: Look into using `crossbar_view` and maybe that will address the issue for why you're not able to use `crossbar_doc:delete`. It seems awfully strange that `delete` didn't work for you and I would rather fix how Context is getting loaded on the read portion.

bpar476: Thanks for the tip. I'll try fix it! :)

noahmehl:
As discussed on the community call (thanks for that by the way!), @lazedo and @jamesaimonetti suggested that we try to merge Navi into Pusher.  My understanding is that there is agreement that Pusher should be more general purpose and its job should be to receive messages and push them (to FCM, GCM, or APNS).  As such there are a few items to figure out:

1. Move the existing message creation out of the Pusher module
2. Combine the code bases
3. Extend the kamailio pusher role

@lazedo if I have something incorrect, or am missing anything please let everyone know.  We'd really appreciate if you would give your recommendation on where to put some of this stuff (outside of Pusher/Navi).

PR #4651, opened 2018-03-12
Kazoo EDoc

icehess:
This is currently working. It is slow and huge in size (around 800MB), but for now it's okay (please agree with me on this).

Moving side bar to it own [web component](https://www.webcomponents.org/)  (or react component) takes more time than I expected. This can be done at later time since the whole sidebar is copied to every html file is consuming  over 500MB ! Also is painfully makes rendering slow.

Search feature is primitive, it doesn't understand non alphanum characters (like `cb_devices` won't work but `cb devices` will). Basically you have to provide a bunch simple of words. You can search base some field like `fun`, `app`, `module`, `type` and `desc` like `fun:*migrate*` (which will match any function name which has migrate, probably taking stemming into account too). Read more about searching [here](https://lunrjs.com/guides/searching.html).

This fine for now until I found a better library for full-text searching.

Generated documents are put in `doc/edoc` folder. Intermediate JSON files containing the each module doc are store temporary into `doc/edoc/tmp` which later will be read again to render the HTML.

Entry point for generating edoc is `ERL_LIBS=deps/:core/:applications/ scripts/kz_docgen.escript`.

You have to build kazoo before, and installing `nodejs` and `sudo npm install -g lunr` (If you want to have search index file generated).


core/kazoo_ast/src/kz_edoc_layout.erl
```diff
@@ -0,0 +1,1240 @@
...
+is_block_element(_) -> false.
+
+-spec par_flush([#xmlElement{}], [exporty_thing()], [exporty_thing()]) -> [exporty_thing()].
+par_flush(Es, [], Bs) ->
+    make_paraghraph(Es, [], Bs);
+par_flush(Es, As, Bs) ->
+    make_paraghraph(Es, [], Bs ++ [{p, As}]).
+
+-spec get_elem(atom(), [#xmlElement{}]) -> [#xmlElement{}].
+get_elem(Name, [#xmlElement{name = Name} = E | Es]) ->
```
jamesaimonetti: Can you look at using/expanding kz_xml for these types of operations? Would be good to expand that functionality for others to build on as well.

icehess: I can, but don't forget that this file is exactly `edoc_layout` module from `edoc` app itself with some changes like return proplist instead of html and specs added and some other minor changes.

And currently it doesn't have any dependency on anything other than erlang itself.

core/kazoo_ast/src/kz_edoc_layout.erl
```diff
@@ -0,0 +1,1240 @@
...
+
+-spec seq(fun((#xmlElement{}) -> exporty_thing()), [#xmlElement{}], string(), [string()]) -> [exporty_thing()].
+seq(F, [E], _Sep, Tail) ->
+    F(E) ++ Tail;
+seq(F, [E | Es], Sep, Tail) ->
+    F(E) ++ [Sep] ++ seq(F, Es, Sep, Tail);
+seq(_F, [], _Sep, Tail) ->
+    Tail.
+
+-spec filter_empty([{atom(), list() | iolist() | binary()}]) -> [{atom(), list() | iolist() | binary() | tuple() | atom() | boolean()}].
```
jamesaimonetti: 2 questions:

1. Why not use `kz_term:proplist()` and `kz_term:proplist_property()` here and in `is_not_empty/1`?
2. Why not use `props:filter_empty/1` (or `props:filter/2` passing in `is_not_empty/1`) instead?

I think the discrepancy in the return type vs the input type is causing dialyzer (via `scripts/dialyze-changed.bash`) to complain that `kz_edoc_layout.erl:353: Function function/3 has no local return`. If you remove the call to `filter_empty/1` in that function then dialyzer is happy. When I changed the types to use our proplist types, dialyzer is happy.

core/kazoo_ast/src/kz_edoc_layout.erl
jamesaimonetti: Is there a reason to avoid Kazoo modules as dependencies though?

icehess:
Closing this for now until I fixed the remaining crash issues in the layout module.

PR #4653, opened 2018-03-13
fix file name for dev VM args

k-anderson:


jamesaimonetti:
Honestly, I'd rather remove those scripts and run the release:

`make build-dev-release` then `make release REL=kazoo_apps` and `make release REL=ecallmgr` get you there (including having reloader run).

PR #4675, opened 2018-03-20
KAZOO-5800: Re-check for call restrictions after shortdial correction

harenson:


applications/callflow/src/cf_route_win.erl
```diff
@@ -375,6 +374,23 @@ get_incoming_security(Call) ->
...
+    DefaultEndpointId = kapps_call:authorizing_id(Call),
+    kapps_call:kvs_fetch(?RESTRICTED_ENDPOINT_KEY, DefaultEndpointId, Call).
+
+-spec include_denied_call_restrictions(kapps_call:call()) -> kapps_call:call().
+include_denied_call_restrictions(Call) ->
+    case kz_endpoint:get(get_endpoint_id(Call), Call) of
+        {'error', _R} ->
+            Call;
+        {'ok', JObj} ->
+            CallRestriction = kz_json:get_value(<<"call_restriction">>, JObj),
```
jamesaimonetti: Use the more specific getters when you can: `get_json_value/2` (maybe with a default `kz_json:new()`?) and `get_ne_binary_value/2` on the next line.

applications/crossbar/priv/couchdb/schemas/kapi.offnet_resource.req.json
```diff
@@ -77,6 +77,9 @@
         "Custom-SIP-Headers": {
             "type": "object"
         },
+        "Denied-Call-Restrictions": {
```
jamesaimonetti: Are they really a string? Or should this be an object? You can hint at it in the kapi module using a validator or you can manually correct this schema.

jamesaimonetti:
@harenson i refactored a bit; mind checking if things are still working on your tests? Will merge if you agree :)

PR #4685, opened 2018-03-21
HELP-35856-4.2: use conference name as ID if ID is missing

jamesaimonetti:
Prior to the dynamic conference profile work, conference names were
used when interacting with FreeSWITCH. The field used was shifted to
the 'id' in the kapps_conference record but the 'name' was not used as
a backup if id was missing when building the kapps_conference record.

What would happen is, a callflow with a dynamic conference doc in the
action (defining "name" but not "id") would get "id" set to
rand_hex(8) instead of the value of "name". Conference dial API would
put another caller in the "name" conference and an API command to put
the original caller in would fail as the rand_hex was used instead.

If "id" isn't defined, we now default to using the "name".


jamesaimonetti:
See master PR #4684

PR #4691, opened 2018-03-26
Revert "Change Erlang version to 19.3.6.5"

lazedo:
Reverts 2600hz/kazoo#4443


efine:
I’m curious why this was reverted. 19.3 has some serious bugs that have affected us already.

lazedo:
@efine check again, it wasn't reverted but this commit made ci fail for any new PRs.
anyhow, 19.3.6.5 should be removed as 19.3 now gets 19.3.6.7

efine:
> check again, it wasn't reverted but this commit made ci fail for any new PRs.

My apologies, I saw commit e85718a, which clearly changes it to 19.3, but that commit was never merged to master, which I missed.

> anyhow, 19.3.6.5 should be removed as 19.3 now gets 19.3.6.7

Not sure what does that, but this statement is not entirely accurate if you consider `kerl`. Running `kerl build 19.3 19.3` with the github backend gets you `19.3`, not `19.3.6.7`. I just verified this; `kerl` pulled the 19.3 tarball from github.

So if anything, we should change it to `19.3.6.7`at some point, but this requires the installation of that version of Erlang in our internal build system before we can do that.

lazedo:
this was already fixed

lazedo:
@efine for your reference, the fix was this https://github.com/2600hz/kazoo/pull/4693

PR #4707, opened 2018-03-27
HELP-35943: log non-jobj and return default

k-anderson:


k-anderson:
This is only a 4.2 change

PR #4712, opened 2018-03-27
KAZOO-5866 handle invalid account_id in url

lazedo:


applications/crossbar/src/api_resource.erl
```diff
@@ -405,15 +405,23 @@ malformed_request(Req, Context, ?HTTP_OPTIONS) ->
...
     case props:get_value(<<"accounts">>, cb_context:req_nouns(Context)) of
-        [<<>>] -> api_util:stop(Req, cb_context:add_system_error(<<"account_id is missing">>, Context));
-        [AccountId] ->
-            Context1 = cb_accounts:validate_resource(Context, AccountId),
+        'undefined' ->
+            {'false', Req, Context};
+        [] ->
+            {'false', Req, Context};
+        [?MATCH_ACCOUNT_RAW(AccountId) | More] ->
+            Context1 = cb_accounts:validate_resource(Context, [AccountId] ++ More),
```
jamesaimonetti: Could just do
```erlang
    [?MATCH_ACCOUNT_RAW(_AccountId) | _]=AccountArgs ->
        Context1 = cb_accounts:validate_resource(Context, AccountArgs),
```

Then don't need to rebuild the list.

lazedo: i'm dumb

PR #4728, opened 2018-03-28
HELP-35945: address race condition with conference profiles

jamesaimonetti:
When dialing out via conference dial, the caller is put into the
conference at the same time as the conf_participant is being spun
up. As such, the conference profile request is missed, delaying the
conference setup until the config request times out.

This adds a default config handler for all conference configs as well
as sets the participant up to respond (hopefully sooner).

Also added was the ability to specify the profile_name on the
conference dial API command, to improve caching for dynamic
conferences.


jamesaimonetti:
see 4.2 in #4727

PR #4738, opened 2018-03-30
handle mixed array of terms

lazedo:


jamesaimonetti:
```erlang
(pdx@pdx.2600hz.com)8> JObj = kz_json:from_list([{<<"foo">>, [kz_json:new()]}]).
{[{<<"foo">>,[{[]}]}]}
(pdx@pdx.2600hz.com)9> kz_json:from_map(kz_json:to_map(JObj)).
{[{<<"foo">>,{[{[]}]}}]}
(pdx@pdx.2600hz.com)10> kz_json:from_map(kz_json:to_map(JObj)) =:= JObj.
false
```

core/kazoo_stdlib/test/kz_json_test.erl
```diff
@@ -832,13 +832,32 @@ set_value_test_() ->
...
+                                                         ])}])).
 
 -define(FROM_MAP_MAP_1, #{options => [{opt1,<<"my-opt1">>},{opt2,<<"my-opt2">>}]}).
 -define(FROM_MAP_MAP_2, #{outer_key => #{inner_key => <<"inner_value">>}}).
 -define(FROM_MAP_MAP_3, #{outer_key => #{inner_key => <<"inner_value">>
                                         ,inner_options => [{opt1,<<"my-opt1">>}
                                                           ,{opt2,<<"my-opt2">>}
                                                           ]}}).
+-define(FROM_MAP_MAP_4, #{outer_key => #{inner_key => <<"inner_value">>
+                                        ,inner_options => [{opt1,<<"my-opt1">>}
```
jamesaimonetti: Is the intent here to be a proplist? In which case `<<"anykey">>` should be treated as if it was really `{<<"anykey">>, true}`?

jamesaimonetti:
If i change `recursive_from_map` to be:

```erlang
    case lists:any(fun is_json_object/1, Res) of
        'true' -> [maybe_tuple_to_json(R) || R  <- Res];
        'false' -> from_list(Res)
    end;
```

instead of the `is_tuple/1` call, then the above works but other tests fail. So...dunno atm

core/kazoo_stdlib/test/kz_json_test.erl
lazedo: not a proplist, and that's exactly what we want to show
[{field, <<"account_id">>}
,<<"&realm=">>
,{field, <<"account_realm">>}
]

lazedo:
@jamesaimonetti we may need to fix `to_map`

core/kazoo_stdlib/test/kz_json_test.erl
jamesaimonetti: I guess my thought is this list for inner_options doesn't have a valid JSON representation then.

lazedo:
@jamesaimonetti this works for me for the from_map(to_map()) example above.
```
recursive_from_map(List) when is_list(List) ->
    Res = [recursive_from_map(Item) || Item <- List],
    case lists:all(fun is_tuple/1, Res)
        andalso Res of
        'false' -> [maybe_tuple_to_json(R) || R  <- Res];
        [?JSON_WRAPPER(_)] -> Res;
        _ -> from_list(Res)
    end;
```

core/kazoo_stdlib/test/kz_json_test.erl
lazedo: inner_options is a list of terms, when all elements are tuple, we transform it with from_list(), when they're not, we need to ensure that each element is a valid json element which a tuple is is not and must be transformed with from_list([tuple]) on the element.

jamesaimonetti:
@lazedo 

```erlang
(pdx@pdx.2600hz.com)4> JObj = {[{<<0>>,[{[]},{[]}]}]}.
{[{<<0>>,[{[]},{[]}]}]}
(pdx@pdx.2600hz.com)5> kz_json:to_map(JObj).
#{<<0>> => [#{},#{}]}
(pdx@pdx.2600hz.com)6> kz_json:from_map(v(5)).
{[{<<0>>,{[{[]},{[]}]}}]}
(pdx@pdx.2600hz.com)7>
(pdx@pdx.2600hz.com)7> JObj =:= v(6).
false
```

We need something more general. Plus I think to_map is doing it right.

jamesaimonetti:
If you `make proper` in the kazoo_stdlib you can test the property yourself:

`proper:quickcheck(kz_json_test:prop_to_map(), 15000).`

jamesaimonetti:
@lazedo I rebased and added a commit to check for raw (non-JSON_WRAPPER) tuples. All tests look to be passing.

PR #4743, opened 2018-04-02
handle when channel goes down while answering

jamesaimonetti:
observed that when a db operation takes longer than expected saving
the slot, the channel_destroy will be waiting for the cf_park when it
goes into b_answer. However, because the code ignored the return of
b_answer, the cf_park process would end up blocking in b_prompt. This
is because the ecallmgr_call_control process was gone and would not
respond with an error (and obviously the channel_destroy wasn't coming
again).

Another issue to address is kapps_call_command:wait_for_noop/X uses
'infinity' for its timeout when receiving events. Should make it big
but not forever.


jamesaimonetti:
See 4.2 PR at #4742

PR #4748, opened 2018-04-03
[4.2] fix confusion with dollars/units and update tests

jamesaimonetti:
the proper tests actually would have caught this but unfortunately we
don't have a place to run them. My thought is to put them in CircleCI
during the running of the release. I'll add that to master though.


jamesaimonetti:
see master version at #4751

PR #4749, opened 2018-04-03
[4.2] HELP-36124: handle crash in cb_cdrs to avoid returning view results

icehess:


applications/cdr/src/cdr_channel_destroy.erl
```diff
@@ -101,7 +107,8 @@ update_pvt_parameters('undefined', _, JObj) ->
             ],
     kz_doc:update_pvt_parameters(JObj, ?KZ_ANONYMOUS_CDR_DB, Props);
 update_pvt_parameters(AccountId, Timestamp, JObj) ->
-    AccountMODb = kz_util:format_account_id(AccountId, Timestamp),
+    CorrectTimetamp = kz_json:get_ne_binary_value(<<"Interaction-Time">>, JObj, Timestamp),
```
jamesaimonetti: This makes no sense to me. I think you should be using get_integer_value/3, no?

applications/cdr/src/cdr_channel_destroy.erl
```diff
@@ -101,7 +107,8 @@ update_pvt_parameters('undefined', _, JObj) ->
...
+    AccountMODb = kz_util:format_account_id(AccountId, CorrectTimetamp),
```
jamesaimonetti: This will crash if "Interaction-Time" is defined as CorrectTimestamp will be a binary and will crash this function call:

```erlang
1> kz_util:format_account_id(kz_binary:rand_hex(16), kz_term:to_binary(kz_time:now_s())).
** exception error: no function clause matching kz_util:format_account_id(<<"bff45dfbc3c51765e882c0b072e2f3a7">>,
                                                                          <<"63690098832">>) (src/kz_util.erl, line 186)
```

PR #4752, opened 2018-04-04
KAZOO-5856: Add validation check for users endpoint in order to make sure the given hotdesk.id is unique

harenson:


applications/crossbar/src/modules_v2/cb_users_v2.erl
```diff
@@ -518,6 +520,39 @@ check_user_schema(UserId, Context) ->
...
+
+-spec is_hotdesk_id_unique(kz_term:api_binary(), kz_term:api_binary(), kz_term:ne_binary()) -> boolean().
+is_hotdesk_id_unique(AccountDb, UserId, HotdeskId) ->
+    ViewOptions = [{'key', HotdeskId}],
+    case kz_datamgr:get_results(AccountDb, ?LIST_BY_HOTDESK_ID, ViewOptions) of
+        {'ok', []} -> 'true';
+        {'ok', [JObj|_]} -> kz_doc:id(JObj) =:= UserId;
+        _Else ->
+            lager:debug("error ~p checking view ~p in ~p", [_Else, ?LIST_BY_HOTDESK_ID, AccountDb]),
+            'false'
```
jamesaimonetti: This is going to be confusing to a client when Couch is temporarily unable to respond, or some other situation. I think differentiating between not unique and system error is worthwhile here. Maybe return `boolean() | {'error', any()} and update Context with the error (crossbar_doc:handle_datamgr_errors).

applications/crossbar/src/modules_v2/cb_users_v2.erl
```diff
@@ -518,6 +520,40 @@ check_user_schema(UserId, Context) ->
...
+
+-spec is_hotdesk_id_unique(kz_term:api_binary(), kz_term:api_binary(), kz_term:ne_binary()) -> boolean() | {'error', any()}.
+is_hotdesk_id_unique(AccountDb, UserId, HotdeskId) ->
+    ViewOptions = [{'key', HotdeskId}],
+    case kz_datamgr:get_results(AccountDb, ?LIST_BY_HOTDESK_ID, ViewOptions) of
+        {'ok', []} -> 'true';
+        {'ok', [JObj|_]} -> kz_doc:id(JObj) =:= UserId;
+        Else ->
+            lager:debug("error ~p checking view ~p in ~p", [Else, ?LIST_BY_HOTDESK_ID, AccountDb]),
+            {'error', Else}
```
jamesaimonetti: Won't `Else` be `{error, Something}`? So won't this become `{error, {error, Something}}`?

applications/crossbar/src/modules_v2/cb_users_v2.erl
```diff
@@ -518,6 +520,40 @@ check_user_schema(UserId, Context) ->
...
+        'true' ->
+            Context;
+        'false' ->
+            lager:debug("Hotdesk.id ~p is already used", [HotdeskId]),
+            non_unique_hotdesk_id_error(Context, HotdeskId);
+        {'error', DatamgrError} ->
+            crossbar_doc:handle_datamgr_errors(DatamgrError, HotdeskId, Context)
+    end.
+
+-spec is_hotdesk_id_unique(kz_term:api_binary(), kz_term:api_binary(), kz_term:ne_binary()) -> boolean() | {'error', any()}.
```
jamesaimonetti: Should be `boolean() | kz_datamgr:data_error().`

applications/crossbar/src/modules_v2/cb_users_v2.erl
harenson: Good catch!

PR #4759, opened 2018-04-05
Fix typo in doc, 'tyre' -> 'true'

Igneous:
Not much of a description necessary here. Just something silly I came across in the crossbar docs.

Cheers.


jamesaimonetti:
Merged, thanks!

PR #4778, opened 2018-04-10
Fixed tiff2pdf call on Alpine dist

sergey-safarov:
On Alpine options must goes before input file


lazedo:
@sergey-safarov doesn't this break other platforms ?

sergey-safarov:
I think no.
Alpine use small, but not so much flexible command line parser. Other platform think use more flexible parser.

lazedo:
@sergey-safarov i'll give a shot and see if it breaks for other platforms

lazedo:
@sergey-safarov @icehess thanks for pr and testing this.

PR #4794, opened 2018-04-13
[4.2] HELP-36339: add default description to debit transaction

icehess:
master: #4793


lazedo:
@icehess i think we should enforce that `description` is supplied instead of adding the default.

PR #4802, opened 2018-04-18
std-tests: rename *_test.erl tests files to eunit's standard *_tests.erl

fenollp:
This is to update & split up https://github.com/2600hz/kazoo/pull/4107 into smaller PRs.
Rebar3 runs EUnit tests found in src/ and test/ but when in test/ the modules have to match that pattern.

I used these commands to create the PR:
```shell
for path in $(git ls-files | \grep _test.erl); do echo $path; git mv "$path" "${path%%_test.erl}"_tests.erl; done
ged 's%(module.+)_test\)%\1_tests)%'

# `ged` is one of my terse/marvelous bash aliases:
ged () { 
    local sprf="$1";
    shift;
    [[ "$sprf" == 's%'* ]] || return 10;
    local p="$(echo "$sprf" | cut -d% -f2)";
    [[ "$p" == '' ]] && return 11;
    if git status --porcelain | grep --color=auto -n '^ ' | grep --color=auto -n .; then
        return 12;
    fi;
    local git_grep='git --no-pager grep -IE';
    local f="$(echo "$sprf" | cut -d% -f4)";
    [[ "$f" == *'i'* ]] && local git_grep="$git_grep -i";
    if $git_grep --files-with-matches "$p" $@; then
        $git_grep --files-with-matches --null "$p" $@ | xargs -P "$(getconf _NPROCESSORS_ONLN)" --null --no-run-if-empty sed -i -r "$sprf";
    else
        return 1;
    fi
}
```
:)


claytonfinney:
The return of the golden boy?

k-anderson:
@fenollp howdy!
"terse/marvelous" lol 

Hope all is well, appreciate the PR!  We will need to figure out why CI seems to think something is wrong for unrelated changes...  Just wanted to say hi ;)

Thanks!

fenollp:
Alright alright alright!
Hi! Everything is great, hope the same to you :)
I fixed the tests. I'll be back for more once that's merged :)

jamesaimonetti:
Merged, et cimer gros

PR #4807, opened 2018-04-20
KAZOO-5887: Allow emergency calls regardless of restrictions

harenson:


applications/ecallmgr/src/ecallmgr_fs_authz.erl
```diff
@@ -416,3 +417,8 @@ rating_req(CallId, Props) ->
                            ,{<<"Authorizing-Type">>, kzd_freeswitch:authorizing_type(Props)}
                             | kz_api:default_headers(?APP_NAME, ?APP_VERSION)
                            ]).
+
+-spec is_emergency_number(kzd_freeswitch:data()) -> authz_reply().
+is_emergency_number(Props) ->
+    <<"emergency">> =:= knm_converters:classify(kzd_freeswitch:callee_id_number(Props))
```
jamesaimonetti: I think you should classify on the Request/To (using `kzd_freeswitch:to_did(Props)` I think).

PR #4812, opened 2018-04-25
[master] allow list service plan editable field without account id in path

icehess:
4.2: #4811

### DIFF
--- a/applications/crossbar/src/modules/cb_service_plans.erl
+++ b/applications/crossbar/src/modules/cb_service_plans.erl
@@ -8,6 +8,7 @@
 -module(cb_service_plans).
 
 -export([init/0
+        ,authorize/2
         ,allowed_methods/0, allowed_methods/1, allowed_methods/2
         ,resource_exists/0, resource_exists/1, resource_exists/2
         ,content_types_provided/1 ,content_types_provided/2, content_types_provided/3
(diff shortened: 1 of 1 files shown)


applications/crossbar/src/modules/cb_service_plans.erl
```diff
@@ -58,14 +59,31 @@
...
+%% allowed to access the resource, or false if not.
+%% @end
+%%------------------------------------------------------------------------------
+-spec authorize(cb_context:context(), path_token()) -> boolean().
+authorize(Context, _) -> is_authorize(Context, cb_context:req_verb(Context), cb_context:req_nouns(Context)).
+
+-spec is_authorize(cb_context:context(), req_verb(), req_nouns()) -> boolean() | {'halt', cb_context:context()}.
+is_authorize(_Context, ?HTTP_GET, [{<<"service_plans">>, ?EDITABLE}]) ->
+    'true';
+is_authorize(_Context, ?HTTP_GET, [{<<"service_plans">>, _} | _]) ->
```
k-anderson: @icehess What is this function clause allowing?

icehess: other service plan path tokens (`[{<<"service_plans">>, ?CURRENT} | _]`, ...)

PR #4813, opened 2018-04-25
handle short-lived calls in call-control - 4.2

lazedo:


applications/ecallmgr/src/ecallmgr_call_control.erl
```diff
@@ -427,7 +437,10 @@ call_control_ready(#state{call_id=CallId
              | kz_api:default_headers(?APP_NAME, ?APP_VERSION)
             ],
     lager:debug("sending control usurp for ~s", [FetchId]),
-    kapi_call:publish_usurp_control(CallId, Usurp).
+    kapi_call:publish_usurp_control(CallId, Usurp);
+call_control_ready('false', _) ->
+    lager:info("call is not in the channels cached, short lived call?"),
```
jamesaimonetti: s/cached/cache

PR #4819, opened 2018-04-29
Added convertapi

sergey-safarov:


applications/fax/src/fax_worker.erl
```diff
@@ -845,11 +847,6 @@ prepare_contents(JobId, RespHeaders, RespContent) ->
 -spec prepare_contents(kz_term:ne_binary(), kz_term:ne_binary(), kz_term:ne_binary(), kz_term:ne_binary()) ->
                               {'ok', kz_term:ne_binary()} |
                               {'error', kz_term:ne_binary()}.
-prepare_contents(<<"image/tiff">>, JobId, RespContent, TmpDir) ->
```
sergey-safarov: TIFF files also must be converted to TIFF group 4 format and set to proper resolution

applications/fax/src/fax_worker.erl
```diff
@@ -865,11 +862,11 @@ prepare_contents(<<"application/pdf">>, JobId, RespContent, TmpDir) ->
     end;
 
 prepare_contents(<<"image/", SubType/binary>>, JobId, RespContent, TmpDir) ->
-    InputFile = list_to_binary([TmpDir, JobId, ".", SubType]),
+    InputFile = list_to_binary([TmpDir, JobId, "-input.", SubType]),
```
sergey-safarov: When input is TIFF file, then input and output file must have different filename

applications/fax/src/fax_worker.erl
```diff
@@ -879,11 +876,26 @@ prepare_contents(<<"image/", SubType/binary>>, JobId, RespContent, TmpDir) ->
...
 
 prepare_contents(<<?OPENXML_MIME_PREFIX, _/binary>> = CT, JobId, RespContent, TmpDir) ->
-    convert_openoffice_document(CT, TmpDir, JobId, RespContent);
+    Secret = ?CONVERTAPI_SECRET,
+    case Secret of
+        'undefined' -> convert_openoffice_document(CT, TmpDir, JobId, RespContent);
+        _ -> convert_msoffice_document(CT, TmpDir, JobId, RespContent, Secret)
+    end;
+
+prepare_contents(<<?OPENOFFICE_MIME_PREFIX, _/binary>> = CT, JobId, RespContent, TmpDir) ->
```
sergey-safarov: Added openoffice format conversion.
May other formats is supported by [ConvertAPI service](https://www.convertapi.com/) but many is died and unknown.
Additional mime format is added as need.

applications/fax/src/fax_worker.erl
```diff
@@ -908,6 +920,41 @@ convert_openoffice_document(CT, TmpDir, JobId, RespContent) ->
...
+    lager:debug("attemting to convert document using convertapi for JobId: ~s", [JobId]),
+    case kz_http:post(kz_term:to_list(RequestURL), Headers, Body) of
+        {'ok', 200, _RespHeaders, RespBody} ->
+            JObj = kz_json:decode(RespBody),
+            case kz_json:is_defined(<<"Files">>, JObj) of
+                'true' ->
+                    lager:debug("jobid ~s converted successfully using convertapi", [JobId]),
+                    [File|_] = kz_json:get_list_value(<<"Files">>, JObj),
+                    FileDataBase64 = kz_json:get_value(<<"FileData">>, File),
+		    prepare_contents(<<"image/tiff">>, JobId, base64:decode(FileDataBase64), TmpDir);
```
sergey-safarov: Resulted ConverAPI TIFF file is have RGB color space and need to prepare to before passing to FreeSwitch/SpanDSP.
Here is simple call other content converted and pass resulted ConverAPI TIFF file.

applications/fax/src/fax.hrl
```diff
@@ -88,9 +88,10 @@
           "-sOutputFile=~s -- ~s > /dev/null 2>&1"
           "&& echo -n success"
         >>).
--define(CONVERT_IMAGE_CMD, <<"convert -density 204x98 "
+-define(CONVERT_IMAGE_CMD, <<"convert ~s -resample 204x98 "
```
sergey-safarov: `-density` keyword is overrides  image density in image file. Correct usage `-resample`
Also as I understand `convert` utility options input file must goes before options of output file.

applications/fax/src/fax.hrl
```diff
@@ -88,9 +88,10 @@
...
                              "-units PixelsPerInch "
-                             "-size 1728x1078 ~s ~s > /dev/null 2>&1"
+                             "-compress group4 "
```
sergey-safarov: Added color space conversion option.

applications/fax/src/fax_worker.erl
```diff
@@ -879,11 +876,26 @@ prepare_contents(<<"image/", SubType/binary>>, JobId, RespContent, TmpDir) ->
     end;
 
 prepare_contents(<<?OPENXML_MIME_PREFIX, _/binary>> = CT, JobId, RespContent, TmpDir) ->
-    convert_openoffice_document(CT, TmpDir, JobId, RespContent);
+    Secret = ?CONVERTAPI_SECRET,
```
jamesaimonetti: I would simply these clauses to not bind `Secret` in the outer scope:

```erlang
case ?CONVERTAPI_SECRET of
    'undefined' -> ...;
    Secret -> ....
end;
```

applications/fax/src/fax_worker.erl
```diff
@@ -908,6 +920,41 @@ convert_openoffice_document(CT, TmpDir, JobId, RespContent) ->
...
+                                         {'ok', kz_term:ne_binary()} |
+                                         {'error', kz_term:ne_binary()}.
+convert_document_convertapi(CT, TmpDir, JobId, RespContent, Secret) ->
+    Extension = kz_mime:to_extension(CT),
+    FileName = list_to_binary([JobId, ".", Extension]),
+    RequestURL = list_to_binary([?CONVERTAPI_URL, "/", Extension, "/to/tiff?ImageResolutionH=204&ImageResolutionV=392&ScaleImage=true&ScaleProportions=true&ScaleIfLarger=true&ImageHeight=4312&ImageWidth=1728&Secret=", Secret]),
+    Boundary = kz_term:to_hex_binary(crypto:strong_rand_bytes(24)),
+    ContentTypeHeaderValue = list_to_binary(["multipart/form-data; boundary=", Boundary]),
+    Headers = [{"Content-Type", ContentTypeHeaderValue}],
+    Body = list_to_binary(["--", Boundary
```
jamesaimonetti: Does `Body` need to be a `binary()`? Or can it remain an `iolist()` into `kz_http`?

lazedo:
i have some reservations concerning this pr. it is adding a feature for a paid service and that is being incorporated inside fax_worker. a `convert api` should be added with some sort of abstraction for which modules could be developed to support vendors like `convertapi`. also some of the conversion assumptions  are changing behaviour and must be explicitly demonstrated as not working, which in this case they should be submitted in different pull request. if these `default`commands are indeed wrong then a replace setting must be done if the existing value is the previous default

bazemoreb:
Hi Luis - I can't comment on the technical piece, but our intent with sponsoring this was to give people the ability to deal with clients that fax. doc files, excel files, etc.  The convertapi has a free basic tier that actually supports a good amount of conversions, and the service supports converting almost any file type imaginable.  My 2 cents here only. ;)

lazedo:
@bazemoreb @sergey-safarov  Hi, thanks for the comments & contribution. i'll try to work on this during this week.

sergey-safarov:
Luis (@lazedo) i updated PR to apply dynamic converter call.
Could you look this update.
In short time i will test PR.

swysor:
@sergey-safarov
FYI, I am currently building a converter library on kazoo/core that will can be used by multiple apps for fax file conversions. You may want to hold off a bit until this gets merged and then rebase your changes off of that branch, I think it mostly wont impact what you have done, but you will be able to call into kz_converter:fax/3 for any local file conversions required. 

Also, thank you for the input on the tiff file conversion formats command fixes, I am testing integrating these changes into the new kz_fax_converter.

sergey-safarov:
Hello Sean @swysor 
Where i be able look your dev branch.

swysor:
@sergey-safarov SO my branch is here https://github.com/2600hz/kazoo/compare/HELP-36774, i still have some work to do but the converter is working and the interface should not change. I will update you when this gets merged. Thank you!

swysor:
The kazoo_convert library is now available via master. If you want to implement the convertAPI via a module there, make sure you use the behavior gen_kz_converter for the module. Then it would be enabled by configuring the fax_converter module in the kazoo_convert config to use the convertapi module instead of the kz_fax_converter module (the default converter module).

jamesaimonetti:
@sergey-safarov have you had a chance to check out all the converter work @swysor has done? Is this PR still necessary in light of that work?

sergey-safarov:
@jamesaimonetti thank for this reminder
I will rebase on master of this PR. But little busy next week.

bazemoreb:
I know @swysor has been working like crazy on faxes - I didn't see that his work was able to convert a word document into a fax-able item, though.  That is what this work was for.

swysor:
It does support docx if you have libreoffice installed. We kind of treat that like an alpha feature though, but it works.

bazemoreb:
Oh interesting - OK, if that is the case then we can cancel this commit @sergey-safarov .  (provided it actually does work with libre, we'll want to test)

swysor:
Sounds good. The docs for this are here https://github.com/2600hz/kazoo/blob/master/core/kazoo_convert/doc/fax_converter.md#openoffice-compatible-to-pdf

sergey-safarov:
@bazemoreb two things
1) libreoffice may break doc formating (you can test this by opening docs on your PC);
2) convert API support many more doc formats than libreoffice.

bazemoreb:
Spoke to the team, and agree with you @sergey-safarov - would be better to not have to run.  Let us know when rebased?

sergey-safarov:
Hello @bazemoreb I ported convertapi feature to current master.
I created https://github.com/2600hz/kazoo/pull/5102 and suggest to continue discussion here.

PR #4820, opened 2018-04-29
Update docs and add stability level to schema properties

icehess:


applications/crossbar/priv/couchdb/schemas/accounts.json
```diff
@@ -73,13 +73,15 @@
         "enabled": {
             "default": true,
             "description": "Determines if the account is currently enabled",
+            "support_level": "supported",
```
lazedo: @icehess didn't we agree on assuming `supported` if not defined ?

icehess: @lazedo for checking yes, but here @claytonfinney and @christopherni have their own script to search monster-ui and extract used fields. This is not complete but it's a base.

icehess: @claytonfinney and @christopherni can explain better.

PR #4822, opened 2018-05-07
KAZOO-5362: Exclude assigned devices from ringing when a user is in a queue

harenson:
Required for https://github.com/2600hz/kazoo-qubicle/pull/40 to work


core/kazoo_endpoint/src/kz_endpoint.erl
```diff
@@ -1131,6 +1131,7 @@ create_sip_endpoint(Endpoint, Properties, #clid{}=Clid, Call) ->
                       ,{<<"Failover">>, maybe_build_failover(Endpoint, Clid, Call)}
                       ,{<<"Metaflows">>, kz_json:get_json_value(<<"metaflows">>, Endpoint)}
                       ,{<<"Endpoint-Actions">>, endpoint_actions(Endpoint, Call)}
+                      ,{<<"Exclude-From-Queues">>, kz_json:get_value(<<"exclude_from_queues">>, Endpoint)}
```
k-anderson: This should not be a new root level property but should be in the existing object:

"qubicle": {
    "enabled": true,
   "presence_id": "7900@test.com",
    "exclude": true
}

k-anderson:
Also, please make sure the schemas are updated :)

harenson:
I am missing to do some changes to this PR, so for the moment this PR is not ready to be merged.

mark2600:
I agree with making it a generic flag

PR #4851, opened 2018-05-16
KAZOO-5898: fixed email being sent for recvfax when user has it disabled

swysor:


applications/fax/src/fax_request.erl
```diff
@@ -565,6 +565,8 @@ notify_failure(JObj, 'undefined', State) ->
     notify_failure(JObj, <<"unknown error">>, State);
 notify_failure(JObj, NonBinary, State) when not is_binary(NonBinary) ->
     notify_failure(JObj, kz_term:to_binary(NonBinary), State);
+notify_failure(_JObj, _Reason, #state{fax_notify='undefined'}) ->
+    ok;
```
jamesaimonetti: tickie the atoms

jamesaimonetti: Maybe consider logging (in debug) the lack of a notify list. Something like `lager:debug("not sending failure notification - no emails to use");`. Adjust for below too (lager returns `'ok'`)

swysor: oh did i forget to tickie? lol

swysor: I decided to put the log line in the place where the decision to not send is actually made so the log is closer to the logic.

lazedo:
@swysor 4.2 please

PR #4853, opened 2018-05-16
HELP-37054: always publish usurp; conditionally publish route_win

jamesaimonetti:
When using conference dial and DIDs, the loopback will cause a
dialplan fetch request to come back around into Kazoo. With multiple
ecallmgrs connected, it is possible that the control process will be
on a separate VM from the conference dial process.

The conference dial code must start a control process on the
initiating node to supply the conference participant process with an
AMQP control queue. Since there's no route_win to publish as well (no
controller queue), this commit conditionally publishes the route_win
if the controller queue is defined but opts to always publish the
usurp (which should teardown the control process started by the
loopback).

log the ecallmgr node

use kz_api

try to set fetch-id on the originate


jamesaimonetti:
See 4.2 in #4852

PR #4854, opened 2018-05-16
 HELP-36774: for faxbox send receipt, only try to decode tiff attachments, otherwise send as is

swysor:


lazedo:
@swysor we need to send the rendered fax to the requestor. a change in `fax_worker` is needed so it will save the result of conversion to tiff (the one that is sent to freeswitch), and the fax_helper will need to get this attachment instead of the original.
this way, if we start adding other processes to the pipe like `cover pages` , `templates`, ...
we will send the actual sent fax to the requestor.

PR #4856, opened 2018-05-17
[master] crossbar: handle file content multipart headers

icehess:
4.2: https://github.com/2600hz/kazoo/pull/4857


icehess:
@efine `cowboy_req:headers(Req1)` is returning the request headers not the current processing part.

`extract_multipart/3` is looping over each part, extracting current part header, calls `get_req_data/5` to extract part's content until all parts are processed (the first `extract_multipart/3` func clause).

PR #4864, opened 2018-05-24
[4.2] build default using system_config_document_schema

icehess:


icehess:
This change suppose to fix the issue that `kapps_config_doc:build_default/1` is being called by `kapps_config:get_category` is not generating correct system config document (it doesn't store default object in `default` key).

In 4.2 this doesn't applied since it's not getting/merging default object from schema.

PR https://github.com/2600hz/kazoo/pull/4865 for master is fixing that issue. Closing this one.

PR #4866, opened 2018-05-24
HELP-37054-4.2: address ownership of conference dial channels

jamesaimonetti:
When there are multiple ecallmgrs connected to a FreeSWITCH server, it
is possible that the conference dial occurs on ecallmgr1 and the fetch
request for the loopback-b channel comes in on ecallmgr2. This results
in the legs associated with the loopback (a and b) and the final
b-leg, are all considered "local" to ecallmgr2.

However, the conference dial needs to start a call control process on
ecallmgr1 to get a control queue for the add_participant API (used to
start a conf_participant process to manage the now-dialed
participant).

This means both ecallmgr1 and ecallmgr2 consider the b-leg to be
"local". Metaflows checks the local ecallmgr's channel cache to
determine if it should repond to channel API actions - since both
ecallmgrs consider the channel local, both metaflow listeners publish
the payloads for konami to execute.

In this particular case, two CHANNEL_PIVOTs are picked up by the
cf_exe process and cause the conference action to be skipped (since
both "set_variables" actions call cf_exe:continue/1), the callflow is
exhausted, and cf_exe tears down the channel.

### DIFF
--- a/applications/ecallmgr/src/ecallmgr_call_control.erl
+++ b/applications/ecallmgr/src/ecallmgr_call_control.erl
@@ -54,6 +54,7 @@
 -export([other_legs/1
         ,update_node/2
         ,control_procs/1
+        ,publish_usurp/3
         ]).
 -export([fs_nodeup/2]).
 -export([fs_nodedown/2]).
@@ -79,17 +80,17 @@
 -record(state, {node :: atom()
                ,call_id :: kz_term:ne_binary()
                ,command_q = queue:new() :: queue:queue()
-               ,current_app :: kz_term:api_binary()
+               ,current_app :: kz_term:api_ne_binary()
                ,current_cmd :: kz_term:api_object()
                ,start_time = os:timestamp() :: kz_time:now()
                ,is_call_up = 'true' :: boolean()
                ,is_node_up = 'true' :: boolean()
                ,keep_alive_ref :: kz_term:api_reference()
                ,other_legs = [] :: kz_term:ne_binaries()
-               ,last_removed_leg :: kz_term:api_binary()
+               ,last_removed_leg :: kz_term:api_ne_binary()
                ,sanity_check_tref :: kz_term:api_reference()
-               ,msg_id :: kz_term:api_binary()
-               ,fetch_id :: kz_term:api_binary()
+               ,msg_id :: kz_term:api_ne_binary()
+               ,fetch_id :: kz_term:api_ne_binary()
                ,controller_q :: kz_term:api_ne_binary()
                ,control_q :: kz_term:api_ne_binary()
                ,initial_ccvs :: kz_json:object()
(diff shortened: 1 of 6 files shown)


applications/ecallmgr/src/ecallmgr_fs_channel.erl
```diff
@@ -766,37 +770,37 @@ props_to_update(Props) ->
...
+      ,{#channel.owner_id, props:get_value(<<"Owner-ID">>, CCVs)}
+      ,{#channel.precedence, kz_term:to_integer(props:get_value(<<"Precedence">>, CCVs, 5))}
+      ,{#channel.presence_id, props:get_value(<<"Channel-Presence-ID">>, CCVs, props:get_value(<<"variable_presence_id">>, Props))}
+      ,{#channel.profile, props:get_value(<<"variable_sofia_profile_name">>, Props)}
+      ,{#channel.realm, props:get_value(<<"Realm">>, CCVs, get_realm(Props))}
+      ,{#channel.reseller_billing, props:get_value(<<"Reseller-Billing">>, CCVs)}
+      ,{#channel.reseller_id, props:get_value(<<"Reseller-ID">>, CCVs)}
+      ,{#channel.resource_id, props:get_value(<<"Resource-ID">>, CCVs)}
+      ,{#channel.to_tag, props:get_value(<<"variable_sip_to_tag">>, Props)}
+      ,{#channel.username, props:get_value(<<"Username">>, CCVs, get_username(Props))}
```
lazedo: @jamesaimonetti i understand the `sort` but for me when reading code, it makes more sense to group by `affinity`., `from_tag` close to `to_tag`. but i'm ok with it.

applications/ecallmgr/src/ecallmgr_call_control.erl
```diff
@@ -360,9 +361,9 @@ handle_call_events(JObj, FetchId) ->
     kz_util:put_callid(kz_json:get_value(<<"Call-ID">>, JObj)),
     case kz_api:event_name(JObj) of
         <<"usurp_control">> ->
-            case kz_json:get_ne_binary_value(<<"Fetch-ID">>, JObj) =:= FetchId of
-                'false' -> gen_listener:cast(self(), {'usurp_control', JObj});
-                'true' -> 'ok'
+            case kz_json:get_ne_binary_value(<<"Fetch-ID">>, JObj) =/= FetchId of
```
lazedo: @jamesaimonetti i don't understand this change

applications/ecallmgr/src/ecallmgr_fs_channel.erl
jamesaimonetti: Yeah, i find myself hating that more and more :) I was trying to see if handling_locally was included in the update proplist, and I was just like, man, sort this list! Lol.

applications/ecallmgr/src/ecallmgr_call_control.erl
jamesaimonetti: I kept reading it as "if JObj's FetchId matches our FetchId then usurp control". It made me go down a wrong path trying to usurp the loopback's call control process and I was really mad. I found this to be easier to parse into English in my head :)

PR #4874, opened 2018-05-29
HELP-37225-4.2: don't consider [] as empty

jamesaimonetti:
Especially since tts/play commands consider [] different from
'undefined' for terminators. It is getting hard to know which API
commands might have a callflow payload (or a terminators attribute in
general); easier to permit a key with [] when it make no difference
than stripping keys with [] when there is a difference.

### DIFF
--- a/core/kazoo_amqp/src/api/kapi_callflow.erl
+++ b/core/kazoo_amqp/src/api/kapi_callflow.erl
@@ -64,5 +64,10 @@ declare_exchanges() ->
 %%------------------------------------------------------------------------------
 -spec publish_resume(kz_term:api_terms()) -> 'ok'.
 publish_resume(JObj) ->
-    {'ok', Payload} = kz_api:prepare_api_payload(JObj, ?RESUME_VALUES, fun resume/1),
+    {'ok', Payload} = kz_api:prepare_api_payload(JObj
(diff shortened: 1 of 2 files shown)


core/kazoo_amqp/src/api/kapi_callflow.erl
```diff
@@ -64,5 +64,10 @@ declare_exchanges() ->
 %%------------------------------------------------------------------------------
 -spec publish_resume(kz_term:api_terms()) -> 'ok'.
 publish_resume(JObj) ->
-    {'ok', Payload} = kz_api:prepare_api_payload(JObj, ?RESUME_VALUES, fun resume/1),
+    {'ok', Payload} = kz_api:prepare_api_payload(JObj
+                                                ,?RESUME_VALUES
+                                                ,[{'formatter', fun resume/1}
+                                                 ,{'remove_recursive', 'false'}
```
lazedo: @jamesaimonetti this went into master but, isn't the below change to `should_strip_from_payload` make this change unnecessary ?

jamesaimonetti: Yes and no. We don't really need to recurse into the Flow object so this saves us time, especially if its a big callflow doc.

PR #4875, opened 2018-05-29
HELP-37225: don't consider [] as empty

jamesaimonetti:
Especially since tts/play commands consider [] different from
'undefined' for terminators. It is getting hard to know which API
commands might have a callflow payload (or a terminators attribute in
general); easier to permit a key with [] when it make no difference
than stripping keys with [] when there is a difference.


jamesaimonetti:
See 4.2 in #4874

core/kazoo_amqp/src/api/kz_api.erl
```diff
@@ -285,7 +285,6 @@ do_empty_value_removal([{K,V}=KV|T], Recursive, Acc) ->
 
 -spec is_empty(any()) -> boolean().
 is_empty('undefined') -> 'true';
-is_empty([]) -> 'true';
```
k-anderson: @jamesaimonetti this could be significant, I would like some more time to look into the ramifications (or further discussion during tomorrows standup).

PR #4883, opened 2018-05-31
[4.2] filter same name test modules

icehess:
Filter test modules which has the same name as their normal module (e.g. `kz_date` and `kz_date_tests`) to avoid running tests in `kz_date_tests` twice.

When running EUnit to test the module `kz_date`, it will also look for the module `kz_date_tests` and run those tests as well.


jamesaimonetti:
I don't think we need a 4.2 for this - its not a bug, per se.

swysor:
It does shave like 6 minutes off of travisCI @jamesaimonetti 💃

PR #4888, opened 2018-06-01
HELP-37171-4.2: check if media is convertible to prompt://

jamesaimonetti:
When overriding prompts with other forms of media, like tone_stream or
HTTP URLs, normalization still occurs as it would for prompts like
"/{ACCOUNT_ID}/{PROMPT}" or "/system_media/{PROMPT}" or
"{PROMPT}". This results in invalid "prompt://" URLs and FreeSWITCH
being unable to play the appropriate media back to the call leg.

Introduce a function that looks for media that isn't
prompt-normalizable and returns a boolean indicating such.

### DIFF
--- a/core/kazoo_media/Makefile
+++ b/core/kazoo_media/Makefile
@@ -4,3 +4,8 @@ PROJECT = kazoo_media
 all: compile
 
 include $(ROOT)/make/kz.mk
+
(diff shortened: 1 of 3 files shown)


jamesaimonetti:
See master at #4889

PR #4891, opened 2018-06-01
KAZOO-5907: check for CIDR or IP, or resolve hostnames

jamesaimonetti:
Add CIDR checks when resolving an address to an IP/CIDR.

Also refactors ecallmgr_maintenance ACL functions to remove
duplication and resolve the supplied address in one place.

Also provides some useful feedback is resolution fails.


jamesaimonetti:
See 4.2 in #4890

PR #2693, opened 2016-10-12
Added support to callflow modules and pivot to utilize custom KVs

danielfinke:
Long lost follow up to both #1328 and #1255.

Addressed concerns about cross-app dependencies by moving custom-KV-related code into kapps_call.
Moved to a generic variable replacer in cf_exe.
Schema doc is empty template since variables (the properties of module's data) are freeform.

PR #4892, opened 2018-06-04
Voxter master pr custom kvs

jamesaimonetti:
Closes #2693 but still wanting @danielfinke to weigh in on changes


danielfinke:
@jamesaimonetti hey! I am wondering if we should open a discussion if this feature can be implemented atop that recent addition of Custom-Application-Vars (CAVs). I haven't looked much into them, but does it appear they are used for a similar purpose?

jamesaimonetti:
@danielfinke indeed! CAVs are included on all subsequent event payloads of the call; however I'm not sure they're included on Pivot requests. Would including them be sufficient for the use case this PR was intended to address?

jamesaimonetti:
Although reading the "variables" description more, it looks like this PR wanted to potentially replace values in the `Data` object with values from the custom vars? CAVs aren't used in callflow modules (apart from setting them).

danielfinke:
@jamesaimonetti part of the use case would be including them in Pivot requests. The other part as you mentioned is to overwrite values in the callflow data payload. I could see using CAVs to still do this - it is the responsibility of the flow to define the variables/keys that are pulled from the Custom-KVs/CAVs to replace the data object's values. Since this is new code, there shouldn't be any backwards-compatibility to worry about; specifically, nobody (other than my team) should already have these variables specified and encounter unexpected behaviour from looking at CAVs

jamesaimonetti:
@danielfinke hmm, i'm still not sure I understand why CAVs (or Custom KVs) should affect an existing callflow action's `data` object. Can you give me an idea of how you are using it in practice?

CAVs are intended to hold state info for folks using pivot/channels API and to not clutter CCVs. Wouldn't your Pivot response include the required changes to the callflow?

danielfinke:
@jamesaimonetti so there's 2 things I'm trying to achieve:

1. state info persisting on a call, shared with pivot requests (so CAVs do this, which is great)
2. allow mutation of a callflow based on variables assigned to a call:
for example we have callflow nodes that look like:
```
"data": {
    "id": "1154736238d5b4d7fde12bb159396cd7"
},
"module": "callflow",
"variables": {
    "id": "CALLFLOW_OVERRIDE"
},
"children": {}
```
When this node is executed, if the CALLFLOW_OVERRIDE Custom KV is set, the value of the Custom KV _replaces_ the id in the data payload. If it is not set, the value in the data payload is retained and used.
What we achieve here is having UI-built callflows where people can define (or not) variables in multiple callflows, have the last module be a "Callflow" callflow module, and redirect the caller to a shared callflow that examines the variables.
An example we have is multiple callflow entry points where the distinction is solely the level of priority the customer should receive - callflow 1 sets the 'Priority Queue' variable to 'x', callflow 2 sets it to 'y', and callflow 3 sets it to 'z' (x, y, z are queue IDs). All 3 callflows then redirect to a common callflow of significant complexity. Instead of having to reproduce this callflow 3 times (once for each entry point) we can just dynamically read the queue ID from the 'Priority Queue' variable at the point that it is needed. Drastically reduces the repetition in making changes to shared callflows at the expense of a little less clarity - we reserve this feature for only certain customers.

jamesaimonetti:
OK, then I would say two PRs - one to expose CAVs on Pivot requests and this one, with any adjustments to make it work for your usecases

jamesaimonetti:
Since there's no movement, and the code no longer will merge cleanly, I'll close this. Please reopen when updates are ready.

PR #4897, opened 2018-06-06
KAZOO-5912: Fixed leg creation when undefined privacy prefs

sergey-safarov:
For more details please looks [KAZOO-5912](https://2600hz.atlassian.net/browse/KAZOO-5912)

### DIFF
--- a/applications/callflow/src/module/cf_resources.erl
+++ b/applications/callflow/src/module/cf_resources.erl
@@ -380,7 +380,7 @@ use_endpoint_prefs(Call) ->
     %% only overwrite the ccvs if privacy has not been set by cf_privacy
     %% or if the call has been configured to overwrite cf_privacy settings
     not kz_privacy:has_flags(kapps_call:custom_channel_vars(Call))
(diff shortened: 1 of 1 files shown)


applications/callflow/src/module/cf_resources.erl
```diff
@@ -372,7 +372,7 @@ handle_channel_destroy(_, JObj) ->
 get_privacy_prefs(Call) ->
     case use_endpoint_prefs(Call) of
         'true' -> check_inception(Call);
-        'false'  -> []
+        _  -> []
```
jamesaimonetti: I'd rather change `update_endpoint_prefs` and do `kapps_call:kvs_fetch(<<"use_endpoint_privacy">>, Call, 'false').` instead, and keep this case clause using `false`.

jamesaimonetti: Err, `(<<"use_endpoint_privacy">>, 'false', Call)`.

sergey-safarov:
Hello @jamesaimonetti 
PR is updated.

jamesaimonetti:
And merged! Thanks :)

PR #4899, opened 2018-06-06
[4.2] Help-36418: Address issues with disabling serialization by outbound number in faxing

swysor:


make/kz.mk
```diff
@@ -79,7 +79,7 @@ ebin/$(PROJECT).app:
 	@mkdir -p ebin/
 	ERL_LIBS=$(ELIBS) erlc -v $(ERLC_OPTS) $(PA) -o ebin/ $(SOURCES)
 	@sed "s/{modules,\s*\[\]}/{modules, \[$(MODULES)\]}/" src/$(PROJECT).app.src \
-	| sed -e "s/{vsn,\([^}]*\)}/\{vsn,\"$(KZ_VERSION)\"}/g" > $@
+	| sed -e "s|{vsn,\([^}]*\)}|\{vsn,\"$(KZ_VERSION)\"}|g" > $@
```
lazedo: are we sure about this ?
` /home/ubuntu/kazoo/_rel/kazoo/releases/4.2-HELP-36418.1/sys.config`
should be `4.2.next`

lazedo: i think the problem can be with those branches starting with 4.2 and the fact that you renamed yours instead of creating a new one and cherry pick the commits in it.

PR #4903, opened 2018-06-07
PISTON-575: ignore non-existent dbs in unassign_from_prev

danielfinke:
When unassigning a number from prev_assigned_to, not_found errors should be ignored when the prev_assigned_to refers to an account that was deleted

### DIFF
--- a/core/kazoo_number_manager/src/knm_phone_number.erl
+++ b/core/kazoo_number_manager/src/knm_phone_number.erl
@@ -332,11 +332,22 @@ split_by_prevassignedto(PNs) ->
                           true ->
                               lager:debug("prev_assigned_to is empty for ~s, ignoring", [number(PN)]),
                               undefined;
-                          false -> kz_util:format_account_db(PrevAssignedTo)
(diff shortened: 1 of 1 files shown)


jamesaimonetti:
Would it be worth changing `split_by_prevassignedto/1` to check `kz_datamgr:db_exists/1` as well in the [case statement](https://github.com/2600hz/kazoo/blob/master/core/kazoo_number_manager/src/knm_phone_number.erl#L325)?

danielfinke:
@jamesaimonetti are you thinking to do that as an alternative or an addition

jamesaimonetti:
I was originally thinking instead of, but it certainly doesn't hurt to have both checks. I wasn't sure if the `{error, not_found}` tuple could come from the doc not being found; as far as I recall we didn't have a way to distinguish between "db not found" and "doc not found" but since it uses `kz_datamgr:del_docs/2`, the `not_found` should be for the DB.

So let's do both. :)

jamesaimonetti:
Thanks!

PR #4910, opened 2018-06-09
[master] HELP-37319: handle huge load view better

icehess:
Ledgers summary was not honoring page size, and timestamp correctly, so it was loading all ledgers in a MODB database causing slowness and pressure on CouchDb and kazoo.

This PR changes ledgers API (`/ledger`) to use use crossbar_view for handling multi database load and time range query (default to 50). Currently I'm not sure how set the page size and next start key in response. This is the summary API, meaning it is returning all categories sums, so if you have 2 service category it returns two object. Since behind scene it limits the db result to 50, and we returning two object there is ambiguity between request page size (default to 50) and response  page size (in this example 2).

Also fixing KAZOO-5191 . There are times that view result is big, and couchdb dropping the connection during parsing buffer in couchbeam, and couchbeam should return proper error. [this fix depends on https://github.com/2600hz/couchbeam/commit/3fcd4679da7d6301304ad715d76e15bf2ae74c84]

4.2: https://github.com/2600hz/kazoo/pull/4911

### How to test

From UI:

Select and account which has a huge amount of the call (ledger: per-minute for example):

1. From UI go to account settings and Usage Charges
2. Use date range and next page, previous page crawl all ledgers.
3. It should returning without any error
4. It should returning all matched result
5. It should be relatively faster than before
6. It should *not* be as slow as and as time consuming as before.

For testing with API:

1, Make request to `/v2/account/{account_id}/ldegers` on an account which has more than 50 ledgers for this month, the API should only returns sum for the 50 ledgers (starting from latest one to older one), check with UI, (manually sum 50 ledgers).
2. make request to same URL and setting `?page_size=2`, it should return sum of the last 2 ledgers.
3. make request to same URL add `?paginate=false`, it should dump the whole ledgers for 1 month, causing slowness and takes a lot of time and probably results in and error or timeout.

### DIFF
--- a/applications/crossbar/priv/api/swagger.json
+++ b/applications/crossbar/priv/api/swagger.json
@@ -29463,6 +29463,67 @@
             },
             "type": "object"
         },
+        "system_config.kazoo_convert": {
+            "description": "Schema for kazoo_convert system_config",
+            "properties": {
+                "convert_command_timeout": {
+                    "default": 120000,
+                    "description": "kazoo_convert convert_command_timeout",
+                    "type": "integer"
+                },
+                "convert_image_command": {
+                    "default": "convert $FROM -resample 204x98 -units PixelsPerInch -compress group4 -size 1728x1078 $TO",
+                    "description": "kazoo_convert convert_image_command",
+                    "type": "string"
+                },
+                "convert_openoffice_command": {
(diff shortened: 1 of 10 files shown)


jamesaimonetti:
rebased, letting it run to see if all is well, then merging

PR #4916, opened 2018-06-13
[WIP] PISTON-549: cb_collections

danielfinke:
Access multiple entities within a single GET request. Later adding PUT/POST/PATCH/DELETE

Sample request URL for GET: `{host}:8000/v1/accounts/{account_id}/collections?path=%2Fv1%2Faccounts%2F%7Bid%7D&arguments=%7B%22id%22%3A%5B%22id1%22%2C%22id2%22%5D%7D`
For GET (future: DELETE) requests, the params for the collection operation are equivalent to the contents of the data object of the (future) body of PUT/POST/PATCH requests, except URL-encoded.

Payload:
```
{
    "data": {
        "path": "/v1/accounts/{id}",
        "arguments": {
            "id":["id1","id2"]
        }
    }
}
```
Properties of "arguments" replace placeholders (e.g. {id}) from the path. They will also replace placeholders in the PUT/POST/PATCH body in the future.

### DIFF
--- a/applications/crossbar/src/modules/cb_collections.erl
+++ b/applications/crossbar/src/modules/cb_collections.erl
@@ -1,0 +1,305 @@
+%%%-----------------------------------------------------------------------------
+%%% @copyright (C) 2018, Voxter Communications Inc
+%%% @doc Multiple operations module
+%%% Handle requests to perform crossbar operations on multiple IDs at once or on
+%%% the results of a view
+%%%
+%%% @author Daniel Finke
+%%% @end
+%%%-----------------------------------------------------------------------------
+-module(cb_collections).
+
+-export([init/0
+        ,allowed_methods/0
+        ,resource_exists/0
+        ,validate/1
+        % ,post/1
+        ]).
+
+-include("crossbar.hrl").
+
+-spec init() -> 'ok'.
+init() ->
+    Bindings = [{<<"*.allowed_methods.collections">>, 'allowed_methods'}
+               ,{<<"*.resource_exists.collections">>, 'resource_exists'}
+               ,{<<"*.validate.collections">>, 'validate'}
+               % ,{<<"*.execute.put.accounts">>, 'put'}
+               % ,{<<"*.execute.post.accounts">>, 'post'}
+               % ,{<<"*.execute.patch.accounts">>, 'patch'}
+               % ,{<<"*.execute.delete.accounts">>, 'delete'}
+               ],
+    cb_modules_util:bind(?MODULE, Bindings).
+
+%%------------------------------------------------------------------------------
+%% @doc This function determines the verbs that are appropriate for the
+%% given Nouns.
+%%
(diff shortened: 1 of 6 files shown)


jamesaimonetti:
Does this work properly with `/v2`?

jamesaimonetti:
Also, it appears you are not choosing an endpoint to do the GET from, just supplying an arbitrary list of IDs. Different endpoint modules have potentially different access controls to documents; how do you propose to make sure those controls are respected? Also of interest is whether token_restrictions can be bypassed this way too, since a given auth token could not be allowed access to `/callflows` for instance.

applications/crossbar/src/modules/cb_collections.erl
```diff
@@ -0,0 +1,305 @@
...
+    %% TODO validate using JSONschema
+    validate_request_path(Context, cb_context:req_value(Context, <<"path">>)).
+
+%%------------------------------------------------------------------------------
+%% @doc
+%% @end
+%%------------------------------------------------------------------------------
+-spec validate_request_path(cb_context:context(), kz_json:api_json_term()) ->
+                                   cb_context:context() | {'error', cb_context:context()}.
+validate_request_path(Context, <<_,_/binary>> = Path) ->
```
jamesaimonetti: There is a ?NE_BINARY macro you could use here

danielfinke:
@jamesaimonetti not ready for /v2/ yet. Will be addressed before merge, but I just haven't got there yet haha
so the chosen endpoint is specified by the path argument:
```
"path": "/v1/accounts/{id}"
```
which has the path placeholder id replaced with the list. It could honestly be anything, i.e.:
```
"path": "/{version}/accounts/someid/devices/{device_id}",
"data": {
    "version": ["v1","v2"],
    "device_id": ["abc","def"]
}
```
That's the plan, anyway. Right now it won't produce only combinations where all placeholders are replaced, it'll return all combinations. But just trying to show the gist of what I'm doing here

I'll take a look into the token_restrictions stuff, but I believe since I am calling the api_flow methods again for each and every path that is produced in cb_collections, some of them should fail if access is denied - which results in a mini-error-payload object for that key in the data object.

jamesaimonetti:
OK, cool. That makes more sense seeing the example. Thanks

danielfinke:
@jamesaimonetti did you have any additional thoughts about the approach I am taking? Or should I keep on going with this

jamesaimonetti:
How does this relate to #3991 ? Seems similar in nature, as far as fetching a subset of documents vs the whole collection.

danielfinke:
@jamesaimonetti there's some similarity at this point, but I'm going for a much more open scope. Specifically once I get into adding POST/PUT there will be a lot of flexibility here

lazedo:
@danielfinke there hasn't been much progress and there are conflicts. please create a new pr if/when you wish to proceed

PR #4917, opened 2018-06-13
[4.2] HELP-37496: don't let dynamic cid down if capture group is empty

icehess:


applications/callflow/src/module/cf_dynamic_cid.erl
```diff
@@ -210,10 +192,17 @@ maybe_strip_features_code(Call, Number) ->
 
 %%------------------------------------------------------------------------------
 %% @doc Lookup callflow and continue with the call if we have a destination number
+%%
 %% @end
 %%------------------------------------------------------------------------------
--spec maybe_route_to_callflow(kz_json:object(), kapps_call:call(), kz_term:ne_binary()) -> 'ok'.
-maybe_route_to_callflow(Data, Call, Number) ->
+-spec maybe_route_to_callflow(kz_json:object(), kapps_call:call(), kz_term:ne_binary(), kz_term:ne_binary()) -> 'ok'.
```
jamesaimonetti: Looks like the third term should be `kz_term:api_ne_binary()`

PR #4919, opened 2018-06-13
HELP-37395-4two: correctly check for hotdesking when doing a 'user' s…

mark2600:
…cope lookup in cf_branch_variable


applications/crossbar/priv/couchdb/schemas/devices.json
```diff
@@ -21,6 +21,20 @@
                     "description": "Enable the call-forwarding parameters if the device is offline",
                     "type": "boolean"
                 },
+                "hotdesk": {
```
jamesaimonetti: I don't think this schema should be under "call_forward" but on the root of the doc, no?

PR #4920, opened 2018-06-13
HELP-37395-master: correctly check for hotdesking when doing a 'user'…

mark2600:
… scope lookup in cf_branch_variable

### DIFF
--- a/core/kazoo_documents/src/kzd_devices.erl
+++ b/core/kazoo_documents/src/kzd_devices.erl
@@ -28,6 +28,7 @@
 -export([enabled/1, enabled/2, set_enabled/2]).
 -export([exclude_from_queues/1, exclude_from_queues/2, set_exclude_from_queues/2]).
 -export([formatters/1, formatters/2, set_formatters/2]).
+-export([hotdesk/1, hotdesk/2, set_hotdesk/2]).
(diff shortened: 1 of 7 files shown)


core/kazoo_documents/src/kzd_devices.erl
```diff
@@ -487,6 +488,23 @@ owner_id(Doc, Default) ->
 set_owner_id(Doc, OwnerId) ->
     kz_json:set_value([<<"owner_id">>], OwnerId, Doc).
 
+-spec is_hotdesked(doc()) -> boolean().
+is_hotdesked(Doc) ->
+    not kz_json:is_empty(kz_json:get_value([<<"hotdesk">>, <<"users">>], Doc, kz_json:new())).
```
jamesaimonetti: We should probably add a `hotdesk` object to the devices schema, no? Something like:

```json
"hotdesk":{
    "type":"object",
    "description":"The hotdesk status of this device",
    "users":{
        "type":"object",
        "description":"The user(s) currently hotdesked into the device",
        "patternProperties":{
            "^[a-zA-Z0-9]{32}$":{
                "type":"object",
                "description":"user-specific hotdesk settings"
            }
        }
    }
}
```

PR #4923, opened 2018-06-15
Fix teletype deregister test

danielfinke:
kz_template:render will render the quotes in the placeholder replacement as `&quot;`, not `\"`


icehess:
You're right, it should replace double quotation with `&quot;`.
But if you see the raw log from Travis CI the teletype render test is failing (there is a problem with the eunit_run escript, it is not exiting with correct exit status so `make eunit` is not failing. This is fixed in another PR).

I don't know why it is not rendered correctly in Travis CI yet. It needs more testing.

danielfinke:
@icehess that's curious. It is failing at the same spot that I was having eunit fail for me prior to my change. That was why I made the change

jamesaimonetti:
@danielfinke is this still failing for you? Could you test again with latest master?

icehess:
@danielfinke @jamesaimonetti This is already fixed in https://github.com/2600hz/kazoo/pull/5141 by putting User agent inside an inline code HTML tag, to force ErlyDTL to not escape the double quote.

danielfinke:
@icehess @jamesaimonetti yeah it looks like it's covered in #5141 so I'm going to close this

PR #4936, opened 2018-06-21
KAZOO-5913: Make voicemail setup wizard and voicemail change PIN prompt to only try 3 times

harenson:


applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -1408,26 +1414,30 @@ change_pin(#mailbox{mailbox_id=Id
...
-            invalid_pin(Box, Call)
+            invalid_pin(Box, Call, Loop)
     end.
 
--spec invalid_pin(mailbox(), kapps_call:call()) ->
+-spec invalid_pin(mailbox(), kapps_call:call(), non_neg_integer()) ->
                          mailbox() |
                          {'error', any()}.
-invalid_pin(Box, Call) ->
+invalid_pin(_Box, Call, Loop) when Loop > 2 -> %% third failure, terminate
```
lazedo: create a `macro` at top `MAX_CHANGE_PIN_TRIES` (or a better name) and use it here instead of  `2`

icehess: or even better make it system configurable and add it to `mailbox` record.

applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -1408,26 +1414,30 @@ change_pin(#mailbox{mailbox_id=Id
...
+    lager:debug("channel hungup after several empty or invalid pins"),
+    _ = kapps_call_command:b_prompt(<<"vm-goodbye">>, Call),
+    {'error', 'channel_hungup'};
```
icehess: Isn't hanging up here too harsh? You can just return `Box` here and it goes back to config menu.
Also instead of playing `vm-goodbye`, play `vm-pin_invalid`.

icehess: @lazedo Config menu has it's own loop and will stop.
I've notice config menu is going to crash if it passes 3 loop. Shouldn't it just return to main menu(and again menu has its own loop)?

lazedo: @icehess whatif its not from config menu ?

icehess: @lazedo invalid_pin is only called by change_pin and config_menu

icehess: @harenson So you can play `vm-pin_invalid` here and return `{'error', 'max_retry'}`.
And in `setup_mailbox/2` check `change_pin/2` result if it is an error play `vm-goodbye` and return `{'error', 'channel_hungup'}`.

@lazedo What do you think?

applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -1408,26 +1422,29 @@ change_pin(#mailbox{mailbox_id=Id
...
     end.
 
--spec invalid_pin(mailbox(), kapps_call:call()) ->
+-spec invalid_pin(mailbox(), kapps_call:call(), non_neg_integer()) ->
                          mailbox() |
                          {'error', any()}.
-invalid_pin(Box, Call) ->
+invalid_pin(_Box, _Call, Loop) when Loop >= ?MAX_INVALID_PIN_LOOPS ->
+    lager:debug("Several empty or invalid pins"),
+    {'error', 'max_retry'};
```
icehess: play
```
_ = kapps_call_command:b_prompt(<<"vm-pin_invalid">>, Call),
{'error', 'max_retry'};
```

PR #4937, opened 2018-06-22
KAZOO-5870: fixed fax attachments from inbound fax getting saved over…

swysor:
… and over and over


swysor:
Hold on until I test this :D

swysor:
Tested it, should be good to go.

PR #4942, opened 2018-06-22
HELP-37633: Add faxbox.smtp_permission_list validation

harenson:


core/kazoo_schemas/src/kz_json_schema_extensions.erl
```diff
@@ -98,6 +98,15 @@ extra_validation(<<"storage.attachment.onedrive.oauth_doc_id">>, Value, State) -
     validate_attachment_oauth_doc_id(Value, State);
 extra_validation(<<"storage.attachment.dropbox.oauth_doc_id">>, Value, State) ->
     validate_attachment_oauth_doc_id(Value, State);
+extra_validation(<<"faxbox">>, Value, State) ->
```
lazedo: @harenson you're doing this for every element in faxbox

core/kazoo_schemas/src/kz_json_schema_extensions.erl
```diff
@@ -98,6 +98,24 @@ extra_validation(<<"storage.attachment.onedrive.oauth_doc_id">>, Value, State) -
     validate_attachment_oauth_doc_id(Value, State);
 extra_validation(<<"storage.attachment.dropbox.oauth_doc_id">>, Value, State) ->
     validate_attachment_oauth_doc_id(Value, State);
+extra_validation(<<"faxbox.smtp_permission_list">>, Values, State) ->
+    F = fun(Val, Acc) ->
```
jamesaimonetti: 1. When the function to use doesn't need access to variables outside its scope (you aren't using Values or State), create a named function and use `fun named_fun/2` as an arg to `lists:foldl/3`.

2. Bare `lists:foldl` calls like below can be hard to figure out what they're doing since you need to inspect the function being used in the fold. Better to call a well-named function that the reader can understand the purpose of the fold.

So I would write this as:

```erlang
extra_validation(<<"faxbox.smtp_permission_list">>, Values, State) ->
    validate_smtp_permission_list(Values, State).

validate_smtp_permission_list(Regexps, State) ->
    case find_invalid_regexps(Regexps) of
        [] -> State;
        FailingRegexps -> ...
    end.
```
Now we see that we want to filter the list of regexps to find the invalid ones so we could write that as

```erlang
find_invalid_regexps(Regexps) ->
    lists:filter(fun is_not_valid_regexp/1, Regexps).

is_not_valid_regex(Regexp) ->
    {'error', _} =:= re:compile(Regexp).
```

So I think we have better variable naming that just `Values` and hopefully we capture what is happening a little more clearly using small functions. Plus now you can write tests for `find_invalid_regexps`.

Let me know if this makes sense or if you have suggestions about how to further improve it.

PR #4946, opened 2018-06-24
[master] ensure cover dir is created + halt eunit_run on error

icehess:
4.2: https://github.com/2600hz/kazoo/pull/4947


icehess:
@lazedo please merge this.

PR #4949, opened 2018-06-25
[master] log error when updating views, add some edoc for system startup

icehess:


applications/crossbar/src/crossbar_maintenance.erl
```diff
@@ -1147,29 +1147,76 @@ set_app_screenshots(AppId, PathToScreenshotsFolder) ->
...
+%% @doc Updates system schemas using files in Crossbar `priv' folder during
+%% start up.
+%%
+%% This is called by {@link db_init/0} during Crossbar start up.
+%% @end
+%%------------------------------------------------------------------------------
 -spec update_schemas() -> 'ok'.
 update_schemas() ->
-    kz_datamgr:suppress_change_notice(),
+    kz_datamgr:enable_change_notice(),
```
lazedo: @icehess move ` kz_datamgr:suppress_change_notice` to `db_init_schemas` before calling `update_schemas`. `enable_change_notice` is by default and usually used after operations that fire `changes`. but we may look at the impact of removing it completely, i think `kz_datamgr:revise_docs_* `  does't update the doc if the content doesn't change.

core/kazoo_apps/src/kapps_maintenance.erl
```diff
@@ -53,10 +53,11 @@
 
 -export([flush_getby_cache/0
         ,flush_account_views/0
-        ,get_all_account_views/0
+        ,get_all_account_views/0 %% tries to read from cache first
+        ,fetch_all_account_views/0 %% read views from file system (use only for debugging)
```
jamesaimonetti: What about `read_all_account_views`; I think we use `fetch` when doing a DB/AMQP type request vs `read` which usually implies a filesystem read? Thoughts?

core/kazoo_data/src/kzs_db.erl
```diff
@@ -222,9 +222,17 @@ add_update_remove_views(Server, Db, CurrentViews, NewViews, ShouldRemoveDangling
 add_views(Server, Db, Add, NewViews) ->
     Views = [props:get_value(Id, NewViews) || Id <- Add],
     {'ok', JObjs} = kzs_doc:save_docs(Server, Db, Views, []),
-    [kz_doc:id(JObj)
+    [Id
```
jamesaimonetti: Can this list comprehension be put in a function; looks identical to the LC below this one.

core/kazoo_data/src/kzs_db.erl
```diff
@@ -222,9 +222,17 @@ add_update_remove_views(Server, Db, CurrentViews, NewViews, ShouldRemoveDangling
...
      || JObj <- JObjs,
-        <<"conflict">> =:= kz_json:get_value(<<"error">>, JObj)
+        Id <- [kz_doc:id(JObj)],
+        Error <- [kz_json:get_value(<<"error">>, JObj)],
+        'true' <- [(Error =/= 'undefined'
```
jamesaimonetti: Ugh, this is ugly. It is clever but hard to read, which I worry will make it hard to understand later.

Could you just make this a more explicit filter+log, something like:

```erlang
[Id || JObj <- JObjs, {Id, <<"conflict">>} <- [save_has_error(JObj)] ].

save_has_error(JObj) ->
    Error = kz_json:get_value(<<"error">>, JObj),
    Id = kz_doc:id(Id),
    save_has_error(Id, Error).

save_has_error(Id, <<"conflict">>=Error) -> {Id, Error};
save_has_error(Id, 'undefined'=Error) -> {Id, Error};
save_has_error(Id, Error) ->
    lager:warning("..."),
    {Id, Error}.
```

Something along these lines anyway. Make it more easy to read, less clever.

PR #4952, opened 2018-06-26
[4.1] KAZOO-5870: fixed fax attachments from inbound fax getting saved over and over and over

swysor:
… and over and over

### DIFF
--- a/applications/fax/src/fax_request.erl
+++ b/applications/fax/src/fax_request.erl
@@ -524,7 +524,7 @@ store_attachment(Pid, #state{call=Call
             end
     end.
 
--spec check_fax_attachment(kz_term:ne_binary(), kz_term:ne_binary(), kz_term:ne_binary())->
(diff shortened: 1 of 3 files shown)


applications/fax/src/fax_request.erl
```diff
@@ -506,7 +508,34 @@ store_attachment(Pid, #state{call=Call
...
+                    timer:sleep(?RETRY_SAVE_ATTACHMENT_DELAY),
+                    gen_server:cast(Pid, 'store_attachment');
+                {'error', _R} ->
+                    lager:debug("error '~p' saving fax attachment on fax id ~s",[_R, FaxId]),
+                    timer:sleep(?RETRY_SAVE_ATTACHMENT_DELAY),
+                    gen_server:cast(Pid, 'store_attachment')
+            end
+    end.
+
+-spec check_fax_attachment(kz_term:ne_binary(), kz_term:ne_binary(), kz_term:ne_binary())->
```
swysor: Need to fix that :(

PR #4954, opened 2018-06-29
HELP-37779-4two: fix comparison when requested props is a subset of e…

mark2600:
…xisting props


mark2600:
So the purpose of this is to fix the following comparison case.  In this example a subscription is first made for a qubicle recipient event exchange with a specific recipient id, then a second request for all recipients '*'.

First set of props:
`[{event_only,true},{account_id,<<"TESTACCOUNT">>},{recipient_id,<<"TEST">>},federate]`

Second set of props:
`[{event_only,true},{account_id,<<"TESTACCOUNT">>
},federate]`

Previously this would be considered to be the same binding (due to the incorrect comparison) and as such the second binding for the generic would never be added.

PR #4974, opened 2018-07-11
[master] KAZOO-5926: check in teletype that the found addresses are really email addresses

icehess:
Checks list of emails to see if they're really an email address. For example in teletype fax utils one of the places ti search for email address is username in user document, and it possible the username is not an email address.

4.2: #4973

### DIFF
--- a/applications/teletype/src/teletype_util.erl
+++ b/applications/teletype/src/teletype_util.erl
@@ -709,14 +709,21 @@ get_address_value(Key, Path, [JObj|JObjs]) ->
 -spec check_address_value(binary() | kz_term:binaries() | kz_json:object() | 'undefined') -> kz_term:api_ne_binaries().
 check_address_value('undefined') -> 'undefined';
 check_address_value(<<>>) -> 'undefined';
-check_address_value(<<_/binary>> = Email) -> [Email];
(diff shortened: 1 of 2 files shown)


applications/teletype/src/teletype_util.erl
```diff
@@ -709,14 +709,21 @@ get_address_value(Key, Path, [JObj|JObjs]) ->
 -spec check_address_value(binary() | kz_term:binaries() | kz_json:object() | 'undefined') -> kz_term:api_ne_binaries().
 check_address_value('undefined') -> 'undefined';
 check_address_value(<<>>) -> 'undefined';
-check_address_value(<<_/binary>> = Email) -> [Email];
+check_address_value(<<_/binary>> = Email) -> check_address_value([Email]);
 check_address_value(Emails) when is_list(Emails) ->
-    case [E || E <- Emails, not kz_term:is_empty(E)] of
+    case [E || E <- Emails,
+               kz_term:is_ne_binary(E),
+               length(binary:split(E, <<"@">>)) == 2
```
jamesaimonetti: Since the expected length is known, you could use `[_, _] =:= binary:split(E, <<"@">>)` for the comparison.

icehess: I can't, I would get error `variable '_' is unbound`.

PR #4977, opened 2018-07-13
KAZOO-5934: fix response content-type when tunneled accept header is used

jamesaimonetti:
Also adds basic testing to the API (PropEr testing to come)

### DIFF
--- a/core/kazoo_apps/src/kapps_maintenance.erl
+++ b/core/kazoo_apps/src/kapps_maintenance.erl
@@ -1156,7 +1156,9 @@ check_release() ->
              ,fun kazoo_proper_maintenance:run_seq_modules/0
              ],
     try lists:foreach(fun(F) -> F() end, Checks) of
-        'ok' -> lager:info("check_release/0 succeeded"), init:stop()
+        'ok' ->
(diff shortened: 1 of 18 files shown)


core/kazoo_apps/src/kapps_maintenance.erl
```diff
@@ -1148,14 +1148,17 @@ init_system() ->
 
 -spec check_release() -> 'ok' | 'error'.
 check_release() ->
-    Checks = [fun kapps_started/0
+    Checks = [fun init_system/0
```
lazedo: you'll need toi to call `kz_datamgr:init_dbs()` prior to `init_system` as in `kapps_controller:initialize_kapps()` or move the call to `kz_datamgr:init_dbs()` to `init_system/0`

jamesaimonetti: `kapps_controller:initialize_kapps/0` is run on startup every time so we should be good there.

lazedo: yes, and it calls `init_system`,  why calling it here too ?

PR #4984, opened 2018-07-16
[4.2] UI-3102: change the default retry value to 3 from 1 

swysor:


lazedo:
@swysor wasn' this fixed on the ui side ? what's the reason for changing here ?

swysor:
I wanted the backend defaults to match the new ui defaults so the documentation is clearer

lazedo:
@swysor but this is a change in current behaviour. the documentation is about the api not the ui usage of the api.

PR #4989, opened 2018-07-17
handle short lived calls in calflow

lazedo:


applications/callflow/src/cf_exe.erl
```diff
@@ -68,7 +69,7 @@
                ,flows = [] :: kz_json:objects()
                ,cf_module_pid :: kz_term:api_pid_ref()
                ,cf_module_old_pid :: kz_term:api_pid_ref()
-               ,status = <<"sane">> :: kz_term:ne_binary()
+               ,status = <<"init">> :: kz_term:ne_binary()
```
jamesaimonetti: Is there a specific reason to keep this a binary? could we change it to an atom (then dialyzer can be more helpful in checking that `status` is a known value).

jamesaimonetti:
Just curious if you've tested a pivot call with this? I haven't looked too closely but want to make sure the `route_resume` path still works.

lazedo:
@jamesaimonetti haven't test it with `pivot`,  but i see no reason why it wouldn't work, tho the protection doesn't apply

PR #4992, opened 2018-07-20
KAZOO-5937: ecallmgr default mutivar separator is configurable now

sergey-safarov:

PR #4993, opened 2018-07-20
KAZOO-5937: ecallmgr default mutivar separator set to "~" (#4993)

sergey-safarov:


applications/ecallmgr/src/ecallmgr_util.erl
```diff
@@ -59,7 +59,7 @@
 
 -define(HTTP_GET_PREFIX, "http_cache://").
 
--define(FS_MULTI_VAR_SEP, ecallmgr_config:get_ne_binary(<<"multivar_separator">>, <<";">>)).
+-define(FS_MULTI_VAR_SEP, ecallmgr_config:get_ne_binary(<<"multivar_separator">>, <<"|">>)).
```
lazedo: @sergey-safarov thanks, you'll need to change the default in `system_config.ecallmgr.json`  [here](https://github.com/2600hz/kazoo/pull/4992/files#diff-ad6fddb22739cbba2b2a453234de7a2dR262) and run `make apis` to update swagger

sergey-safarov:
Lius @lazedo , done

sergey-safarov:
@lazedo value `"|"` is breaks IVR menu when on callee endpoint is enabled call recording

sergey-safarov:
@lazedo  suggest use as default value symbol `"~"`.
Tested on my servers

lazedo:
@sergey-safarov yes, `~` would be fine as `| ` is used for bridge separator

marodru:
I have tried to put it to the config as ~, it broke down inbound calls... so please really test carefully...
Intention was to fix tone_stream ringback with two "();" sections which is failing in 4.2 (Australian ring back tone).

sergey-safarov:
@marodru are you speack about updating ringback sting?
Is correctly I understand that you suggest use `(%(400,200,400,450);%(400,2200,400,450));` instead `%(400,200,400,450);%(400,2200,400,450)`

marodru:
@sergey-safarov I'm stating that setting multivar_separator to ~ broke calls so looks like not only | is unsafe.
about tone_stream - I'll try your notation with ()

lazedo:
@marodru can you show some evidence of your statement about breaking inbound calls with `~` ?

marodru:
@lazedo I will need to collect it properly in the lab when I reproduce it.
@sergey-safarov No, () don't work, @lazedo could you please help here? after 4.2 ringback values with more than one %(); block don't work, instead default US ringbacks are used despite ringback and transfer tones are set, when it comes to playing them, it uses defaults. One block works ok as expected.

lazedo:
@marodru please post your freeswitch logs (gist or pastebin or similar) with `~` set as default separator

sergey-safarov:
@marodru this not works because ringback have `";"` symbol threaded as separator. As not ringback description.
This is reason why need to redefine separator.

marodru:
@sergey-safarov Exactly. It did not work with another separator set in config though. According to logs ; was used anyway

lazedo:
@sergey-safarov please update the PR by changing `|` to `~` , it works as intended

sergey-safarov:
Hello Luis @lazedo 
PR is updated

PR #4995, opened 2018-07-23
[4.2] Fixed UK ringback and made configurable multivar separator

sergey-safarov:
Also  default default multivar separator set to `"~"`


lazedo:
thanks @sergey-safarov

PR #5003, opened 2018-07-25
check fetch_id for short-lived-calls

lazedo:


applications/callflow/src/cf_route_req.erl
```diff
@@ -176,11 +176,12 @@ send_route_response(Flow, JObj, Call) ->
...
     lager:info("callflow not ready after 5 tries, exiting");
 wait_for_running(Call, N) ->
+    FetchId = kapps_call:fetch_id(Call),
     receive
-        'channel_destroy' ->
+        {'channel_destroy', FetchId} ->
             cf_exe:hard_stop(Call),
             lager:info("received channel destroy while setting up callflow executor, exiting")
-    after 2000 ->
+    after 250 ->
```
jamesaimonetti: I wonder if we could keep the longer timeout but have cf_exe send a message to this process when the gen_listener bindings are setup? Maybe store this `pid()` in the kapps_call kvs store and when cf_exe gets the `{'gen_listener', {'is_consuming', 'true'}}` cast it sends back a message to this receive clause.

jamesaimonetti: Right, but this process is basically polling the cf_exe; why not have `cf_exe` tell this process when it is running so this process can exit without the extra checking?

Mostly I think the polling is unnecessary except in cases where cf_exe is really slow in getting its AMQP bindings established.

PR #5005, opened 2018-07-25
[4.2] QF: fixed slow api responses, better result messages for transmission and receive status

swysor:


applications/crossbar/src/modules/cb_faxes.erl
```diff
@@ -313,11 +313,17 @@ validate_inbox_fax_action(Action, Id, Context) ->
 
 -spec put(cb_context:context()) -> cb_context:context().
 put(Context) ->
-    maybe_save_attachment(crossbar_doc:save(Context)).
+    NewContext = crossbar_doc:save(Context),
+    _ = kz_util:spawn(fun maybe_save_attachment/1, [NewContext]),
+    %crossbar_util:response_202(<<"processing fax attachments">>, cb_context:doc(NewContext), NewContext).
```
lazedo: @swysor try to use the commented line with this change in crossbar_util:response_202/3
``` 
response_202(Msg, JTerm, Context) ->
    create_response('accepted', Msg, 202, JTerm, Context).
```

lazedo: i think the change to actually return `202` with proper envelope status needs careful analysis.
we can return the `201` for now (ie, return NewContext)

PR #5007, opened 2018-07-26
KAZOO-5938 - 4.2

pramsky:
should_lookup_cnam() does a check to see if use_stepswitch_cnam() is set to true in the couchdoc. If so, then incoming cnam requests are done against the stepswitch cnam service.

### DIFF
--- a/core/kazoo_number_manager/src/carriers/knm_vitelity.erl
+++ b/core/kazoo_number_manager/src/carriers/knm_vitelity.erl
@@ -102,10 +102,7 @@ disconnect_number(Number) ->
 %%------------------------------------------------------------------------------
 -spec should_lookup_cnam() -> boolean().
 should_lookup_cnam() ->
-    case kapps_config:get_ne_binary(?KNM_VITELITY_CONFIG_CAT, <<"use_stepswitch_cnam">>) of
-        <<"true">> -> 'true';
-        _          -> 'false'
-    end.
+    kapps_config:get_is_true(?KNM_VITELITY_CONFIG_CAT, <<"use_stepswitch_cnam">>, false).
 
 %%------------------------------------------------------------------------------
 %% @doc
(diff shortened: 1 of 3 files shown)


pramsky:
Not sure why check-unstaged is erroring ? Running it on my local does not throw that error

lazedo:
@pramsky check the errors [here](https://circleci.com/gh/2600hz/kazoo/12998) , your schema needs to add the type and there's also formatting issues (run `make fmt`)

pramsky:
Ahh got it, looking at the results, it looks like i need to tab 4 instead of tab 2

lazedo:
@pramsky i added the `string` type to schema. before we can consider merging this into 4.2 branch, we will need a PR for master

pramsky:
I'll do the update for master and issue a new PR. Thanks for your help Luis!

core/kazoo_number_manager/src/carriers/knm_vitelity.erl
```diff
@@ -96,11 +96,16 @@ disconnect_number(Number) ->
...
 %%------------------------------------------------------------------------------
-%% @doc
+%% @doc Check to see if use_stepswitch_cnam is defined in the couchdoc. If it is
+%% set to true, then incoming calls will use stepswitch for cnam
 %% @end
 %%------------------------------------------------------------------------------
 -spec should_lookup_cnam() -> boolean().
-should_lookup_cnam() -> 'false'.
+should_lookup_cnam() ->
+    case kapps_config:get_ne_binary(?KNM_VITELITY_CONFIG_CAT, <<"use_stepswitch_cnam">>) of
```
icehess: you can use `kapps_config:get_is_true/2,3` which returns boolean `true` or `false`, so you don't need to use `case`.

pramsky: That works great, updating PR now. Thanks!!!!

PR #5013, opened 2018-07-26
kz_network_utils: Fixed false error message for ping to IPv6 loopback

sergey-safarov:
When on host is not used IPv6 address then kazoo reports false error message.
That PR is fix this.
Login if hosts not have IPv6 address, then host not able to ping any IPv6 address.

### DIFF
--- a/core/kazoo/src/kz_network_utils.erl
+++ b/core/kazoo/src/kz_network_utils.erl
@@ -51,6 +51,8 @@
         ,set_option_usevc/2
         ]).
 
+-export([get_local_ip/0]).
+
 -include_lib("kernel/include/inet.hrl").
 -include_lib("kernel/src/inet_dns.hrl").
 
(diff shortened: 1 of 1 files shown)


core/kazoo/src/kz_network_utils.erl
```diff
@@ -123,6 +125,25 @@ is_cidr(Address) ->
         'error':'invalid_cidr' -> 'false'
     end.
 
+-spec get_local_ip() -> [].
+get_local_ip() ->
+{ok, Descriptions} = inet:getifaddrs(),
+IPs = lists:map(fun(Description) ->
```
icehess: You can write the whole functions (and below function) for getting interface IP addresses like this:

```erlang
get_interfaces_ip() ->
    case inet:getifadrrs() of
        {'ok', IFs} ->
            [IP || {_, IFOpts} <- IFs,
                   {'addr', IP} <- IFOpts,
                   {'flags', Flags} <- IFOpts,
                   'up' <- Flags
            ];                                                                                                                             
        {'error', _Reason} ->
            lager:debug("failed to get interfaces: ~p", [_Reason]),
            []
    end.
```

icehess:
Also can you check https://github.com/2600hz/kazoo/pull/5006 for another approach and see that fix your false error message problem.

sergey-safarov:
Thanks @icehess think #5006 will fix this.

PR #5019, opened 2018-07-30
add tokens cache

lazedo:


core/kazoo_auth/src/kz_auth_client.erl
```diff
@@ -304,3 +315,58 @@ add_claims(Claims) ->
...
+        {'ok', 200, _RespHeaders, RespBody} -> {ok, RespBody};
+        {'ok', 401, _RespHeaders, _RespXML} -> {error, 'not_authorized'};
+        {'ok', 404, _RespHeaders, _RespXML} -> {error, 'not_found'};
+        {'ok', _Code, _RespHeaders, _RespXML} -> {error, 'other'};
+        {'error', _}=Error -> Error
+    end.
+
+-spec cached_token(map()) -> map().
+cached_token(#{subject := #{'_id' := Id}} = Map) ->
+    case kz_cache:fetch_local(?TOKENS_CACHE, {client, Id}) of
```
jamesaimonetti: I don't follow why the 'expires_absolute' flag is needed when `kz_cache:peek_local` takes care of not bumping the timestamp used for expiration. Can you not use `peek_local` here instead of making the change to kazoo_caches?

core/kazoo_auth/src/kz_auth_client.erl
```diff
@@ -304,3 +315,56 @@ add_claims(Claims) ->
 
 add_claim(<<"iat">>) -> {iat, kz_time:current_unix_tstamp()-500};
 add_claim(<<"exp">>) -> {exp, kz_time:current_unix_tstamp()+(2 * ?MILLISECONDS_IN_SECOND)}.
+
+-spec request(kz_term:ne_binary(), map()) -> kz_http:ret().
+request(URL, Token) ->
+    request('get', URL, <<>>, Token).
+
+-spec request(atom(), kz_term:ne_binary(), binary(), map()) -> {ok|error, any(), kz_term:proplist()}.
```
jamesaimonetti: Return type is wrong. Should be `kz_http:req()` I think.

core/kazoo_auth/src/kz_auth_client.erl
```diff
@@ -304,3 +315,56 @@ add_claims(Claims) ->
...
+request(Verb, URL, Body, #{token := #{authorization := Authorization}}) ->
+    Options = [{headers_as_is, true}, {ssl, [{versions, ['tlsv1.2']}]}],
+    {ok,{_,_, Host, _, _, _}} = http_uri:parse(URL),
+    Headers = [{<<"host">>, Host}
```
jamesaimonetti: can we standardize the headers? You have two keys as binaries and one as a string, `"host"` is all lower case while the others use capital letters.

lazedo: @jamesaimonetti that's intended (  `{headers_as_is, true}` )

core/kazoo_auth/src/kz_auth_client.erl
lazedo: fixing

PR #5035, opened 2018-08-06
Fix kazoo_number_manager_maintenance fix_account_numbers

kalda341:
Currently the sync button in Monster UI will actually remove valid numbers from the list of available numbers, but clicking it again will result in them coming back. This corrects that.


core/kazoo_number_manager/src/kazoo_number_manager_maintenance.erl
```diff
@@ -260,7 +260,7 @@ fix_account_numbers(AccountDb = ?MATCH_ACCOUNT_ENCODED(A,B,Rest)) ->
                             %% timer:sleep(?TIME_BETWEEN_ACCOUNTS_MS),
                             gb_sets:subtract(Leftovers, AuthoritativePNs)
                     end
-                   ,fun (Set1, Set2) -> gb_sets:union(Set1, Set2) end
+                   ,fun (Set1, Set2) -> gb_sets:intersection(Set1, Set2) end
```
jamesaimonetti: No need for the fun (not sure why its there); can just be `fun gb_sets:intersection/2`

kalda341: Good catch!

PR #5037, opened 2018-08-06
HELP-37990: added optional quantifiers for faxbox, vmbox and conference

k-anderson:
There is no master variant as it will be included in KAZOO-5564.

### DIFF
--- a/applications/crossbar/priv/couchdb/account/services.json
+++ b/applications/crossbar/priv/couchdb/account/services.json
@@ -26,7 +26,7 @@
             "reduce": "_sum"
         },
         "voicemails": {
-            "map": "function(doc) { if (doc.pvt_type != 'vmbox' || doc.pvt_deleted || doc.enabled == false) return; emit('mailbox', 1); }",
+            "map": "function(doc) { if (doc.pvt_type != 'vmbox' || doc.pvt_deleted || doc.enabled == false) return; emit('vmbox', 1); }",
             "reduce": "_sum"
         }
     }
--- a/applications/crossbar/priv/couchdb/account/services.json
+++ b/applications/crossbar/priv/couchdb/account/services.json
@@ -2,8 +2,16 @@
     "_id": "_design/services",
     "language": "javascript",
     "views": {
+        "conferences": {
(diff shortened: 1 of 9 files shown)


applications/crossbar/priv/couchdb/account/services.json
```diff
@@ -16,6 +24,10 @@
         "users": {
             "map": "function(doc) { if (doc.pvt_type != 'user' || doc.pvt_deleted || doc.enabled == false) return; emit(doc.priv_level || 'user', 1); }",
             "reduce": "_sum"
+        },
+        "voicemails": {
+            "map": "function(doc) { if (doc.pvt_type != 'vmbox' || doc.pvt_deleted || doc.enabled == false) return; emit('mailbox', 1); }",
```
jamesaimonetti: Why mailbox here? The other changes appear to follow the `pvt_type`. Also, why not a "generic" map of `if(deleted/not_enabled) return; emit(pvt_type, 1);`? Then query based on pvt_type and reduce to get the count?

k-anderson: @jamesaimonetti I am not doing a generic pvt_type map to control churn on the services doc and only selectively quantity things that are billable for the cluster (as controlled in system_config).  It also makes more logical groupings for the items.  However, I have come around to using the pvt_type 'vmbox' rather than 'mailbox' so I will make that change now :)

jamesaimonetti: works for me :)

core/kazoo_services/src/services/kz_service_conferences.erl
```diff
@@ -0,0 +1,57 @@
...
+    Routines = [fun reconcile_faxbox/2],
+    lists:foldl(fun(F, S) ->
+                        F(S, JObj)
+                end
+               ,reconcile(Services)
+               ,Routines
+               ).
+
+-spec reconcile_faxbox(kz_services:services(), kz_json:object()) -> kz_services:services().
+reconcile_faxbox(Services, JObj) ->
```
lazedo: faxbox ?

PR #5038, opened 2018-08-06
Call recording doc update

jamesaimonetti:


jamesaimonetti:
@lazedo one question I wasn't sure of was if call recording follows call-forwarded devices? I assume so but would like confirmation and would like to explicitly state it in the doc.

PR #5039, opened 2018-08-06
[4.2] Fixed application kazoo_convert not starting automatically

swysor:


core/kazoo_documents/src/kazoo_documents.app.src
```diff
@@ -7,13 +7,12 @@
  ,{applications, [ kernel
                  , stdlib
                  , crypto
-
+                 , kazoo_convert
```
lazedo: @swysor i think you should add this to `fax` and `teletype`. i don't like it being part of `kazoo_documents`

PR #5040, opened 2018-08-06
Fixed application kazoo_convert not starting automatically

swysor:


applications/fax/src/fax_maintenance.erl
```diff
@@ -430,23 +429,3 @@ next_key(<<>>) ->
...
-load_smtp_attachment(DocId, Filename, FileContents) ->
-    CT = kz_mime:from_filename(Filename),
-    case kz_datamgr:open_cache_doc(?KZ_FAXES_DB, DocId) of
-        {'ok', JObj} ->
-            case kzd_fax:save_outbound_fax(?KZ_FAXES_DB, JObj, FileContents, CT) of
-                {'ok', _Doc} -> io:format("attachment ~s for docid ~s recovered~n", [Filename, DocId]);
-                {'error', E} -> io:format("error attaching ~s to docid ~s : ~p~n", [Filename, DocId, E])
-            end;
-        {'error', E} -> io:format("error opening docid ~s for attaching ~s : ~p~n", [DocId, Filename, E])
-    end.
```
lazedo: @swysor why are we removing this ?

swysor: It's not used anywhere.

lazedo: its a maintenance command. if smtp processing fails a system notify is sent and a admin can recover with this maintenance command

swysor: Oh right, ok, I will restore that.

lazedo: the attachment is extracted from smtp and put into tmp dir. if it fails to upload to bigcouch a system alert is sent.
the fax doc is created but has no attachment, this allow a admin to retry saving the attachment to bigcouch.

swysor:
Ah ok thanks for the info!

applications/fax/src/fax_maintenance.erl
swysor: Fixed

PR #5042, opened 2018-08-08
[4.2] Twiml gather finishondtmf

jamesaimonetti:
Empty finishOnKey means no DTMF cancel of the gather


lazedo:
@jamesaimonetti should last 2 commits from https://github.com/2600hz/kazoo/pull/5041 be here too ?

PR #5043, opened 2018-08-08
HELP-38014: work around bottleneck in teletype handle_event

swysor:


applications/teletype/src/teletype_bindings.erl
```diff
@@ -53,7 +53,14 @@ start_modules([]) ->
 notification(JObj) ->
     kz_util:put_callid(JObj),
     {EventCategory, EventName} = kz_util:get_event_type(JObj),
+    ShouldHandle = teletype_util:should_handle_notification(JObj),
```
jamesaimonetti: What happens if we make this an assertion instead?

```erlang
notification(JObj) ->
    'true' = teletype_util:should_handle_notification(JObj),
    ...
```

So the spawned process will just crash immediately if it won't handle the payload.

icehess: I don't agree, it produce not necessary CRASH REPORT log for a case which teletype is not the default notify app.

icehess:
@swysor Can you add this to your PR please:

```diff
diff --git a/core/kazoo_apps/src/kapps_notify_publisher.erl b/core/kazoo_apps/src/kapps_notify_publisher.erl
index b3c3cdc4a6..d3c6ba675b 100644
--- a/core/kazoo_apps/src/kapps_notify_publisher.erl
+++ b/core/kazoo_apps/src/kapps_notify_publisher.erl
@@ -303,13 +303,12 @@ find_reason_from_jsons(Reason, JObjs, Map) ->
         Val -> maps:update_with(Reason, fun(List) -> [Val|List] end, [Val], Map)
     end.
 
--spec cast_to_binary(any()) -> kz_term:api_ne_binary().
+-spec cast_to_binary(any()) -> kz_term:ne_binary().
 cast_to_binary(Error) ->
     try kz_term:to_binary(Error)
     catch
         _:_ ->
-            lager:debug("failed to convert notification failure reason to binary: ~p", [Error]),
-            'undefined'
+            kz_term:to_binary(io_lib:format("~p", [Error]))
     end.
 
 %%------------------------------------------------------------------------------
```

applications/teletype/src/teletype_bindings.erl
jamesaimonetti: We have assertions in other event handlers for validating the payloads `true = kapi_whatever:payload_v(JObj),` so it isn't unprecedented. But I don't have a burning desire to see this change; just suggesting it for happy-case programming vs defensive programming.

PR #5046, opened 2018-08-09
PISTON-598: add provider support to stepswitch_cnam + Telnyx mod

danielfinke:
Default handler for stepswitch_cnam will not handle the Telnyx CNAM lookup API - made the module generic and allowing of provider modules. Kept old code as default.


applications/stepswitch/src/cnam_providers/stepswitch_cnam_telnyx.erl
```diff
@@ -0,0 +1,84 @@
...
+-spec get_http_options(nonempty_string()) -> kz_term:proplist().
+get_http_options(Url) ->
+    Defaults = [{'connect_timeout', ?HTTP_CONNECT_TIMEOUT_MS}
+               ,{'timeout', 1500}
+               ],
+    maybe_enable_ssl(Url, Defaults).
+
+-spec maybe_enable_ssl(nonempty_string(), kz_term:proplist()) -> kz_term:proplist().
+maybe_enable_ssl("https://" ++ _, Props) ->
+    [{'ssl', [{'verify', 'verify_none'}]}|Props];
```
jamesaimonetti: Does telnyx not provider a proper TLS certificate that Kazoo can verify against? Testing the domain seems like the cert is trustable: https://www.ssllabs.com/ssltest/analyze.html?d=data.telnyx.com&latest

Were there issues here? Or is it mostly holdover from the opencnam module? If so, let's test verifying the peer's cert.

jamesaimonetti:
@danielfinke thanks, merged!

PR #5057, opened 2018-08-14
Google Speech TTS support for kazoo_speech

remcovanvugt:
This is a first version of a module to support Google Speech TTS API's in Kazoo.
It currently only works for cb_media in crossbar, and not for cf_tts in callflows.
This is because Google returns a base64 encoded string inside a JSON payload. 

This seems to be handled in kz_media_tts_cache, but this fails because google sends charset=UTF-8 along with the type in the reponse payload. I added the scenario to handle this gracefully for now. I will probably update this later to add support for using cf_tts.


core/kazoo_media/src/kz_media_tts_cache.erl
```diff
@@ -171,8 +171,13 @@ handle_info({'http', {ReqID, 'stream', Bin}}, #state{kz_http_req_id=ReqID
                                    ,timer_ref=start_timer()
                                    }};
         <<"application/json">> ->
+            lager:debug("JSON response: ~s", [Bin]),
+            {'noreply', State, 'hibernate'};
+
+        <<"application/json; charset=UTF-8">> ->
```
lazedo: you can change this to one clause with  `<<"application/json", _/binary>>` tho is strange the content-type comes that way

jamesaimonetti: Better would be to use something like we did in [crossbar](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/src/modules/cb_modules_util.erl#L105) and parse the header. You get back something like `{{<<"application">>, <<"json">>, '*' | proplist()}, _Weight, _Extensions}`

remcovanvugt:
Thanks for the pointers, I will update this PR shortly

jamesaimonetti:
@remcovanvugt hey, just checking in to see if you had a chance to make the changes asked? Need any help?

sergey-safarov:
Hello @remcovanvugt @jamesaimonetti 
Could you wait some time and i will post my PR for Google TTS

I resolved TTS from callflow too

jamesaimonetti:
@remcovanvugt is this PR still relevant or is the contribution from Sergey working for you?

remcovanvugt:
@jamesaimonetti The contribution from Sergey works, this PR can be closed. Thanks!

PR #5059, opened 2018-08-15
[4.2] HELP-37107: fixed kz_notify_resend to use a timer, added debug logs t…

swysor:
…o instrument this module


lazedo:
@swysor did you try to add `?TIME_BETWEEN_CYCLE` to the missing clauses ?

swysor:
I did not, Hesaam wanted to switch to using the timer method since we know that works everywhere else we use it. I have tested this and its been working reliably today. We decided to make the change when we observed that even instrumented, this module was not producing any logs, despite being a legit process in kz_globals.

lazedo:
@swysor the current implementation has flaws because not every `handle_cast` & `handle_info` return the proper tuple with `?TIME_BETWEEN_CYCLE` which will cancel the timer.

PR #5064, opened 2018-08-17
Toggle whether to publish missing rate system alerts

jamesaimonetti:


applications/hotornot/src/hon_rater.erl
```diff
@@ -91,6 +91,16 @@ get_rate_data(RateReq, _AuthType) ->
             get_rate_data(RateReq, ToDID, FromDID, Rates)
     end.
 
+-spec maybe_send_system_alert(kapi_rate:req(), kz_term:ne_binary(), kz_term:ne_binary()) -> 'ok'.
+maybe_send_system_alert(RateReq, FromDID, ToDID) ->
+    Direction = kapi_rate:direction(RateReq),
+    case hotornot_config:should_publish_alert(Direction) of
+        'true' ->
+            kz_notify:system_alert("no rate found for ~s to ~s", [FromDID, ToDID]);
```
lazedo: can we add `Direction` in the notification ?

PR #5065, opened 2018-08-17
HELP-37837: Fix tasks app startup race condition issue

harenson:


applications/tasks/src/knm_number_crawler.erl
```diff
@@ -98,6 +98,9 @@ handle_call(_Request, _From, State) ->
 handle_cast('stop', State) ->
     lager:debug("crawler has been stopped"),
     {'stop', 'normal', State};
+handle_cast({'$proxy_stop', Reason}, State) ->
+    lager:debug("~p stopping with reason: ~p", [?MODULE, Reason]),
+    {stop, Reason, State};
```
jamesaimonetti: plz tickie all the atoms

lazedo: `handle_cast({'$proxy_stop', Reason}, State) ->` are these really necessary ?

applications/tasks/src/knm_port_request_crawler.erl
```diff
@@ -82,6 +82,9 @@ handle_call(_Request, _From, State) ->
 handle_cast('stop', State) ->
     lager:debug("crawler has been stopped"),
     {'stop', 'normal', State};
+handle_cast({'$proxy_stop', Reason}, State) ->
```
lazedo: same here, will skip the others

core/kazoo_globals/src/kz_globals.erl
```diff
@@ -574,6 +581,10 @@ do_amqp_unregister(Global, Reason) ->
               ],
     lager:debug("deleting ~p", [Name]),
     ets:delete(?TAB_NAME, Name),
+    kz_util:spawn(fun() ->
+                          timer:sleep(500), % Give time to other nodes to register the name
+                          gen_listener:call(?SERVER, {'amqp_delete', Global, Reason})
+                  end),
```
lazedo: is this necessary ?

core/kazoo_globals/src/kz_globals.erl
```diff
@@ -150,8 +150,13 @@ where_is(Name) ->
 -spec lookup_name(kz_global:name()) -> kz_global:global() | 'undefined'.
 lookup_name(Name) ->
     case ets:lookup(?TAB_NAME, Name) of
-        [Global] -> Global;
-        [] -> 'undefined'
+        [Global] ->
+            case is_process_alive(kz_global:pid(Global)) of
+                'true' -> Global;
+                'false' -> 'undefined'
```
icehess: If process is not alive shouldn't we delete it from ETS table?

jamesaimonetti:
@harenson can you rebase and address @lazedo's comments?

lazedo:
@harenson i think the only change that makes sense is the `lookup_name`. the other changes need explanation of their need.

PR #5076, opened 2018-08-27
KAZOO-5564: refactor kazoo_services

k-anderson:


lazedo:
@jamesaimonetti `ci` fails because `crossbar_services` spawns process to process updates & audit log when the account is already deleted

PR #5078, opened 2018-08-27
HELP-39096: use auth_user timestamp for attachment filenames in cb_faxes

swysor:


applications/crossbar/src/modules/cb_faxes.erl
```diff
@@ -643,13 +643,38 @@ set_fax_binary(Context, Content, ContentType, Filename) ->
...
+                        lager:debug("bad TZ: ~p", [Timezone]),
+                        UtcTime
+                end,
+    TimeString = kz_time:pretty_print_datetime(LocalTime),
+    re:replace(kz_term:to_lower_binary(TimeString), <<"\\s+">>, <<"_">>, [{'return', 'binary'}, 'global']).
+
+
+-spec get_timezone(cb_context:context()) -> kz_term:ne_binary().
+get_timezone(Context) ->
+    {'ok', UserDoc} = kz_datamgr:open_cache_doc(kz_util:format_account_db(cb_context:auth_account_id(Context)), cb_context:auth_user_id(Context)),
```
swysor: Should I maybe not let it crash here? Wondering if this would break if I tried to download with an API key

PR #5080, opened 2018-08-28
[4.2] HELP-39082: fixed call dropping when deleting a not configured tempor…

swysor:
…ary greeting


jamesaimonetti:
See master at #5081

PR #5082, opened 2018-08-29
Prepend after privacy

sudomabider:
When privacy is enabled the prepend gets overwritten after anonymization. This fixes it by applying the prepend again afterwards.

### DIFF
--- a/core/kazoo_endpoint/src/kz_attributes.erl
+++ b/core/kazoo_endpoint/src/kz_attributes.erl
@@ -21,6 +21,7 @@
 -export([get_account_external_cid/1]).
 -export([maybe_get_assigned_number/3]).
 -export([maybe_get_account_default_number/4]).
+-export([maybe_prefix_cid/5]).
 -export([get_flags/2]).
 -export([process_dynamic_flags/2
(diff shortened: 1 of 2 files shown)


lazedo:
@sudomabider thanks for your effort. however, this doesn't seem right. can you describe what your use case is and/or what you're  trying to fix ?

it seems to me that your change will override the applied privacy settings if there's a prepend (which was already applied before privacy)

`kz_endpoint:get_clid` calls `kz_attributes:caller_id` which applies the prepend, then privacy is applied.

maybe your use case needs a fix elsewhere

sudomabider:
@lazedo Thanks for the reply. 

So our expectation is that the prepend should be preserved after privacy. This used to be the case before 4.1.  Since 4.1 (including in 4.2) the prepend now gets removed by privacy. And it's causing some issues.

DrJacobHolden:
Any progress on this?
@lazedo this pull request is just fixing what appears to be a bug. Currently when prepend is applied to a private number the prepend is removed by privacy, i.e. you cannot have a prepend apply to private numbers without this fix.

PR #5087, opened 2018-08-30
HELP-38834: ignore soft deleted documents in open_doc

swysor:


core/kazoo_data/src/kzs_doc.erl
```diff
@@ -32,7 +32,11 @@
                       {'ok', kz_json:object()} |
                       data_error().
 open_doc(#{server := {App, Conn}}, DbName, DocId, Options) ->
-    App:open_doc(Conn, DbName, DocId, Options).
+    Response = App:open_doc(Conn, DbName, DocId, Options),
+    case props:get_value('ignore_soft_delete', Options, 'false') of
```
jamesaimonetti: Hmm, `ignore_soft_delete` doesn't seem immediately obvious in what it should do. What about using `deleted` because you can already fetch a doc that has been hard-deleted: `kz_datamgr:open_doc(Db, DocId, [{'deleted', 'true'}])`. Maybe we extend it to also include fetching soft-deleted docs.

```erlang
    handle_opened_doc(App:open_doc(Conn, DbName, DocId, Options), Options).

handle_opened_doc({'error', _}=E, _Options) -> E;
handle_opened_doc({'ok', Doc}=OK, Options) ->
    handle_opened_doc(Doc, kz_doc:is_deleted(Doc), props:is_true('deleted', Options)).

handle_opened_doc(Doc, 'false',  _ReturnDeleted) -> {'ok', Doc};
handle_opened_doc(Doc, 'true', 'true') -> {'ok', Doc};
handle_opened_doc(Doc, 'true', 'false') -> {'error', 'not_found'}.
```

Is the filter-by-doc-type really needed? I think the caller should be explicit when they potentially want the hard/soft-deleted doc.

My 2 cents anyway... :)

core/kazoo_data/src/kzs_doc.erl
```diff
@@ -32,7 +32,21 @@
...
+handle_opened_doc({'error', _}=Error, _Options) -> Error;
+handle_opened_doc({'ok', Doc}, Options) ->
+    handle_opened_doc(Doc, kz_doc:is_deleted(Doc), props:is_true('deleted', Options)).
+
+-spec handle_opened_doc({'ok', kz_json:object()}|data_error(), boolean(), boolean()) ->
+                               {'ok', kz_json:object()} | data_error().
+handle_opened_doc(Doc, 'false',  _ReturnDeleted) -> {'ok', Doc};
+handle_opened_doc(Doc, 'true', 'true') -> {'ok', Doc};
+handle_opened_doc(Doc, 'true', 'false') ->
+    lager:info("denying access to soft deleted doc: ~s", [kz_doc:id(Doc)]),
```
jamesaimonetti: `s/soft/a` since `kz_doc:is_deleted/1` checks hard- and soft-deleted properties

PR #5097, opened 2018-09-05
Calculate ringing seconds properly based on total call duration minus…

vladimirralev:
… the billable minus the time it takes to start ringing

taking progressmsec is wrong because it is the time from INVITE to 180 as speced here https://freeswitch.org/confluence/display/FREESWITCH/mod_xml_cdr

and ringing secs are already documented to be the actual ring duration

    "ringing_seconds": {
      "type": "string",
      "required": false,
      "name": "Ringing (seconds)",
      "description": "How many seconds the leg was ringing (pre-answer)"
    },

### DIFF
--- a/applications/ecallmgr/src/ecallmgr_call_events.erl
+++ b/applications/ecallmgr/src/ecallmgr_call_events.erl
@@ -1089,9 +1089,9 @@ get_billing_seconds(Props) ->
 -spec get_ringing_seconds(kz_term:proplist()) -> kz_term:ne_binary().
 get_ringing_seconds(Props) ->
     RingingSeconds =
-      props:get_integer_value(<<"variable_duration">>, Props) -
-      get_billing_seconds(Props) -
-      props:get_integer_value(<<"variable_progresssec">>, Props),
+        props:get_integer_value(<<"variable_duration">>, Props)
+        - get_billing_seconds(Props)
+        - props:get_integer_value(<<"variable_progresssec">>, Props),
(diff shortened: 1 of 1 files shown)


applications/ecallmgr/src/ecallmgr_call_events.erl
```diff
@@ -1086,6 +1086,12 @@ get_billing_seconds(Props) ->
         Billmsec -> kz_term:to_binary(kz_term:ceiling(Billmsec / 1000))
     end.
 
+-spec get_ringing_seconds(kz_term:proplist()) -> kz_term:api_binary().
```
jamesaimonetti: Return should be `kz_term:ne_binary()` since `'undefined'` can't be returned.

vladimirralev: Ah ok. Done.

vladimirralev:
I was trying to pass the circleci tests, I thought the staging check script scanned for diff marks and got confused by the minus signs  but that's not the case. Please advise.

jamesaimonetti:
@vladimirralev you need to run `make fmt` to align your code properly. You can run it from the root of the project or from the `/applications/ecallmgr` directory.

vladimirralev:
Got it. I think its ready now.

lazedo:
```
if ( answered_time == 0 ) then
 length_of_phone_ringing = hangup_time - created_time
else   
 length_of_phone_ringing = answered_time - created_time
end
```

PR #5099, opened 2018-09-06
KAZOO-5964: Convert tiff group3 to tiff group4 due to issues with tiff2pdf 4.0.3 inverting colors

swysor:


swysor:
Does not fix this... :(

PR #5102, opened 2018-09-09
Added convertapi support

sergey-safarov:


swysor:
Circle is failing due to the type int he schema not matching the default 

/home/ubuntu/2600hz/kazoo/applications/crossbar/priv/couchdb/schemas/system_config.kazoo_convert.convertapi.json:1: bad default type
default does not match type
at: timeout
type: string
default: 60

Should be type: number

swysor:
The test that is timing out, can you reproduce that issue on your end or does it only time out in CI?

core/kazoo_convert/src/kz_fax_converter.erl
```diff
@@ -176,6 +186,10 @@ image_to_tiff(FromPath, #{<<"from_format">> := <<"image/tiff">>, <<"tmp_dir">> :
 image_to_tiff(FromPath, Options) ->
     convert_file(?CONVERT_IMAGE_COMMAND, FromPath, <<".tiff">>, Options).
 
+-spec image_to_pdf(kz_term:ne_binary(), map()) -> fax_converted().
```
swysor: Is there a use case for this in faxing?

core/kazoo_convert/src/kz_fax_converter.erl
```diff
@@ -370,8 +384,9 @@ format_output(FilePath, #{<<"output_type">> := 'binary'}) ->
     end;
 format_output(FilePath, #{<<"output_type">> := 'path'}) ->
     {'ok', FilePath};
-format_output(FilePath, _Options) ->
-    {'ok', FilePath}.
+format_output(FilePath, Options) ->
+    read_file({'file', FilePath}, Options).
```
swysor: This changes the default format from 'path' to 'binary', whats the need for this? You can always ask for binary if you need binary.

core/kazoo_convert/src/kz_fax_converter.erl
sergey-safarov: For faxing is not.
I simple can see that all types of converters to `TIFF` and `PDF` is exist except of image_to_pdf.
This is reason why i added this.

core/kazoo_convert/src/kz_fax_converter.erl
sergey-safarov: Current master implementations is always return `path` and ignores `binary` options.
I missed default value. Will update to
```
format_output(FilePath, _) ->
    {'ok', FilePath};
format_output(FilePath, #{<<"output_type">> := 'binary'}=Options)
    read_file({'file', FilePath}, Options).
```

sergey-safarov: yes, i revert back this my change

core/kazoo_convert/src/kz_fax_converter.erl
swysor: Since we don't have a use case for this, can you please revert this? I explicitly excluded cases that were not relevant for faxing in the kz_fax_converter as we plan on adding and moving more converters ro this lib in the future (various media conversions like wav to mp3, and other types of files for teletype notifications, port request documents etc..).  

I would prefer to keep the fax_converter cases specific to use cases related to faxing only. For faxing this type of conversion would always be from tiff to pdf and never other types of image to pdf since the PDF is always generated from the tiff we sent/received via the fax_smtp and fax_request.

sergey-safarov: Ok I also will remove unneeded commits.

sergey-safarov:
@swysor think pr is ready to squash and merge

applications/crossbar/priv/api/swagger.json
```diff
@@ -30191,6 +30191,58 @@
...
+                        "1.4",
+                        "1.5",
+                        "1.6",
+                        "1.7"
+                    ],
+                    "type": "string"
+                },
+                "resolution": {
+                    "default": 200,
+                    "description": "Whith document resolution is used for converted files",
```
swysor: Typo here

sergey-safarov:
Hello @swysor
Typo is fixed.
May be rename `core/kazoo_convert/src/providers` to `core/kazoo_convert/src/fax_providers`?

swysor:
Yeah, i think src/providers works ok actually, that way we can use that dir for any of the providers we make as we extend this for all the other convert providers for media and other things. Do you want to move the kz_fax_converter into the providers dir as well? 

I am going to update the docs a bit and push a commit for that. Do you want to move the kz_fax_converter under providers in the meantime?

core/kazoo_convert/src/providers/kz_fax_converter_convertapi.erl
```diff
@@ -0,0 +1,295 @@
...
+%%% @copyright (C) 2018-, 2600Hz
+%%% @doc
+%%% @end
+%%%-----------------------------------------------------------------------------
+-module(kz_fax_converter_convertapi).
+
+-export([convert/4
+        ,read_metadata/1
+        ]).
+
```
swysor: Can you modify this to implement the behavior?

-behaviour(gen_kz_converter).

sergey-safarov:
Hello @swysor 
Is pr ready to merge?

PR #5104, opened 2018-09-10
Kazoo 5564 docs

k-anderson:


core/kazoo_services/src/kz_services_plans.erl
```diff
@@ -355,7 +355,7 @@ merge_to_single(PlansJObjs, JObj) ->
 
 -spec merge_to_single_fold(kz_json:object(), kz_json:object()) -> kz_json:object().
 merge_to_single_fold(JObj, Merged) ->
-    kz_json:merge(Merged, JObj).
+    kz_json:merge_recursive(Merged, JObj).
```
jamesaimonetti: Curious why merge_recursive was the right choice over merge?

k-anderson: @jamesaimonetti to allow parameters from each merge.strategy to combine vs taking the highest priority base object.  For example, each strategy has plans.devices.XXX, where XXX is unique per strategy, want to include each in the final plan.

PR #5111, opened 2018-09-13
[master] KAZOO-5970: service invoice template

icehess:


applications/crossbar/priv/couchdb/schemas/kapi.notifications.service_added.json
```diff
@@ -39,6 +39,9 @@
         "HTML": {
             "type": "string"
         },
+        "Items": {
+            "type": "string"
```
jamesaimonetti: `Items` don't appear to be strings but arrays of objects?

PR #5114, opened 2018-09-13
some config fixes

lazedo:
* applies the defaults honoring existing values
* fixes get_node_value crash
* excludes more fields from built schema
* allow removal of values


core/kazoo_apps/src/kapps_config_doc.erl
```diff
@@ -64,7 +65,7 @@ apply_schema_defaults_to_default(Id, Config) ->
 
 -spec schema_defaults(kz_term:ne_binary()) -> kz_json:object().
 schema_defaults(Id) ->
-    kz_json_schema:default_object(kapps_config_util:system_schema(Id)).
+    kz_json_schema:default_object(kapps_config_util:system_config_document_schema(Id)).
```
jamesaimonetti: Can we add a test that showed the error using `system_schema/1` that is resolved with using `system_config_document_schema/1`?

lazedo: @jamesaimonetti that's what this is supposed to fix. however needs further work on configs api.

PR #5124, opened 2018-09-17
konami: trunkstore added metaflow support

sergey-safarov:


applications/konami/src/konami_util.erl
```diff
@@ -41,3 +43,35 @@ send_break_req(CallId) ->
           ],
     lager:debug("attempting to break ~s", [CallId]),
     kz_amqp_worker:cast(API, fun kapi_metaflow:publish_action/1).
+
```
jamesaimonetti: Kazoo applications (those apps under `/applications`) should not call code that lives in other Kazoo applications. If this is accepted, trunkstore would depend on konami being present to avoid crashing.

The proper thing to do would be to put this in an appropriate core library; perhaps `kz_endpoint`?

applications/callflow/src/cf_route_win.erl
```diff
@@ -279,19 +279,32 @@ maybe_start_metaflow(Call) ->
 -spec maybe_start_metaflow(kapps_call:call(), kz_term:api_binary()) -> kapps_call:call().
 maybe_start_metaflow(Call, 'undefined') ->
     maybe_start_endpoint_metaflow(Call, kapps_call:authorizing_id(Call)),
+    maybe_start_ts_metaflow(Call, kapps_call:authorizing_id(Call)),
     Call;
 maybe_start_metaflow(Call, App) ->
     lager:debug("metaflow app ~s", [App]),
     Call.
 
+-spec maybe_start_ts_metaflow(kapps_call:call(), kz_term:api_binary()) -> 'ok'.
```
jamesaimonetti: Metaflows don't really seem appropriate to apply to a downstream PBX system. Wouldn't that system be the one offering (or not) the ability to take action on DTMFs? Trunkstore really only concerns itself with trunking a simple failover.

lazedo:
@sergey-safarov first pass, you are creating inter-app dependencies andalso core -> app dependencies. 
also, trunkstore soon to be rewritten will not have freeswitch in the path, adding konami capabilities to trunkstore is not a good idea.

applications/trunkstore/src/ts_util.erl
```diff
@@ -92,6 +93,38 @@ constrain_weight(W) when W > 100 -> 100;
...
+    case kz_cache:fetch_local(?CACHE_NAME
+                             ,{'lookup_trunk', TrunkId, AccountId}
+                             )
+    of
+        {'ok', _}=Resp ->
+            lager:info("cache hit for ~s", [TrunkId]),
+            Resp;
+        {'error', 'not_found'} ->
+            Options = [{'key', TrunkId}],
+            case kz_datamgr:get_results(AccountDb, ?TS_VIEW_CROSSBAR_LISTING, Options) of
```
jamesaimonetti: Why query the view? Why not just open the doc? The view's key is just `_id` anyway.

jamesaimonetti: To add, why is this function even needed? Nothing in trunkstore uses it, the doc is already cached via kz_datamgr.

core/kazoo_endpoint/src/kz_endpoint.erl
```diff
@@ -869,38 +867,11 @@ maybe_start_metaflows(Call, Endpoints, _CallId) ->
     of
         'false' -> 'ok';
         'undefined' ->
-            [maybe_start_metaflow(Call, Endpoint) || Endpoint <- Endpoints],
+            [konami_util:maybe_start_metaflow(Call, Endpoint) || Endpoint <- Endpoints],
```
jamesaimonetti: No core libraries can add a dependency on a Kazoo application, as that would require the app be installed alongside core.

sergey-safarov:
Hello @jamesaimonetti and @lazedo 
Use case - over trunk is connected callcenter PBX.
Agent must mask/unmask call recording using DTMF codes. I not want create account level metaflow and want use metaflow only call received by external callcenter agents.

applications/callflow/src/cf_route_win.erl
sergey-safarov: trunstore + metaflow may used if we want move functionality to kazoo from old PBX.
I used this functionality only for one customer and not sure that will be used for other.
I can simple close this PR as not wanted.

core/kazoo_endpoint/src/kz_endpoint.erl
sergey-safarov: Understand

sergey-safarov:
PR is closed as not wanted

PR #5129, opened 2018-09-17
KAZOO-5977: Added phonebook provider to porting apis

swysor:


applications/crossbar/src/modules/phonebook.erl
```diff
@@ -0,0 +1,195 @@
...
+handle_resp({'ok', 200, _, Resp}, _) ->
+    lager:debug("phonebook success ~s", [Resp]);
+handle_resp({'ok', Code, _, Resp}, AccountId) ->
+    lager:warning("phonebook error ~p. ~s", [Code, Resp]),
+    create_alert(kz_json:decode(Resp), AccountId);
+handle_resp(_Error, _) ->
+    lager:error("phonebook fatal error ~p", [_Error]).
+
+
+create_alert(JObj, AccountId) ->
```
jamesaimonetti: muh spec

applications/crossbar/src/modules/phonebook.erl
```diff
@@ -0,0 +1,195 @@
...
+    kapps_config:get_is_true(?MOD_CONFIG_CAT, <<"enabled">>, 'false').
+
+-spec req_from_phonebook(cb_context:context()) -> boolean().
+req_from_phonebook(Context) ->
+    case cb_context:req_header(Context, <<"User-Agent">>) of
+        <<"phonebook">> -> 'true';
+        _ -> 'false'
+    end.
+
+send_to_phonebook(Context) ->
```
jamesaimonetti: spec

applications/crossbar/src/modules/phonebook.erl
```diff
@@ -0,0 +1,195 @@
...
+req_uri('comment', AccountId, PortId) ->
+    phonebook_uri([<<"accounts">>, AccountId, <<"ports">>, <<"in">>, PortId, <<"notes">>]);
+req_uri('cancel_port_in', AccountId, PortId) ->
+    phonebook_uri([<<"accounts">>, AccountId, <<"ports">>, <<"in">>, PortId]).
+
+-spec phonebook_uri(iolist()) -> iolist().
+phonebook_uri(ExplodedPath) ->
+    Url = kapps_config:get_binary(?MOD_CONFIG_CAT, <<"phonebook_url">>),
+    Uri = kz_util:uri(Url, ExplodedPath),
+    binary:bin_to_list(Uri).
```
jamesaimonetti: `kz_term:to_list/1` instead, only for consistency

applications/crossbar/src/modules/phonebook.erl
```diff
@@ -0,0 +1,195 @@
...
+        _ -> 'false'
+    end.
+
+send_to_phonebook(Context) ->
+    cb_context:resp_status(Context) =:= 'success'
+        andalso phonebook_enabled().
+
+-spec create_port_in(kz_json:object(), kz_term:ne_binary()) -> 'ok'.
+create_port_in(JObj, AuthToken) ->
+    send_req('create_port_in'
```
jamesaimonetti: So you don't really use this atom for much in `send_req/5`. Meaning `send_req` isn't really a generic function since each clause just matches the atom in the first arg to do its logic.

Instead, I would use these named functions to create the specific pieces (URL, req body/headers) then call a generic make_request with the necessary bits. Does that make sense?

applications/crossbar/src/modules/phonebook.erl
```diff
@@ -0,0 +1,195 @@
...
+                            ,{<<"value">>, <<"admins">>}
+                            ])
+         ,kz_json:from_list([{<<"type">>, kz_services_reseller:get_id(AccountId)}
+                            ,{<<"value">>, <<"admins">>}
+                            ])
+         ],
+
+    Title = <<"Phonebook Error">>,
+    Msg = <<"Error trying to handle portin">>,
+    {'ok', AlertJObj} = kapps_alert:create(Title, Msg, From, To, Props),
```
icehess: check for possible `{error, Error}` if alerts is disable

applications/crossbar/src/modules/phonebook.erl
```diff
@@ -0,0 +1,195 @@
...
+    {ok, _} = kapps_alert:save(AlertJObj),
```
icehess: ignore the return

icehess: I don't who uses the alerts, but why not sending a `system_alert` notification instead?

PR #5130, opened 2018-09-17
KAZOO-5564: addressing some findings on sandbox

k-anderson:


applications/crossbar/src/modules/cb_ledgers.erl
```diff
@@ -152,6 +159,23 @@ validate(Context, ?CREDIT) ->
     cb_context:validate_request_data(<<"ledgers">>, Context);
 validate(Context, ?DEBIT) ->
     cb_context:validate_request_data(<<"ledgers">>, Context);
+validate(Context, ?TOTAL) ->
+    case kz_ledgers:total_sources(cb_context:account_id(Context)) of
+        {'error', _R} ->
+            lager:debug("failed to fetch ledgers total: ~p", [_R]),
+            cb_context:add_system_error('unspecified_fault', Context);
```
icehess: Why not passing `_R` here. it already have crossbar errors.

applications/crossbar/src/modules/cb_transactions.erl
```diff
@@ -99,7 +99,11 @@ resource_exists(_) -> 'true'.
 %%------------------------------------------------------------------------------
 -spec authorize(cb_context:context(), path_token()) -> boolean() | {'stop', cb_context:context()}.
 authorize(Context, Path) ->
-    authorize_request(Context, Path, cb_context:req_verb(Context)).
+    try authorize_request(Context, Path, cb_context:req_verb(Context))
```
icehess: Is `try catch` necessary?

core/kazoo_services/src/kz_services_plans.erl
```diff
@@ -306,7 +317,7 @@ merge_override(ServicesJObj, Overrides) ->
 %%------------------------------------------------------------------------------
 -type mergable() :: plans() | plans_list().
 -spec merge(mergable()) -> kz_services_plan:plan().
-merge([Head|Tail]=Plans) when is_list(Plans) ->
+merge([Head|Tail]=Plans) ->
```
icehess: what if the list is empty? Also I think its better to check if it is list by `is_list`.

applications/braintree/src/sdk/braintree_util.erl
```diff
@@ -135,15 +135,30 @@ error_to_props({'no_payment_token'=Reason, Details}) ->
...
-                                         ,Details
-                                         ,?DEFAULT_ERROR_MESSAGE
-                                         ),
-    [{<<"Message">>, Message}
-    ,{<<"Status">>, <<"error">>}
-    ,{<<"Reason">>, Key}
-    ,{<<"Details">>, Errors}
-    ];
+    case kz_json:get_ne_binary_value([Key, <<"message">>]
+                                    ,Details
```
jamesaimonetti: Is there a numeric code or something else to key off of here, instead of text?

core/kazoo_services/src/kz_services.erl
```diff
@@ -429,14 +475,18 @@ ratedeck_id(Services) ->
...
 plan_ratedeck_id(Services) ->
-    Plans = kz_services_plans:foldl(fun(_BookkeeperHash, PlansList, Plans) ->
-                                            PlansList ++ Plans
-                                    end
-                                   ,[]
-                                   ,plans(Services)
-                                   ),
-    Plan = kz_services_plans:merge(Plans),
-    kz_services_plan:ratedeck_id(Plan).
+    case kz_services_plans:foldl(fun(_BookkeeperHash, PlansList, Plans) ->
```
jamesaimonetti: We've been trying to move folds like this into their own nicely-named function to make it easier to know what's being done in the fold. For instance, maybe `case collect_plans(plans(Services) of`. Also, using a named fun for the fold is preferred if no state is being used from outside the fun()'s scope.

core/kazoo_services/src/kz_services.erl
```diff
@@ -455,14 +505,18 @@ ratedeck_name(Services) ->
...
 plan_ratedeck_name(Services) ->
-    Plans = kz_services_plans:foldl(fun(_BookkeeperHash, PlansList, Plans) ->
-                                            PlansList ++ Plans
-                                    end
-                                   ,[]
-                                   ,plans(Services)
-                                   ),
-    Plan = kz_services_plans:merge(Plans),
-    kz_services_plan:ratedeck_name(Plan).
+    case kz_services_plans:foldl(fun(_BookkeeperHash, PlansList, Plans) ->
```
jamesaimonetti: Same here

core/kazoo_services/src/kz_services_quantities.erl
```diff
@@ -510,7 +510,7 @@ calculate_faxbox_updates(JObj, Updates) ->
     case kz_doc:type(JObj) =:= <<"faxbox">> of
         'false' -> Updates;
         'true' ->
-            Key = [<<"faxes">>, <<"faxbox">>],
+            Key = [<<"faxes">>, <<"mailbox">>],
```
jamesaimonetti: If you want, these functions appear to be building a `kz_json:flat_proplist()`, essentially `[{path(), json_term()}]`. Could be nice to have a more specific return type used.

Also, unrelated, but I notice `calculate_qubicle_*_updates/2`; I wonder, could we leverage AMQP to do these calculations and let the apps that want to be quantifiable bind in and be called to calculate things? Then it becomes a `kz_amqp_worker:collect` with a binding key like `services.calculate.{PVT_TYPE}` and the payload is the JObj. Then each app that wants to be "billable" has a listener for the pvt_type's they are interested in, also allowing custom apps to add billing for their doc types.

core/kazoo_services/src/kz_services.erl
```diff
@@ -455,14 +505,18 @@ ratedeck_name(Services) ->
...
+                                 end
+                                ,[]
+                                ,plans(Services)
+                                )
+    of
+        [] -> 'undefined';
+        Plans ->
+            Plan = kz_services_plans:merge(Plans),
+            kz_services_plan:ratedeck_name(Plan)
+    end.
```
jamesaimonetti: Unrelated to here but Dialyzer complains:

```
  kz_services.erl:804: The call kz_services:commit_updates(Account::any(),Current::any(),Proposed::any(),[]) will never return since the success typing is ('undefined' | <<_:8,_:_*8>>,[{[any()]}],[{[any()]}],{[{atom() | <<_:8,_:_*8>>,_}]}) -> #kz_services{account_id::'undefined' | <<_:8,_:_*8>>,services_jobj::{[{_,_}]},current_services_jobj::{[{_,_}]},plans::'undefined' | kz_services_plans:plans(),account_updates::{[{_,_}]},cascade_updates::{[{_,_}]},manual_updates::{[{_,_}]},invoices::'undefined' | kz_services_invoices:invoices(),audit_log::{[{_,_}]},dirty::boolean()} and the contract is (kz_term:ne_binary(),kz_services_quantities:billables(),kz_services_quantities:billables(),kz_json:object()) -> services() 
```
I think you need to use `kz_json:new()` instead of `[]` for AuditLog; otherwise update the spec for `commit_updates/4`.

jamesaimonetti:
@k-anderson i rebased and added a couple commits handling dialyzer complaints and a little spelling fix ;)

core/kazoo_services/src/kz_services.erl
jamesaimonetti: This should be fixed in my commit

applications/crossbar/src/modules/cb_ledgers.erl
k-anderson: @icehess because we can't guaranty the type of _R and should be a rare occurrence

core/kazoo_services/src/kz_services_plans.erl
k-anderson: @icehess trying to enforce never merging no plans, would result in the requester doing work it doesn't need to.  Also, why the keep the guard?  Its guaranteed by the function clause...

applications/crossbar/src/modules/cb_ledgers.erl
k-anderson: @icehess Because the type of _R is not guaranteed and should be a rare occurrence

applications/crossbar/src/modules/cb_transactions.erl
k-anderson: @icehess safety 3rd ;)

applications/crossbar/src/modules/cb_ledgers.erl
```diff
@@ -302,14 +302,18 @@ maybe_convert_units(_, Value) -> Value.
 process_action(Context, ?CREDIT, Ledger) ->
     case kz_ledger:credit(Ledger) of
         {'error', Reason} ->
-            crossbar_util:response('error', Reason, Context);
+            cb_context:setters(Context, [{fun cb_context:set_resp_status/2, 'error'}
```
icehess: The standard way is to use `cb_context:add_system_error/2,3,4` or in the case of `kz_datamgr` use `crossbar_doc:handle_datamgr_errors/2,3`.

applications/crossbar/src/modules/cb_ledgers.erl
```diff
@@ -302,14 +302,18 @@ maybe_convert_units(_, Value) -> Value.
...
+                                        ,{fun cb_context:set_resp_data/2, Reason}
+                                        ]);
         {'ok', SavedLedger} ->
             maybe_impact_reseller(Context, SavedLedger)
     end;
 process_action(Context, ?DEBIT, Ledger) ->
     case kz_ledger:debit(Ledger) of
         {'error', Reason} ->
-            crossbar_util:response('error', Reason, Context);
+            cb_context:setters(Context, [{fun cb_context:set_resp_status/2, 'error'}
```
icehess: same, use `crossbar_doc:handle_datamgr_errors/2,3`.

applications/crossbar/src/modules/cb_ledgers.erl
```diff
@@ -323,15 +327,19 @@ maybe_impact_reseller(Context, Ledger) ->
 
 -spec maybe_impact_reseller(cb_context:context(), kz_ledger:ledger(), boolean(), kz_term:api_binary()) -> cb_context:context().
 maybe_impact_reseller(Context, Ledger, 'false', _ResellerId) ->
-    crossbar_util:response(kz_ledger:public_json(Ledger), Context);
+    cb_context:setters(Context, [{fun cb_context:set_resp_status/2, 'success'}
```
icehess: Please use `crossbar_doc:handle_json_success/2`.

PR #5134, opened 2018-09-19
PISTON-615: kz_time:decr_timeout doesn't handle ms, so don't use ms

danielfinke:


danielfinke:
Looks like the failed checks here are not related to the contents of this PR

jamesaimonetti:
I agree, merged.

PR #5138, opened 2018-09-20
PISTON-613: do not reduce statuses since recent is not used

danielfinke:
Most-Recent is never used at this time.

Removing reduce for all agents status request so that the behaviour matches that of the agent/<agent_id> payload. Will add an implementation of Most-Recent in a subsequent update.


jamesaimonetti:
Merged, thanks!

PR #5140, opened 2018-09-21
HELP-39529: ensure comparisons of FROM to permissions list is ensuring lower case 

swysor:
…sions list and from email

from email is always lowercase on receipt of a message, but the permissions list could have upper case characters making the comparison fail. 
https://github.com/2600hz/kazoo/blob/master/applications/fax/src/fax_smtp.erl#L111

### DIFF
--- a/applications/fax/src/fax_smtp.erl
+++ b/applications/fax/src/fax_smtp.erl
@@ -516,7 +516,8 @@ check_empty_permissions(#state{errors=Errors
 
 -spec match(binary(), binary()) -> boolean().
 match(Address, Element) ->
-    re:run(Address, Element) =/= 'nomatch'.
+    Address =:= kz_term:to_lower_binary(Element)
+        orelse re:run(Address, Element) =/= 'nomatch'.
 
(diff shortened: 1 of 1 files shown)


applications/fax/src/fax_smtp.erl
```diff
@@ -516,7 +516,8 @@ check_empty_permissions(#state{errors=Errors
 
 -spec match(binary(), binary()) -> boolean().
 match(Address, Element) ->
-    re:run(Address, Element) =/= 'nomatch'.
+    Address =:= kz_term:to_lower_binary(Element)
```
k-anderson: @swysor any thought to normalize Address too?

kz_term:to_lower_binary(Address) =:= kz_term:to_lower_binary(Element)

swysor: its normalized on ingress

swysor: https://github.com/2600hz/kazoo/blob/a77772f3513097e649fc3953c983915fe03efe71/applications/fax/src/fax_smtp.erl#L111

PR #5142, opened 2018-09-21
PISTON-618: move timeouts/gproc sends to info - resolve infinite loops

danielfinke:
Conflicts:
	applications/acdc/src/acdc_agent_fsm.erl


danielfinke:
Are the tests that failed for MakeBusy fixed in master? I don't believe they are caused by this PR.

PR #5149, opened 2018-09-24
QF: only send one request on create, not on save, strip pvt_ stuff

swysor:


jamesaimonetti:
@swysor I added some fixes from Dialyzer; let me know if you have any questions on the changes made. I'll merge your commit separate from mine if you agree on the changes.

PR #5153, opened 2018-09-26
spell check templates

jamesaimonetti:


icehess:
Can you please text version of the templates too, for example `port_scheduled.text` has the same issue :dagger:

PR #5157, opened 2018-09-27
KAZOO-5952: do not default settings in conference callflow

swysor:
Since callflow always wins over conference config when merging settings, remove the default values which would supersede the values in the conference so new conference callflows that are created will use the conference settings, not the callflow settings.


swysor:
Since callflow always wins over conference config when merging settings, remove the default values which would supersede the values in the conference so new conference callflows that are created will use the conference settings, not the callflow settings.

PR #5161, opened 2018-09-28
Refresh views

icehess:


applications/conference/src/conference_maintenance.erl
```diff
@@ -1,41 +0,0 @@
-%%%-----------------------------------------------------------------------------
-%%% @copyright (C) 2011-2018, 2600Hz
-%%% @doc
-%%% @author Karl Anderson <karl@2600hz.org>
-%%% @end
-%%%-----------------------------------------------------------------------------
--module(conference_maintenance).
```
jamesaimonetti: I would not delete the files even though they're empty. We'll need to add stuff here to interact with the conference app. In fact, there's a cache used by the app that should have a flush/0 added here! So let's keep the maintenance modules even if emptied by these changes.

core/kazoo_apps/src/kapps_maintenance.erl
```diff
@@ -256,7 +307,29 @@ get_database_sort(Db1, Db2) ->
...
+    _ = kz_datamgr:refresh_views(Database),
+    _ = kazoo_bindings:map(binding({'refresh', Database}), [Database]),
+    'ok'.
+
+-spec refresh_account_db(kz_term:ne_binary()) -> 'ok'.
+refresh_account_db(Database) ->
+    AccountDb = kz_util:format_account_id(Database, 'encoded'),
+    AccountId = kz_util:format_account_id(Database, 'raw'),
+    _ = kz_datamgr:refresh_views(AccountDb),
+    kapps_account_config:migrate(AccountDb),
```
lazedo: why calling `migrate` here ?

core/kazoo_apps/src/kapps_maintenance.erl
```diff
@@ -256,7 +307,29 @@ get_database_sort(Db1, Db2) ->
...
+refresh_views(Database, 'modb') ->
+    kazoo_modb:refresh_views(Database),
+    _ = kazoo_bindings:map(binding({'refresh', Database}), [Database]),
+    'ok';
+refresh_views(Database, 'account') ->
+    refresh_account_db(Database);
+refresh_views(Database, _) ->
+    _ = kz_datamgr:refresh_views(Database),
+    _ = kazoo_bindings:map(binding({'refresh', Database}), [Database]),
+    'ok'.
```
lazedo: why not handle this in a simpler way ? call `kz_datamgr:refresh_views`  and then `kazoo_bindings:map(binding({'refresh', Database}), [Database]),` ?

applications/crossbar/src/modules/cb_accounts.erl
```diff
@@ -1381,10 +1371,9 @@ set_trial_expires(JObj) ->
     Expires = kz_time:now_s() + TrialTime,
     kzd_accounts:set_trial_expiration(JObj, Expires).
 
-
 -spec load_initial_views(cb_context:context()) -> 'ok'.
 load_initial_views(Context)->
-    _ = kz_datamgr:refresh_views(cb_context:account_db(Context)),
+    _ = kapps_maintenance:refresh(cb_context:account_db(Context)),
```
lazedo: the intent here is to refresh_views, so a call to kz_datamgr:refresh_views seems appropriate since `kapps_maintenance:refresh` may do other things other than refreshing_views

applications/crossbar/src/modules/cb_accounts.erl
```diff
@@ -1381,10 +1371,9 @@ set_trial_expires(JObj) ->
...
     _ = kazoo_number_manager_maintenance:update_number_services_view(cb_context:account_db(Context)),
```
lazedo: `kapps_maintenance:refresh` calls bindings and this is called indirectly

core/kazoo_apps/src/kapps_maintenance.erl
icehess: I moved this from `kapps_maint_listener` here. Should I moved to migrate function?

lazedo: yes, move to migrate

core/kazoo_apps/src/kapps_maintenance.erl
icehess: I don't understand, this is doing what you told.

lazedo: you handle it differently for `modb` and  `account`

lazedo: to be clear, it is ok to handle `account` and `modb` differently in `refresh` but `refresh_views` should be simple as the last `function` clause (i was trying to show that but the lines showed only these)

core/kazoo_number_manager/src/kazoo_number_manager_maintenance.erl
```diff
@@ -163,6 +166,28 @@ print_convert_carrier_failure(Target, TotalLength, KOs) ->
 convert_carrier_module_number(Num, Target) ->
     convert_carrier_module_numbers([Num], Target).
 
+-spec init_dbs() -> 'ok'.
+init_dbs() ->
+    init_dbs([?KZ_MEDIA_DB
```
lazedo: why `?KZ_MEDIA_DB` ?

core/kazoo_media/src/kazoo_media_maintenance.erl
```diff
@@ -239,17 +240,19 @@ maybe_retry_upload(ID, AttachmentName, Contents, Options, Retries) ->
...
+            Result = kz_datamgr:db_create(?KZ_MEDIA_DB),
+            lager:debug("~s database is created: ~p", [?KZ_MEDIA_DB, Result]);
+        'true' -> 'ok'
+    end,
+    _ = kapps_maintenance:refresh(?KZ_MEDIA_DB),
     'ok'.
 
+-spec register_views() -> 'ok'.
+register_views() ->
+    kz_datamgr:register_views_from_folder('kazoo_media').
```
lazedo: this is not being called at startup

icehess: `kazoo_media_app` is call `kapps_maintenance:bind_and_register_views` which doing the bind and calling register function.

core/kazoo_apps/src/kapps_maintenance.erl
icehess: because account have to maps refresh_account binding, previously there was no map for modb binding

core/kazoo_media/src/kazoo_media_maintenance.erl
lazedo: `true` and this is one of the reasons that i would like to see the `register_views` binding in a separate pull request. this way, you're turning a `utility` into a `must use` which i don't like.
on `app` startup
1) create any DB specific to the `app`
2) register views specific to the `app`
3) do other bindings

```
kapps_maintenance:bind_and_register_views('kazoo_media', 'kazoo_media_maintenance', 'register_views'),
kapps_maintenance:bind('migrate', 'kazoo_media_maintenance', 'migrate'),
kazoo_media_maintenance:refresh()
```
should be

```
kazoo_media_maintenance:init_dbs()
kapps_maintenance:bind('register_views', 'kazoo_media_maintenance', 'register_views'),
kapps_maintenance:bind('migrate', 'kazoo_media_maintenance', 'migrate'),

```
in kazoo_media_maintenance
```
init_dbs() ->
    _ = kz_datamgr:db_create(?KZ_MEDIA),
    _ = register_views().
```

this should be for all apps, i understand that `kapps_maintenance:bind_and_register_views` is like `killing 2 birds with one stone` but  will keep things simpler

core/kazoo_apps/src/kapps_maintenance.erl
lazedo: but you are calling the binding for `refresh_account` from refresh_views and it should be called from `refresh`.
also, in refresh_views we're calling the binding for `refresh` and it should be `refresh_views`, and `refresh` binding should be called from `refresh`

PR #5165, opened 2018-09-29
move interactions to own view

lazedo:


icehess:
`applications/cdr/src/cdr_listener.erl` needs to update to use `interactions`

PR #5174, opened 2018-10-01
KAZOO-5984: crossbar view should stop when map function returns error

icehess:


applications/crossbar/src/crossbar_view.erl
```diff
@@ -864,26 +867,53 @@ limit_with_last_key('true', PageSize, _ChunkSize, TotalQueried) ->
...
-%% Filter function can be arity 1 (operating on list of JObjs) and
-%% arity 2 (classical Erlang foldl function, e.g. (JObj, Acc)).
+%% @doc Apply filter/mapper function if provided while keep maintaining
+%% the result order.
+%%
+%% Mapper function can be arity 1 (operating on a list of JObjs) and
+%% arity 2 of `(JObj, Acc)'.
+%%
+%% The result set is already reversed by {@link last_key/4} for finding
+%% the last key when so data that is passed to mapper/filter function is already
```
jamesaimonetti: This reads awkwardly. Is the comment to alert developers that the result set is in reverse order when passed tot he mapper/filter function?

"Take note: Because of the call to `last_key/4` the  result set is in reverse order when passed to the mapper/filter function"

Something like that maybe?

PR #5177, opened 2018-10-02
dollars problem

icehess:
I don't know, do not use Outlook, Gmail I think? Use Mail.app? Everything seems fine, ErlyDTL rendered it okay, mimemail encode it correctly, gen_smtp_client send it correctly, Gmail or outlook or whatever received it correctly, but they can't doing their job of showing it properly.

Mail.app (on iOS and Mac) just render it just as what it should be.

### DIFF
--- a/applications/teletype/priv/templates/service_added.html
+++ b/applications/teletype/priv/templates/service_added.html
@@ -63,11 +63,11 @@
                                     <table role="presentation" border="0" cellpadding="0" cellspacing="0" align="center" width="100%">
                                         <tr>
                                             <td valign="middle" colspan="2" style="padding:10px 0;font-family:'Open Sans',sans-serif;color:#555555;">
-                                                <b>{{srv_item}}</b><br><span style="color:#999999;font-size:13px"><small>Quantity: {{item_props.quantity}}<br>Rate: &dollar;{{item_props.rate}}{% if item_props.discounts != "0.00" %}<br>Discounts: &dollar;{{item_props.discounts}}{% endif %}{% if item_props.changes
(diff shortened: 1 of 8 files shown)


applications/teletype/priv/templates/service_added.html
```diff
@@ -82,7 +82,7 @@ <h2 style="margin:0;padding:0;text-align:center;font-family:'Open Sans',sans-ser
                         <table role="presentation" aria-hidden="true" border="0" cellpadding="0" cellspacing="0" align="center" width="100%" style="max-width:560px;border:1px solid #dedede;border-top-style:none;">
                             <tr>
                                 <td align="right" bgcolor="#ffffff" colspan="2" style="padding:10px 20px;font-family:'Open Sans',sans-serif;color:#999999;border-right:1px solid #dedede;"><small>TOTAL</small></td>
-                                <td align="right" bgcolor="#ffffff" colspan="1" style="padding:10px 20px;font-family:'Open Sans',sans-serif;color:#555555">&dollar;{{invoice.total}}</td>
+                                <td align="right" bgcolor="#ffffff" colspan="1" style="padding:10px 20px;font-family:'Open Sans',sans-serif;color:#555555">{{invoice.total}} USD</td>
```
jamesaimonetti: I wonder if we should include a "currency" macro so people can override their default currency from "USD" to "EUR" or "YEN" or whatever.

icehess: If kazoo services provides the currency in the payload yes, just like transaction (braintree).

PR #5178, opened 2018-10-02
Revert endpoint ID

vccs:
Reverting incorrectly modified endpoint ID variable. Please push to the rpm as soon as possible, much appreciated, thanks.


vccs:
Thank's!

PR #5179, opened 2018-10-02
fix hook name in find_webhooks

lazedo:


core/kazoo_documents/src/kzd_accounts.erl
```diff
@@ -723,7 +723,10 @@ fetch(AccountId, 'accounts') ->
                             {'ok', doc()} |
                             kz_datamgr:data_error().
 open_cache_doc(Db, AccountId) ->
-    kz_datamgr:open_cache_doc(Db, AccountId, [{'cache_failures','false'}]).
+    Options = [{'cache_failures','false'}
+              ,{'deleted','true'}
```
jamesaimonetti: I'm worried that in a future time, an account will be fetched that is soft-deleted and will be treated as existing. I wonder if we introduce `kzd_accounts:exhume_if_necessary(AccountId)`? Make it explicit that the caller knows they could potentially get a soft/hard deleted account doc.

lazedo: @jamesaimonetti i agree to add more `options` to  `fetch` (no_cache, deleted, ..). but  `now` we havge side effects because of this pr #5087

PR #5195, opened 2018-10-05
HELP-39798: find conference username from sip_req_uri

jamesaimonetti:
When bridging a remote conference participant, the 'conference' username is not always being found via the `Hunt-Destination-Number` FreeSWITCH variable:

```
|{CALL_ID}|ecallmgr_fs_authz:91(<0.29383.2451>) is destination number 'conference': undefined
```

If missing, check the `variable_sip_req_uri` which has the `conference@FS:11000` (when it is a bridged conference participant).


jamesaimonetti:
And just to clarify, `Hunt-Destination-Number` is in the `mod_kazoo` event filter list.

PR #5202, opened 2018-10-08
use username instead of presence_id for Referred-By - 4.3

lazedo:


applications/ecallmgr/src/ecallmgr_fs_transfer.erl
```diff
@@ -99,14 +99,18 @@ transfer_set_callid(UUID, _) -> UUID.
...
+                       ,realm=Realm
+                       }} -> <<"<sip:", Username/binary, "@", Realm/binary, ">">>;
         _Else -> 'undefined'
     end;
 transfer_referred(UUID, _) ->
     case ecallmgr_fs_channel:fetch_other_leg(UUID, 'record') of
-        {'ok', #channel{presence_id='undefined'}} -> 'undefined';
-        {'ok', #channel{presence_id=PresenceId}} -> <<"<sip:", PresenceId/binary, ">">>;
+        {'ok', #channel{username='undefined'}} -> 'undefined';
+        {'ok', #channel{username=Username
```
jamesaimonetti: It may be worth adding a clause for `#channel{realm='undefined'}` to avoid crashing (just in case the Realm hasn't been populated yet.

PR #5203, opened 2018-10-08
[4.3] Better iso8601

icehess:
master: https://github.com/2600hz/kazoo/pull/5197


lazedo:
@icehess can you create your branch from 4.3 instead of master ?

PR #5207, opened 2018-10-10
PISTON-624: handle timeout arriving in ready due to member hangup while waiting for connect_req_req

danielfinke:


jamesaimonetti:
@danielfinke merged

PR #5223, opened 2018-10-16
PISTON-629: update old/last actions before cf module spawn

danielfinke:
Fixes cf_exe logging the wrong module in `cf module <mod> down normally` when cf modules use cf_exe:set_call/1. Those modules receive a version of the kapps_call object before its cf_old_action and cf_last_action kvs are updated and end up reverting it when calling cf_exe:set_call/1.

### DIFF
--- a/applications/callflow/src/cf_exe.erl
+++ b/applications/callflow/src/cf_exe.erl
@@ -695,55 +695,56 @@ do_launch_cf_module(#state{call=Call
     Module = <<"cf_", (kz_json:get_ne_binary_value(<<"module">>, Flow))/binary>>,
     Data = kz_json:get_json_value(<<"data">>, Flow, kz_json:new()),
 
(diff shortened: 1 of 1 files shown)


danielfinke:
@jamesaimonetti an example is a callflow that goes:
Language > Prepend > Play Media
You should see "cf module undefined down normally" after the first two since they use cf_exe:set_call/1.
If you do something like:
Menu > 1 > Language > Prepend > Play Media
You'll see "cf module cf_menu down normally" after the Language/Prepend instead of their respective modules

jamesaimonetti:
@danielfinke thanks! Probably some other places we don't properly track updates to the call record...will see if I can write something to track this better, and will test your PR out later today and merge.

jamesaimonetti:
@danielfinke sorry for the delay, merged

PR #5182, opened 2018-10-02
Dialyzing hard of callflows

jamesaimonetti:

PR #5230, opened 2018-10-17
Dialyzing hard of callflows (#5182) - 4.3

lazedo:
* update kazoo json type usages

* don't bind unless necessary

* will filter undefined while setting

* fix some more json specs/types, use kz_datamgr:update_doc/3

* fix type for config_key()

* more type updates

* a few more dialyzer-related updates

* refactor out the try/catch

* more refactoring to appease the dialyzer

* cleanup docs for dynamic CID

* more dialyzer updates

* clean up spec for kz_json:from_list

* update specs

* revert to from_list/1

Conflicts:
	core/kazoo_amqp/src/kz_amqp_worker.erl


jamesaimonetti:
LGTM

PR #5231, opened 2018-10-17
Master pr limit extension length

aisconn:
Minimum extension length

User wants to place a minimum required length on callflow number, vmbox number, user username and device SIP username to avoid 1-digit or 2-digit extensions causing issues or errors.

Default minimum length for all except callflows is 3, although it should be configurable. Callflows default to minimum length of 1 for use cases such as dialling 0 rings reception. Configured minimum length only applies to new or updated values, not unchanged.


applications/crossbar/src/modules/cb_callflows.erl
```diff
@@ -187,9 +187,12 @@ request_patterns(Context) ->
 validate_request(CallflowId, Context) ->
     case request_numbers(Context) of
         [] -> validate_patterns(CallflowId, Context);
-        OriginalNumbers
-          when is_list(OriginalNumbers) ->
-            validate_callflow_schema(CallflowId, normalize_numbers(Context, OriginalNumbers));
+        OriginalNumbers when is_list(OriginalNumbers) ->
+            Validators = [{fun validate_min_length_numbers/3, [OriginalNumbers]}
```
jamesaimonetti: At the moment, I feel like you could just edit the schema with the system_config-defined length and then validate as before. Then you don't need this custom validator code to check length, right?

aisconn: For this change, the default minimum length is defined in system_config, but can be overridden at an account level. Relying on the system_config schema alone for validation will not allow the minimum length to be overridden for a particular account, which is what the kapps_account_config:get_global() call within the new validator code gives us. 

Also we don't want to enforce minimum length for existing callflow numbers. So the unchanged check within the validator code is needed to ensure the user can update and save a callflow without being blocked by validation of an existing number that does not meet minimum length. 

Does that make sense? Is there an easier way of doing this, using validate_callflow_schema, that I've missed?

lazedo:
@aisconn i really don't like to add these extra validators this way. the solution must be more generic and not enforced in each place we want to provide a system configured limit and not a schema.

applications/crossbar/src/modules/cb_callflows.erl
jamesaimonetti: My initial preference is to put this behind a config toggle; if an account opts in then all callflow changes will be subject to it.

My pseudocode would then be:

```erlang
    OriginalNumbers ->
        validate_numbers(CallflowId, Context, OriginalNumbers, kapps_config:is_true(?APP_NAME, <<"enforce_minimum_length">>, 'false'))
end.

validate_numbers(CallflowId, Context, OriginalNumbers, 'false') ->
    validate_callflow_schema/2...;
validate_numbers(CallflowId, Context, OriginalNumbers, 'true') ->
    MinLength = kapps_account_config:get(),
    Schema = kz_json_schema:get(<<"callflows">>),
    Updated = add_min_length_check(Schema, MinLength),
    validate_callflow_schema/2...
```

So if the sysadmin opts into min-length checks, adjust the schema to use the account's configured min length and apply them to any callflow changes (new or edits).

Thoughts?

applications/crossbar/src/modules_v1/cb_users_v1.erl
```diff
@@ -509,7 +509,10 @@ load_user(UserId, Context) -> crossbar_doc:load(UserId, Context, ?TYPE_CHECK_OPT
 %%------------------------------------------------------------------------------
 -spec validate_request(kz_term:api_binary(), cb_context:context()) -> cb_context:context().
 validate_request(UserId, Context) ->
-    prepare_username(UserId, Context).
+    Validators = [{fun is_username_min_length/2, []}
```
jamesaimonetti: `_v1` modules are considered deprecated. These edits should be in `_v2` by default, `_v1` if you really want them there.

applications/crossbar/src/modules/cb_callflows.erl
aisconn: Thanks for the suggestion, I can see your point. I'm going to try this approach and will get back to you with an updated solution

applications/crossbar/src/modules_v1/cb_users_v1.erl
aisconn: I'll address this with updated solution too

applications/crossbar/src/modules/cb_callflows.erl
aisconn: Just to clarify, should the config toggle be at a system level or at an account level? I realise the example is just pseudocode but kapp_configs module accesses system_config, while it sounds like the decision to enforce minimum length should be at an account config level. Can you confirm your preference?

jamesaimonetti: @aisconn System-level config for whether to enforce the minimum or not; account-level config for the minimum itself.

aisconn:
I understand this is now part of a bigger discussion, but updating the PR with the changes based on your feedback anyway. Thanks for all advice.

lazedo:
@aisconn closing this. we will work around this in 5.x series. the idea behind is to have defaults applied in the schema validator not on modules. thank you for your effort on this

PR #5232, opened 2018-10-18
Updates to Pivot and ClickToCall

jamesaimonetti:


noahmehl:
This is really important to us :) As it fixes a UTF-8 encoding problem with Pivot <SAY/> verbs :)

PR #5237, opened 2018-10-19
Search directory by first name, last name, or both.

aisconn:
Search a user directory by first name, last name or both names. Searching by first name will not try to match to a last name, if first name not found. Searching by last name will not try to match to a first name, if last name not found. Only when 'both' is specified will both first and last names be searched.

sort_by is then used to sort the list of users by first or last name.

### DIFF
--- a/applications/callflow/src/module/cf_directory.erl
+++ b/applications/callflow/src/module/cf_directory.erl
@@ -79,6 +79,7 @@
 
 -define(PROMPT_ENTER_PERSON_LASTNAME, <<"dir-enter_person_lastname">>). %% Please enter the first few letters of the person's lastname
 -define(PROMPT_ENTER_PERSON_FIRSTNAME, <<"dir-enter_person_firstname">>). %% Please enter the first few letters of the person's firstname
+-define(PROMPT_ENTER_PERSON_NAME, <<"dir-enter_person_name">>). %% Please enter the first few letters of the person's name
 -define(PROMPT_FIRSTNAME, <<"dir-first_name">>). %% first name
 -define(PROMPT_LASTNAME, <<"dir-last_name">>). %% last name
 -define(PROMPT_SPECIFY_MINIMUM, <<"dir-specify_minimum">>). %% You need to specify a minimum of two digits
@@ -106,6 +107,7 @@
 -type directory_users() :: [directory_user()].
 
 -record(directory, {sort_by = 'last' :: 'first' | 'last'
+                   ,search_fields = 'both' :: 'first' | 'last' | 'both'
                    ,min_dtmf :: pos_integer()
                    ,max_dtmf :: non_neg_integer()
                    ,confirm_match = 'false' :: boolean()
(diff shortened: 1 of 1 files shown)


applications/callflow/src/module/cf_directory.erl
```diff
@@ -434,36 +445,61 @@ get_directory_user(U, CallflowId) ->
...
-    queue:to_list(
-      lists:foldl(fun(U, Q) -> maybe_queue_user(U, Q, DTMFs, Size, 'last') end
-                 ,queue:new()
-                 ,Users
-                 )
-     );
-filter_users(Users, DTMFs, 'first') ->
+-spec sort_users(directory_users(), 'first' | 'last') -> directory_users().
+sort_users(Users, Order) ->
+    lists:sort(fun(User1, User2) ->
```
jamesaimonetti: We typically like to break things out into named functions. At first glance I would try something like:

```erlang
sort_users(Users, 'first') ->
    lists:sort(fun sort_by_first/2, Users);
sort_users(Users, 'last') ->
    lists:sort(fun sort_by_last/2, Users).
```

Now a reader can see that the sort function used is toggled by the `Order`.

Now the sort functions themselves can pick apart the `directory_user` record as they see fit:

```erlang
sort_by_first(#directory_user{first_name=AFirst, last_name=ALast}, #directory_user{first_name=AFirst, last_name=BLast}) ->
    ALast < BLast;
sort_by_first(#directory_user{first_name=AFirst}, #directory_user{first_name=BFirst}) ->
    AFirst < BFirst.
```

Similar structure for `sort_by_last/2`. 

What do you think?

applications/callflow/src/module/cf_directory.erl
```diff
@@ -79,6 +79,7 @@
 
 -define(PROMPT_ENTER_PERSON_LASTNAME, <<"dir-enter_person_lastname">>). %% Please enter the first few letters of the person's lastname
 -define(PROMPT_ENTER_PERSON_FIRSTNAME, <<"dir-enter_person_firstname">>). %% Please enter the first few letters of the person's firstname
+-define(PROMPT_ENTER_PERSON_NAME, <<"dir-enter_person_name">>). %% Please enter the first few letters of the person's name
```
jamesaimonetti: Everything looks good except for this prompt? Maybe you meant this one? https://github.com/2600hz/kazoo-sounds/blob/master/kazoo-core/prompts.txt#L46

aisconn: Yes, you're right. I was still using our own repo for kazoo-sounds when testing. I'll fix that

aisconn: Hi James. I'm adding a prompt for dir-enter_person_name in kazoo sounds to support this change (PR coming). We only have this file in English. Is there a need to support the prompt in other languages?

aisconn: @jamesaimonetti see https://github.com/2600hz/kazoo-sounds/pull/7

jamesaimonetti:
@aisconn thanks for your patience and collaboration! Merged

PR #5243, opened 2018-10-22
HELP-39208: when transfering via API/DTMF, use authz context

jamesaimonetti:


jamesaimonetti:
See 4.3 at #5244

core/kazoo_endpoint/src/kz_endpoint.erl
```diff
@@ -1718,8 +1719,9 @@ encryption_method_map(CCVs, Endpoint) ->
 
 -spec set_sip_invite_domain(ccv_acc()) -> ccv_acc().
 set_sip_invite_domain({Endpoint, Call, CallFwd, CCVs}) ->
+    SipRealm = get_sip_realm(Endpoint, Call, kapps_call:request_realm(Call)),
```
lazedo:
```suggestion
    SipRealm = get_sip_realm(Endpoint, kapps_call:account_id(Call), kapps_call:request_realm(Call)),
```

PR #5245, opened 2018-10-23
[Master] Phonebook: allow patching of port documents and move push to phonebook to staging state change

swysor:


applications/crossbar/doc/ref/port_requests.md
```diff
@@ -75,6 +75,16 @@ curl -v -X POST \
     http://{SERVER}:8000/v2/accounts/{ACCOUNT_ID}/port_requests/{PORT_REQUEST_ID}
 ```
 
+## Patch
```
jamesaimonetti: This does not appear to be in ../port_requests.md with examples.

swysor: fixed

PR #5253, opened 2018-10-25
Update knm_vitelity_util.erl

IgorrG:
Due to Vitelity API gone accessible only via HTTPS default URL should be changed in sources.

Vitelity message: "Please be advised, there are upcoming changes to the Vitelity API. HTTP requests have been shut off as of Monday Oct 15 at 10 AM MDT and all API requests are being redirected to HTTPS. As of Wednesday, Nov 14 at 10 AM MST, there will no longer be support for the redirects and all API requests must be done using HTTPS."


jamesaimonetti:
Hey @IgorrG thanks! I added a migrate step to move people over automatically if they have the default http URL still configured. Merged!

PR #5255, opened 2018-10-25
[4.3] when using update_doc/3, don't include {_rev, null}

jamesaimonetti:
Previous versions used kz_doc:delete_revision/1 before calling
kz_datamgr:ensure_saved/2 to clear the _rev from the doc. In
deprecating kz_datgmgr:ensure_saved/2 the initial refactor chose to
refactor this to using {_rev, null} in the {update, [...]} list passed
to kz_datamgr:update_doc/3.

However, as these calls were being made to replicate a doc from an
account database to an aggregate database (like sip_auth or accounts),
the effect was that the save would fail with conflict (because the
original doc's _rev didn't match the aggregate doc's _rev), the
aggregate doc would be fetched and updates applied (nulling _rev
again). Saving would then conflict again as no _rev was supplied but
the doc exists...and you got yourself a loop.

### DIFF
--- a/applications/crossbar/src/modules/cb_faxboxes.erl
+++ b/applications/crossbar/src/modules/cb_faxboxes.erl
@@ -541,14 +541,16 @@ oauth_req(Doc, OAuthRefresh, Context) ->
 faxbox_doc_save(Context) ->
     Ctx2 = crossbar_doc:save(Context),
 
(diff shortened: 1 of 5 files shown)


jamesaimonetti:
See master in #5254

PR #5266, opened 2018-11-02
Move account creation/deletion to core module

jamesaimonetti:
Allow other apps to create/delete accounts as necessary.


jamesaimonetti:
@swysor I had to move a lot of provisioner stuff around too; do you have a dev setup for the provisioner backends actually in play that you can test against?

PR #5267, opened 2018-11-04
Voicemail callback

sergey-safarov:
This PR allow make callback to `caller_id_number` of voicemail message.
By default is used key `8` for callback.
PR is created to callback from offnet also as voicemail box owner. If `authorizing_id` or `owner_id` is `undefined`, then callback is initiated with account permissions.
Designed to use/apply dialplans (device/user/account) - please check implementation.
Now no new announcements is create/used. Please look currently defined, may be you can suggest other announcements that will be more appropriate.

### DIFF
--- a/applications/callflow/src/module/cf_voicemail.erl
+++ b/applications/callflow/src/module/cf_voicemail.erl
@@ -924,7 +924,7 @@ play_messages([H|T]=Messages, PrevMessages, Count, #mailbox{seek_duration=SeekDu
                 Number ->
                     lager:info("caller chose to callback number ~s", [Number]),
                     case maybe_branch_call(Call, Number, Box) of
-                        'ok' -> 'complete';
+                        'ok' -> 'ok';
                         _ -> play_messages(Messages, PrevMessages, Count, Box, Call)
                     end
             end;
--- a/applications/callflow/src/module/cf_voicemail.erl
+++ b/applications/callflow/src/module/cf_voicemail.erl
@@ -149,6 +149,7 @@
               ,delete = <<"7">> :: kz_term:ne_binary()
               ,rewind = <<"5">> :: kz_term:ne_binary()
               ,fastforward = <<"8">> :: kz_term:ne_binary()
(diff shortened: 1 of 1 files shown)


applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -104,37 +104,38 @@
...
 
-                                        %% Post playback
+                                    %% Post playback
               ,keep = <<"1">> :: kz_term:ne_binary()
               ,replay = <<"2">> :: kz_term:ne_binary()
               ,forward = <<"3">> :: kz_term:ne_binary()
               ,prev = <<"4">> :: kz_term:ne_binary()
               ,next = <<"6">> :: kz_term:ne_binary()
               ,delete = <<"7">> :: kz_term:ne_binary()
+              ,callback = <<"8">> :: kz_term:ne_binary()
```
sergey-safarov: As options we can use `5` key for callback

jamesaimonetti:
I debate whether this feature should be behind a feature flag.

We're also going to have to figure out how to balance different features wanting to use the same menu DTMF (see newer PR for voicemail fastforward/rewind).

In any case, @sergey-safarov can you rebase and we can have that bigger discussion?

sergey-safarov:
PR is rebased.
About DTMF keys usage
How about?
```erl
              ,keep = <<"1">> :: kz_term:ne_binary()
              ,replay = <<"2">> :: kz_term:ne_binary()
              ,forward = <<"3">> :: kz_term:ne_binary()
              ,prev = <<"4">> :: kz_term:ne_binary()
              ,callback = <<"5">> :: kz_term:ne_binary()
              ,next = <<"6">> :: kz_term:ne_binary()
              ,rewind = <<"7">> :: kz_term:ne_binary()
              ,delete = <<"8">> :: kz_term:ne_binary()
              ,fastforward = <<"9">> :: kz_term:ne_binary()
```
As options for delete we can also use `0, *, #`

PR #5272, opened 2018-11-06
[4.3] Phonebook: allow patching of port documents and move push to phonebook to staging state change 

swysor:
QF: moved the port_request push to phonebook to inside state change to pending

make fmt

fixed docs

make apis

added example to docs

fix return value

bad args for cb_context:load()

there is already a loader funciton here

there is already a loader funciton here

fixed global listing by number

removed restriction on superduper admin user to support api_keys

### DIFF
--- a/applications/crossbar/src/modules/cb_port_requests.erl
+++ b/applications/crossbar/src/modules/cb_port_requests.erl
@@ -18,13 +18,12 @@
         ,validate/1, validate/2, validate/3, validate/4
         ,get/3
         ,put/1, put/3
-        ,patch/3
+        ,patch/3, patch/2
         ,post/2, post/4
         ,delete/2, delete/4
         ,authority/1
-
         ,acceptable_content_types/0
-
+        ,validate_request/2
         ]).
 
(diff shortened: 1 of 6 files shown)


applications/crossbar/src/modules/cb_port_requests.erl
```diff
@@ -770,26 +788,38 @@ get_account_names(Keys) ->
...
+                                                       || AnAccountId <- kapps_util:account_descendants(AccountId)
+                                                      ]));
+        _ -> load_global_summary_by_number(Context, E164)
     end.
 
+-spec load_global_summary_by_number(cb_context:context(), kz_term:ne_binary()) -> cb_context:context().
+load_global_summary_by_number(Context, Number) ->
+    load_summary(Context, {'keymap', Number}, ?PORT_REQ_NUMBERS).
+
+-spec load_account_summary_by_number(cb_context:context(), kz_term:proplist()) -> cb_context:context().
```
jamesaimonetti: `Keys` is not a proplist but `kz_term:ne_binaries()`

PR #5277, opened 2018-11-09
ACDc developments

N4COM:
- Created new strategy: "ring_all". It allows to ring all agents like a ring group.
- Added the possibility to have more than one winner.
- Added event member_call_satisfied for telling to agents that the request has been satisfied
- Bug fix


jamesaimonetti:
@N4COM A couple things to fix from CircleCI (not sure on MakeBusy just yet). But, we generally prefer a PR against master with backport PRs as necessary. Can you open one for master as well?

PR #5283, opened 2018-11-12
Added feature to emit aleg uuid into bleg

sergey-safarov:
Useful when carrier use sip_trace software like homer.

### DIFF
--- a/applications/ecallmgr/src/ecallmgr_fs_xml.erl
+++ b/applications/ecallmgr/src/ecallmgr_fs_xml.erl
@@ -41,6 +41,7 @@
 -include("ecallmgr.hrl").
 
 -define(DEFAULT_USER_CACHE_TIME_IN_MS, ?MILLISECONDS_IN_HOUR). %% 1 hour
+-define(ALEG_CALLID_HEADER, kapps_config:get_ne_binary(?APP_NAME, <<"aleg_callid_header">>)).
 
 -spec acl_xml(kz_json:object()) -> {'ok', iolist()}.
 acl_xml(AclsJObj) ->
@@ -427,6 +428,13 @@ route_resp_ringback(JObj) ->
             MsgId = kz_api:msg_id(JObj),
             Stream = ecallmgr_util:media_path(Media, 'extant', MsgId, JObj),
             action_el(<<"set">>, <<"ringback=", (kz_term:to_binary(Stream))/binary>>)
+    end.
+
+-spec maybe_emmit_aleg_callid() -> kz_types:xml_el() | 'undefined'.
+maybe_emmit_aleg_callid() ->
+    case ?ALEG_CALLID_HEADER of
+        'undefined' -> 'undefined';
+        HeaderName -> action_el(<<"set">>, <<"sip_h_", HeaderName/binary, "=${uuid}">>)
     end.
 
 -spec route_resp_ccvs(kz_json:object()) -> kz_types:xml_el().
@@ -1415,6 +1423,7 @@ route_resp_park_xml(JObj, DialplanContext) ->
             ,unset_custom_sip_headers()
             ,route_resp_set_control_info(DialplanContext)
             ,route_resp_fire_route_win(JObj, DialplanContext)
+            ,maybe_emmit_aleg_callid()
             ,route_resp_park()
             ],
     [E || E <- Exten, E =/= 'undefined'].
--- a/applications/crossbar/priv/api/swagger.json
+++ b/applications/crossbar/priv/api/swagger.json
@@ -30861,6 +30861,10 @@
                     "description": "ecallmgr acls",
                     "type": "object"
                 },
+                "aleg_callid_header": {
+                    "description": "Header to emit aleg uuid into bleg",
+                    "type": "string"
+                },
                 "allow_endless_recording": {
                     "default": false,
                     "description": "ecallmgr allow_endless_recording",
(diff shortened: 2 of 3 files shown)


applications/ecallmgr/src/ecallmgr_fs_xml.erl
```diff
@@ -404,6 +406,13 @@ route_resp_ringback(JObj) ->
             action_el(<<"set">>, <<"ringback=", (kz_term:to_binary(Stream))/binary>>)
     end.
 
+-spec maybe_emmit_aleg_callid() -> kz_types:xml_els().
+maybe_emmit_aleg_callid() ->
+    case ?ALEG_CALLID_HEADER of
+        'undefined' -> [];
+        HeaderName -> [action_el(<<"set">>, <<"sip_h_", HeaderName/binary, "=${uuid}">>)]
```
k-anderson: @sergey-safarov Isn't freeswitch's UUID parameter the call-id of that leg?  Not sure how ${uuid} on the leg issuing the fetch request (usually the aleg) will translate to the aleg. Futher, custom SIP headers should start with "X-", iirc sip_h_* does not automatically do that.  Finally, this should be optional and default to disabled because it will expand the size of the UDP packet and fragmentation can be a real issue for some networks if it exceeds the MTU.

My untested recommendation would be something like this:

maybe_emmit_other_leg_callid() ->
    case  kapps_config:get_is_true(?APP_NAME, <<"emmit_other_leg_callid">>, 'false') of
        'false' -> [];
        'true' -> [action_el(<<"set">>, <<"sip_h_X-Other-Leg-Call-ID=${Other-Leg-Unique-ID}">>)]
    end.

However, I am not sure if the var replacement will be dynamic or only when the sip_h_* var is set.  Might need some tweaking but perhaps this helps show what I am thinking :)

Let me know!

k-anderson:
@lazedo no worries, what issues?

@sergey-safarov Had the thought that a better parameter to send for correlation would be the interaction-id :)

applications/ecallmgr/src/ecallmgr_fs_xml.erl
sergey-safarov: @k-anderson here `ALEG_CALLID_HEADER` is not param in fetch request/responce.
`ALEG_CALLID_HEADER` ecallmgr configuration param.
```erlang
-define(ALEG_CALLID_HEADER, kapps_config:get_ne_binary(?APP_NAME, <<"aleg_callid_header">>)).
```
This value is semi permanent for all calls and not affect fetch request/response.

`ALEG_CALLID_HEADER` is optional already. Because header default value is `undefined` (feature is disabled). In this case FreeSwitch dialplan not generate additional XML elements.

Var replacement is dynamic. Works well.

P.S. Commit about 'X-' headers will be added.

sergey-safarov:
> just putting a hold on this, there's some other issues that i would like to address here

Hi Luis
If possible, please share some info about issue with me.

sergey-safarov:
> Had the thought that a better parameter to send for correlation would be the interaction-id :)

That is not for companies that use `Homer` call monitoring/troubleshooting tool. This tools is use A-leg Call-ID for call correlation.
https://github.com/sipcapture/homer/wiki/Using-Homer-5#session-correlation

If exist other tool like Homer that use ` interaction-id`. I can extend PR to support both cases. But need to know what is tool.

lazedo:
@sergey-safarov we need to address the fact that the correlation-id may already be set and we don't want to override it. i'm experimenting this. `sip_h_*` is fine, no need to restrict to `sip_h_X-*`. there's also some changes in kamailio scripts for this as part of the tests

applications/ecallmgr/src/ecallmgr_fs_xml.erl
sergey-safarov: Also about UDP size.
All X-headers will be striped when call is routed to phone via kamailio.
[Kamailio config link](https://github.com/2600hz/kazoo-configs-kamailio/blob/55091d53807d6d9034c692499eafb20e957077e3/kamailio/default.cfg#L682)

For carrier this not matter because they not use NAT or have properly configured NAT equpment.

lazedo:
@sergey-safarov `correlation-id` is a generic term, it applies to both `Interaction-ID` and  `aleg_callid_header`.  no action needed

bazemoreb:
@sergey-safarov - anything we can do to get this one worked out and compatible with Homer?

sergey-safarov:
Hi Brandon @bazemoreb 
I guess Luis want complete PRs https://github.com/2600hz/kazoo-configs-kamailio/pull/91 and https://github.com/2600hz/kazoo-configs-kamailio/pull/92 before merge this PR.

Luis @lazedo could you review and merge this PR before kamailio config is finished?

P.S. This PR is rebased on master

bazemoreb:
Perfect, thanks @sergey-safarov

jamesaimonetti:
@lazedo anything needed to move this along?

sergey-safarov:
Hi Luis @lazedo 
Kamailio PRs is merged. Could you make this PR review?

lazedo:
@sergey-safarov i think this is already addressed here https://github.com/2600hz/kazoo/blob/master/applications/ecallmgr/src/ecallmgr_fs_xml.erl#L1407

sergey-safarov:
Thanks Luis @lazedo 
I will test your implementation and will give feedback.

sergey-safarov:
I tested master branch and `freeswitch 1.8.6`
All works as expected.
Closing this PR

PR #5291, opened 2018-11-15
fix relx script for macOS

icehess:


noahmehl:
I had to cherry pick this into master to work, but maybe that was my local env...

I was able to make build-dev-release :)

PR #5301, opened 2018-11-28
Handle new device only on successful save

jamesaimonetti:
When a device is dry-run'd with a 402 accept charges, the
'maybe_add_mobile_mdn/1' call would still add the number to the
account. On the subsequent PUT to accept charges, since the number now
exists, the PUT will fail and the device will not be created.

Instead, we check that the save was successful (since 402 is an
'error') before handling the new device. This is more in line with how
it used to work when crossbar_services:maybe_dry_run/2 took a callback
function if not dry-running.


jamesaimonetti:
See 4.3 at #5302

PR #5340, opened 2018-12-11
[MASTER] QF: Fixed porting error response

swysor:


applications/crossbar/src/modules/cb_port_requests.erl
```diff
@@ -382,10 +384,20 @@ patch(Context, Id, NewState=?PORT_CANCELED) ->
...
-        {'error', _} ->
-            cb_context:add_system_error('datastore_fault', <<"unable to cancel port request from carrier">>, Context)
+        {'error', {Code, Response}} ->
+            handle_phonebook_error(Context, Code, Response);
+        {'error', Message} ->
+            cb_context:add_system_error('datastore_fault', Message, Context)
     end.
 
+-spec handle_phonebook_error(cb_context:context(), integer(), kz_json:object()) -> cb_context:context().
+handle_phonebook_error(Context, Code, Response) ->
```
k-anderson: @swysor We should set a parameter that indicates it a pass-through error and where it originated so that API consumers don't try to parse "data" as a kazoo format

PR #5344, opened 2018-12-13
KAZOO-6017: Update CCVs to update presence id

remcovanvugt:
Presence is not being tracked for hotdesked users on outbound calls. Adding the Presence-ID to the CCVs avoids the Presence-ID defaulting to the phones SIP username.

### DIFF
--- a/applications/callflow/src/cf_route_win.erl
+++ b/applications/callflow/src/cf_route_win.erl
@@ -268,6 +268,7 @@ update_ccvs(Call) ->
               [{<<"Hold-Media">>, kz_attributes:moh_attributes(<<"media_id">>, Call)}
               ,{<<"Caller-ID-Name">>, CIDName}
               ,{<<"Caller-ID-Number">>, CIDNumber}
+              ,{<<"Presence-ID">>, kz_attributes:presence_id(Call)}
                | get_incoming_security(Call)
               ]),
     kapps_call:set_custom_channel_vars(Props, Call).


jamesaimonetti:
Hey @remcovanvugt, what code is looking in the CCVs for Presence-ID? My cursory review shows Presence-ID typically being a top-level key but I easily could have missed where the CCVs are consulted.

remcovanvugt:
Hi @jamesaimonetti This is probably just a matter of naming conventions. Presence-ID is actually a top level key. It just that cf_route_win has a function that is named update_ccvs to update various call related properties which are all top-level keys (Caller-ID-Name, Caller-ID-Number). None of these are actually in Custom-Channel-Vars. So I just assumed this was a good place to put this fix.

Please correct me if I'm wrong :)

jamesaimonetti:
@remcovanvugt i'm still not sure I believe this is the fix.

The problem, as I understand it, is the presence_id being generated is using the SIP username of the device instead of the currently-hotdesked user's presence_id, yes?

I would look at the `kz_endpoint:merge_attribute/4` and see what the `Owner` variable is set to when merging the account/user/device attributes. You may want to define a custom function for the `presence_id` key instead that checks the device's hotdesk'd users for presence_id.

remcovanvugt:
Hi @jamesaimonetti I'm not sure if we are on the same frequency here.

The problem is, when a user hotdesked on a phone **starts** a call to whatever destination the presence id is never updated to reflect the currently hotdesked users presence id. It sticks to the default of the device. As far as I know, kz_endpoint is never involved when a user dials **out**? Or am I missing that somewhere?

So I don't see why merge_attributes in kz_endpoint would be able to solve this? As far as I know kz_endpoint is only involved in constructing endpoints for bridging to them? That works totally fine without this fix, the presence id is correctly updated.

That's why I did put this here, because other call related attributes are set here such as language, caller id, etc. It works fine here, but there may as well be a better place to fix this which I can't see because of a misunderstanding on my side in the way outbound calls get setup?

remcovanvugt:
@lazedo Sorry for the delay. Did you guys already start on a fix for this? Otherwise I will have a stab at it..

PR #5356, opened 2018-12-19
[master] qf: fixed comments not being applied in response from phonebook

swysor:


core/kazoo_number_manager/priv/couchdb/views/port_requests.json
```diff
@@ -97,7 +97,7 @@
         "port_in_numbers": {
             "map": [
                 "function(doc) {",
-                "  if (doc.pvt_type != 'port_request' || doc.pvt_deleted || doc.pvt_port_state == 'unconfirmed' || doc.pvt_port_state == 'completed' || doc.pvt_port_state == 'rejected' || doc.pvt_port_state == 'canceled') return;",
+                "  if (doc.pvt_type != 'port_request' || doc.pvt_deleted || doc.pvt_port_state == 'completed' || doc.pvt_port_state == 'canceled') return;",
```
icehess: Why changing this? Keep in mind that this view is also used by stepswitch and knm find number to complete a port request when a call is coming to one of the port's number.

swysor: I will just make a new view then

swysor: New view added

PR #5357, opened 2018-12-20
HELP-40948: disable transfer follows when starting recording by config

jamesaimonetti:
When the caller is an authorized endpoint, prior to this change the
flag to have recordings follow transfers was always set, regardless of
whether the endpoint had recordings configured to be started
implicitly.

However, when the recording isn't started by config but is started via
callflow action (`cf_record_call`), the resulting recording will not
follow transfers which is unexpected behaviour.

This change now only sets the follow flag to false if the
recording-by-config is happening.

When starting recordings-by-action, always follow transfers.

### DIFF
--- a/applications/callflow/src/module/cf_record_call.erl
+++ b/applications/callflow/src/module/cf_record_call.erl
@@ -29,8 +29,9 @@ handle(Data0, Call) ->
 
 -spec handle(kz_json:object(), kapps_call:call(), kz_term:ne_binary()) -> kapps_call:call().
 handle(Data, Call, <<"start">>) ->
-    Call1 = kapps_call:kvs_store('recording_follow_transfer', 'true', Call),
-    lager:info("starting call recording via action"),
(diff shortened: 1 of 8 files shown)


jamesaimonetti:
See 4.3 at #5358  and 4.2 at #5359

applications/callflow/src/module/cf_record_call.erl
```diff
@@ -29,7 +29,9 @@ handle(Data0, Call) ->
 
 -spec handle(kz_json:object(), kapps_call:call(), kz_term:ne_binary()) -> kapps_call:call().
 handle(Data, Call, <<"start">>) ->
-    cf_exe:update_call(kapps_call:start_recording(Data, Call));
+    Call1 = kapps_call:kvs_store('recording_follow_transfer', 'true', Call),
```
lazedo: can we get the desired behavior from Data object (we can still have it default to true) ?

jamesaimonetti: That's a good point, I'll add that

PR #5360, opened 2018-12-21
[master] some services/payment fixes

icehess:
* Return reason why account is/is not in good standing (jonny5 flat rate, crossbar requests, kazoo number manager)
* Don't try to update payment token if account doesn't have a plans (maintenance command only)
* Change wording in service added when item is added to plan

4.3: https://github.com/2600hz/kazoo/pull/5361


core/kazoo_number_manager/src/knm_numbers.erl
```diff
@@ -701,10 +701,10 @@ check_creditably(_Services, _Quotes, Amount) when Amount =< 0 ->
 check_creditably(Services, _Quotes, Amount) ->
     Options = #{amount => kz_currency:dollars_to_units(Amount)},
     case kz_services:is_good_standing(Services, Options) of
-        'true' -> 'ok';
-        'false' ->
+        {'true', _} -> 'ok';
```
jamesaimonetti: The spec for this function is wrong, takes a `boolean() | integer()` in the last argument, no?

applications/crossbar/src/crossbar_services.erl
```diff
@@ -82,9 +82,12 @@ check_creditably(Context, _Services, _Quotes, Amount) when Amount =< 0 ->
 check_creditably(Context, Services, _Quotes, Amount) ->
     Options = #{amount => kz_currency:dollars_to_units(Amount)},
     case kz_services:is_good_standing(Services, Options) of
-        'true' -> Context;
-        'false' ->
-            cb_context:add_system_error('no_credit', Context)
+        {'true', _} -> Context;
```
jamesaimonetti: Similar spec issue here with `boolean() | number()` I think.

applications/teletype/priv/templates/service_added.html
```diff
@@ -64,7 +64,7 @@ <h2 style="margin:0;padding:0;text-align:center;font-family:'Open Sans',sans-ser
...
+                                                %}<br>{% with type=item_props.changes.type %}{% if type == "plan_new" %}<span style="color:#07861d;">New service added to your plan</span>{% elif type == "plan_removed" %}<span style="color:#de0a0a">{{item_props.changes.quantity}} item(s) stopped billed by your service plan</span>{% elif type == "added" %}<span style="color:#07861d">{{item_props.changes.quantity}} item(s) added to your service</span>{% elif type == "removed" %}<span style="color:#de0a0a">{{item_props.changes.quantity}} item(s) removed from your service</span></small></span>{% endif %}{% endwith %}{% endif %}
```
k-anderson: @icehess 
Should update the verbiage for removal of a service plan item as well.

Currently:
"{{item_props.changes.quantity}} item(s) stopped billed by your service plan"
should make it match the added, perhaps:
"Service removed from your plan"

core/kazoo_services/src/kz_services.erl
```diff
@@ -683,30 +691,30 @@ is_good_standing_fold(Services, _Options, []) ->
     lager:debug("account ~s ran out of good funs, the ugly"
                ,[account_id(Services)]
                ),
-    'false';
+    {'false', <<"has not enough credit or no payment token">>};
```
k-anderson: @icehess the "has" in these errors is odd.  It makes it so they are not stand alone:

{'error', "has not enough credit or no payment token"} vs {'error', "not enough credit or no payment token"}

I suspect you are doing that because logging/API errors but why not just make the 'has' part of that format?

(PS> applies to the other reasons like "has exceeded the maximum postpay amount", additionally in that case they "will" exceeded if allowed to debit the proposed amount...)

core/kazoo_services/src/kz_services.erl
```diff
@@ -739,18 +747,18 @@ has_good_balance(Services, #{amount := Amount}=Options) ->
...
 has_good_balance(Balance, Amount, 'true', MaxPostPay) ->
     case (Balance - Amount) > MaxPostPay of
         'true' ->
             {'true', <<"has enough postpay balance">>};
         'false' ->
-            Msg = io_lib:format("has exceed the maximum postpay amount,"
-                                " curr_balance: ~b amount: ~b max_postpay: ~b proposed_balance: ~b"
-                               ,[Balance, Amount, MaxPostPay, Balance - Amount]
+            Msg = io_lib:format("has exceeded the maximum postpay amount,"
+                                " current balance: ~b max postpay: ~b proposed balance: ~b"
```
k-anderson: @icehess maybe there is a cleaner way to convey this?

Msg = io_lib:format("debit of ~b from ~b exceeds the maximum postpay amount ~b", [Amount, Balance, MaxPostPay]).

core/kazoo_services/src/kz_services.erl
```diff
@@ -739,18 +747,18 @@ has_good_balance(Services, #{amount := Amount}=Options) ->
 has_good_balance(Balance, Amount, 'false', _) when (Balance - Amount) > 0 ->
     {'true', <<"has positive balance">>};
 has_good_balance(Balance, Amount, 'false', _) when (Balance - Amount) =< 0 ->
-    Msg = io_lib:format("has negative balance, curr_balance: ~b amount: ~b proposed_balance: ~b"
-                       ,[Balance, Amount, Balance - Amount]
+    Msg = io_lib:format("has negative balance, current balance: ~b proposed balance: ~b"
```
k-anderson: @icehess maybe there is a cleaner way to convey this?

Msg = io_lib:format("debit of ~b from ~b results in a negative balance", [Amount, Balance]),

applications/crossbar/src/crossbar_services.erl
icehess: But `boolean() | number()` is correct, no?

PR #5377, opened 2019-01-03
[master] SSSU-73: date format and improve port's templates, store ported numbers in another field

icehess:
**Changes to port templates:**

* change date format for porting to show AM/PM and timezone
* conditionally show scheduled or request port date for some port templates
* improve wording
* show port request important details
* for comment or reason show author if author field is exists otherwise get user info, if failed show "An agent"
* fix billing address fields
* fix encoding issue in Email subject (e.g. `&` is encoded to `&amp;`)
* do not add ports form and bill pdf attachments. Only adds them to port_request_admin

**Changes to Kazoo Number Manager:**

* Preserved ported numbers to `ported_numbers` field in port request document after the port is transitioned to `completed` state so Teletype `ported` template can show the numbers (UI can use this to show ported numbers too)

### DIFF
--- a/applications/teletype/src/teletype_port_utils.erl
+++ b/applications/teletype/src/teletype_port_utils.erl
@@ -12,6 +12,7 @@
 -export([fix_port_request_data/2]).
 
 -include("teletype.hrl").
+-include_lib("kazoo_number_manager/include/knm_port_request.hrl").
 
 -spec is_comment_private(kz_json:object()) -> boolean().
 is_comment_private(DataJObj) ->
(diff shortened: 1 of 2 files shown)


applications/teletype/src/teletype_port_utils.erl
```diff
@@ -110,7 +110,10 @@ fix_port_request_data(JObj, DataJObj) ->
 
 -spec fix_numbers(kz_json:object(), kz_json:object()) -> kz_json:object().
 fix_numbers(JObj, _DataJObj) ->
-    NumbersJObj = kz_json:get_value(<<"numbers">>, JObj, kz_json:new()),
+    NumbersJObj = kz_json:get_json_value(<<"numbers">>
```
k-anderson: @icehess its possible for there to be both numbers and ported_numbers on the document...

core/kazoo_number_manager/src/knm_port_request.erl
```diff
@@ -466,7 +466,10 @@ is_in_account_and_in_service(AccountId, PN) ->
 %%------------------------------------------------------------------------------
 -spec clear_numbers_from_port(kz_json:object()) -> {ok, kz_json:object()}.
 clear_numbers_from_port(PortReq) ->
-    Cleared = kz_json:set_value(?NUMBERS_KEY, kz_json:new(), PortReq),
+    Update = [{?NUMBERS_KEY, kz_json:new()}
```
k-anderson: @icehess would want to be able to find ports by number in either list...

IE: when attempting to find the port request for 555 111 2222 should check in both ?NUMBERS_KEY and "ported_numbers".

icehess: `all_port_in_numbers` and `listing_by_number` views are updated to emit number in both `numbers` and `ported_numbers` key. Both is only used by cb_port_requests to list port requests.

PR #5401, opened 2019-01-08
KAZOO-6007: preserve null on provisioner payloads but remove from db

k-anderson:
Wanted to start a discussion before adding kz_json tests and master PR.


applications/crossbar/src/modules/provisioner_util.erl
```diff
@@ -131,6 +131,15 @@ ensure_mac_cleansed(JObj) ->
             kzd_devices:set_mac_address(JObj, cleanse_mac_address(MacAddress))
     end.
 
+-spec get_new_device(cb_context:context()) -> kzd_devices:doc().
+get_new_device(Context) ->
+    JObj = cb_context:doc(Context),
```
jamesaimonetti: Could `JObj` be populated with default values from JSON schema validation that might be necessary later?

k-anderson: @jamesaimonetti not at this time, no

PR #5403, opened 2019-01-08
PISTON-653: JSON-encode kz_json CAVs when set on a channel

danielfinke:
We would like to be able to have our pivot users set more complex structures for CAVs (also would apply to CCVs) using JSON. kapps_call and kapps_call_command can already pass these through to ecallmgr, however they are discarded because of their type. This PR encodes them as binary and they are then retrievable in blackhole events, CDRs, etc. We are leaving it up to the user to decode them

### DIFF
--- a/applications/ecallmgr/src/ecallmgr.hrl
+++ b/applications/ecallmgr/src/ecallmgr.hrl
@@ -182,6 +182,7 @@
 %% message
 -define(CHANNEL_VAR_PREFIX, "ecallmgr_").
 -define(APPLICATION_VAR_PREFIX, "cav_").
+-define(JSON_APPLICATION_VAR_PREFIX, "json_cav_").
 
 -define(CCV(Key), <<?CHANNEL_VAR_PREFIX, Key/binary>>).
 -define(GET_CCV(Key), <<"variable_", ?CHANNEL_VAR_PREFIX, Key/binary>>).
@@ -195,6 +196,9 @@
 -define(GET_CAV(Key), <<"variable_", ?APPLICATION_VAR_PREFIX, Key/binary>>).
 -define(SET_CAV(Key, Value), <<?APPLICATION_VAR_PREFIX, Key/binary, "=", Value/binary>>).
 -define(GET_CAV_HEADER(Key), <<"variable_sip_h_X-", ?APPLICATION_VAR_PREFIX, Key/binary>>).
+
+-define(JSON_CAV(Key), <<?JSON_APPLICATION_VAR_PREFIX, Key/binary>>).
+-define(GET_JSON_CAV(Key), <<"variable_", ?JSON_APPLICATION_VAR_PREFIX, Key/binary>>).
 
 -define(CREDS_KEY(Realm, Username), {'authn', Username, Realm}).
 
--- a/applications/ecallmgr/src/ecallmgr_call_command.erl
+++ b/applications/ecallmgr/src/ecallmgr_call_command.erl
@@ -472,13 +472,14 @@ get_fs_app(Node, UUID, JObj, <<"set">>) ->
         'true' ->
             ChannelVars = kz_json:to_proplist(kz_json:get_json_value(<<"Custom-Channel-Vars">>, JObj, kz_json:new())),
             CallVars = kz_json:to_proplist(kz_json:get_json_value(<<"Custom-Call-Vars">>, JObj, kz_json:new())),
-            AppVars = kz_json:to_proplist(kz_json:get_json_value(<<"Custom-Application-Vars">>, JObj, kz_json:new())),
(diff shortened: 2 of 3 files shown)


danielfinke:
@lazedo since it's serialized to binary it comes back as binary from FS -> ecallmgr in events. At this point it leaves it up to the user to decode the JSON string. I'm up for looking into a way to trigger an automatic decode when it gets back to ecallmgr.

I wouldn't suggest we attempt a decode on every CCV/CAV as it comes in as that would be a lot of overhead, but perhaps introduce a new FS variable prefix such as "json_cav_" which would be used if the input var was a JObj. Then the incoming event side would know to decode it. Thoughts?

jamesaimonetti:
Personally, my thought on CAVs are that they're solely for the external clients to read and interpret. Kazoo should pass the encoded JSON along but not attempt to decode it. The receiving blackhole/pivot/webhook client can then decode at their leisure. To Kazoo, it should remain an encoded binary.

My 2 cents anyway.

I also would caution using CCVs for this. Again, personally, I would prefer CCV values stay as simple strings/integers/booleans. Persuasion possible but gut reaction is "I don't like" :)

danielfinke:
@jamesaimonetti I'm open to suggestions for alternatives. Currently, one could use kvs on a kapps_call to contain some information including JObjs. Our bigger picture however includes the other recent PR #5404. What our customer's use case is:
1. Caller calls in and hits pivot
2. Pivot uses cf_set_variables with the intention of setting a complex, nested structure of their choice
3. Eventually the caller is bridged to some agent
4. That agent is using an interface that has websocket bindings for calls belonging to that agent
5. The customer wants to see the complex data structure they set on the caller's CAVs on the agent's CAVs <- this step is why #5404 was required (allows the websocket binding to be simpler on the client - they don't have to bind to all calls and filter all calls not belonging to the user)

jamesaimonetti:
@danielfinke yeah I figured the use case was pretty close to this :)

I think you'll need to include a schema update to `callflows.set_variables` as it only allows strings as values.

More of my thoughts:

1. Hide ability behind a feature flag
  A client could potentially (well, let's be real - will eventually) include a large enough JSON object which would balloon the size of the data transferred between Kazoo and FreeSWITCH as well as the AMQP payloads (since CAVs are included on all call events published to RabbitMQ). On a busy enough system this could overwhelm the broker and saturate the NICs. System operators should need to opt-in to allow their clients to do this to avoid unpleasant surprises.

2. Restrict the size of the object allowed?
    If restricted, Pivot will fail to process the response due to validation errors but doesn't communicate back to the responding server that this pivot response will not be able to proceed. This is handled by the cf_pivot's child (if any). The validation error will be in the pivot debug API but that's some indirection to figure out why a call was failing. And maybe that's okay, the price you pay for using CAVs as a database?

3. Do not allow this on CCVs in any case. I think these should stay simple and restricted for Kazoo's internal purposes.

danielfinke:
@jamesaimonetti just to clarify, are you treating CAVs and CCVs as two distinct things or the same as they are backed by the same thing: FS channel vars? The PR does restrict the JSON-encoding to just CAVs, which means they don't interfere with Kazoo's internal functions, but they are subject to the same other limitations that you mentioned. I guess what I'm wondering is are you saying you think they should not be allowed in any case for just CCVs, or CCVs _and_ CAVs?

jamesaimonetti:
@danielfinke within the broader Kazoo ecosystem, yes, I think treat CAVs and CCVs separately. For storing them in the channel/leg vars in FreeSWITCH, you're right that they are encoded the same way. My comment is more a philosophical one that Kazoo itself not use JSON objects in CCVs.

In practice, esp with private Kazoo applications, people can do as they please and use JSON objects in CCVs - I don't think ecallmgr should encode them differently (other than the var prefix) from CAVs.

Does that clarify it?

lazedo:
@danielfinke @jamesaimonetti

i think the `size` concern because of `json` is not a valid argument because there is no limit presently in the number of  `CAV's` (as simple types) a `pivot` can set, which means thousands of `CAV's` would still eventually be set. so this `size` issue should be addressed maybe in another PR before these changes go into a stable branch. one of the ways to restrict size would be to enforce schema strict checking on these `json_cavs_`.

for the encoding/decoding part, this should be handled as a whole, ie, serialize as binary to freeswitch, deserialize as json from freeswitch. for that, the approach to set something like `json_cav_` prefix is the preferred method. the upcoming changes to `mod_kazoo` (custom serialization of  events as json/map) will support the proper encoding from freeswitch to handling application (ecallmgr or other).

to proceed with this pr, the move to `json_cav_` and a proper fs decode from `props` in ecallmgr_call_events is required.

let's leave `Custom-Application-Vars` as `simple types`.

hope this all makes sense

jamesaimonetti:
@danielfinke can you rebase this branch against latest master? @lazedo any objections remaining with the changes here?

lazedo:
@jamesaimonetti no, just waiting on the rebase

jamesaimonetti:
Merged

PR #5404, opened 2019-01-08
PISTON-652: add Export-All option for kapps_call_command:set

danielfinke:
Allow CCVs and CAVs to be exported

### DIFF
--- a/applications/ecallmgr/src/ecallmgr_call_command.erl
+++ b/applications/ecallmgr/src/ecallmgr_call_command.erl
@@ -474,10 +474,7 @@ get_fs_app(Node, UUID, JObj, <<"set">>) ->
             CallVars = kz_json:to_proplist(kz_json:get_json_value(<<"Custom-Call-Vars">>, JObj, kz_json:new())),
             AppVars = kz_json:to_proplist(kz_json:get_json_value(<<"Custom-Application-Vars">>, JObj, kz_json:new())),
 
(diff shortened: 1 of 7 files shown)


applications/ecallmgr/src/ecallmgr_call_command.erl
```diff
@@ -474,9 +474,15 @@ get_fs_app(Node, UUID, JObj, <<"set">>) ->
             CallVars = kz_json:to_proplist(kz_json:get_json_value(<<"Custom-Call-Vars">>, JObj, kz_json:new())),
             AppVars = kz_json:to_proplist(kz_json:get_json_value(<<"Custom-Application-Vars">>, JObj, kz_json:new())),
 
+            Command = case kz_json:get_boolean_value(<<"Export-All">>, JObj) of
```
lazedo: this can return `undefined` and will throw a function_clause exception.
i would prefer if you move this to function, something like
```
Command  = get_set_command_action(JObj),

...
get_set_command_action(JObj) ->
     case get_set_action(kz_json:get_boolean_value(<<"Export-All">>, JObj, 'false') of
            'true' -> <<"kz_export">>;
             'false' -> <<"kz_multiset">>
     end.

```

danielfinke: OK, broke that out into it's own function. Good catch on the default value

PR #5407, opened 2019-01-09
Add UTF-8 support for json objs values

harenson:


core/kazoo_stdlib/src/kz_json.erl
```diff
@@ -1425,3 +1425,9 @@ exec(Funs, ?JSON_WRAPPER(_)=JObj) ->
 exec_fold({F, K, V}, C) when is_function(F, 3) -> F(K, V, C);
 exec_fold({F, V}, C) when is_function(F, 2) -> F(V, C);
 exec_fold(F, C) when is_function(F, 1) -> F(C).
+
+-spec utf8_binary(any()) -> kz_term:ne_binary().
```
jamesaimonetti: Spec should be `utf8_binary(json_term()) -> json_term().` `any()` means I could pass a PID, Reference or Port value - those are not allowed.

core/kazoo_stdlib/src/kz_json.erl
```diff
@@ -257,7 +257,7 @@ are_equal(JObj1, JObj2) ->
 %%------------------------------------------------------------------------------
 -spec from_list([set_value_kv()] | flat_proplist()) -> object() | flat_object().
 from_list(L) when is_list(L) ->
-    ?JSON_WRAPPER(props:filter_undefined(L)).
+    ?JSON_WRAPPER([{K, utf8_binary(V)} || {K, V} <- props:filter_undefined(L)]).
```
jamesaimonetti: I would just add the filter here directly rather than calling out to `props`. As-is, you'll traverse the list to filter, then traverse the new list to map the `utf8_binary` function. Just make life easy and `[{K, utf8_binary(V)} || {K, V} <- L, V =/= 'undefined'].` One pass, filter and map :)

core/kazoo_stdlib/test/kz_json_tests.erl
```diff
@@ -1104,6 +1104,39 @@ from_map_test_() ->
      }
     ].
 
+utf8_binary_values_test_() ->
+    Props = [{<<"name">>, <<"Loïc">>}],
```
jamesaimonetti: Any good Colombian cartoon characters that have UTF8 names we could use?

core/kazoo_stdlib/src/kz_json.erl
harenson: Nice tip, thanks :)

PR #5411, opened 2019-01-11
[4.3] qf: billable quantity diff is jobj now

icehess:
master: https://github.com/2600hz/kazoo/pull/5410


jamesaimonetti:
wuz merged

PR #5420, opened 2019-01-15
kazoo_attachments: Fixed storing recordings to external url

sergey-safarov:
Required cherry-pick into 4.3 branch

### DIFF
--- a/core/kazoo_attachments/src/kz_att_http.erl
+++ b/core/kazoo_attachments/src/kz_att_http.erl
@@ -83,7 +83,7 @@ format_verb(<<"PUT">>) ->
     'put';
 format_verb(<<"put">>) ->
     'put';
-format_verb(Value) ->
+format_verb(Value) when not erlang:is_binary(Value) ->
     format_verb(kz_term:to_binary(Value)).
 
 -spec fetch_attachment(gen_attachment:handler_props()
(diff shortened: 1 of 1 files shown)


core/kazoo_attachments/src/kz_att_http.erl
```diff
@@ -82,7 +82,9 @@ format_verb(<<"post">>) ->
 format_verb(<<"PUT">>) ->
     'put';
 format_verb(<<"put">>) ->
-    'put'.
+    'put';
+format_verb(Value) ->
```
lazedo: @sergey-safarov what `verb` are you trying to handle besides `POST`and `PUT` ?

sergey-safarov: In my case i got crash because `format_verb` is called with `put` argument as `text, not as `binary`.
I will add one commit.

lazedo: @sergey-safarov `Verb` comes from settings and it should be a `binary` not `text`

sergey-safarov: `Account/tenant` or `system_config` database?
I have defined recording at account object like
```json
  "call_recording": {
    "endpoint": {
      "inbound": {
        "onnet": {
          "enabled": true,
          "time_limit": 3600,
          "url": "http://217.12.247.220",
          "format": "mp3"
        }
      }
    }
  }
```
and get crash
one more commit is pushed.

sergey-safarov:
PR is rebased

lazedo:
@sergey-safarov the problem comes from recording routines when using `legacy` settings for saving the recordings. this was exposed by a recent change to `kz_att_http`.

check if this #5421 fixes your crashes

sergey-safarov:
I close this because more preferred fix is created.

PR #5428, opened 2019-01-15
PISTON-660: set_accepting_charges to handle dry_run in v1 phone_numbers

danielfinke:
Most of this is pulled from cb_phone_numbers_v2

### DIFF
--- a/applications/crossbar/src/modules_v1/cb_phone_numbers_v1.erl
+++ b/applications/crossbar/src/modules_v1/cb_phone_numbers_v1.erl
@@ -211,27 +211,32 @@ validate_3(Context, ?HTTP_GET, Number, ?IDENTIFY) ->
 
 -spec post(cb_context:context(), path_token()) -> cb_context:context().
 post(Context, ?COLLECTION) ->
-    set_response(collection_process(Context), <<>>, Context);
+    CB = fun() -> ?MODULE:post(cb_context:set_accepting_charges(Context), ?COLLECTION) end,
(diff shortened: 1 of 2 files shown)


applications/crossbar/src/modules_v1/cb_phone_numbers_v1.erl
```diff
@@ -211,27 +211,32 @@ validate_3(Context, ?HTTP_GET, Number, ?IDENTIFY) ->
 
 -spec post(cb_context:context(), path_token()) -> cb_context:context().
 post(Context, ?COLLECTION) ->
-    set_response(collection_process(Context), <<>>, Context);
+    CB = fun() -> ?MODULE:post(cb_context:set_accepting_charges(Context), ?COLLECTION) end,
+    set_response(collection_process(Context), Context, CB);
```
k-anderson: @danielfinke Howdy, so this is an oversight.  The 'v1' APIs should NOT trigger dry_run and therefore should never need the callback...

Without testing anything ;) I think all the callback additions should be changed to a cleaner version of:

Context1 = cb_context:set_accepting_charges(Context),
set_response(collection_process(Context1), <<>>, Context1);

IE: set_accepting_charges to true for all v1 requests. in cb_phone_numbers.  Possibly could also be done more generically in the api_util / api_resource modules...  That is to say, one of the big differences between the v1 and v2 APIs was the addition of a 402 Payment Required on v2, so for backward compatibility v1 is never supposed to return 402.

PR #5434, opened 2019-01-18
[master] KAZOO-6024: determine reseller id properly for authz request by number

icehess:
4.3: https://github.com/2600hz/kazoo/pull/5433

### DIFF
--- a/applications/ecallmgr/src/ecallmgr_fs_authz.erl
+++ b/applications/ecallmgr/src/ecallmgr_fs_authz.erl
@@ -181,6 +181,9 @@ authz_response(JObj, Props, CallId, Node) ->
                 'true' -> authorize_account(JObj, Props, CallId, Node);
                 'false' ->
                     %% Set the following CCVs so that we can see why the call was barred in CDRs
+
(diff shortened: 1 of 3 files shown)


applications/ecallmgr/src/ecallmgr_fs_authz.erl
```diff
@@ -181,8 +181,24 @@ authz_response(JObj, Props, CallId, Node) ->
...
+                    %% (in case the call is inbound from carrier)
+                    AccountId = kz_json:get_value(<<"Account-ID">>
+                                                 ,JObj
+                                                 ,props:get_value(?GET_CCV(<<"Account-ID">>), Props)
+                                                 ),
+                    ResellerId = kz_json:get_value(<<"Reseller-ID">>
+                                                  ,JObj
+                                                  ,props:get_value(?GET_CCV(<<"Account-ID">>)
+                                                                  ,Props
+                                                                  ,kz_services_reseller:get_id(AccountId)
```
lazedo: that's a lot of defaults.
you're making ecallmgr aware of kz_services which is not good, ecallmgr should deal with props / json.
i think you can set the variables that come out of j5 either authorized or not.

PR #5440, opened 2019-01-22
KAZOO-6031: add system_config parameter to enforce good_standing

k-anderson:
Master PR #5441

### DIFF
--- a/core/kazoo_services/src/kz_services.erl
+++ b/core/kazoo_services/src/kz_services.erl
@@ -616,31 +616,59 @@ summary(?NE_BINARY=Account) ->
                    ],
     summary(fetch(Account, FetchOptions));
 summary(Services) ->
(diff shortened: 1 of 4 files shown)


core/kazoo_services/src/kz_services.erl
```diff
@@ -673,13 +696,19 @@ is_good_standing(Thing) ->
 %% * All other cases the account is not in good standing
 %% @end
 %%------------------------------------------------------------------------------
+%% @equiv is_good_standing(Thing, 0)
```
icehess: This is probably breaking edoc. `equiv` is only shown for this function and all other comments are ignored. And since you moved the comments for `is_good_standing/2` here, eodc is not generating any doc for arity 2.

k-anderson: @icehess I will fix this on master, but accept here to allow the build to continue.

PR #5441, opened 2019-01-22
KAZOO-6031: add system_config parameter to enforce good_standing

k-anderson:


k-anderson:
As per @icehess on the 4.3 PR, regarding moving the is_good_standing/0:

This is probably breaking edoc. equiv is only shown for this function and all other comments are ignored. And since you moved the comments for is_good_standing/2 here, eodc is not generating any doc for arity 2.

lazedo:
@icehess can you fix here or in a followup pr ?

icehess:
@lazedo pr updated

PR #5446, opened 2019-01-23
[master] check port support email first

icehess:


k-anderson:
@icehess Also, just saw this while reviewing...

get_port_submitter_emails(DataJObj) ->
    Keys = [[<<"port_request">>, <<"customer_contact">>]
           ,[<<"port_request">>, <<"notifications">>, <<"email">>, <<"send_to">>]
           ],


Seems like we should prioritize notifications.email.send_to over customer_contact, assuming that the notifications.email.send_to is the optional email on the last screen of the port request wizard because that entry is supposed to be used only if the user wants to override the default....  Put another way, where does customer_contact come from? :)

icehess:
I don't know what is `customer_contact` actually. I'=m just copy pasting from old code to new one.

PR #5454, opened 2019-01-23
PISTON-656: Added e164_dest to offnet org cdrs

bradfordben:
e164_destination is missing from CDR's for offnet origination calls

PR #5455, opened 2019-01-23
PISTON-657:Added e164_originator to offnet-origination cdrs

bradfordben:
Added e164_originator to offnet-origination cdrs

### DIFF
--- a/applications/stepswitch/src/stepswitch_util.erl
+++ b/applications/stepswitch/src/stepswitch_util.erl
@@ -70,26 +70,11 @@ convert_to_e164_format(Number, ConfigDoc, ConfigKey) ->
 -spec get_inbound_origination(kz_json:object()) -> kz_term:ne_binary().
 get_inbound_origination(JObj) ->
     {Number, _} = kapps_util:get_origination(JObj, ?APP_NAME, <<"inbound_user_origination_field">>),
-    case verify_is_did(Number) of
+    case knm_converters:is_reconcilable(Number) of
         'true' ->
             convert_to_e164_format(Number, ?SS_CONFIG_CAT, <<"assume_inbound_origination_e164">>);
         'false' ->
             'undefined'
-    end.
-
-%%------------------------------------------------------------------------------
-%% @doc Verify the DID is a number or a number beginning with +
-%% @end
-%%------------------------------------------------------------------------------
--spec verify_is_did(kz_term:ne_binary()) -> boolean().
-verify_is_did(<<$+, Did/binary>>) ->
-    verify_is_did(Did);
-verify_is_did(Did) ->
-    try
-        _ = binary_to_integer(Did),
-        'true'
-    catch error:badarg ->
-            'false'
     end.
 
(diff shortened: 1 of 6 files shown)


applications/stepswitch/src/stepswitch_util.erl
```diff
@@ -49,11 +50,48 @@ get_realm(JObj) ->
...
     end.
 
+%%------------------------------------------------------------------------------
+%% @doc Get the origination DID of a inbound call if its a number, else return undefined
+%% @end
+%%------------------------------------------------------------------------------
+-spec get_inbound_origination(kz_json:object()) -> kz_term:ne_binary().
+get_inbound_origination(JObj) ->
+    {Number, _} = kapps_util:get_origination(JObj, ?APP_NAME, <<"inbound_user_origination_field">>),
+    case verify_is_did(Number) of
```
k-anderson: @bradfordben would be better to use:

knm_converters:is_reconcilable(Number)

instead of:

verify_is_did/1

The knm_converters:is_reconcilable/1 is oddly named but basically answers the question "is this string a number that could be routed outside kazoo (IE: it is NOT an internal extension, or is in the format of a public number)?".  Since that is under the system admin's control it would match internationalization rules rather than just attempting to conver any integer.  Put another way, the logic would become "if the inbound user origination field username (number) is not an internal kazoo extension then attempt to convert it to e164".

applications/stepswitch/src/stepswitch_util.erl
```diff
@@ -49,11 +50,48 @@ get_realm(JObj) ->
...
+        'false' ->
+            'undefined'
+    end.
+
+%%------------------------------------------------------------------------------
+%% @doc Verify the DID is a number or a number beginning with +
+%% @end
+%%------------------------------------------------------------------------------
+-spec verify_is_did(kz_term:ne_binary()) -> boolean().
+verify_is_did(<<$+, Did/binary>>) ->
```
k-anderson: @bradfordben Above I am recommending removing this but I thought it worth mentioning that if the number starts with a plus then it was provided in e164 format.  In that case it would be preferable to return as is and not attempt a double conversion... IE:

verify_is_did(<<$+, _/binary>> = E164) -> E164;

k-anderson:
Master PR #5454

applications/stepswitch/src/stepswitch_util.erl
bradfordben: Good point 👍

applications/stepswitch/src/stepswitch_util.erl
bradfordben: Great thanks, sounds good. Il make that change now and resubmit the PR.

bradfordben:
@k-anderson Not sure why the make-busy tests are failing. The failures dont seam to be related.

jamesaimonetti:
Merged, sorry for the delay

bradfordben:
Thanks 👍

PR #5457, opened 2019-01-24
HELP-40835: option to base64-encode data

jamesaimonetti:


core/kazoo_attachments/src/kz_att_http.erl
```diff
@@ -101,8 +101,11 @@ fetch_attachment(HandlerProps, DbName, DocId, AName) ->
     case kz_json:get_value(<<"url">>, HandlerProps) of
         'undefined' -> kz_att_error:new('invalid_data', Routines);
         Url ->
-            AttSettings = kz_json:get_value(<<"handler_props">>, HandlerProps),
-            handle_fetch_attachment_resp(fetch_attachment(Url), AttSettings, Routines)
+            Base64Encoded = kz_json:is_true(<<"base64_encode_data">>, HandlerProps, 'false'),
```
lazedo: this is not needed for decoding, httpc will decode if needed and the server can send or not the attachment encoded, it does not depend on what we saved (still useful for debugging purposes)

core/kazoo_attachments/src/kz_att_http.erl
```diff
@@ -116,7 +119,15 @@ handle_fetch_attachment_resp({'ok', <<"\r\n", _/binary>>=Multipart}
...
-    {'ok', base64:decode(Body)};
+    try base64:decode(Body) of
+        Decoded ->
+            lager:debug("decoded base64 response"),
+            {'ok', Decoded}
+    catch
+        'error':_R ->
+            lager:debug("response was not base64 encoded"),
+            {'ok', Body}
+    end;
```
lazedo: i think this is not needed, the `httpc:request` called from kz_http does the decoding if needed

jamesaimonetti: I don't believe this to be true. At least, when I updated the testing in pqc_cb_storage I was getting the encoded results back before I patched this. After this change I get the raw MP3 back. Might be because of the pqc_httpd not sending headers that httpc needs to auto-decode but we'll definitely have servers that are incomplete in this way.

lazedo: actually, there's no protocol handling for base64-encoded binary as the body of a http request (no header that can be set on response to indicate it). to use base64-encoded binary, one would have to use mime encoding (event with only one part) and provider Content-Transfer-Encoding in that part. so the pqc_httpd needs to decode the parts on post/put and return the binary image as-is on get.

lazedo: i would prefer to support well behaved http servers and this could be a blocker (compat issues later) when we start to stream the attachment instead of return the all binary in memory.

PR #5479, opened 2019-01-31
jonny5 onnet-termination exempt from billing

bputtick:


applications/crossbar/doc/ref/limits.md
```diff
@@ -14,6 +14,7 @@ Key | Description | Type | Default | Required | Support Level
 `burst_trunks` | The number of two-way, flat-rate trunks used only if no other trunks are available | `integer()` |   | `false` | `beta`
 `calls` | A hard limit for the total number calls | `integer()` |   | `false` | `beta`
 `inbound_trunks` | The number of inbound, flat-rate trunks | `integer()` |   | `false` | `supported`
+`onnet_exempt` | Determines if onnet-terminated calls should be exempt from billing (kazoo account to kazoo account calls) | `boolean()` | `false` | `false` | `beta`
```
jamesaimonetti: There are two versions of the "docs"; here in the `doc/ref` folder KAZOO auto-generates a "reference" doc with what is auto-detected from schemas and source code changes.

Up a directory, you should find `doc/limits.md`; that is the real doc that the docs site displays when built. So you should migrate any changes here to that doc to keep them in sync, as well as any additional explanatory text needed to make sense of what's changing in the code.

applications/jonny5/src/j5_authz_req.erl
```diff
@@ -198,21 +198,36 @@ maybe_authorize(Request, Limits) ->
...
-        <<"emergency">> ->
-            lager:debug("allowing emergency call"),
-            j5_request:authorize(<<"limits_disabled">>, Request, Limits);
-        <<"tollfree_us">> when CallDirection =:= <<"outbound">> ->
-            lager:debug("allowing outbound tollfree call"),
-            j5_request:authorize(<<"limits_disabled">>, Request, Limits);
-        _Else -> maybe_hard_limit(Request, Limits)
+        'false' ->
+            lager:debug("The resource type is ~s",[ResourceType]),
+            case  maybe_authorize_onnet(ResourceType, Limits) of
```
jamesaimonetti: Generally speaking we try to avoid more than 2 levels of nesting, preferring function clauses to case clauses.

For instance, the clause checking `Classification` should probably become a `maybe_authorize_by_classificiation(Request, Limits, Classification)` and do the checking of `Classification` in the function clauses.

applications/jonny5/src/j5_authz_req.erl
```diff
@@ -198,21 +198,36 @@ maybe_authorize(Request, Limits) ->
...
             j5_per_minute:authorize(Request, Limits);
-        <<"emergency">> ->
-            lager:debug("allowing emergency call"),
-            j5_request:authorize(<<"limits_disabled">>, Request, Limits);
-        <<"tollfree_us">> when CallDirection =:= <<"outbound">> ->
-            lager:debug("allowing outbound tollfree call"),
-            j5_request:authorize(<<"limits_disabled">>, Request, Limits);
-        _Else -> maybe_hard_limit(Request, Limits)
+        'false' ->
+            lager:debug("The resource type is ~s",[ResourceType]),
```
jamesaimonetti: Minor nit but we start log lines with lowercase

core/kazoo_documents/src/kzd_limits.erl.src
```diff
@@ -74,6 +75,18 @@ inbound_trunks(Doc, Default) ->
 set_inbound_trunks(Doc, InboundTrunks) ->
     kz_json:set_value([<<"inbound_trunks">>], InboundTrunks, Doc).
 
+-spec onnet_exempt(doc()) -> boolean().
```
jamesaimonetti: Similar to the `ref` vs `doc` situation, KAZOO auto-generates these `.erl.src` files from the schemas. Any changes here should be migrated to the `.erl` version and committed.

bputtick:
Thanks, lazedo.

1. Ok. just made more sense keeping it with its calling function
2. maybe_authorize_exception seemed a bit of a mess and mixing classification and AuthType checking, splitting them out so that ResourceType could be matched in preference of classifications meant that it was only checking for mobile hence it being renamed to maybe_authorize_mobile, in-line with all the other chained checks.
3. Noted.

The aim here was to allow billing to be bypassed for inter-account calling that would otherwise match external call routing. I might be missing your point, but the resource_type infers direction, as far as I understand. The resource type is either (onnet|offnet)-(origination|termination),  So we could have {"ignore_resource_types": {"onnet-termination": true }} - would that be better?

lazedo:
true for `direction` missed that.
`{"ignore_resource_types": {"onnet-termination": true }} - would that be better?` its better, but we need a better naming, we are replying `i authorized this as being a trunk without actually consuming one` and it should be something like `{..., "whatever_name_comes_out_of_this": ["onnet-termination"], ...}` (list of resource types)

`maybe_authorize_exception`, we're trying to move away from chained functions when it comes to scope and use a routines list to be folded instead.

bputtick:
I agree the object->array_list is a better model. As for the naming of the key, I am open to suggestions?

I had just adapted the coding style with the rest of the source file; which is all chained - can I have an example of a routine list you would like me to replicate please?

Forgive me, I read my first line of Erlang 2 weeks ago.

bputtick:
Perhaps invert the logic to match resource type?

jamesaimonetti:
To clarify point 2 (and @lazedo can clarify further), a list of routines might look like https://github.com/2600hz/kazoo/blob/master/applications/callflow/src/cf_route_win.erl#L212

I think in this case, you could define the list of checks being made for whether to authz and fold over the functions until one authorizes the call (or none do and the call remains unauthorized). If you want to take a stab at that great; if you need more explanation let us know too!

bputtick:
@jamesaimonetti great :)

The log levels for these tests were all debug before?

jamesaimonetti:
@bputtick merged manually. There were a couiple spec issues that Dialzyer complained about - took care of those. Great stuff!

PR #5483, opened 2019-02-03
if request voicemail transcribe, then recording in wav format 

sergey-safarov:
Google ASR us not support mp3 audio files.
For other engines  ASR over wav file will give more accurate results.

If requested voicemail transcribe, then
1) recording is made using wav format;
2) kazoo-app converts wav to mp3 and attach as secondary attachment;
3) kazoo-app try try transcribe voicemail, if result successful, then wav file is deleted.

For file conversion is used ffmpeg. Now test is fails. Think it because host is not have ffmpeg installed.

To implement this conversion is extended kazoo_convert for audio files. Now implemented wav to mp3 conversion.

For conversion is used ffmpeg. FFMPEG also supports video formats. May rename `kz_audio_converter.erl` to `kz_av_converter.erl`? AV is abbreviation of "audio video".


core/kazoo_documents/src/kzd_box_message.erl
```diff
@@ -120,23 +123,55 @@ new(AccountId, Props) ->
                               ),
     Description = props:get_value(<<"Description">>, Props, <<"voicemail message with media">>),
 
-    DocProps = [{<<"_id">>, MsgId}
+    DocProps = props:filter_undefined(
```
jamesaimonetti: The `filter_undefined` is done in `kz_json:from_list/1` so no need to add it here and recurse the list twice.

core/kazoo_documents/src/kzd_box_message.erl
```diff
@@ -120,23 +123,55 @@ new(AccountId, Props) ->
...
                                  ]
                                 ).
 
+%%------------------------------------------------------------------------------
+%% @doc
+%% @end
+%%------------------------------------------------------------------------------
+-spec gen_attachment_filename(kz_term:proplist()) -> kz_term:ne_binary().
+gen_attachment_filename(Props) ->
+    case props:get_value(<<"Transcribe-Voicemail">>, Props)
```
jamesaimonetti: I think using `props:is_true/2` in place of `get_value` would make this clearer

core/kazoo_documents/src/kzd_box_message.erl
```diff
@@ -120,23 +123,55 @@ new(AccountId, Props) ->
...
+        andalso props:is_defined(<<"Required-Format">>, Props)
+        andalso props:get_value(<<"Required-Format">>, Props) =/= <<"wav">> of
```
jamesaimonetti: I think instead of checking if the key is defined and that the value is not `wav`, you could accomplish the same thing with `props:get_ne_binary_value(<<"Required-Format">>, Props, <<"wav">>) =/= <<"wav">>`. If the key doesn't exist, `<<"wav">>` is returned and the check is false.

core/kazoo_voicemail/src/kvm_message.erl
```diff
@@ -87,13 +84,60 @@ new(Call, Options) ->
...
             end
     end.
 
+-spec maybe_convert_attachment(kapps_call:call(), kz_term:ne_binary(), integer(), kz_term:proplist()) -> {'ok', kapps_call:call()}.
+maybe_convert_attachment(Call, MediaId, Length, Options) ->
+    AccountId = kapps_call:account_id(Call),
+    Db = kvm_util:get_db(AccountId, MediaId),
+    {'ok', MediaDoc} = kz_datamgr:open_doc(Db, MediaId),
+    case kzd_box_message:transcribe_filename(MediaDoc) =/= 'undefined'
+        andalso convert_attachment(Db, MediaDoc) of
```
jamesaimonetti: Trying to move away from returning non-booleans on the right side of `andalso` expressions. I would prefer to see a function call to handle the clauses:

```erlang
    {'ok', MediaDoc} = kz_datamgr:open_cache_doc(Db, MediaId),
    maybe_convert_attachment(Call, MediaId, Length, Options, MediaDoc, kzd_box_message:transribe_filename(MediaDoc)).

maybe_convert_attachment(Call, MediaId, Length, Options, _MediaDoc, 'undefined') ->
    notify_and_update_meta(...);
maybe_convert_attachment(Call, MediaId, Length, Options, _MediaDoc, _Filename) ->
    convert_attachment(Call, MediaId, Length, Options, MediaDoc).

convert_attachment(Call, MediaId, Length, Options, MediaDoc) ->
    Db = kvm_util:get_db(kapps_call:account_id(Call), MediaId),
    Conversion = convert_attachment(Db, MediaDoc),
    handle_conversion(Call, MediaId, ...)
```

Anyway, hopefully that gets the idea across. Now you have a function for explicitly doing the conversion while the `maybe_` function doesn't do anything apart from testing conditions for conversion.

core/kazoo_voicemail/src/kvm_util.erl
```diff
@@ -370,6 +370,16 @@ maybe_transcribe(AccountId, MediaId, 'true') ->
     end;
 maybe_transcribe(_, _, 'false') -> 'undefined'.
 
+-spec transcribe_filename(kz_json:object()) -> kz_term:amy_ne_binary().
+transcribe_filename(MediaDoc) ->
+    TranscribeFilename = kzd_box_message:transcribe_filename(MediaDoc),
+    case TranscribeFilename =:= 'undefined'
+        andalso  kz_doc:attachment_names(MediaDoc) of
```
jamesaimonetti: Same thing here about not using the right side of `andalso` to return non-booleans

core/kazoo_voicemail/src/kvm_util.erl
```diff
@@ -388,6 +399,16 @@ maybe_transcribe(Db, MediaDoc, Bin, ContentType) ->
             'undefined'
     end.
 
+-spec maybe_delete_transcribe_attachment(kz_term:ne_binary(), kz_json:object()) -> 'ok' | {'error', any()}.
+maybe_delete_transcribe_attachment(Db, MediaDoc) ->
+    Filename = kzd_box_message:transcribe_filename(MediaDoc),
+    case Filename =/= 'undefined'
+        andalso kz_datamgr:delete_attachment(Db, kzd_box_message:get_msg_id(MediaDoc), Filename) of
```
jamesaimonetti: Same thing here about non-booleans

jamesaimonetti:
@sergey-safarov I think you will need to edit `.circleci/config.yml` to install ffmpeg. You can add it to the apt-get install line here: https://github.com/2600hz/kazoo/blob/master/.circleci/config.yml#L90

sergey-safarov:
FFMPEG also supports video formats.
May rename `kz_audio_converter.erl` to `kz_av_converter.erl`? AV is abbreviation of "audio video".

sergey-safarov:
This PR update is finished.

sergey-safarov:
Thank Luis @lazedo for revew
> the transcribe provider should invoke the convert library to obtain a desired format, if the format is already `wav` in the case of google asr a conversion is not needed and the attachment will be returned as saved.

If voicemail box settings is defines use mp3, than transcribe result will be not good for Google ASR. Because this ASR engine not works with mp3 compressed audio.
For human ear uncompressed audio and `mp3` compressed audio looks as same.But for Google ASR engine this will be a trick.

sergey-safarov:
Also little statistic info
1) voicemail related files changed - 4
2) converted related files added/changed - 14.

May split this to two different PR?
First PR to converter. And when this accepted then we can return to voicemail related files update.

bazemoreb:
Hey @lazedo - so yeah, as Sergey mentioned, basically the issue is that to transcribe properly we have to take a WAV file that has been recorded as one. If we convert from MP3 -> WAV and then transcribe it is terrible to transcribe.

lazedo:
@bazemoreb i understand, configure the extension to `wav` and that's it, no need for these changes

bazemoreb:
Right, the issue is that if a client enables transcription but forgets to enable WAV, the email as a whole will fail.  @sergey-safarov  - do we have a way of just fixing that, and also of somehow setting the default to WAV when transcription is enabled for the best experience?

sergey-safarov:
Yes, Brandon I can add this check.
Luis, is need to add such behaviour for Google ASR only? Or for all ASR engines?

sergey-safarov:
Created PR https://github.com/2600hz/kazoo/pull/5529.
This is paused until new PR is merged.

k-anderson:
@sergey-safarov PR #5529 has been merged :)

sergey-safarov:
Hi guys
This PR contains audio files converter. Also same converted may be used as video converter.
Are you want use this functionality in future? Or need to drop this PR?

sergey-safarov:
From `ci/circleci:release` log is not obviously what test is failed.
I will pause fixing test after rebase on current master.
Will continue after confirmation of feature requirements.

jamesaimonetti:
@sergey-safarov want to check in and see what the status is here? Can you rebase and push the updated branch for CI to run?

sergey-safarov:
Hi @jamesaimonetti 
PR is rebased about 3 hours ago.
All test is completed except Travis. I looked other PRs and found many with not completed Travis tests.

lazedo:
@sergey-safarov @bazemoreb hello, what do you want to do with this pr ?

sergey-safarov:
Here implemented converter using `ffmpeg`.
This converter may be used for audio and video conversion between may media types.

So need reopen PR with only this converter.
I close this PR and open some time later

lazedo:
thanks @sergey-safarov

PR #5484, opened 2019-02-04
PISTON-675: add support for extra headers in ecallmgr_util:send_cmd

danielfinke:
Also add hold-bleg: true for ecallmgr_fs_transfer:attended/3 so hold
music works


lazedo:
@danielfinke hey, the changes to ecallmgr_call_command and ecallmgr_util are `ok`. can you split the commits and move them to a separate PR ? the changes to fs_transfer needs debating if we should always set or allow the dialplan to send that option (event if defaulted to true), currently it doesn't set so some users may actually find that a feature

danielfinke:
@lazedo OK, I removed the ecallmgr_fs_transfer change from this PR. Will put it in as a separate PR

jamesaimonetti:
merged

PR #5495, opened 2019-02-06
[4.3] HELP-41845: generate CSVs providing a consistent cell ordering

jamesaimonetti:
Rather than ouputting rows in an ad-hoc manner, apply a consistent
ordering of JSON paths to cells when processing JSON objects that may
vary in the paths present.


applications/crossbar/src/api_resource.erl
```diff
@@ -1071,28 +1084,33 @@ process_chunk(#{context := Context
                }=ChunkMap) ->
     case cb_context:resp_data(Context) of
         0 ->
+            lager:debug("resp data is 0, resetting"),
```
icehess: Can you say something like
```erlang
lager:debug("(chunked) ~s did not send any resp data", [_ToFun]),
```

Same for log lines below too, add `(chunked)` and how many row/data was sent instead.

PR #5496, opened 2019-02-06
HELP-41845: generate CSVs providing a consistent cell ordering

jamesaimonetti:
Rather than ouputting rows in an ad-hoc manner, apply a consistent
ordering of JSON paths to cells when processing JSON objects that may
vary in the paths present.


jamesaimonetti:
See 4.3 in #5495

icehess:
CDRs is generating the CSV itself and it is not using `kz_csv`. So `api_resource/api_util` still need to handle the case which the csv is generate by the module and should stream the binary as is.

I think binary response data was used to be handled in `api_util:csv_body/2`.

icehess:
CDRs is still broken

```
01:13:27.649 [error] |0000000000|Undefined:Undefined(<0.3029.0>) Lager event handler error_logger_lager_h exited with reason {'EXIT',{{case_clause,[api_resource,<0.22048.0>,1,<0.22049.0>,function_clause,[{kz_json,flatten,[<<"id,call_id,caller_id_number,caller_id_name,callee_id_number,callee_id_name,duration_seconds,billing_seconds,timestamp,hangup_cause,other_leg_call_id,owner_id,to,from,direction,request,authorizing_id,cost,dialed_number,calling_from,datetime,unix_timestamp,rfc_1036,iso_8601,iso_8601_combined,call_type,rate,rate_name,bridge_id,recording_url,media_recordings,media_server,call_priority,interaction...">>],...},...]]},...}}
```

jamesaimonetti:
@icehess Can you provide the tests that caused the error you posted? Hard to know what your commit is trying to fix without seeing the erroring test.

icehess:
> @icehess Can you provide the tests that caused the error you posted? Hard to know what your commit is trying to fix without seeing the erroring test.

This the command I used to download CDRs. I set page_size to 3 and chunk_size to 1 so I expect to receive each row in one chunk. The `--raw` option in curl is printing the result of each chunk response.

```bash
curl -s --raw -H "Content-Type: application/json" -H "Accept: application/json" --basic --user "582a712a47bf254d56b0feab130f6bfd:`echo -n admin:admin | md5sum | awk '{print $1}'`" "`dockpen apps 8000 url`/v2/accounts/27e11003227e820ff58ea37e35de2e8b/cdrs/interaction?page_size=3&chunk_size=1&accept=text/csv"
```

icehess:
Btw I found out that `pqc_cb_rates` is failing.

PR #5498, opened 2019-02-07
PISTON-677: look for ignore_xref like rebar3 does

danielfinke:
`rebar xref` supports the 'ignore_xref' attribute on modules. Since the root Makefile uses the check-xref escript, it doesn't handle the attribute. Add support in this PR


jamesaimonetti:
@danielfinke merged, thanks

PR #5501, opened 2019-02-07
PISTON-625: do not pass pids as start_link args. They get re-used on restart

danielfinke:


jamesaimonetti:
Merged, thanks @danielfinke

PR #5502, opened 2019-02-09
PISTON-609: send registration notifications from local zone only

danielfinke:


jamesaimonetti:
@lazedo i think it is in reference to [this](https://forums.2600hz.com/forums/topic/10142-oldest-registrar-handling-registerderegister-notifications/?tab=comments#comment-56429)

lazedo:
@jamesaimonetti thanks.
@danielfinke what happens if there are no `ecallmgr` nodes available in the zone ? from the forum topic, seems odd that the databases are not available in one zone, what happens if the all DBs in the zone goes down ? i think your setup is defeating the goal of segregated zones that operate independently but with a common database. but other may have different opinions.
a `must` have for this PR (if @jamesaimonetti and @k-anderson agree) would be to `opt-in` for the reasons above

danielfinke:
@lazedo you raise a good point about not having any ecallmgr nodes in the zone that the registration took place in. Though, is it likely that you'd ever have Kamailio registrars running that register to another zone (i.e. that zone maybe doesn't have any kapps at all)?
What I will do to handle this anyway is first check if there are ecallmgr nodes in the zone of the reg_success. If so, retain the behaviour in this PR. If not, retain the original behaviour.

Yes, we can have problems where the DBs in an account's zone go down. But it's pretty much a requirement to rely on replication only at this point. You can't cluster databases across the WAN due to latency, failure to meet quorum, etc.

danielfinke:
Those changes have been made to keep more in line with the original behaviour

danielfinke:
@lazedo @jamesaimonetti hey guys, this one's been around for a while. Are you guys happy with the approach now?

jamesaimonetti:
@lazedo anything left to do here?

PR #5503, opened 2019-02-09
system storage api update

lazedo:
* validates attachment handlers once
* probes system attachment handlers
* changes caching of dataplans to avoid concurrency load

### DIFF
--- a/core/kazoo_attachments/src/kz_att_onedrive.erl
+++ b/core/kazoo_attachments/src/kz_att_onedrive.erl
@@ -17,6 +17,9 @@
 
 -define(GRAPH_HTTP_OPTIONS, [{'version', "HTTP/1.1"}
                             ,{'ssl', [{'versions', ['tlsv1.2']}]}
+                            ,{'follow_redirect', 'true'}
+                            ,{'max_redirect', 5}
+                            ,'with_body'
                             ]).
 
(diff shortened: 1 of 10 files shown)


core/kazoo_attachments/src/kz_att_onedrive.erl
```diff
@@ -143,8 +146,10 @@ do_fetch_attachment({'ok', #{'token' := #{'authorization' := Authorization}}},
                     DriveId, ContentId, HandlerProps, DbName, DocId, AName) ->
     Headers = [{<<"Authorization">>, Authorization}],
     Url = ?DRV_FETCH_URL(DriveId, ContentId),
-    lager:debug("getting ~p", [Url]),
-    case kz_http:get(Url, Headers, ?GRAPH_HTTP_OPTIONS) of
+    lager:debug("getting ~p : ~p", [Url, Headers]),
+    %% case kz_http:get(Url, Headers, ?GRAPH_HTTP_OPTIONS) of
+    %% SEEMS HTTPC IS BROKEN IN OTP 21.2.5
```
jamesaimonetti: What's the issue you found?

lazedo: not conclusive, but it fails consistently, needs further debugging, (maybe related to size of header content , url encoding, [insert yours here]). same thing in `aws` which was using `httpc` by default and the upgrade of `erlcloud` worked well

PR #5513, opened 2019-02-14
HELP-41542: improved interaction_id queries

jamesaimonetti:


applications/crossbar/src/modules/cb_cdrs.erl
```diff
@@ -587,25 +604,26 @@ load_legs(<<Year:4/binary, Month:2/binary, "-", _/binary>> = DocId, Context) ->
...
+              ,{'range_keymap', 'nil'}
+              ,{'databases', [MODB]}
+              ,'include_docs'
               ],
-    crossbar_doc:load_view(?CB_INTERACTION_LIST_BY_ID
-                          ,Options
-                          ,Context
-                          ,fun normalize_leg_view_results/2
-                          ).
+    crossbar_view:load_modb(Context, ?CB_INTERACTION_LIST_BY_ID, Options).
```
icehess: @jamesaimonetti can you add a clause to handle invalid interaction id?

applications/crossbar/src/modules/cb_cdrs.erl
```diff
@@ -587,25 +604,26 @@ load_legs(<<Year:4/binary, Month:2/binary, "-", _/binary>> = DocId, Context) ->
...
-    Options = ['include_docs'
-              ,{'startkey', [InteractionId]}
-              ,{'endkey', [InteractionId, kz_json:new()]}
+load_legs(<<BinTimestamp:11/binary, "-", _Key/binary>>=InteractionId, Context) ->
+    MODB = kazoo_modb:get_modb(cb_context:account_id(Context), kz_term:to_integer(BinTimestamp)),
+
+    Options = [{'mapper', fun normalize_leg_view_results/2}
+              ,{'range_start_keymap', [InteractionId]}
+              ,{'range_endkey_keymap', [InteractionId, kz_json:new()]}
+              ,{'range_keymap', 'nil'}
```
icehess: why setting this ti `nil` ? it makes start/end key `undefined`

jamesaimonetti: If `nil`:

```
20:58:09.317 [debug] |63c87d62d3-63717397089308|crossbar_view:744(<0.5491.0>) kz_datamgr:get_results(<<"account%2F45%2F47%2F99e78c9b6ed6f21a6377909c9b52-201902">>, <<"interactions/interaction_listing_by_id">>, [{limit,51},descending,{range_endkey_keymap,[<<"63717397088-bb8a5dba">>,{[]}]},include_docs]) 
```

With no `nil`:

```
21:08:13.826 [debug] |ad8265db7d-63717397693817|crossbar_view:744(<0.7750.0>) kz_datamgr:get_results(<<"account%2F35%2Fac%2Fc29caf10a6a4354c7eae781d0460-201902">>, <<"interactions/interaction_listing_by_id">>, [{limit,51},{startkey,63717397693},{endkey,[<<"63717397693-05f46540">>,63714715693]},descending,{range_endkey_keymap,[<<"63717397693-05f46540">>,{[]}]},include_docs])
```

Notice that startkey becomes a timestamp which borks everything.

icehess: Okay, so if you want to not add timestamp pass the start/end keymap as a function arity 1 like this:

```erlang
load_legs(<<BinTimestamp:11/binary, "-", _Key/binary>>=InteractionId, Context) ->
    MODB = kazoo_modb:get_modb(cb_context:account_id(Context), kz_term:to_integer(BinTimestamp)),

    Options = [{'mapper', fun normalize_leg_view_results/2}
              ,{'range_start_keymap',fun(_) -> [InteractionId] end}
              ,{'range_end_keymap', fun(_) -> [InteractionId, kz_json:new()] end}
              ,{'databases', [MODB]}
              ,'include_docs'
              ],
    crossbar_view:load_modb(Context, ?CB_INTERACTION_LIST_BY_ID, Options).
```

Notice the name of the key is `range_end_keymap`

icehess:
@jamesaimonetti When you are here can you please add `no_filter` to `CB_SPECIFIC_VIEW_OPTIONS` macro in `crossbar_view`

applications/crossbar/src/modules/cb_cdrs.erl
jamesaimonetti: Seems like a convulated way to make this happen. I tried just setting startkey/endkey but those got overwritten. Why doesn't crossbar_view let me explicitly set start/end? This all feels too clever for its own good.

icehess: There was a lot of mistakes setting start/end key in the past times, every module developed their own function which most of the was not working at all.

I gathered all logic needed from all modules in here so the keys are set properly for range query.

But this was my opinion, feel free to change if you like.

jamesaimonetti: I think the logic is needed for more complex interactions but the simple case has been lost. The spec for keymap() allows for the simple list but then surprises the user by appending the timestamp. There is no "range_startkey", only "range_start_keymap" so that's what I used.

What I include `startkey` and `endkey` without range map stuff:

```
kz_datamgr:get_results(<<"account%2F0b%2F8a%2F766cd57b545dee927b33d1453648-201902">>, <<"interactions/interaction_listing_by_id">>, [{limit,51},{startkey,63717406736},{endkey,63714724736},descending,include_docs])
```
The keys are replaced with the scalar timestamp.

So here we are, needing anonymous functions to feed static list of startkey/endkey.

applications/crossbar/src/crossbar_view.erl
```diff
@@ -34,7 +34,7 @@
 -include("crossbar.hrl").
 
 -define(CB_SPECIFIC_VIEW_OPTIONS,
-        ['ascending', 'databases', 'mapper'
+        ['ascending', 'databases', 'mapper', 'filter'
```
icehess: `s/filter/no_filter/`

applications/crossbar/src/modules/cb_cdrs.erl
icehess: Well we can get first `startkey/endkey` like how non-range query is implemented.

But my concern is that people start passing those and they won't implement the descending or other range query stuff, they don't test it and no one is noticing or testing so it breaks the query.

And don't forget `keymap()` is for `load` (non-range query) and `range_keymap()` for `load_range` and `load_modb`. And most range views have timestamp at the end, and crossbar view is considering all options and query string values to consider how to get timestamp and how to create start/end key in what direction and etc...

So yes I agree, there is a lot cleverness and surprises happening here. I don't like it, but it makes life easier if you understand it, and I don't have any idea to how avoid these surprises and also avoid make this thing more complex, especially with chunking stuff too.

PR #5529, opened 2019-02-19
When set transcribe, then force used wav media_extension

sergey-safarov:


lazedo:
@sergey-safarov as mentioned before, the changes to cf_voicemail are not desired. the changes to cb_vmboxes are ok (minor spelling) but they should be `opt-in` and the extension should be configurable and not static `wav`, you can use the same configuration field for opt-in and extension, `value` /= `undefined` => `opt-in`

sergey-safarov:
Thanks Luis @lazedo 
`cf_voicemail` changes is reverted.
transcribe extension check also now optional and configurable.

lazedo:
@sergey-safarov thanks, can you please squash your commits ? thanks

sergey-safarov:
@lazedo commits is squashed

PR #5532, opened 2019-02-20
Use value from prompt results to determine attachments

aisconn:
Fixing 500 error returned from prompts API when loading an single existing prompt

### DIFF
--- a/applications/crossbar/src/modules/cb_media.erl
+++ b/applications/crossbar/src/modules/cb_media.erl
@@ -675,7 +675,7 @@ load_media_docs_by_prompt(Context, PromptId, 'undefined') ->
                             ,[{'startkey_fun', fun(Ctx) -> prompt_start_key(Ctx, PromptId) end}
                              ,{'endkey', [PromptId, kz_json:new()]}
                              ,{'reduce', 'false'}
-                             ,{'include_docs', 'false'}
+                             ,'include_docs'
                              ]
                             ,cb_context:set_account_db(Context, ?KZ_MEDIA_DB)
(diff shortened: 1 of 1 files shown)


jamesaimonetti:
@aisconn Hi! I'm not sure this change is acceptable as-is. Because the view query uses `include_docs` the `doc` of the view results will include (if present) the `_attachments` key  which is what `kz_doc:attachments/1` uses to fetch existing attachments. Because the code calls the `/1` version, `undefined` is returned if there are no attachments which allows setting the `has_attachments` flag.

Can you share the steps to reproduce the 500 error you're attempting to fix as well as the logs of the failing query?

aisconn:
@jamesaimonetti Ye sure. Sorry I didn't give more detail initially. 
I get 500 when using GET /v2/accounts/{ACCOUNT_ID}/media/prompts/{PROMPT_ID} for prompts I know have attachments. I traced the issue to the normalize_prompt_results/2.

Looking at the options set for the view query in load_media_docs_by_prompt/3 functions, 'include_docs' is 'false'. And so when the results reach the filter function, there is no <<"doc">> to access in the JObj.

Should the correct fix be to set 'include_docs' to 'true' for the view query?

If you still think this should not be changed, I can reproduce the issue and include log output. Just let me know

jamesaimonetti:
@aisconn you're right on the money with the `include_docs` - I improperly assumed since normalize was getting `<<"doc">>`. I think you are right to change `include_docs` to be `true` (or just the atom instead of the tuple) for this case.

PR #5535, opened 2019-02-21
WIP HELP-42311-Master: get a db  list for all current kzd_connections

mk1s:
It was reported that `migrate/0`, `migrate/1` were only returning the system databases when configured to use multiple bigcouch clusters for storage.

Currently `db_info/0` only returns a listing of the databases of the system bigcouch connection.

This change now grabs all current kz_dataconnections and merges their databases into a single listing.


core/kazoo_data/src/kz_datamgr.erl
```diff
@@ -345,7 +345,26 @@ db_exists_all(DbName) ->
...
-    kzs_db:db_info(kzs_plan:plan()).
+    Connections = get_connections(),
+    Databases = build_db_list(Connections),
+    {'ok', Databases}.
+
+-spec get_connections() -> map().
+get_connections() ->
+    [#{tag => Tag, server => kz_dataconnections:get_server(Tag)}
+     || {_, _, _, _, _, _, _, Tag}
+            <- ets:tab2list('kz_dataconnections')].
```
jamesaimonetti: Drop the trailing `]` down

core/kazoo_data/src/kz_datamgr.erl
```diff
@@ -345,7 +345,26 @@ db_exists_all(DbName) ->
...
+
+-spec build_db_list(map()) -> kz_term:ne_binaries().
```
jamesaimonetti: `Connections` isn't a `map()` it is `[map()]`

core/kazoo_data/src/kz_datamgr.erl
jamesaimonetti: I feel like this should be a call to kz_dataconnections instead of a read from the ETS table.

mk1s: Ignore this for now, this was a poc to discuss and not the acutal implementation.

 I've renamed with WIP for now and will update when its ready for review.

jamesaimonetti: Meaning `kz_datamgr` doesn't need to know about the implementation details of how connections are tracked (via ETS or in a gen_server or whatever). `kz_datamgr` should ask kz_dataconnections for a list of active ones and go from there.

Also, have you tested with a storage plan added? Would be a good test to make sure it isn't included in the listing or crash the call somehow.

jamesaimonetti: OK thanks

icehess:
Hi @mk1s. please rebase to master to fix the circelci issue. Also there is a conflict on `core/kazoo_data/src/kzs_plan.erl`.

PR #5543, opened 2019-02-23
Update users.md

spditner:
Add in object details for the do-not-disturb setting


jamesaimonetti:
Hi @spditner the markdown tables are auto-generated from the schema files (in this case the [users](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/priv/couchdb/schemas/users.json) schema. Since the `do_not_disturb` key is defined in the users.json the table contains both `do_not_disturb` and `do_not_disturb.enabled` (since that's the only sub-property defined in the schema).

The definitions are when a schema references a sub-schema with `$ref` - we build a separate table and generate that link.

So if you want to propose breaking out the dnd properties into a separate table, you'll need to create a dnd schema and `$ref` it in the `users` schema.

Let us know what you want to do; I'm not opposed to the change but as-is this PR won't pass CI checks or be merged.

jamesaimonetti:
@spditner just wanted to check in and see if you had any thoughts or questions on what I posted earlier?

jamesaimonetti:
@spditner feel free to open a new PR if you want to make the changes suggested. I'll close this one for now.

PR #5557, opened 2019-03-06
[4.3] help-42525: fixed faxbox not handling unset of parameters when saving…

swysor:
… to faxes db


applications/crossbar/src/modules/cb_faxboxes.erl
```diff
@@ -548,15 +548,11 @@ faxbox_doc_save(Context) ->
                                ),
 
     DocId = kz_doc:id(ToSave),
-    Ctx3 = crossbar_doc:update(cb_context:set_account_db(Ctx2, ?KZ_FAXES_DB)
-                              ,DocId
-                              ,kz_json:to_proplist(kz_json:flatten(ToSave))
-                              ),
-
-    case cb_context:resp_status(Ctx3) of
```
swysor: Should I still do this part?

icehess:
master: https://github.com/2600hz/kazoo/pull/5556

PR #5558, opened 2019-03-07
KAZOO-6058: expand good standing checks and item options

k-anderson:


k-anderson:
Master PR #5559

icehess:
also you need to rebase to fix circelci

PR #5564, opened 2019-03-08
PISTON-695: shared ringback set btwn ecallmgr_fs_transfer/bridge

danielfinke:
Share ringback set command between ecallmgr_fs_transfer and ecallmgr_fs_bridge so that ecallmgr_fs_transfer:attended will have a ringback even when the transferring leg did not have the ringback variables set yet. It seems that channels originated using kapi_resource:publish_originate_req/1 to predefined endpoints will not have that var set because it is only set by either kapps_call_command:bridge or the XML dialplan

### DIFF
--- a/applications/ecallmgr/src/ecallmgr_fs_transfer.erl
+++ b/applications/ecallmgr/src/ecallmgr_fs_transfer.erl
@@ -22,6 +22,8 @@ attended(Node, UUID, JObj) ->
 
     ReqURI = <<TransferTo/binary, "@", Realm/binary>>,
 
+    ecallmgr_call_command:set_ringback(Node, UUID, JObj),
+
     Vars = [{<<"Ignore-Early-Media">>, <<"ring_ready">>}
            ,{<<"Simplify-Loopback">>, <<"false">>}
            ,{<<"Loopback-Bowout">>, <<"true">>}
--- a/applications/ecallmgr/src/ecallmgr_call_command.erl
+++ b/applications/ecallmgr/src/ecallmgr_call_command.erl
@@ -9,6 +9,7 @@
 
 -export([exec_cmd/4]).
 -export([fetch_dialplan/4]).
+-export([set_ringback/3]).
 
 -ifdef(TEST).
 -export([get_conference_flags/1
@@ -1747,3 +1748,20 @@ add_page_conference_app(Dialplan, ConferenceName) ->
     [{"application", <<"conference ", ConferenceName/binary>>}
      | Dialplan
     ].
+
+-spec set_ringback(atom(), kz_term:ne_binary(), kz_json:object()) -> 'ok'.
+set_ringback(Node, UUID, JObj) ->
+    case kz_json:get_first_defined([<<"Ringback">>
+                                   ,[<<"Custom-Channel-Vars">>, <<"Ringback">>]
+                                   ]
+                                  ,JObj
+                                  )
+    of
+        'undefined' ->
+            {'ok', Default} = ecallmgr_util:get_setting(<<"default_ringback">>),
+            ecallmgr_fs_command:set(Node, UUID, [{<<"ringback">>, kz_term:to_binary(Default)}]);
+        Media ->
+            Stream = ecallmgr_util:media_path(Media, 'extant', UUID, JObj),
+            lager:debug("bridge has custom ringback: ~s", [Stream]),
+            ecallmgr_fs_command:set(Node, UUID, [{<<"ringback">>, Stream}])
+    end.
--- a/applications/ecallmgr/src/ecallmgr_fs_bridge.erl
+++ b/applications/ecallmgr/src/ecallmgr_fs_bridge.erl
@@ -34,7 +34,7 @@ call_command(Node, UUID, JObj) ->
 
             {'ok', Channel} = ecallmgr_fs_channel:fetch(UUID, 'record'),
 
-            _ = handle_ringback(Node, UUID, JObj),
+            ecallmgr_call_command:set_ringback(Node, UUID, JObj),
             _ = maybe_early_media(Node, UUID, JObj, Endpoints),
             _ = maybe_b_leg_events(Node, UUID, JObj),
             BridgeJObj = add_endpoints_channel_actions(Node, UUID, JObj),
@@ -79,23 +79,6 @@ unbridge(UUID, JObj) ->
 %% @doc Bridge command helpers
 %% @end
 %%------------------------------------------------------------------------------
--spec handle_ringback(atom(), kz_term:ne_binary(), kz_json:object()) -> 'ok'.
-handle_ringback(Node, UUID, JObj) ->
-    case kz_json:get_first_defined([<<"Ringback">>
-                                   ,[<<"Custom-Channel-Vars">>, <<"Ringback">>]
-                                   ]
-                                  ,JObj
-                                  )
-    of
-        'undefined' ->
-            {'ok', Default} = ecallmgr_util:get_setting(<<"default_ringback">>),
-            ecallmgr_fs_command:set(Node, UUID, [{<<"ringback">>, kz_term:to_binary(Default)}]);
-        Media ->
-            Stream = ecallmgr_util:media_path(Media, 'extant', UUID, JObj),
-            lager:debug("bridge has custom ringback: ~s", [Stream]),
-            ecallmgr_fs_command:set(Node, UUID, [{<<"ringback">>, Stream}])
-    end.
-
 -spec maybe_early_media(atom(), kz_term:ne_binary(), kz_json:object(), kz_json:objects()) -> ecallmgr_util:send_cmd_ret().
 maybe_early_media(Node, UUID, JObj, Endpoints) ->
     Separator = ecallmgr_util:get_dial_separator(JObj, Endpoints),


jamesaimonetti:
@danielfinke doesn't loopback-b set it during the route_reply work?

applications/ecallmgr/src/ecallmgr_fs_transfer.erl
```diff
@@ -22,6 +22,8 @@ attended(Node, UUID, JObj) ->
 
     ReqURI = <<TransferTo/binary, "@", Realm/binary>>,
 
+    ecallmgr_call_command:set_ringback(Node, UUID, JObj),
```
k-anderson: @danielfinke code looks good but I am wondering... prior to this change if the ringback was set on during the bridge or early in the call it would persist through the transfer.  As implemented if the ringback is not present on the AMQP request the ringback will be reset to default, removing the potential for any existing ringback on the channel to persist.  That is to say, with this change during a non-SIP transfer ringback is set or reset but can't be maintained.

The same argument could be made with regards to how we (and would continue after this PR) set ringback during bridge.  I would argue that during a bridge you are effectively initializing a channel and specifically with the endpoint parameters so I feel that would align with expectations.  I also feel a mid-call transfer would not intuitively be expected to reset an existing ringback unless explicitly requested.  Using another random parameter as an example, I would not expect the channel to reset its MOH preferences during a transfer unless explicitly part of the request.

Long story shorter, I am thinking that when calling set_ringback during transfer it should negate setting of the default if not provided on the AMQP request.  However, I suspect not setting the default might not correct your issue?

P.S>
However, I have identified why some of these PRs are tricky!  I was going to play with this on my dev box but I don't know exactly what your call scenario is.  I see that you describe the problem as originating a call to predefined endpoints (not 100% sure I understand that terminology, I am assuming that is on-net device as compared to external numbers) then transfer.  That doesn't provide a lot of specifics for me to reproduce the issue ;)  On all the PRs if there was one or more call scenarios that described what steps you take to produce the problem, what Kazoo currently does and what you are expecting that would help tremendously.  A lot of times we are trying to infer the purpose from code which leads to much back and forth ;)  Any time we are required to infer it means we can't quickly appreciate the full issue and impact of the change and so often we are under time constraints so move to the next with every intention to return when we have more time but that delays things.  Thats just the reality I am seeing on some of these, so the lower you can make that initial cognitive load by bootstraping us into your thought process the better! :)

danielfinke: @k-anderson OK, let me try to address the purpose a little more clearly.

With regards to "predefined" endpoints, I mean those that are provided, for example, by kz_endpoints:by_owner_id/3 or defined as a loopback, like:
```
                   [{<<"Invite-Format">>, <<"loopback">>}
                   ,{<<"Route">>,  Extension}
                   ,{<<"To-DID">>, Extension}
                   ,{<<"To-Realm">>, kapps_call:account_realm(Call)}
                   ,{<<"Custom-Channel-Vars">>, kz_json:from_list(CCVs)}
                   ])),
```
or similar.

The situation that has the issue I am trying to address is the loopback case. I have not confirmed if it is present with a list of kz_endpoints, but I don't think it would be.

The scenario can be set up like this:

1. Place a call C1 from device D1 to a callflow F1 that answers the call and puts it on hold.
2. Originate a call C2 to device D2 using kapi_resource:publish_originate_req/1, with Resource-Type: originate and Application-Name: bridge (so, a bridge done without kapps_call_command:bridge, such as you would do in the case of call queues because the bridge has to wait a while). Now, you have a leg for D2 that did not have its ringback set in the route_resp (which is what @lazedo was talking about) because it was originated by a list of endpoints, not through callflows, nor was it set by ecallmgr_fs_bridge (because kapps_call_command:bridge wasn't used). D2 has nothing in its FS vars for ringback.
3. D2 executes an ecallmgr_fs_transfer:attended/3 via whatever means to callflow F2. FreeSWITCH att_xfer performs a switch_ivr_originate a executes the loopback provided by the attended fun, resulting in loopbacks A & B and legs to any devices or offnet resources that are bridged to during callflow F2 (let's call those Cx).
4. The agent is getting its audio from C2. That call will not be bridged to the loopbacks (if doing bowout) nor any of Cx, until the far end answers. So, even if the ringback is set/exported during the callflow, it will not be set on C2 until the call is answered, which is too late, obviously.

So my solution was to set it on the call C2 based on settings passed into the attended xfer JObj. This puts the onus on the proc publishing the transfer to compute the desired ringback, which is imperfect, but it's something.

I didn't account, however, for the fact that there might already be ringback settings on the leg doing the attended xfer from any other scenario than my above one. So a modification would be to check somehow if it is already set on the UUID. If so, don't overwrite it. Or maybe that should also still be the responsibility of the proc publishing the transfer. If they send Ringback: undefined, they don't want to set it because it's already set.

I hoooope that is a lot clearer 🙂

danielfinke: @k-anderson hey, anything more I could answer to help out with this?

lazedo:
@danielfinke so why don't you change your originate_req / bridge referred in `2` ? i would welcome a PR that would accept dialplan vars in ecallmgr_originate if they're missing

danielfinke:
@lazedo there are a couple reasons why I have not made the change on the originate requestor side. First is that it means every originate requestor will have to compute the expected ringback setting. Second is related to the set_ringback function (previously ecallmgr_fs_bridge:handle_ringback). It checks ecallmgr_util:get_setting(<<"default_ringback">>) which isn't a behaviour that should be pulled out of ecallmgr into the originator I would think. Part of the problem here is that the ringback processing code is never executed for an originate to a list of endpoints. The originate string looks like this in FS:
```
originate({call_timeout=15,originate_timeout=15,origination_caller_id_number='1002',origination_caller_id_name='1002 Guy',ignore_early_media='true',ignore_display_updates='true',ecallmgr_Account-ID='63463bc16b7af744f88c05567219c839',ecallmgr_Authorizing-ID='0f9905c70943e36781c743e936530f47',ecallmgr_Call-Interaction-ID='63721616677-10a54fdf',ecallmgr_Ecallmgr-Node='kazoo_apps@kazoo41-1.dev',ecallmgr_Fetch-ID='46a68917325836e072957b002785d0a6',ecallmgr_Member-Call-ID='FibgmWNkhJO7oWyXXkcDmHbtoZWqejqa',ecallmgr_Original-Caller-ID-Name='1002 Guy',ecallmgr_Original-Caller-ID-Number='1002',ecallmgr_Request-ID='602ceee73c4e',ecallmgr_Retain-CID='true',effective_caller_id_number='1002',effective_caller_id_name='1002 Guy',local_var_clobber='true'}[^^!ecallmgr_Username='1001'!ecallmgr_Realm='realm.dev'!presence_id='1001@realm.dev'!origination_callee_id_number='2001'!origination_callee_id_name='1001 Guy'!origination_uuid='b7dd247c2e77eff53d0711c0e64b3cb8-0a1fde86ac5e801b57ae58510710b31f-f97bd7eb'!ignore_completed_elsewhere='false'!leg_timeout='15'!sip_h_X-KAZOO-AOR=sip:1001@realm.dev!sip_h_X-KAZOO-INVITE-FORMAT=username!ecallmgr_Authorizing-ID='b5741fa122467b6abff99876cca48106'!ecallmgr_Authorizing-Type='device'!ecallmgr_Owner-ID='0a1fde86ac5e801b57ae58510710b31f'!ecallmgr_Account-ID='63463bc16b7af744f88c05567219c839'!sdp_secure_savp_only='false'!sip_invite_domain='realm.dev'!absolute_codec_string='^^:PCMU:PCMA'!effective_callee_id_number='2001'!effective_callee_id_name='1001 Guy'!origination_uuid=b7dd247c2e77eff53d0711c0e64b3cb8-0a1fde86ac5e801b57ae58510710b31f-f97bd7eb]sofia/sipinterface_1/1001@realm.dev;fs_path=sip:192.168.99.100:5060  'set:intercept_unbridged_only=true,intercept:FibgmWNkhJO7oWyXXkcDmHbtoZWqejqa' inline )
```
Because this isn't a loopback, it's not going to get a route_resp, and because it's an originate, it doesn't use the ecallmgr_fs_bridge:handle_ringback behaviour. It would have had to have been determined by the originate requestor. But as I said it seems weird to pull that out of ecallmgr since it controls the default ringback (ecallmgr_util:get_setting(<<"default_ringback">>))

jamesaimonetti:
@lazedo any more feedback or objections?

lazedo:
@jamesaimonetti the same objections

jamesaimonetti:
@danielfinke any update here?

danielfinke:
@jamesaimonetti I have addressed why the proposed changes don't work to solve the problem

jamesaimonetti:
@danielfinke a rebase will be necessary here.

danielfinke:
@jamesaimonetti @lazedo noticed that the latest master depends on some new stuff in FS master. There appears to be a customized version of kz_att_xfer. Can either of you guys summarize what's been customized in that?
Sorry for the late follow-up. Just returned from vacation.

lazedo:
@danielfinke closing here. we don't want the share. if you have a problem with originate, lets talk after the breakening

PR #5567, opened 2019-03-11
PISTON-694: better arg join in ecallmgr_fs_transfer:attended/3

danielfinke:


jamesaimonetti:
I don't see a difference, performance wise, so if MakeBusy passes this time I'll merge.

danielfinke:
@jamesaimonetti sounds good! Do note there is a slight behaviour change (improvement). The old version would retain empty strings between commas (e.g. 'thing=value,,thing2=value2' for 'undefined' values such as in CID. So now those will be filtered out during the flatten

jamesaimonetti:
Merged!

PR #5570, opened 2019-03-12
[master] help-42655: fixed views getting saved to system_data with multilines …

swysor:
…in them

### DIFF
--- a/core/kazoo_data/src/kz_datamgr.erl
+++ b/core/kazoo_data/src/kz_datamgr.erl
@@ -1675,7 +1675,7 @@ maybe_register_view({<<"_design/", Name/binary>>, View}, App, {ClassId, ViewMaps
     log_register_views(Name, DocId, App, ViewMaps),
 
     Update = [{<<"kazoo">>, kz_json:from_list([{<<"view_map">>, ViewMaps}])}
-             ,{<<"view_definition">>, kz_json:delete_key(<<"kazoo">>, View)}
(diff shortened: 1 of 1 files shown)


core/kazoo_data/src/kz_datamgr.erl
```diff
@@ -1675,7 +1675,7 @@ maybe_register_view({<<"_design/", Name/binary>>, View}, App, {ClassId, ViewMaps
     log_register_views(Name, DocId, App, ViewMaps),
 
     Update = [{<<"kazoo">>, kz_json:from_list([{<<"view_map">>, ViewMaps}])}
-             ,{<<"view_definition">>, kz_json:delete_key(<<"kazoo">>, View)}
+             ,{<<"view_definition">>, maybe_adapt_multilines(kz_json:delete_key(<<"kazoo">>, View))}
```
icehess: This is not fixing root cause of the problem. This just register views.

Register view is reading the json file from applications folder on start up and save it to `system_data`. Then `kapps_maintenance:refresh_view` is calling `kz_datamgr:refresh_views` which eventually is calling `maybe_adapt_multilines` function.

icehess: ok, nvm. It is good to get ride of multiline anyway (although we lose the ability to quickly compare registered view with actual view, but it's fine).

And of course it doesn't hurt to `maybe_adapt_multiline` to `maybe_create_view`, right?

PR #5575, opened 2019-03-12
PISTON-692: clear push on device if reg'd without push params

danielfinke:


lazedo:
@danielfinke Hi, please provide a good context for this use case. this is really not something we want to do cause it will break some implementations we know of. you can clear the `push` with the api, just send a patch with push : null , does that work for you ?

danielfinke:
@lazedo we have a client who cannot rely on their users to "do something" e.g. send an API request, visit a URL, etc. to clear their push settings when they switch from registering on a push device to a traditional SIP device

lazedo:
@danielfinke shouldn't they be using another device ?  we have at least 2 cases where this would not work.

danielfinke:
@lazedo I don't disagree, but at the same time, we've run into this problem a number of times. It's not clear to users that they need to disable push to stop notifications from being sent to the other physical/push device. I'm wondering what your uses cases are that would be limited by this change?

core/kazoo_documents/src/kzd_devices.erl
```diff
@@ -650,6 +651,22 @@ provision_id(Doc, Default) ->
 set_provision_id(Doc, ProvisionId) ->
     kz_json:set_value([<<"provision">>, <<"id">>], ProvisionId, Doc).
 
+-spec push(doc()) -> kz_json:api_object().
+push(Doc) ->
```
jamesaimonetti: generally speaking, "public" fields like this (meaning no `pvt_`) should be defined in the associated schema. Even if the "push" object sub-properties aren't defined at least the key itself should be.

applications/pusher/src/pusher_listener.erl
```diff
@@ -67,50 +67,71 @@ handle_push(JObj, _Props) ->
...
+        _ ->
+            AccountDb = kz_util:format_account_db(AccountId),
+            {'ok', _} = kz_datamgr:save_doc(AccountDb, kzd_devices:delete_push(Doc)),
+            lager:debug("clearing push object for ~s: ~s", [AccountId, AuthorizingId])
+    end.
+
+-spec maybe_update_push_token(kz_term:ne_binary(), kz_term:ne_binary(), kz_json:object(), kz_json:object(), kz_term:proplist(), kz_json:api_object()) -> 'ok'.
+maybe_update_push_token(_, _, _, _, _, 'undefined') -> 'ok';
+maybe_update_push_token(AccountId, AuthorizingId, UA, JObj, Params, Doc) ->
+    Push = kz_json:get_value(<<"push">>, Doc),
```
jamesaimonetti: Use the accessor

danielfinke: hey. I just copy-pasted that part 😄
I'll make sure to update it

danielfinke:
@jamesaimonetti I have addressed your comments. Just wondering about the .src version of kzd_devices: should I be trying to add anything (e.g. delete_push/1) or only to the .erl file? Also it seems there is something in place that is lowercasing the keys in the autogenerated .src file. Suggestions of what to do with that file for this PR?

danielfinke:
@jamesaimonetti any thoughts about the recent changes & `kzd_devices.erl.src`?

jamesaimonetti:
@danielfinke the .erl.src of kazoo_document modules are generated from the JSON schema. Not sure what you are asking re: the lowercasing of keys? I think only the function names are lowercased while the JSON keys remain as-is. Any additional functionality should only be added to the .erl module; the .src is just a check for the developer/reviewer to see that new functionality exists and should be migrated to the .erl module.

jamesaimonetti:
Oh, I see in the CI errors re: case. Hmm, I think because it is unusual to have JSON schemas define "Foo-Bar" instead of "foo_bar" the generator code probably just assumes (or normalizes to) "-" should be "_". You'll either need to figure out where the conversion is happening and address the generator OR rename the fields to "token_*" and map those in the pusher code to "Token-*".

lazedo:
@danielfinke we're not merging this. if you want to remove the `push` properties for a device, use   `patch` on `device` with `{"push" : null}`

PR #5576, opened 2019-03-12
KAZOO-6038: Fix cb_devices_v2.erl to allow mobile devices modification

harenson:


applications/crossbar/src/modules_v2/cb_devices_v2.erl
```diff
@@ -903,16 +903,28 @@ update_provision(Context) ->
...
+    NewMDN = get_mdn(cb_context:doc(Context)),
+    %% Since the next 2 values are only used in 1 place each, get their values only if
+    %% they are actually going to be used. Otherwise, no need to calculate them ;)
+    OldMDN = fun() -> kz_json:get_ne_value(?KEY_MDN, DbDoc) end,
+    IsSuperAdmin = fun() -> cb_context:is_superduper_admin(Context) end,
     case get_device_type(Context) =:= <<"mobile">>
-        andalso kz_term:is_not_empty(MDN)
+        andalso kz_term:is_not_empty(NewMDN)
+        andalso (kz_term:is_empty(DbDoc) %% Request is to create a device
+                 orelse (OldMDN() =/= NewMDN    %% mdn has been changed.
```
icehess: **right** hand side of `orelse` executes *only  if* the **left** hand side is `false`, so no need for anonymous function.

Also `cb_context:is_superduper_admin` is always set in `context` record and accessing it is always constants.

icehess:
This needs to backported to 4.3 too.

applications/crossbar/src/modules_v2/cb_devices_v2.erl
```diff
@@ -903,16 +903,28 @@ update_provision(Context) ->
     sync_sip_data(Context, 'false').
 
 %%------------------------------------------------------------------------------
-%% @doc
+%% @doc There are only 2 possible scenarios for adding/updating a mdm:
```
jamesaimonetti: s/mdm/MDN/

Couple places you have mdm instead of MDN

applications/crossbar/src/modules_v2/cb_devices_v2.erl
```diff
@@ -903,16 +903,28 @@ update_provision(Context) ->
...
 %% @end
 %%------------------------------------------------------------------------------
 -spec maybe_add_mdn(kz_term:api_binary(), cb_context:context()) -> cb_context:context().
 maybe_add_mdn(_DeviceId, Context) ->
-    MDN = get_mdn(cb_context:doc(Context)),
+    DbDoc = cb_context:fetch(Context, 'db_doc'),
+    NewMDN = get_mdn(cb_context:doc(Context)),
+    %% Since the next 2 values are only used in 1 place each, get their values only if
+    %% they are actually going to be used. Otherwise, no need to calculate them ;)
+    OldMDN = fun() -> kz_json:get_ne_value(?KEY_MDN, DbDoc) end,
```
jamesaimonetti: I agree with @icehess that the anonymous functions really don't help anything here.

What I would change is breaking up the `andalso` clauses into well-named functions.

jamesaimonetti:
@harenson please backport to 4.3

PR #5545, opened 2019-02-26
PISTON-688: add callee/caller ID to ecallmgr_fs_channel

danielfinke:

PR #5577, opened 2019-03-12
PISTON-688: add callee/caller ID to ecallmgr_fs_channel - 4.3

danielfinke:
Backporting https://github.com/2600hz/kazoo/pull/5545


lazedo:
@danielfinke this needs to be fully tested before accepted into 4.3

danielfinke:
@lazedo no problem! I just wanted to make sure it was up for consideration to be backported

PR #5578, opened 2019-03-12
[4.3] PORT-4: list agent ports

icehess:
master: #5509


icehess:
Please don't merge it yet, tests are (silently) failing.

PR #5580, opened 2019-03-13
[master] KAZOO-6045: more intuitive parameter names

k-anderson:


core/kazoo_services/src/kz_services_item.erl
```diff
@@ -380,7 +380,6 @@ public_json(Item) ->
           ,{<<"minimum">>, undefine_empty(minimum(Item))}
           ,{<<"maximum">>, undefine_empty(maximum(Item))}
           ,{<<"prorate">>, kzd_item_plan:prorate(item_plan(Item))}
-          ,{<<"unprorate">>, kzd_item_plan:unprorate(item_plan(Item))}
```
jamesaimonetti: Did you want to add `prorate_removals` here instead or just remove the tuple entirely?

PR #5583, opened 2019-03-14
PISTON-698: set originate timeout for ecallmgr_fs_transfer:attended/3

danielfinke:
Set a high timeout for att_xfer. By default the FS implementation of att_xfer will perform an originate that uses the current timeout channel vars on the call making the transfer. This can backfire if that timeout is lower than the desired timeout on the other side of the att_xfer loopback. So here we set it to a high value and rely on the bridges on the other side of the loopback to control the ringing timeout.

### DIFF
--- a/applications/ecallmgr/src/ecallmgr_fs_transfer.erl
+++ b/applications/ecallmgr/src/ecallmgr_fs_transfer.erl
@@ -31,6 +31,7 @@ attended(Node, UUID, JObj) ->
            ,{<<"Outbound-Caller-ID-Name">>, kz_json:get_ne_binary_value(<<"Caller-ID-Name">>, JObj)}
            ,{<<"Outbound-Callee-ID-Number">>, TransferTo}
            ,{<<"Outbound-Callee-ID-Name">>, TransferTo}
+           ,{<<"Timeout">>, 60}
            ],
 
     Props = add_transfer_ccvs_to_vars(CCVs, Vars),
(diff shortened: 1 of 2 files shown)


lazedo:
@danielfinke not seeing the relevance of this. can you give an example of this happening ? att_xfer uses loopback, is this early-media related in some way ?

danielfinke:
@lazedo 
Originate a call from D1 to D2 with a ring timeout of 10s.
Answer the call on D2.
Have D2 perform an attended transfer to some callflow that rings D3 but with a ring timeout of, say, 20s (anything higher than the 10s of the original call).
Don't answer on D3. You'll find the attended transfer will be cancelled after 10s instead of after the expected 20s. This is because FS's att_xfer performs a switch_ivr_originate and that uses the currently set ring timeout on the channel - 10s.

jamesaimonetti:
@lazedo any more comments or concerns about merging?

lazedo:
@jamesaimonetti this pr expects to solve an issue with `att_xfer ` which in my opinion is not correct. this is being addressed in another way using `scope` variables.

lazedo:
@jamesaimonetti but i'm not opposing to merging it, just that its going to be removed later

jamesaimonetti:
@danielfinke is this still useful to merge, considering @lazedo's comments on eventual replacement with `scope` variables?

danielfinke:
@jamesaimonetti I'm OK either way - we're using this internally to solve the problem for now

lazedo:
@danielfinke going to merge this `as-is` and will change to take the `Timeout` from the payload

PR #5584, opened 2019-03-14
PISTON-693: blackhole registration module

danielfinke:
For discussion.
Related to [https://forums.2600hz.com/forums/topic/10425-tracking-registration-state-events-registerderegister/](https://forums.2600hz.com/forums/topic/10425-tracking-registration-state-events-registerderegister/)

### DIFF
--- a/applications/blackhole/src/blackhole.hrl
+++ b/applications/blackhole/src/blackhole.hrl
@@ -16,6 +16,7 @@
                          ,'bh_object'
                          ,'bh_fax'
                          ,'bh_conference'
+                         ,'bh_registration'
                          ]).
 
 -define(COMMAND_MODULES, ['bh_events'
--- a/applications/blackhole/src/modules/bh_registration.erl
+++ b/applications/blackhole/src/modules/bh_registration.erl
@@ -1,0 +1,49 @@
+%%%-----------------------------------------------------------------------------
+%%% @copyright (C) 2019, Voxter Communications Inc.
+%%% @doc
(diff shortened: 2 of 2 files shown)


k-anderson:
@danielfinke I am very concerned on the scalability of this approach.  For some very large accounts (100s of devices) with 5 minute registrations this can be significant per second updates.  If multiple connections for that account were established that would be a major amplifier of a high rate of messages sent, potentially producing a lot large overhead for a use-case to detect when the registered device count changes... Unless I misunderstand the intention :)

Thinking about the issue and reading the forum discussion (which I will crosspost) I do think that moving the suppression logic into the handler is the best option.  While I agree with James that we want to protect the message bus, and keep it as lean/efficient as possible to maintain scalability and failover overhead, I feel all the alternatives to add this functionality are worse.  I also feel these events have potential uses for future applications.  Another point, I don't think majority of the devices have their deregistration suppressed so I don't believe a significant portion of registrations aren't published on loss.

One much larger approach could be some type of 'batching'. For example, every 5 seconds the registrar could publish a summary of changes...  This is primarily around avoiding a huge load spike / flood if there were connectivity or kamailio issues in a zone.  Not sure I like that approach but throwing it out there to see if it triggers other thoughts / discussions :)

lazedo:
@danielfinke closing this, but keep the branch as in 5.x it will surely be merged. as we need to move forward with the breakening on apps into separate repo, i'm going to close this for now.

PR #5585, opened 2019-03-14
KAZOO-6064: [master] ask bookkeepers for early running recurring charges

icehess:
Updating how  early bill (collect recurring charges 5 days before due date) is running.

It sets some maker document in current modb (the month before due date) and send amqp message to bookkeepers of invoices to collect money.

This marker also used to check before running bookkeeper on new month modb to prevent double charge.


core/kazoo_services/src/kz_services_recurring.erl
```diff
@@ -0,0 +1,379 @@
...
+%%%=============================================================================
+
+%%------------------------------------------------------------------------------
+%% @doc
+%% @end
+%%------------------------------------------------------------------------------
+-spec get_previous_month(kz_time:gregorian_seconds()) -> kz_time:date().
+get_previous_month(DueTimestamp) ->
+    {{DueYear, DueMonth, _}, _} = calendar:gregorian_seconds_to_datetime(DueTimestamp),
+    calendar:datetime_to_gregorian_seconds({kz_date:normalize({DueYear, DueMonth - 1,1}), {0, 0, 1}}).
```
jamesaimonetti: `calendar` returns gregorian seconds but the type (and expected return where called is a date 3-tuple). How does this work?

PR #5593, opened 2019-03-18
Voicemail Fast Forward and Rewind

noahmehl:
This is a first stab at integrating the uuid_fileman seek command in relation to call flows voicemail module to add the feature of fast forward and rewind of playback of voicemails.


noahmehl:
I also want to note that in our testing, even if we can get kazoo to issue the correct FS command: uuid_fileman $uuid seek:+/- value, the voicemail doesn't actually seek.  In my cursory testing, even issuing the command via fs_cli doesn't work either.  I haven't had a chance to test via vanilla FS build/configs.

grahamsnz:
@noahmehl Adding fast-forward and rewind to voicemail playback would be great. Some of our users have very long messages and have requested this enhancement. I personally prefer using a web interface to jump around voicemail messages. However, I know that a good number of users are traditional phone users that never use the web interface. These users want fast-forward and rewind available when using their phones to access and playback voicemail messages.

grahamsnz:
@noahmehl Adding fast-forward and rewind to voicemail playback would be great. Some of our users have very long messages and have requested this enhancement. I personally prefer using a web interface to jump around voicemail messages. However, I know that a good number of users are traditional phone users that never use the web interface. These users want fast-forward and rewind available when using their phones to access and playback voicemail messages.

applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -133,6 +146,8 @@
               ,prev = <<"4">> :: kz_term:ne_binary()
               ,next = <<"6">> :: kz_term:ne_binary()
               ,delete = <<"7">> :: kz_term:ne_binary()
+              ,rw = <<"5">> :: kz_term:ne_binary()
```
jamesaimonetti: we try not to abbreviate names - please use `rewind` and `fast_forward` (esp since `rw` makes me think file I/O `read/write`).

PR #5594, opened 2019-03-18
use Erlang/OTP 21 compatible stacktrace macro everywhere

icehess:


jamesaimonetti:
@icehess lol, i was working on this this weekend :) I have a script for auto-converting code since other apps in community/private may need conversion too.

icehess:
> @icehess lol, i was working on this this weekend :) I have a script for auto-converting code since other apps in community/private may need conversion too.

hehe, I was working on this on Friday, but didn't make a PR.
What do you think about `kz_util:log_stacktrace/1`. I personally don't like. Do you have suggestion?

icehess:
@jamesaimonetti Can you update scripts/generate-schemas.escript too?

```
$ ag stacktrace scripts/
scripts/generate-schemas.escript
23:            ST = [{M, _F, _A, Props}|_] = erlang:get_stacktrace(),
```

jamesaimonetti:
@lazedo can you check MakeBusy's RabbitMQ? Kazoo getting connection refused (but working fine locally). 

Also what will it take to get it building with OTP 21 instead?

make/deps.mk
```diff
@@ -53,62 +44,130 @@ ifeq ($(USER),travis)
...
+dep_lager_syslog = git https://github.com/erlang-lager/lager_syslog 3.0.3
 
-### build
-dep_parse_trans = git https://github.com/lazedo/parse_trans
+dep_meck = git https://github.com/eproxus/meck 0.8.13
+# used in tests for kazoo_voicemail, crossbar, teletype, and other deps
 
-dep_horse = git https://github.com/ninenines/horse
+# dep_meta = git https://github.com/efcasado/meta 0.1.3
+# appears unused
```
lazedo: this is used

lazedo:
@jamesaimonetti the `ini` settings are being loaded but not used in the amqp connection. works locally because you have rabbitmq running in localhost. try to change your config.ini to another url and see why its not being used.

make/deps.mk
```diff
@@ -18,20 +14,16 @@ DEPS = amqp_client \
 	exml \
 	fcm \
 	folsom \
-	fs_event \
-	fs_sync \
 	gen_smtp \
 	getopt \
 	gproc \
 	hep \
-	horse \
```
icehess: Isn't this being used by `kz_{term,binary,json,map,time}_tests`?

make/deps.mk
jamesaimonetti: Where? How? `meta:` isn't found anywhere in the code?

make/deps.mk
jamesaimonetti: Only if you specifically run `make perf` - which most folks are not doing. There's a comment on the `dep_horse` line to that effect.

PR #5596, opened 2019-03-18
Kazoo 6066 master

k-anderson:


applications/crossbar/priv/couchdb/schemas/system_config.services.json
```diff
@@ -27,6 +27,11 @@
             "description": "services get_items_from_quantities",
             "type": "boolean"
         },
+        "hydrate_cascade_commits": {
+            "default": true,
+            "description": "services hydrate_cascade_commits",
```
jamesaimonetti: Can we try getting a reasonable description instead of auto-generated? Can we add some docs in `kazoo_services` around what this means?

PR #5598, opened 2019-03-18
cb_rates: need to search rate by number more precisely

onnet:


jamesaimonetti:
@onnet nice to see you around :) Merged

PR #5599, opened 2019-03-19
crossbar analogue to kazoo-applications status

onnet:


applications/crossbar/doc/system_status.md
```diff
@@ -0,0 +1,20 @@
...
+## Fetch
+
+> GET /v2/system_status
+
+```shell
+curl -v -X GET \
+    -H "X-Auth-Token: {AUTH_TOKEN}" \
+    http://{SERVER}:8000/v2/system_status
+```
+
```
jamesaimonetti: Two things:

1. Can you add a sample (anonymized) JSON response payload
2. Can you add a warning that this API requires a superduper admin auth token?

onnet: Hi @jamesaimonetti! Comments have been addressed.

jamesaimonetti:
Fixed up the CI complaints. Thanks @onnet, merged!

PR #5600, opened 2019-03-19
HELP-41363: Add HotDesk-Current-ID CCV when device is hotdesked

harenson:
Depends on https://github.com/2600hz/kazoo-configs-kamailio/pull/97 to work.

### DIFF
--- a/applications/registrar/src/reg_authn_req.erl
+++ b/applications/registrar/src/reg_authn_req.erl
@@ -103,7 +103,8 @@ create_ccvs(#auth_user{doc=JObj}=AuthUser) ->
       ,{<<"Register-Overwrite-Notify">>, AuthUser#auth_user.register_overwrite_notify}
       ,{<<"Pusher-Application">>, kz_json:get_value([<<"push">>, <<"Token-App">>], JObj)}
        | (create_specific_ccvs(AuthUser, AuthUser#auth_user.method)
-          ++ generate_security_ccvs(AuthUser))
+          ++ generate_security_ccvs(AuthUser)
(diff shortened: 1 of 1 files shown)


applications/registrar/src/reg_authn_req.erl
```diff
@@ -502,3 +503,10 @@ maybe_enforce_security({#auth_user{doc=JObj}=User, Acc}) ->
 -spec maybe_set_encryption_flags({auth_user(), kz_term:proplist()}) -> {auth_user(), kz_term:proplist()}.
 maybe_set_encryption_flags({#auth_user{doc=JObj}=User, Acc}) ->
     {User, encryption_method_map(Acc, JObj)}.
+
+-spec maybe_add_hotdesk_current_id(auth_user()) -> kz_term:proplist().
+maybe_add_hotdesk_current_id(#auth_user{doc=JObj}) ->
```
jamesaimonetti: Honestly I would just call `hotdesk_ids/1` directly and pattern match:

```erlang
case kzd_devices:hotdesk_ids(JObj, []) of
    [] -> [];
    [Id|_]= -> [{...}]
end.
```

harenson: Explicit is better than implicit ;)

jamesaimonetti: Sorry, I meant you don't need the `is_hotdesked/1` call at all. Just the pattern match on `hotdesk_ids/1` will tell you all you need to know.

jamesaimonetti:
@harenson please open a 4.3 PR, cherry-picking the commit to master

PR #5602, opened 2019-03-20
Kazoo 6067 master

k-anderson:


applications/crossbar/priv/api/swagger.json
```diff
@@ -29068,6 +29140,11 @@
         "system_config.callflow.resources": {
             "description": "Schema for callflow.resources system_config",
             "properties": {
+                "default_asserted_identity": {
+                    "default": false,
+                    "description": "callflow resources default_asserted_identity",
```
jamesaimonetti: Update descriptions to be useful to humans plz

applications/stepswitch/src/stepswitch_resources.erl
```diff
@@ -708,6 +718,17 @@ gateway_to_endpoint(DestinationNumber
         ,{<<"Outbound-Caller-ID-Number">>, CIDNumber}
         ,{<<"Outbound-Caller-ID-Name">>, CIDName}
         ,{<<"SIP-Invite-Parameters">>, sip_invite_parameters(Gateway, OffnetJObj)}
+        ,{<<"Privacy-Method">>, PrivacyMethod}
+        ,{<<"Privacy-Hide-Name">>
```
jamesaimonetti: Could these be nested keys under a `Privacy` top-level key instead? Then you could do `Should-Hide-Name` and `Should-Hide-Number` which is more self-evident that they are boolean toggles.

k-anderson:
I merged the 4.3 PR with the make-busy fix since I merged the 4.3 PR of this originally :)

PR #5607, opened 2019-03-21
KAZOO-6070: send notifications for all errors and get the last four f…

k-anderson:
Master PR #5606


core/kazoo_services/src/modules/kz_services_payment_tokens.erl
```diff
@@ -51,14 +51,18 @@ defaults(Thing) ->
 
 -spec default(kz_services:services() | kz_term:ne_binary(), kz_term:ne_binary()) -> kz_json:api_object().
 default(Thing, Bookkeeper) ->
+    Now = kz_time:now_s(),
     Pays = kz_json:filter(fun({_TokenId, Token}) ->
                                   kz_json:is_true(<<"default">>, Token)
+                                      andalso kz_json:get_integer_value(<<"expiration">>, Token, Now) > Now
```
icehess: @k-anderson it shouldn't be `>=` in case `expiration` is `Now`?

PR #5610, opened 2019-03-22
 Avoid passing 'undefined' params to kz_json:merge_recursive when fetching account's dataplan

harenson:


core/kazoo_data/src/kzs_plan.erl
```diff
@@ -401,10 +401,15 @@ fetch_account_dataplan_merged(AccountJObj) ->
...
-            PlanJObj = kz_json:merge_recursive(SystemJObj, fetch_dataplan(PlanId)),
-            kz_json:merge_recursive(PlanJObj, AccountJObj)
+            account_dataplan_merged(SystemJObj, fetch_dataplan(PlanId), AccountJObj)
     end.
 
+-spec account_dataplan_merged(kz_json:object(), kz_term:api_object(), kz_json:object()) -> kz_json:object().
+account_dataplan_merged(SystemJObj, 'undefined', AccountJObj) ->
+    kz_json:merge_recursive(SystemJObj, AccountJObj);
+account_dataplan_merged(SystemJObj, Dataplan, AccountJObj) ->
+    kz_json:merge_recursive([SystemJObj, Dataplan, AccountJObj]).
```
lazedo: this is a change in behaviour and not correct. if the AccountJObj does have a pvt_plan_id then it does not have a dataplan.
this should be `kz_json:merge_recursive(SystemJObj, Dataplan)`

lazedo: nm my comment above, i missed the first merge

harenson: 👍

PR #5612, opened 2019-03-22
PISTON-701: always publish dereg notif, handler will filter suppression

danielfinke:
Context for this change here: https://forums.2600hz.com/forums/topic/10425-tracking-registration-state-events-registerderegister/

Still need to decide on whether this should be opt-in/opt-out/configurable at all. Also, the suppress_unregister record field and the Suppress-Unregister-Notify prop key I don't think are needed anymore nor have a purpose. So I could also remove those.


lazedo:
@danielfinke removing the `Suppress-Unregister-Notify` is not an option. you can try to listen registration.success. in 5.x we will publish de-registration scoped to proxy

PR #5617, opened 2019-03-23
PISTON-710: don't do give away when a graceful shutdown in progress

danielfinke:
During acdc shutdown, such as when restarting it through kapps_controller, there is a hang due to a race condition in the ETS table give_away code. As acdc_stats is shutdown, it sends an ETS-TRANSFER of the tables it was writing back to the original acdc_stats_etsmgr procs. This event will always reach the acdc_stats_etsmgr procs before they terminate. However, this means that the give_away message sent by the acdc_stats_etsmgr procs has a chance to come back to them before they are terminated. If this happens, they will try to run find_ets_mgr/2 which asks the acdc_stats_sup for the stats_srv. This is a deadlock, because acdc_stats_sup is in the midst of its supervision tree shutdown procedure.

In the attached log snippet, deadlocked ETS mgr procs are indicated by `give away $tableName: ok‌` log lines. They are sequentially killed in reverse-start-order as they hit the default 5s timeout for ?WORKER_NAME_ARGS children.

The solution was to detect that the acdc_stats proc exited due to reason shutdown and suppress the give_away by discarding the ref in the state.

[acdc_stats_etsmgr-deadlocks.log](https://github.com/2600hz/kazoo/files/2999867/acdc_stats_etsmgr-deadlocks.log)

### DIFF
--- a/applications/acdc/src/acdc_stats_etsmgr.erl
+++ b/applications/acdc/src/acdc_stats_etsmgr.erl
@@ -25,6 +25,7 @@
 
 -record(state, {table_id :: ets:tid() | 'undefined'
                ,etssrv :: kz_term:api_pid()
+               ,last_etssrv :: kz_term:api_pid()
                ,give_away_ref :: kz_term:api_reference()
                }).
 -type state() :: #state{}.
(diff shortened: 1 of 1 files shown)


jamesaimonetti:
@danielfinke you might also look at replacing `acdc_stats_etsmgr` with `kazoo_etsmgr_srv` since the code looks almost identical and the unique bits could be handled with config options to start_link/init. I'll merge this for now though but something to consider.

jamesaimonetti:
Merged

danielfinke:
@jamesaimonetti thanks for the suggestion! I was able to do so: https://github.com/2600hz/kazoo/pull/5646

PR #5618, opened 2019-03-25
KAZOO-6073: [master] api to get summary grouped by sub-accounts

icehess:


applications/crossbar/doc/ledgers.md
```diff
@@ -108,6 +108,84 @@ curl -v -X GET \
...
+```
+
+**Sample response:**
+
+```json
+{
+    "auth_token": "{AUTH_TOKEN}",
+    "data": {
+        "{ACCOUNT_ID_1}": {
+            "per-minute-voip": {
```
k-anderson: @icehess I think we should put the ledgers under an object (such as 'ledgers') to seperate the list of ledger sources from the 'account' container/object.

icehess: done. Also I add total just in case :)

PR #5619, opened 2019-03-25
PISTON-711: start_key off by one, resulting in missed records paginating

danielfinke:
When a query is made through a CouchDB view, the supplied start_key and end_key must have the same length as those in the result set (note: with no grouping applied!). If this is not the case, some results may be filtered out as they are sorted outside the selected range. In Kazoo’s case, this can mean that the next_start_key produced by viewing the last result’s key can be wrong when the group level is not equal to the length of the view key in the underlying view definition.

The following is an example of what can happen here:

GET https://{{baseURL}}/v2/accounts/{{account_id}}/users/{{user_id}}/cdrs/interaction?page_size=10

The view backing this query (interactions/interaction_listing_by_owner) has keys of length 4 (owner_id, interaction_time, interaction_key, channel_created_time). cb_cdrs however uses a group_level of 3. This means that the result set will have keys of length 3. Now assume our last result has a key of `["0a1fde86ac5e801b57ae58510710b31f", 63719832457, "98fe29cc"]` (these are owner_id, interaction_time, and interaction_key, omitting the channel_created_time due to the group_level). This will be returned as the next_start_key. It is one element beyond the page_size of the original req. One would assume in the next query:

GET http://{{baseURL}}/v2/accounts/{{account_id}}/users/{{user_id}}/cdrs/interaction?page_size=10&start_key={{start_key}}

that the result set begins with the result specified with the start_key. It does not. It begins at the record one after the start_key record, causing lost results.

This is because the sort order is descending for this request, and therefore results with keys that sort greater than the start_key will be filtered out. By submitting a start_key of `length(ViewKeys)-1`, CouchDB interprets the 4th missing element as `null` which sorts low (https://docs.couchdb.org/en/stable/ddocs/views/collation.html#collation-specification). This means that any records that have the first 3 elements of their key matching the start_key but have _anything_ other than null/false/true as their 4th element of their key will be filtered out.

In the interaction example, since the last element of the key is an integer (channel_created_time), the record that should have been returned by the next_start_key is going to be skipped.

The solution in this PR addresses this by padding the length of the start_key used for the query to a specific length. This must be set on the query options ahead of time, by knowing what the definition of the view is (what length keys are emitted). The change will automatically detect the sort order: ascending sorts mean the end_key should be padded with the "high sort" option of an empty JSON object, and the start_key should be padded with the "low sort" option of false (I used false because there is currently nothing in Kazoo for representing `null` in JSON). Descending sorts do the opposite, to ensure the start_key is padded as high as possible and the end_key is padded as low as possible.

There are likely to be other Crossbar requests that are subject to the issue, but I am not going to be finding them as part of this PR.

### DIFF
--- a/applications/crossbar/src/crossbar_view.erl
+++ b/applications/crossbar/src/crossbar_view.erl
@@ -46,6 +46,9 @@
         ,'created_to', 'created_from', 'max_range'
         ,'range_end_keymap', 'range_key_name', 'range_keymap', 'range_start_keymap'
         ,'should_paginate'
+
+         %% start/end key length fixer
+        ,'key_min_length'
         ]).
 
 -type direction() :: 'ascending' | 'descending'.
@@ -104,7 +107,10 @@
                     {'range_keymap', range_keymap()} |
                     {'range_key_name', kz_term:ne_binary()} |
                     {'range_start_keymap', range_keymap()} |
-                    {'should_paginate', boolean()}
+                    {'should_paginate', boolean()} |
+
+                    %% start/end key length fixer
+                    {'key_min_length', pos_integer()}
                    ].
 
 -type load_params() :: #{chunk_size => pos_integer()
@@ -198,7 +204,9 @@ build_load_params(Context, View, Options) ->
 
             UserMapper = props:get_value('mapper', Options),
 
-            {StartKey, EndKey} = start_end_keys(Context, Options, Direction),
+            StartEnd = start_end_keys(Context, Options, Direction),
+            KeyMinLength = props:get_value('key_min_length', Options),
+            {StartKey, EndKey} = expand_min_max_keys(StartEnd, Direction, KeyMinLength),
 
             Params = LoadMap#{has_qs_filter => HasQSFilter
                              ,mapper => crossbar_filter:build_with_mapper(Context, UserMapper, HasQSFilter)
@@ -237,7 +245,9 @@ build_load_range_params(Context, View, Options) ->
 
             case time_range(Context, Options, TimeFilterKey) of
                 {StartTime, EndTime} ->
-                    {StartKey, EndKey} = ranged_start_end_keys(Context, Options, Direction, StartTime, EndTime),
+                    StartEnd = ranged_start_end_keys(Context, Options, Direction, StartTime, EndTime),
+                    KeyMinLength = props:get_value('key_min_length', Options),
+                    {StartKey, EndKey} = expand_min_max_keys(StartEnd, Direction, KeyMinLength),
                     Params = LoadMap#{end_time => EndTime
                                      ,has_qs_filter => HasQSFilter
                                      ,mapper => crossbar_filter:build_with_mapper(Context, UserMapper, HasQSFilter)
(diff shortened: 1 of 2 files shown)


jamesaimonetti:
I'm going to try restating the problem as I understand it; let me know if its close:

When querying a view with a 4-element key `[a, b, c, d]` with a `group_level` < key length, KAZOO gets keys back of the `group_level` length (`[a, b, c]` in this case). Then, when supplying the `startkey` via next_start_key as `[a, b, c]` there is an implicit conversion by CouchDB to `[a, b, c, null]` and since the view is queried descending, the startkey is more like `[a, b, b, {}]` and so any results under `[a, b, c, a-c]` are missed (since they are < d).

Close?

jamesaimonetti:
Also, effectively no startkey means `[{}, {}, {}, {}]` when descending and `[null, null, null, null]` when ascending.

jamesaimonetti:
Also what about the note further down on Couch collation: `Beware that {} is no longer a suitable “high” key sentinel value. Use a string like "\ufff0" instead.` We do use it a bit in kapps_util, kazoo_number_manager, hotornot and faxes. Not sure it matters; just curious if you had considered it as a "better" sentinel value?

danielfinke:
> When querying a view with a 4-element key [a, b, c, d] with a group_level < key length, KAZOO gets keys back of the group_level length ([a, b, c] in this case). Then, when supplying the startkey via next_start_key as [a, b, c] there is an implicit conversion by CouchDB to [a, b, c, null] and since the view is queried descending, the startkey is more like [a, b, b, {}] and so any results under [a, b, c, a-c] are missed (since they are < d).

> Also, effectively no startkey means [{}, {}, {}, {}] when descending and [null, null, null, null] when ascending.

@jamesaimonetti you got it, exactly.

> Also what about the note further down on Couch collation: Beware that {} is no longer a suitable “high” key sentinel value. Use a string like "\ufff0" instead. We do use it a bit in kapps_util, kazoo_number_manager, hotornot and faxes. Not sure it matters; just curious if you had considered it as a "better" sentinel value?

Yeah I should probably use it instead. It's no longer considered a suitable "high" key sentinel value because you can have objects in the key and {} will sort lower than any populated object. It appears that either this was not the case in CouchDB 1/BigCouch or maybe objects were not allowed in the keys.

danielfinke:
To be clear, what I said you got is the case once this PR is applied. Omitted keys are assumed to be null in every case

jamesaimonetti:
Just building a test in pqc_cb_cdrs to elicit the error, then applying your PR to check that it is fixed so we can avoid regressions here.

icehess:
@danielfinke This is great, thanks. The idea sounds correct. I was suspecting for awhile that CouchDB is skipping but I couldn't setup a proper test case to debug this.

As for higher value you can use something like `<<16#fff0/utf8>>` since `\ufff0` won't translate to a higher unicode value in erlang land. [example](https://github.com/2600hz/kazoo/blob/9a835384775088730fc5281d9af54105eb3d1954/applications/crossbar/src/modules/cb_resource_selectors.erl#L269)

danielfinke:
@jamesaimonetti @icehess switched to the high sentinel value of `<<16#fff0/utf8>>`

jamesaimonetti:
OK so the trigger for the bug appears to be if an interaction_key has 2+ legs and is split between the page requested and the next page. Loading the next page will then skip any legs of the interaction_key not loaded in the prior page.

I'm still seeing the errors in my test after applying your patch, though I'm not sure if it is the tests that are wrong or the code here. Out of time today but will look at it more tomorrow.

danielfinke:
@jamesaimonetti you happen to spend any more time on this? Anything I could help out with?

jamesaimonetti:
Sadly was waylaid by other things; haven't determined if the tests are testing what I think they should be. Worked on them a bit today though, hoping to find some more time later too.

jamesaimonetti:
@danielfinke I've added my commit to test this on https://github.com/2600hz/kazoo/commits/711-test

Assuming CI passes, should get this merged today.

jamesaimonetti:
Merged

PR #5626, opened 2019-03-26
Kazoo script

lazedo:
* allow per node vm.args/sys.config
* remove code path as its not needed

### DIFF
--- a/rel/kazoo
+++ b/rel/kazoo
@@ -40,6 +40,7 @@ ERL_OPTS="{{ erl_opts }}"
 RUNNER_LOG_DIR="${RUNNER_LOG_DIR:-/tmp/log}"
 VM_NODE_FILES_DIR="${VM_NODE_FILES_DIR:-/tmp}"
 NAME_TYPE="${NODE_NAME_TYPE:--name}"
+NODE_STARTUP_FILE_PATH=${NODE_STARTUP_FILE_PATH:-/etc/kazoo/core}
 
 
 # start/stop/install/upgrade pre/post hooks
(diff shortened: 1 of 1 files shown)


rel/ci.vm.args
```diff
@@ -21,7 +21,7 @@
 ##-heart
 
 # GC more often
--env ERL_FULLSWEEP_AFTER 10
+# -env ERL_FULLSWEEP_AFTER 10
```
k-anderson: @lazedo lets remove the line rather than comment out.  Git has a history feature ;)

rel/kazoo
```diff
@@ -40,6 +40,7 @@ ERL_OPTS="{{ erl_opts }}"
 RUNNER_LOG_DIR="${RUNNER_LOG_DIR:-/tmp/log}"
 VM_NODE_FILES_DIR="${VM_NODE_FILES_DIR:-/tmp}"
 NAME_TYPE="${NODE_NAME_TYPE:--name}"
+NODE_STARTUP_FILE_PATH=${NODE_STARTUP_FILE_PATH:-/etc/kazoo/core}
```
k-anderson: @lazedo as per the standup, this looks fine but makes me nerves and we agreed I owe you a shot as well as a beer (at least) if someone needs this functionality.  ;)

lazedo: i pushed GC change in different PRs. the 4.3 of this PR should be merged after next release.

rel/ci.vm.args
lazedo: pushed in #5628 & #5629

PR #5641, opened 2019-03-31
add Rate-ID to cb_rates by number search result

onnet:
It is good to have rate_id in search result in order to able to edit/delete found rate.


lazedo:
@onnet thanks and welcome back :)

PR #5643, opened 2019-04-01
[master] help-43115: fixed limits default settings so soft limits are not on by default 

swysor:


swysor:
Need to add schemas, please hold onto this one until I update.

PR #5649, opened 2019-04-02
Set prefetch to config's value if unnamed queue is used

jamesaimonetti:
When a "random" queue name is used to consume events for a specific
process (vs a "named" queue which generally is for sharing events
among consumers), set the prefetch value to what's in the
config.ini. This setting increases the number of messages in-flight
between the broker and the consumer and speeds up throughput.


jamesaimonetti:
Backport of #5644

PR #5650, opened 2019-04-02
PISTON-731: Changed duration_seconds in cdr to int from string to match all other 'seconds' value types

bradfordben:
duration_seconds in a cdr doc is stored as a string value when billing_seconds is stored as an int. 

As a result every time a cdr is loaded through CB, the string value must be converted to an int. 

Converting it to a int before saving also helps follow consistency.


bradfordben:
Thanks 👍

PR #5651, opened 2019-04-02
PISTON-724: Handled crossbar 'storage' validation when no 'attachment' is supplied in the request body

bradfordben:
If a storage object is created through CB without a 'attachment' the request will crash

I have added a case to handle wen no 'attachment' is supplied and skip validation of the attachment settings

### DIFF
--- a/applications/crossbar/src/modules/cb_storage.erl
+++ b/applications/crossbar/src/modules/cb_storage.erl
@@ -401,7 +401,7 @@ maybe_check_storage_settings(Context, ReqVerb) when ReqVerb =:= ?HTTP_PUT
     case cb_context:resp_status(Context) of
         'success' when ValidateSettings ->
             lager:debug("validating storage settings"),
-            Attachments = kz_json:get_json_value(<<"attachments">>, cb_context:doc(Context)),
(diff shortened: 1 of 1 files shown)


lazedo:
@bradfordben can't you do this with https://github.com/2600hz/kazoo/pull/5651/files#diff-94b118f9598944135d11dbe22b487207L407 ?

bradfordben:
You can but i feel its not acceptable to force the user to append the URL with "validate_settings=false" and also set "sup kzs_plan disallow_validation_overrides" when you are not defining an "attachment"

In the CB "storage" spec you will see that "attachment" is not a required field but when you post a "storage" object it will try and validate the "attachment" and if its not set the code will crash as there is no function to handle this clause of 'undefined'

So this solution just handles the case where there is no "attachment" is defined in the storage object and skips the validation of the "attachment" as its not necessary when its not present.

lazedo:
you can send it as an empty object, schema may change to require it.

lazedo:
or you can change here https://github.com/2600hz/kazoo/pull/5651/files#diff-94b118f9598944135d11dbe22b487207L404 to default to kz_json:new()

bradfordben:
Thanks, ive made the change as you suggested and set the default value to a empty kz_json. Its a cleaner fix.

lazedo:
@bradfordben thanks

PR #5664, opened 2019-04-04
HELP-43123: build ring group members in order

jamesaimonetti:
Process the endpoints specified in the ring group's Data config in
reverse order (using foldr); when the builders return their endpoints,
they are prepended to the accumulator and return to the configured
order of members.


jamesaimonetti:
@lazedo because we process the list of `Builders` in reverse order, the `receive` will block until the last Builder finishes, then the 2nd to last, until the first one is done. So if the last one takes longer, the other Builder responses will queue up in the mailbox and will be selectively received when their turn comes, preserving order.

PR #5666, opened 2019-04-05
[master] help-42518: handle transfer where transferrer does not specify referred-by header

swysor:


swysor:
> i think this is not valid as `Caller-Username` would have the value of the `transferee` not the `transferor`. what is the motivation for this change besides the fix that was provided in `libwebphone` ?

The libwebphone fix I tried works fine if the username is all lowercase. For some reason libwebphone lowercases the contact header so that won't work if any uppercase chars appear in the username. also this client is using sip.js not the jssip.js in master libwebphone so they had to change that on their end but it didn't work due to the mismatch of the username. I could probably figure out how to get at the From header which does contain an uppercase username, but for some silly reason the library lowercases the contact header userpart. 

In my tests the caller-username was correct, but I should find a better way to get that if it could vary based on the call, as I guess if the phone is called it would not show the right username.  I thought the caller-username would be the username of the person sending the REFER from the debugging I did, but that could be wrong. I will do some more testing.

PR #5669, opened 2019-04-06
[master] Allows admin to set routes field in the CSV for tasks rates import.

bputtick:
routes field in the CSV can be a string of either a single route "^\\+?44.+$" or a json array string '["^\\\\+?123$","^\\\\+?456$"]'

Falls back to kzd_rates:set_default_route/1 using prefix.


bputtick:
included some validation to ensure fields are imported using the correct type

jamesaimonetti:
Added tests to check that the JSON was deserialized properly and returned on subsequent fetches.

PR #5675, opened 2019-04-09
HELP-43124: [master] don't reverse bridge strings again when failover option is set

icehess:
Also refactoring use failover if all devices unregister part

4.3: https://github.com/2600hz/kazoo/pull/5672


jamesaimonetti:
@icehess any way you can roll those endpoint payloads into a test here? A test that triggers the reversal before your changes and shows the right order after the changes are applied?

jamesaimonetti:
@icehess tests?

icehess:
@jamesaimonetti Sorry just saw this. I did test the ring group as described in that ticket (tough with not same exact device/user/callflow documents but same scenario) and it was in order.

jamesaimonetti:
@icehess right, i'm asking if you can create the two endpoints (generalized of course) and test the ecallmgr_util function calls that they build the endpoints in order (so a unit test).

PR #5676, opened 2019-04-09
use os:system_time

lazedo:
erlang:system_time gives vm time which is not good for exchanging timestamps with other nodes namely freeswitch & kamailio on which we already have to remove gregorian

### DIFF
--- a/core/kazoo_stdlib/src/kz_time.erl
+++ b/core/kazoo_stdlib/src/kz_time.erl
@@ -85,7 +85,7 @@ current_tstamp() ->
 
 -spec current_unix_tstamp() -> unix_seconds().
 current_unix_tstamp() ->
-    erlang:system_time('seconds').
+    os:system_time('second').
 
 -spec gregorian_seconds_to_unix_seconds(integer() | string() | binary()) -> integer().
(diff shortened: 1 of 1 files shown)


jamesaimonetti:
We may or may not want to adjust our usage of kz_time when timing how long things take. If we want to time things we should be using `erlang:monotonic_time()` as Start and End and calc the diff from those:

```erlang
(pdx-pr@pdx.2600hz.com)5> Start = erlang:monotonic_time().
-576385307291607138
(pdx-pr@pdx.2600hz.com)6> End = erlang:monotonic_time().
-576385296211111299
(pdx-pr@pdx.2600hz.com)7> End-Start.
11080495839
(pdx-pr@pdx.2600hz.com)9> erlang:convert_time_unit(End-Start, native, second).
11
(pdx-pr@pdx.2600hz.com)10> erlang:convert_time_unit(End-Start, native, millisecond).
11080
```

I wonder if we add `kz_time:start_timing()` and `kz_time:end_timing(Start, [second | millisecond |...])` so the code can be clear that it is timing an operation. `start_timing()` just returns `erlang:monotonic_time()` and `end_timing/{1,2}` does the calculation and conversion.

PR #5681, opened 2019-04-10
More private AMQP worker pools

jamesaimonetti:


applications/callflow/src/cf_route_req.erl
```diff
@@ -20,6 +20,7 @@
 
 -spec handle_req(kapi_route:req(), kz_term:proplist()) -> 'ok'.
 handle_req(RouteReq, Props) ->
+    'true' = kapi_route:req_v(RouteReq),
```
lazedo: if you're adding here (which makes sense as we don't want to crash the worker), remove it line 27

jamesaimonetti: Ah, I was just comparing how ts_route_req and cf_route_req started and noticed we didn't have the validator earlier. I should have checked deeper in the code; thanks!

jamesaimonetti: Meaning, I'm an idiot :)

PR #5682, opened 2019-04-11
bputtick - resource_consuming doesn't count itself

bputtick:
When j5_hard_limits calculates resource_consumption_at_limit/1 the current channel isn't counted as a consuming resource so usage needs to be increased by 1 for the assumption of the current channel.


bputtick:
Actually, there's a bigger issue here...

```
j5_request:218(<0.25344.8>) account a90eb3d44e189ea5c15201253b644f3f authorized channel: limits_disabled
j5_hard_limit:19(<0.25344.8>) authorizing hard_limits
j5_hard_limit:42(<0.25344.8>) calls_limit -1:1
j5_channels:182(<0.25344.8>) Tab: [{channel,<<"29a8b0de42a9de5a28e643a86aacd8a3@umis.lonpbx.clearhost.co.uk">>,undefined,<<"inbound">>,<<"a90eb3d44e189ea5c15201253b644f3f">>,undefined,false,undefined,undefined,false,false,63722227425,undefined,undefined,undefined,undefined,undefined,undefined,undefined,undefined,undefined,undefined,undefined,false}]
j5_hard_limit:53(<0.25344.8>) resource_consumption_limit -1:1
```
Given the above log (with extra debug) we can see the channel is updated with `limits_disabled` but jonny5 hasn't noticed this yet, so just increasing the usage by 1 means this channel is then incorrectly counted.

bputtick:
Should the channels be syncronized with j5_channels before checking hard limits or can we assume that the current channel always be excluded from resource_consumption and therefore we can check
j5_request:account_billing/1 to see if it's NOT undefined or <<"limits_disabled">> and only then increment usage by 1?

or

do we just accept that resource_consuming is likely to be 1 out?

k-anderson:
Howdy, I am sorry but I don't follow the logic here.  If the accounts limits are disabled add one to the resource consuming channel count?  Could you provide the call scenario where the channel is unaccounted for?

Based on the description I would have expected the following to be changed to:
https://github.com/2600hz/kazoo/blob/master/applications/jonny5/src/j5_channels.erl#L217
```
count_unique_calls(ets:select(?TAB, MatchSpec)) + 1.
```

bputtick:
Hi

It's adding 1, if << AccountBilling>> is **not** disabled and **not** undefined.

At the point resource_consumption_at_limit/1 is enumerating channels it doesn't take in to account the channel that is about to be created.

So if the limit is 4 and there are currently 4 other channels already in use, then the call is allowed creating the 5th channel. So therefore the pending channel should be assumed in the deny calculation.

Barnaby

bputtick:
> Based on the description I would have expected the following to be changed to:
> https://github.com/2600hz/kazoo/blob/master/applications/jonny5/src/j5_channels.erl#L217
> 
> ```
> count_unique_calls(ets:select(?TAB, MatchSpec)) + 1.
> ```

This would mean that when 0 channels are in use it incorrectly shows 1 channel?

bputtick:
For clarity.. the channel enumeration is correct, it's the deny calculation that should consider the channel that is about to be created should the call be allowed.

lazedo:
@bputtick how about changing `should_deny(Limit, Used) -> Used > Limit.` to `should_deny(Limit, Used) -> Used >= Limit.` ?

bputtick:
> @bputtick how about changing `should_deny(Limit, Used) -> Used > Limit.` to `should_deny(Limit, Used) -> Used >= Limit.` ?

That's exactly what I did to start with, but for reasons that I now can't remember this didn't work in other scenarios.

It could be that because both inbound_channels_per_did_at_limit and calls_at_limit also use should_deny and these may correctly enumerate the 'used' channels and therefore should_deny will cause a premature deny?

The other aspect is that I added a check for AccountBilling is neither "limits_disabled" nor "undef" before incrementing by 1.

I wish I commented the reason for this now.. I'll have to try and remember the reasoning again for this...

bputtick:
The first point is certainly true, I can see from my local debug that

j5_hard_limit:43(<0.14754.29>) calls_limit Used: 1 Limit: -1

bputtick:
For the second point, if the channel leg to be created has has limits_disabled flag or is undefined then it should not be included in the used enumeration. 

The only other consideration is why I am not just checking for limits_enforced, again I am sure I had a reason...

lazedo:
@bputtick please reopen after the repo split

PR #5684, opened 2019-04-12
bputtick - rework get_limit/3 logic to correctly parse pvt/public

bputtick:
There were some failure cases that got the wrong value (IMHO)

Defining a pvt_ limit key should be authoritative and when a public key is defined the lower value of the two wins. If neither are set then the system default should be used.

If `{"pvt_resource_consuming_calls": 10}` is defined without a public version, then the system default was winning.

The current logic numerically compares `PrivateValue < PublicValue` when `PrivateValue` is defined.
`PublicValue` is derived from `get_public_limit/3` which returns the system default or the passed Default parameter if it is 'undefined'.

Given that `-1` is defined as "unlimited", when the system_default and/or parameter is `-1` `PrivateValue < PublicValue` is evaluated as `10 < -1` and the default prevails ignoring the pvt set by the admin on that account.

Moreover, if `{"pvt_resource_consuming_calls": -1, "resource_consuming_calls": 10 }` is defined, I would expect the public key to win as it's lower than "unlimited"; but again this is not the case as `-1` is lower than `10`.    

I have a suggested logic update on the PR that respects `-1` as "unlimited" and only uses the a default if neither public or private vars are defined.

Further more `get_public_limit/3` returns `0` if `-1` is defined as the public var, presumably prohibiting the public var being set to unlimited. This would have an undesired effect causing hard_limit to take immediate effect.  

Again, I am suggesting that, if it is desire is to prohibit "unlimited" on a public limit key, then the system default is used rather than 0.

Thoughts on a postcard...


swysor:
hey @bputtick, 
I added some unit tests and ran them against this PR and this does address the issue you reported. Good catch on this one, this code has been like this forever so really appreciate that you found this and addressed it!  

https://github.com/2600hz/kazoo/pull/5696/files

Typically we would handle this sort of very nested logic in a more explicit way (self documenting code) such as separate function clauses, but since this is just a small stylistic change in this case I will merge it to get the bug corrected quickly.  

Thank you!

PR #5685, opened 2019-04-12
bputtick - kzd_limits has the wrong config macro set

bputtick:


bputtick:
already fixed - https://github.com/2600hz/kazoo/pull/5643

PR #5686, opened 2019-04-13
bputtick - allow reseller and superduper admin to update leaked pvt fields

bputtick:
There is no current way for an admin or superadmin to configure Post Pay via crossbar for a particular account.

cb_limits currently exposes (leaks) `pvt_allow_postpay` and `pvt_max_postpay_amount` as public vars or uses the system default if they do not exist. They are discarded on POST. 

kzd_limits only permits the private versions via `get_limit_boolean/2` and `get_limit_units/2` with the only exception of `allow_prepay`

I propose that if a valid reseller admin or super admin updates the doc via a POST with these public vars set, the pvt_ versions are updated and the public vars are discarded as normal.

The only tweak on the existing functionality is that the vars are only leaked if the PVTs exist and not fall back to the system default to avoid leaking the system defaults in to the doc.

This concept might want to be extended to the other PVT only fields?

A system_config flag probably should be set to enable this behaviour over the default.


mark2600:
After talking with the team we do not want to go in this direction for two primary reasons.  First, it further blends public and private fields which while we do that on some APIs we are working to undo that decision as it confuses developers who submit changes for those for parameters thinking it will impact the behavior.  We have been discussing moving them as part of a v3 API to an attribute on the response envelope outside the data object (such as: {"private": {"created": xxxx, "allow_postpay": false}, "data": { ... }}) Second, while not thrilled with setting a precedence to change pvt fields via the APIs, we do believe that it really should be its own explicitly endpoint.  For example, accounts/{account_id}/limits/predominate

bputtick:
I agree that it's a bit of a mess, but there's no other work around and I need my resellers to set this from the UI, so if I am in the minority who needs this I will leave this committed locally :)

PR #5687, opened 2019-04-13
bputtick -  add billing_seconds to ledger metadata for per_minute billing

bputtick:


k-anderson:
@bputtick Howdy! I might be overlooking something but isn't that redundant?
https://github.com/2600hz/kazoo/blob/b4f34efecf9a9fee52a6b1955fd85f42e082b156/applications/jonny5/src/j5_per_minute.erl#L128

bputtick:
> @bputtick Howdy! I might be overlooking something but isn't that redundant?
> 
> [kazoo/applications/jonny5/src/j5_per_minute.erl](https://github.com/2600hz/kazoo/blob/b4f34efecf9a9fee52a6b1955fd85f42e082b156/applications/jonny5/src/j5_per_minute.erl#L128)
> Line 128 in [b4f34ef](/2600hz/kazoo/commit/b4f34efecf9a9fee52a6b1955fd85f42e082b156)
>  ,{fun kz_ledger:set_usage_quantity/2, Seconds}

I should probably have added it as `<<“billing_duration”>>`

They’re slightly different. The ledger quantity isn’t the billing duration. The call can be 15 seconds long and the quantity would be 60 seconds if that’s the rates minimum or 0 seconds if the rate_nocharge_time is greater that 15 seconds. 

Barnaby

k-anderson:
@bputtick ah, yup ok that makes sense!  I had a feeling I was missing something :)

PR #5694, opened 2019-04-16
Create a single cache listener for conf changes

jamesaimonetti:
Rather than each cache starting a separate AMQP queue for conf (doc
changes) bindings, create one VM-wide listener and have each cache
process cluster bind for events using the local kazoo_bindings
server.

When the cache listener receives a payload, it distributes the event
to the relevant cache processes to flush the cache(s) impacted by the
changed doc.


lazedo:
@jamesaimonetti amazing!

PR #5695, opened 2019-04-16
Added the ability for parent call-restrictions to be inherited

Spencerhutch:
When a child account sets their call-restrictions to `inherit`, this change will allow that restriction to be inherited from their parent accounts. 

A child call-restriction can be inherited from any parent account in the account hierarchy.


jamesaimonetti:
Merged

PR #5696, opened 2019-04-16
QF: added limits tests for the tristate_integer cases

swysor:


k-anderson:
@swysor please add a test case where the pvt_ is positive but smaller than the public value.

PR #5698, opened 2019-04-20
Added media converter and additional config

arekfryz:
Google does not support mp3, media should be converted. Additional config options let you get more accurate recognition. Works with v1 and v1p1beta1 of Cloud Speech-to-Text API


lazedo:
@arekfryz thanks. the pr should target master. maybe you also want to look at #5483

arekfryz:
> #5483

Hi, I looked at 5483 - I'm doing the same conversions here.. but in simple way. I'm using that for voicemail transcriptions.

PR #5700, opened 2019-04-22
PISTON-675: hold-bleg - ecallmgr_fs_transfer:attended/3 hold music works

danielfinke:
Make sure to put the bleg on hold during attended transfer consult. This change was previously in https://github.com/2600hz/kazoo/pull/5484

### DIFF
--- a/applications/ecallmgr/src/ecallmgr_fs_transfer.erl
+++ b/applications/ecallmgr/src/ecallmgr_fs_transfer.erl
@@ -13,7 +13,8 @@
 
 -include("ecallmgr.hrl").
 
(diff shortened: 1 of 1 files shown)


k-anderson:
@lazedo @jamesaimonetti What is going on here :) James is commit spec fixes and Luis said hold, then it looks like everyone is happy but CI is failing (with an error similar to the one I saw on my PRs) and Luis said the changes are approved...

I suspect that we could resolve the conflict (its due to the change James made) and accept this?

danielfinke:
@k-anderson looks like James has tweaked/committed this change in f1acd35, so I will close this

PR #5704, opened 2019-04-23
PISTON-680: remove unused X-KAZOO-PUSHER-Invite-Format, extra_headers

danielfinke:
`extra_headers` and `X-KAZOO-PUSHER-INVITE-FORMAT` were used in a prior version of the Kamailio configs but are no longer used


danielfinke:
Looks like these test failures are unrelated to the content of this PR

PR #5707, opened 2019-04-23
PORT-23: [master] check old key for setting is_private first

icehess:
4.3: https://github.com/2600hz/kazoo/pull/5706

### DIFF
--- a/applications/crossbar/src/modules/cb_port_requests.erl
+++ b/applications/crossbar/src/modules/cb_port_requests.erl
@@ -1051,8 +1051,8 @@ filter_private_comments(Context, JObj) ->
             kzd_port_requests:set_comments(JObj, Comments)
     end.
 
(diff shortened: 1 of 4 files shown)


applications/crossbar/src/modules/cb_port_requests.erl
```diff
@@ -598,7 +598,7 @@ load_port_request(Context, Id) ->
         'false' -> Context2;
         'true' ->
             Doc = cb_context:doc(Context2),
-            IsPrivateKeys = [<<"is_private">>, <<"superduper_comment">>],
+            IsPrivateKeys = [<<"superduper_comment">>, <<"is_private">>],
```
jamesaimonetti: This should be put into `kzd_comment:is_private` instead and `kzd_comment` used in this module.

k-anderson: @icehess I like that suggestion :) any chance you could do that tomorrow?

PR #5708, opened 2019-04-24
list rates according to keys list built out of provided prfix

onnet:


jamesaimonetti:
@onnet thanks! I added some tests to the automated suite to check this new feature. Merged!

PR #5710, opened 2019-04-24
KAZOO-6086: add jobj_return type

icehess:
Add `kazoo_data:jobj_return()` and `kazoo_data:jobjs_return()` types.

### DIFF
--- a/applications/tasks/src/kz_account_crawler.erl
+++ b/applications/tasks/src/kz_account_crawler.erl
@@ -184,7 +184,7 @@ crawl_account(AccountId) ->
     OpenResult = kz_datamgr:open_doc(?KZ_ACCOUNTS_DB, AccountId),
     check_then_process_account(AccountId, OpenResult).
 
--spec check_then_process_account(kz_term:ne_binary(), {'ok', kzd_accounts:doc()} | {'error',any()}) -> 'ok'.
+-spec check_then_process_account(kz_term:ne_binary(), kazoo_data:jobj_return()) -> 'ok'.
 check_then_process_account(AccountId, {'ok', AccountJObj}) ->
     case kz_doc:is_soft_deleted(AccountJObj)
         orelse not kzd_accounts:is_enabled(AccountJObj) of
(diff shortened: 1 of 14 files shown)


core/kazoo_apps/src/kapps_account_config.erl
```diff
@@ -300,7 +300,7 @@ get(Account, Category) ->
         {'error', _} -> kz_doc:set_id(kz_json:new(), kapps_config_util:account_doc_id(Category))
     end.
 
--spec load_config_from_system(kz_term:api_binary(), kz_term:ne_binary()) -> kazoo_data:get_results_return().
+-spec load_config_from_system(kz_term:api_binary(), kz_term:ne_binary()) -> kazoo_data:jobj_return().
```
jamesaimonetti: `get_results_return()` allows for JSON terms (like arrays and whatnot) which are valid results from views. It is not a given that a view request will return just objects.

applications/tasks/src/kz_account_crawler.erl
```diff
@@ -184,7 +184,7 @@ crawl_account(AccountId) ->
     OpenResult = kz_datamgr:open_doc(?KZ_ACCOUNTS_DB, AccountId),
     check_then_process_account(AccountId, OpenResult).
 
--spec check_then_process_account(kz_term:ne_binary(), {'ok', kzd_accounts:doc()} | {'error',any()}) -> 'ok'.
+-spec check_then_process_account(kz_term:ne_binary(), kazoo_data:jobj_return()) -> 'ok'.
```
jamesaimonetti: I think this removes nuance when reading the spec - it isn't just a JSON object, it is the account doc. Specificity over generality in types is preferred.

icehess: @jamesaimonetti That's true. But we can look at it in another way that it hides its true type behind an excessive description (it that a correct term? :). We did this excessive description in many places (not particularly doc/jobj) and makes developer more confused, it doesn't say what is really for example `doc` ? Is it `kz_json:object` or a record or a map or binaries or what? Sometimes we care about it we deep diving in the code and find out it is really jobj, but sometimes we assume and won't look it up.

Anyway ho, w about this:
```erlang
kazoo_data:jobj_return(kzd_accounts:doc())
```

icehess: @jamesaimonetti Do you want me to close this or you're agree with change this to `kazoo_data:jobj_return(kzd_accounts:doc())`?

mark2600:
We talked about this today and decided to close this.

PR #5713, opened 2019-04-25
PISTON-747: Created function to add commas to lists

DustinBrett:
Created intermediate function to add comma separation to lists before converting to binary.

### DIFF
--- a/core/kazoo_csv/src/kz_csv.erl
+++ b/core/kazoo_csv/src/kz_csv.erl
@@ -278,12 +278,19 @@ jobjs_to_file(JObjs, CellOrdering) ->
     File = <<"/tmp/json_", (kz_binary:rand_hex(11))/binary, ".csv">>,
     jobjs_to_file(JObjs, {File, CellOrdering}).
 
+-spec maybe_convert_cell_to_binary(kz_json:get_key(), kz_json:object()) -> binary().
+maybe_convert_cell_to_binary(Path, JObj) ->
+    case kz_json:get_value(Path, JObj, ?ZILCH) of
+        List when is_list(List) -> list_to_binary(lists:join(",", List));
+        Value -> cell_to_binary(Value)
+    end.
+
 -spec jobj_to_file(kz_json:object(), file_return()) -> file_return().
 jobj_to_file(JObj, {File, CellOrdering}) ->
     FlatJObj = kz_json:flatten(JObj),
     NewOrdering = maybe_update_ordering(CellOrdering, FlatJObj),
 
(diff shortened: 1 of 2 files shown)


jamesaimonetti:
@DustinBrett can you make a test in kz_csv_tests that shows what was erroring and what this patch fixes?

DustinBrett:
> @DustinBrett can you make a test in kz_csv_tests that shows what was erroring and what this patch fixes?

I could make a test sure. The issue wasn't an error but more of a formatting problem. The `/v2/accounts/.../phone_numbers` endpoint was returning an array (`features_available`) which kz_csv was flattening and combining into 1 string. This resulted in rows looking like `"carrier_nameforce_outboundprependringback"` when the data was:

```
"features_available": [
  "carrier_name",
  "force_outbound",
  "prepend",
  "ringback"
]
```

Do you think this is something I should make a test for? Or do you have a different suggestion for solving this?

Thanks.

jamesaimonetti:
@DustinBrett yeah, if you could add it so we can check that arrays of strings are encoded into the CSV properly. Maybe something akin to `variable_json_test/0`?

DustinBrett:
> @DustinBrett yeah, if you could add it so we can check that arrays of strings are encoded into the CSV properly. Maybe something akin to `variable_json_test/0`?

Ok sure, will do. I should have this PR updated early next week. Thanks for the feedback.

DustinBrett:
> @DustinBrett yeah, if you could add it so we can check that arrays of strings are encoded into the CSV properly. Maybe something akin to `variable_json_test/0`?

@jamesaimonetti I've added a test based on the suggestion you gave me. The JSON would be:

```json
[
    {
        "a": ["x", "y"]
    },
    {
        "a": [],
        "b": ["x", "y", "z"]
    },
    {
        "a": [
            { "x": 1 },
            { "y": 2 }
        ]
    }
]
```

And the expected results:

```
"a","b"
"x,y"
"","x,y,z"
"",""
```

I had 1 question, the 3rd JSON entry where I test an array of objects, the objects get turned into nothing with the current code, but it's not related to my test/changes (I believe) so I wanted to know if I should adjust that test or leave it or if I should add anymore tests.

Thanks.

jamesaimonetti:
@DustinBrett nothing needed for this PR for the nested objects. I wonder if future work would flatten them, and join the keys using '.'.

```
JSON = {
        "a": [
            { "x": 1 },
            { "y": 2 }
        ]
   },
Flat = {["a", "x"]: 1
        ,["a", "y"]: 2
       }
CSV = "a.x,a.y\n1,2"
```

If it comes up in practice I would look at this as a solution but YAGNI so merged!

PR #5726, opened 2019-04-26
PISTON-397: handle agent channels that are destroyed quickly

danielfinke:
also includes PISTON-352: track endpoints in acdc_agent_fsm better

Have `acdc_agent_listener` monitor all `CHANNEL_DESTROY`s so that it can have `acdc_agent_handler` tell `acdc_agent_fsm` to move out of the outbound state if the agent wasn't bound to the call fast enough (when the outbound call is hung up quickly)


jamesaimonetti:
Merged

PR #5731, opened 2019-04-26
AMQP pool for kzc_recording processes

jamesaimonetti:
Hopefully stemming the tide of leaking processes but also capping the ability for those processes to leak too much before poolboy errors start occurring.

Another option is to subscribe to the necessary call events with a shared listener and distribute, a la the cache listener rework. So a kzc_recording would start up and bind using kazoo_bindings. The kzc_listener could also cache channel destroys for a few seconds to ensure the kzc_recording process gets notified.

@lazedo thoughts on the second approach vs what's in the PR here?


lazedo:
@jamesaimonetti i think we should target this for 4.3 , for master there's #5819. note that `mkbusy` is failing in recording related tests

jamesaimonetti:
@lazedo yeah saw your work on moving it out. I'll take a look at this later this week or next week once ledger/services rollover stuff is done.

jamesaimonetti:
Closing for 4.3 instead

PR #5733, opened 2019-04-29
[master] KZOO-1: fax cleanup task added to handle stale fax docs in the faxes db

swysor:


applications/crossbar/priv/couchdb/schemas/system_config.tasks.fax_cleanup.json
```diff
@@ -0,0 +1,28 @@
...
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "_id": "system_config.tasks.fax_cleanup",
+    "description": "Schema for tasks.fax_cleanup system_config",
+    "properties": {
+        "chunk_size": {
+            "default": 32,
+            "description": "The number of documents to process in one fetch operation",
+            "type": "integer"
+        },
+        "default_age_out": {
```
jamesaimonetti: I prefer to add the unit of time to the name of the variable - `default_age_out_s` and `per_chunk_pause_ms` in this case.

applications/tasks/src/modules/kt_fax_cleanup.erl
```diff
@@ -0,0 +1,306 @@
...
+    maps:get(AccountId, Accounts, 'false').
+
+-spec retries_exhausted(kzd_fax:doc()) -> boolean().
+retries_exhausted(Doc) ->
+    kzd_fax:retries(Doc, 0) =:= kzd_fax:attempts(Doc, 0).
+
+-spec next_start_key(kz_json:objects()) -> {kz_term:ne_binary(), kz_json:objects()}.
+next_start_key([]) -> [];
+next_start_key(Docs) ->
+    {Results, [StartDoc]} = lists:split(length(Docs) -1, Docs),
```
jamesaimonetti: I wonder what the execution comparison of this compared to:

```erlang
[StartDoc | RevResults] = lists:reverse(Docs),
{kz_doc:id(StartDoc), lists:reverse(RevResults)}.
```

What might be a "worst-case" length for `Docs`?

jamesaimonetti: I was curious so I tested.

With lists of 10,000 elements, the average diff is 100 microseconds between the two approaches. Ignore for now :)

doc/mkdocs/mkdocs.yml
```diff
@@ -350,6 +350,7 @@ pages:
     - 'applications/tasks/doc/kazoo_bill_early_task.md'
     - 'applications/tasks/doc/kz_notify_resender.md'
     - 'applications/tasks/doc/maintenance.md'
+    - 'applications/tasks/doc/fax_cleanup.md'
```
jamesaimonetti: Alphabetize this

PR #5734, opened 2019-04-29
use kzs_term:to_atom/2 to prevent crashes

mk1s:


lazedo:
@mk1s please backport to 4.3

PR #5748, opened 2019-05-06
Allow larger media files to be uploaded

extremerotary:
Update the media system schema file to allow media files up to 11Mb to be uploaded


jamesaimonetti:
Thanks @extremerotary, merged

PR #5749, opened 2019-05-06
[Master] - prevent db listing to upgrade to node local connection in bigcouch

mk1s:


lazedo:
@mk1s this needs careful testing. please make sure you create a multi node bigcouch cluster where not all databases exist on every server (n=1). i think the query to `dbs` in admin port is related to that

PR #5753, opened 2019-05-06
PISTON-773: rework some acdc_agent_util status code

danielfinke:
- consider <<"unknown">> same as <<"logged_out">> for recent statuses
- use most_recent_ets_statuses/2 for most_recent_ets_status/2, just
  filter afterwards
- set endkey for most_recent_db_status/2 so other agents don't count for
  prev status of a new agent
- reintroduce db statuses, but cache them at the account level since ETS
  can be trusted after start
- most_recent_ets_statuses/2 will pull from _all_ acdc nodes instead of
  just the first to reply. Sorting omitted - we'll trust whichever node
  is first to answer for each agent


jamesaimonetti:
Merged!

PR #5756, opened 2019-05-07
Ecallmgr streams

lazedo:
decouple event streams from processing


lazedo:
there's still work to be done in kazoo_directory and ecallmgr_fs_xml (for kz endpoints) , these will be addressed in subsequent pr

jamesaimonetti:
@lazedo i rebased and added a couple commits around dialyzer-discovered discrepencies. I need FreeSWITCH with updated mod_kazoo to test though.

PR #5757, opened 2019-05-07
PISTON-780: add max_slot_number to cf_park: continue after >max parked

danielfinke:
We have a customer who would like the ability to add a callflow module that parks the call traversing the callflow but only allow up to a certain number (in our case 6) parked calls at a time and do something else if they are full. So we are using the direct_park method of parking the call and have created max_slot_number which is the highest slot number that may be parked by the cf mod. If the auto-generated slot number exceeds the max_slot_number, the callflow will continue to the next module.

### DIFF
--- a/applications/callflow/src/module/cf_park.erl
+++ b/applications/callflow/src/module/cf_park.erl
@@ -34,6 +34,7 @@
 -define(PARK_DELAY_CHECK_TIME_KEY, <<"valet_reservation_cleanup_time_ms">>).
 -define(PARK_DELAY_CHECK_TIME, kapps_config:get_integer(?MOD_CONFIG_CAT, ?PARK_DELAY_CHECK_TIME_KEY, ?MILLISECONDS_IN_SECOND * 3)).
 -define(PARKING_APP_NAME, <<"park">>).
+-define(MAX_SLOT_NUMBER_KEY, <<"max_slot_number">>).
 
 %%------------------------------------------------------------------------------
 %% @doc Entry point for this module sends an arbitrary response back to the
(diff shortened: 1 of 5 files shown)


applications/callflow/src/module/cf_park.erl
```diff
@@ -118,8 +119,11 @@ handle_nomatch_with_empty_referred_to(Data, Call, PresenceType, ParkedCalls, Slo
 
 -spec direct_park(kz_term:ne_binary(), kz_json:object(), kz_json:object(), kz_json:object(), kapps_call:call()) -> 'ok'.
 direct_park(SlotNumber, Slot, ParkedCalls, Data, Call) ->
-    case save_slot(SlotNumber, Slot, ParkedCalls, Call) of
+    MaxSlotNumber = kz_json:get_integer_value(?MAX_SLOT_NUMBER_KEY, Data),
+    case save_slot(SlotNumber, MaxSlotNumber, Slot, ParkedCalls, Call) of
         {'ok', _} -> parked_call(SlotNumber, Slot, Data, Call);
+        {'error', 'max_slots_exceeded'} ->
+            cf_exe:continue(Call);
```
lazedo: use `cf_exe:continue/2` with a key (maybe `MAX_SLOTS` ?) so that next action knows the reason

danielfinke: @lazedo great idea! I have added this in the last commit

danielfinke:
Did a rebase an hour ago to address a circle: analyze issue with respect to dialyzer

applications/callflow/src/module/cf_park.erl
lazedo: @danielfinke was thinking in a known key like `MAX_SLOTS` not the number by itself

danielfinke: @lazedo yup, that's what I did. Added:
`-define(MAX_SLOT_EXCEEDED, 'max_slot_exceeded').`
and `?MAX_SLOT_EXCEEDED` is the key that is used for `cf_exe:continue/2`

lazedo: ah! got fooled by kz_term:to_binary ( `kz_term:to_binary(?MAX_SLOT_EXCEEDED)` ).
can we put this in uppercase, please ? `kz_term:to_upper_binary` instead of `kz_term:to_binary`. most continue keys are in uppercase as the returned values from commands

danielfinke: @lazedo OK, I have uppercased the continue key

jamesaimonetti:
Merged

PR #5765, opened 2019-05-09
Address content type changes

jamesaimonetti:
Other dialyzer fixes as well


lazedo:
@jamesaimonetti merging this as the `release` seems fixed

PR #5775, opened 2019-05-10
[master] Webhook payload samples

icehess:


applications/crossbar/src/modules/cb_webhooks.erl
```diff
@@ -356,6 +394,51 @@ summary_available(Context) ->
...
+        {'error', 'not_found'} ->
+            crossbar_util:response_bad_identifier(WebhookName, Context);
+        {'error', _} ->
+            cb_context:add_system_error('datastore_fault', Context)
+    end.
+
+-spec fetch_webhook_samples(kz_term:ne_binary() | kz_term:ne_binaries()) ->
+                                   {'ok', kz_json:objects()} |
+                                   {'error', any()}.
+fetch_webhook_samples(<<"webhooks_all">>) ->
```
jamesaimonetti: I think having to fetch `/v2/.../webhooks/samples/all` is a bit strange vs listing available samples via `/v2/.../webhooks/samples/` then querying the specific sample you want to see via `/webhooks/samples/{SAMPLE_ID}`.

lazedo:
@icehess can you address @jamesaimonetti comment and rebase ? thanks

applications/crossbar/src/modules/cb_webhooks.erl
```diff
@@ -193,6 +237,8 @@ validate_webhooks(Context, ?HTTP_PATCH) ->
 -spec validate(cb_context:context(), path_token()) -> cb_context:context().
 validate(Context, ?PATH_TOKEN_ATTEMPTS) ->
     summary_attempts(Context, 'undefined');
+validate(Context, ?PATH_TOKEN_SAMPLES) ->
+    fetch_webhook_samples(Context, <<"webhooks_all">>);
```
lazedo: shouldn't this return a list of available webhooks instead of `all` webhooks ?

icehess: GET `/webhooks` is returning all available webhooks. This one is for fetching all samples in request per what aux team (or ui tema? idk) requested

lazedo: it should be the list of samples not all the samples

PR #5786, opened 2019-05-14
HELP-39757: allow ledger amounts to include fractional values

jamesaimonetti:


applications/crossbar/src/modules/cb_ledgers.erl
```diff
@@ -247,16 +252,19 @@ put(Context, Action) ->
            ,kz_json:get_ne_json_value(<<"metadata">>, ReqData, kz_json:new())
            }
           ,{fun kz_ledger:set_dollar_amount/2
-           ,abs(kz_json:get_integer_value(<<"amount">>, ReqData, 0))
+           ,Dollars
+           }
+          ,{fun kz_ledger:set_unit_amount/2
```
k-anderson: @jamesaimonetti why set both units and dollars?

core/kazoo_documents/src/kzd_ledgers.erl
```diff
@@ -255,19 +259,19 @@ set_account_name(Doc, AccountName) ->
...
 %%------------------------------------------------------------------------------
--spec unit_amount(doc()) -> kz_term:api_integer().
+-spec unit_amount(doc()) -> kz_term:api_number().
 unit_amount(Doc) ->
     unit_amount(Doc, 'undefined').
 
 -spec unit_amount(doc(), Default) -> kz_currency:units() | Default.
 unit_amount(Doc, Default) ->
-    kz_json:get_integer_value(?AMOUNT, Doc, Default).
+    kz_json:get_number_value(?AMOUNT, Doc, Default).
```
k-anderson: @jamesaimonetti why number value here?  this should be in units and units should only be integers.

core/kazoo_documents/src/kzd_ledgers.erl
```diff
@@ -255,19 +259,19 @@ set_account_name(Doc, AccountName) ->
...
 
 %%------------------------------------------------------------------------------
 %% @doc
 %% @end
 %%------------------------------------------------------------------------------
--spec set_unit_amount(doc(), integer()) -> doc().
+-spec set_unit_amount(doc(), number()) -> doc().
```
k-anderson: @jamesaimonetti actually, the second argument should be kz_currency:units().  Same as above tho, would not expect anything other than an integer here.

core/kazoo_documents/src/kzd_ledgers.erl
```diff
@@ -293,6 +297,22 @@ set_dollar_amount(Doc, Amount) ->
     Units = kz_currency:dollars_to_units(Amount),
     kz_json:set_value(?AMOUNT, Units, Doc).
 
+%%------------------------------------------------------------------------------
+%% @doc
+%% @end
+%%------------------------------------------------------------------------------
+-spec amount(doc()) -> kz_term:api_number().
+amount(Doc) ->
```
k-anderson: @jamesaimonetti I was attempting to keep the requests for the amount to explicitly be for units or dollars, always storing units on the document so that the upstream code is more explicit.

applications/crossbar/src/modules/cb_ledgers.erl
jamesaimonetti: Because we have them...

core/kazoo_documents/src/kzd_ledgers.erl
jamesaimonetti: Yeah, I didn't realize you were changing how the kzd modules work - generally they map accessor functions to schema fields. So when I saw `<<"amount">>` i changed it to be number instead of integer (and was curious why you called it unit_amount at first.

The trouble is, this module now breaks expectations when using it because we can't fetch "amount" and expect it to be the schema's definition of "amount" - it is now overridden to represent the "units".

core/kazoo_documents/src/kzd_ledgers.erl
jamesaimonetti: See other comment about breaking expectations

k-anderson:
@jamesaimonetti looks good to me, once CI has completed this should be ready! :)

PR #5788, opened 2019-05-14
add tokens consumed and remaining stats to crossbar response

mk1s:
Adds new fields to crossbar response envelope fields for debugging:

`crossbar_tokens_consumed`: the token cost of the request
`crossbar_tokens_remaining`: remaining tokens for this client

### DIFF
--- a/applications/crossbar/src/api_util.erl
+++ b/applications/crossbar/src/api_util.erl
@@ -1569,6 +1569,11 @@ maybe_set_pull_response_stream({FileLength, TransportFun}, Req, Context)
 maybe_set_pull_response_stream(Other, Req, Context) ->
     {Other, Req, Context}.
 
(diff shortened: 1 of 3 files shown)


applications/crossbar/src/api_util.erl
```diff
@@ -1582,6 +1582,8 @@ do_create_resp_envelope(Context) ->
     Resp = case cb_context:response(Context) of
                {'ok', RespData} ->
                    [{<<"auth_token">>, cb_context:auth_token(Context)}
+                   ,{<<"crossbar_tokens_remaining">>, cb_modules_util:tokens_remaining(Context)}
```
k-anderson: @mk1s This is awesome, but I would prefer:
```
"tokens": {
   "remaining": 100,
   "consumed": 3
}
```

Thoughts?

lazedo:
@mk1s please rebase. thanks

PR #5791, opened 2019-05-16
Prevent infinite loops when undefined is passed to flush_cache_doc or…

vladimirralev:
… when doc doesn't have ID

Every now and then it happens that when you create a new doc it may not have an id yet and if you hit a DB error such as conflict flush_cache_doc is called in an infinite loop. Very difficult to notice it, so I added a guard to detect it and print an error so people can be aware.


jamesaimonetti:
@vladimirralev thanks! I cleaned up the specs a bit but good catch. Merged

PR #5806, opened 2019-05-21
fix record_min_sec & rename folder

lazedo:


lazedo:
i think dialyzer fails because of rename

PR #5807, opened 2019-05-21
HELP-43911 merge

harenson:


core/kazoo_provisioner/src/provisioner_v5.erl
```diff
@@ -218,27 +220,51 @@ set_owner(Device) ->
...
-                               ,{<<"line_keys">>, settings_line_keys(Device)}
-                               ,{<<"combo_keys">>, settings_combo_keys(Device)}
-                               ]),
-    kz_json:from_list(Props);
-settings(Device, _) ->
-    Props = props:filter_empty([{<<"lines">>, settings_lines(Device)}
-                               ]),
-    kz_json:from_list(Props).
+-spec settings(doc()) -> kz_json:object().
+settings(Doc) ->
```
jamesaimonetti: ok, so settings/1 should be called within the context of a device doc which should get the user/account settings and do its thing and if a user, should fetch owned_by devices and loop over them and if the property is already set at the account level in provisioner then don't send it on the device update.

core/kazoo_provisioner/src/provisioner_v5.erl
```diff
@@ -267,11 +265,20 @@ settings_lines(Device) ->
...
+    AccountId = kz_doc:account_id(Device),
+    {'ok', Account} = kzd_accounts:fetch(AccountId),
+    {'ok', Owner} = case kzd_devices:owner_id(Device) of
+                        'undefined' -> {'ok', kz_json:new()};
+                        OwnerId -> kzd_users:fetch(AccountId, OwnerId)
+                    end,
+    Enabled = kzd_devices:enabled(Device)
+        andalso kzd_accounts:is_enabled(Account)
+        andalso kzd_users:enabled(Owner),
+    DisplayName = case kzd_accounts:name(Account) of
```
jamesaimonetti: I think you want to take the user's name into account. I would maybe create a list of possible names, `[NameFromUser, NameFromDevice, NameFromAccount]` for display name.

jamesaimonetti: Then go until you get a non-empty name. See `kz_endpoint:merge_attribute_caller_id/4` for an idea.

PR #5812, opened 2019-05-24
[4.3] fix repeated key and not fullfilling page size

icehess:
Fixes some issues in `crossbar_view`:

*) When crossbar module is filtering some result, crossbar view was querying the next db. It should query the same db again if the last is key is not undefined.
*) In situation which there are some result with the same key, it should stop querying and return.


lazedo:
@icehess ping

lazedo:
@icehess since this never left `draft` and its a bit old, i'm closing it. reopen if needed after the split

PR #5815, opened 2019-05-27
PISTON-795: add crossbar.cdrs maximum_range override

danielfinke:
There is already max_range config for all crossbar listing views, but it would be nice to extend this for solely CDRs as it is common to request longer CDR time ranges. This is added in this PR


jamesaimonetti:
Merged, good call.

PR #5818, opened 2019-05-28
Moved config storage of cluster_id

yonkoma:
Moved from provsioner.cluster_id to cluster.cluster_id
Changed other uses of cluster_id to also use new consolidated location


applications/crossbar/src/modules/phonebook.erl
```diff
@@ -17,6 +17,7 @@
 -include_lib("kazoo_number_manager/include/knm_port_request.hrl").
 
 -define(MOD_CONFIG_CAT, <<"crossbar.phonebook">>).
+-define(MOD_CONFIG_CLUSTER, <<"cluster">>).
```
lazedo: move this to `crossbar.hrl`

core/kazoo_provisioner/src/provisioner_v5.hrl
```diff
@@ -4,6 +4,7 @@
 -include_lib("kazoo_stdlib/include/kz_databases.hrl").
 
 -define(MOD_CONFIG_CAT, <<"provisioner">>).
+-define(MOD_CONFIG_CLUSTER, <<"cluster">>).
```
lazedo: delete this

yonkoma: Where should provisioner_v5.erl reference that from then?

lazedo: include `crossbar.hrl`

lazedo: a better way would be to create an accessor ( `kzd_cluster` ) in kazoo_documents and use it

lazedo: it would also be used in the webhooks pr

core/kazoo_documents/src/kzd_cluster.erl
```diff
@@ -0,0 +1,23 @@
...
+
+-spec id() -> kz_term:ne_binary().
+id() ->
+    case kapps_config:get_ne_binary_value(?MOD_CONFIG_CLUSTER, <<"cluster_id">>) of
+        'undefined' ->
+            ClusterId = kz_binary:rand_hex(16),
+            {'ok', _JObj} = kapps_config:set_default(?MOD_CONFIG_CLUSTER, <<"cluster_id">>, ClusterId),
+            ClusterId;
+        ClusterId -> ClusterId
+    end.
```
lazedo: does `kapps_config:get_ne_binary_value(?MOD_CONFIG_CLUSTER, <<"cluster_id">>, kz_binary:rand_hex(16), <<"default">>)` not work ? i think it will update and return it if its not created.

yonkoma: Possibly, I just modified already existing code.
It also looks like it's `kapps_config:get_ne_binary`.

lazedo: you're correct, that's a  `kz_json` thing

yonkoma: What is last parameter (`<<"default">>`) in that call for?

lazedo: it's the node, otherwise you'll get/set from node()

PR #5822, opened 2019-05-29
[master] Audit summary per day

icehess:
4.3: https://github.com/2600hz/kazoo/pull/5821


core/kazoo_modb/priv/couchdb/views/services.json
```diff
@@ -10,7 +10,143 @@
...
+                "          'addition': changedBillable > 0 ? changedBillable : 0,",
+                "          'last_timestamp': doc.pvt_created,",
+                "          'quantity': item.billable || 0,",
+                "          'removal': changedBillable < 0 ? Math.abs(changedBillable) : 0,",
+                "          'sum_quantity': false",
+                "        };",
+                "        emit(day, result);",
+                "      }",
+                "    }",
+                "  } else if ((doc.pvt_type === 'ledger') && (doc.source.service === 'per-minute-voip')) {",
```
lazedo: is there a reason to only include per-minute-voip ledger ?

core/kazoo_modb/priv/couchdb/views/services.json
```diff
@@ -10,7 +10,143 @@
...
+                "        emit([item.item, day], {",
+                "          'addition': changedBillable > 0 ? changedBillable : 0,",
+                "          'last_timestamp': doc.pvt_created,",
+                "          'quantity': item.billable || 0,",
+                "          'removal': changedBillable < 0 ? Math.abs(changedBillable) : 0,",
+                "          'sum_quantity': false",
+                "        });",
+                "      }",
+                "    }",
+                "  } else if ((doc.pvt_type === 'ledger') && (doc.source.service === 'per-minute-voip')) {",
```
lazedo: same here

core/kazoo_modb/priv/couchdb/views/services.json
icehess: because the UI needs to show per-minute-voip, and audit log is not generated for per-minute.

lazedo: what the UI shows should not determine the api, i don't see a reason not to include other ledgers here.

core/kazoo_modb/priv/couchdb/views/services.json
```diff
@@ -10,7 +10,143 @@
     "language": "javascript",
     "views": {
         "audit_logs_by_creation": {
-            "map": "function(doc) { if ( doc.pvt_deleted || doc.pvt_type != 'audit_log') return; emit(doc.pvt_created, {id: doc._id, status: doc.status, reason: doc.reason, message: doc.message, type: doc.audit.changes.type, summary: doc.invoice.summary}); }"
+            "map": [
+                "function(doc) {",
+                "  if ( doc.pvt_deleted || doc.pvt_type != 'audit_log') return;",
+                "  emit(doc.pvt_created, {",
```
jamesaimonetti: If two audit logs are created at the same second, this could lead to the lost entries when paginating splits on this timestamp. I would make it a composite key of `[doc.pvt_created, doc._id]` to ensure we can enumerate all audit log docs with identical pvt_created fields.

PR #5823, opened 2019-05-29
PISTON-769: do not rely on Hangup-Cause, will be set if transferred to vm

danielfinke:
We have noticed that when somebody is transferred to a voicemail box, the recording review prompt will not be played if the press a DTMF to save the voicemail. This is because the Hangup-Cause prop is set in the RECORD_STOP call event. It is set due to variable_bridge_hangup_cause being set (it is set when the bridged B leg of the call is down: https://lists.freeswitch.org/pipermail/freeswitch-users/2015-January/110430.html). Switching to kapps_call_command:b_channel_status/1 ensures that we get the true state of the call.


lazedo:
@danielfinke this adds unnecessary traffic to amqp.

danielfinke:
@lazedo do you have another idea?

lazedo:
@danielfinke check #5824 for master. your change looks good for 4.3

danielfinke:
@lazedo OK cool. Is that something that can be backported to 4.3 or does it depend on things unique to master

lazedo:
@danielfinke no backport of #5824

danielfinke:
@lazedo OK, going to close this and make a new PR for 4.3

PR #5826, opened 2019-05-30
move media proxies out of core

lazedo:


lazedo:
next iteration will deal with file/tts cache and streaming

PR #5831, opened 2019-05-31
HELP-44059: [master] flush registration when device is deleted using api_v1

icehess:
4.3: https://github.com/2600hz/kazoo/pull/5830


lazedo:
i think were going to deprecate `v1`, not sure if this should be merged @k-anderson

lazedo:
we can deprecate later

PR #5835, opened 2019-06-03
[4.3] PISTON-769: do not rely on Hangup-Cause, will be set if transferred to vm

danielfinke:
4.3 version of https://github.com/2600hz/kazoo/pull/5823


jamesaimonetti:
Merged

PR #5839, opened 2019-06-06
Random commits

lazedo:


applications/crossbar/src/api_resource.erl
```diff
@@ -916,6 +923,11 @@ send_file(Req, Context) ->
     lager:debug("run: send_file"),
     api_util:create_pull_response(Req, Context, fun api_util:create_resp_file/2).
 
+-spec to_xml(cowboy_req:req(), cb_context:context()) -> api_util:pull_file_response_return().
+to_xml(Req, Context) ->
+    lager:debug("run: send_file"),
```
jamesaimonetti: Should this be 'run: to_xml' ?

PR #5840, opened 2019-06-07
OpenAPI 3

icehess:
This is an attempt to generate valid OpenAPI 3. Swagger 2.0 is not supporting JSON Schema Draft 4 (anyOf, allOf, ....), so we need to use OpenApi 3.

Currently this generates exact replica of Swagger2, fixes (OpenAPI-ify json schema) and saves the specs to its own yaml file (per api endpoint).

More important thing that this spec file is VALID! This has been test by [openapi-spec-validator](https://github.com/p1c2u/openapi-spec-validator), [speccy](https://github.com/wework/speccy) and [swagger-cli](https://github.com/APIDevTools/swagger-cli).


core/kazoo_stdlib/src/kz_maps.erl
```diff
@@ -73,15 +73,18 @@ merge_right(_K, {'both', _Left, Right}) -> {'ok', Right}.
...
     case maps:get(Key, Map, {?MODULE, Default}) of
         {?MODULE, Default} ->
             Default;
-        NestedMap ->
-            get(Rest, NestedMap, Default)
+        #{} = NestedMap ->
+            get(Rest, NestedMap, Default);
+        Value when Rest =:= [] ->
+            Value;
+        _ -> Default
```
lazedo: why this change ? the current behaviour is the same as for kz_json, ie, it will crash if you try to get with list of keys where current key is not a map. if that's the case, then add a new method (find?) that will try/catch the get and return `Default`

make/deps.mk
```diff
@@ -169,6 +170,10 @@ dep_trie = hex 1.7.5
 # dep_wsock = git https://github.com/madtrick/wsock 1.1.7
 # appears unused
 
+dep_yamerl = git https://github.com/icehess/yamerl 6f52ab8a93d677773102ab8dee050e97f3e8cdd9
```
lazedo: lets create 2600hz/erlang-yamerl and apply any changes there

core/kazoo_stdlib/src/kz_maps.erl
icehess: Honestly I don't remember why, I changed it months ago. Anyway it seems that I don't use it anywhere in this PR, so i'm reverting it.

lazedo:
@icehess i cloned yamerl to 2600hz/erlang-yamerl, change this and refer v0.7.0 release

PR #5852, opened 2019-06-11
Password reset tokens will now expire 1 hour after creation

Spencerhutch:
Previously, password reset tokens would always be valid and would never be cleaned up. 

This work is for cleaning up previous password reset tokens when a new one is requested. Also with this work a password reset token that was requested over an hour ago will be invalidated.

### DIFF
--- a/applications/crossbar/src/modules/cb_user_auth.erl
+++ b/applications/crossbar/src/modules/cb_user_auth.erl
@@ -39,7 +39,7 @@
             Ok -> Ok
         end).
 -define(RESET_PVT_TYPE, <<"password_reset">>).
--define(RESET_TIMEOUT, 3600).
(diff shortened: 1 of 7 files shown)


applications/crossbar/src/modules/cb_user_auth.erl
```diff
@@ -37,6 +39,7 @@
             Ok -> Ok
         end).
 -define(RESET_PVT_TYPE, <<"password_reset">>).
+-define(RESET_TIMEOUT, 3600).
```
icehess: Can you please make this timeout configurable:

```suggestion
-define(RESET_TIMEOUT, kapps_config:get_pos_integer(<<"crossbar.auth">>, <<"password_reset_expiry_s">>, ?SECONDS_IN_HOUR)).
```

jamesaimonetti:
@Spencerhutch can you make the timeout configurable as suggested (and probably rebase against master). Then we can get this merged!

jamesaimonetti:
Please re-open with corrections when ready!

PR #5853, opened 2019-06-11
Build on macos

corporealshift:
Add documentation for building kazoo on mac osx (for development). Add a configurable timeout around file conversion (default 10 seconds). Move an endif to fix some missing macro definitions.


jamesaimonetti:
@corporealshift thanks! Timely too as another dev was setting up their MacOS environment. Manually merged.

PR #5857, opened 2019-06-13
PISTON-835:Posible fix for kz_json:to_map/1 for kz_json:objects() with the same keys

bradfordben:
There is a issue with to_map in kz_json.

```kz_json:to_map([{[{<<"A">>, 1}, {<<"B">>, 2}]}, {[{<<"A">>, 8}, {<<"B">>, 9}]}])```

Will return:

```#{<<"A">> => 8,<<"B">> => 9}```

Possible solution is to return:

```[#{<<"A">> => 8,<<"B">> => 9},#{<<"A">> => 1,<<"B">> => 2}]```

But im open to suggestions. Other option is to just remove the case that deals with lists and cause a crash.

### DIFF
--- a/core/kazoo_stdlib/src/kz_json.erl
+++ b/core/kazoo_stdlib/src/kz_json.erl
@@ -685,9 +685,9 @@ recursive_to_proplist(Else) -> Else.
 %% @end
 %%------------------------------------------------------------------------------
 
--spec to_map(object() | objects()) -> map().
+-spec to_map(object() | objects()) -> map() | list(map()).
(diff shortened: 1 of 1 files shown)


lazedo:
how about using `jiffy:decode(kz_json:encode(J), [return_maps])` ?

bradfordben:
Looks good. I forgot to look at possible jiffy functions 👍
I have updated the code and pushed a new commit. 
See you at KazooCon

jamesaimonetti:
Added a unit test as well. Merged manually

bradfordben:
Thanks 👍

PR #5862, opened 2019-06-19
user defined blacklist

sergey-safarov:
Added user defined blacklist.
Blacklist may used as white list.
Blacklist applied to callflow module/actions. Also added blacklist cache.
Defined `skip_human` and `ask_human` new blacklist actions.

Feature wants to resolve FS-11894
https://freeswitch.org/jira/browse/FS-11894

### DIFF
--- a/applications/callflow/src/callflow.hrl
+++ b/applications/callflow/src/callflow.hrl
@@ -37,6 +37,7 @@
 -define(LIST_BY_NUMBER, <<"callflows/listing_by_number">>).
 -define(LIST_BY_PATTERN, <<"callflows/listing_by_pattern">>).
 
+-define(NO_MATCH_BL, <<"no_match">>).
 -define(NO_MATCH_CF, <<"no_match">>).
 
 -define(DEFAULT_TIMEOUT_S, ?BRIDGE_DEFAULT_SYSTEM_TIMEOUT_S).
@@ -48,6 +49,8 @@
 -define(RESTRICTED_ENDPOINT_KEY, <<"Restricted-Endpoint-ID">>).
 
 -define(RESOURCE_TYPES_HANDLED, [<<"audio">>, <<"video">>]).
+-define(BLACKLIST_PATTERNS_CACHE_KEY(AccountId, OwnerId), {'bl_patterns', AccountId, OwnerId}).
+-define(BLACKLIST_NUMBER_CACHE_KEY(AccountId, OwnerId, Number), {'bl_number', AccountId, OwnerId, Number}).
 -define(CF_FLOW_CACHE_KEY(Number, AccountId), {'cf_flow', Number, AccountId}).
 -define(CF_PATTERN_CACHE_KEY(AccountId), {'cf_patterns', AccountId}).
 
--- a/applications/callflow/src/callflow.hrl
+++ b/applications/callflow/src/callflow.hrl
@@ -6,6 +6,13 @@
 -include_lib("kazoo_call/include/kapps_call_command_types.hrl").
 -include_lib("kazoo_documents/include/kazoo_documents.hrl").
 
+-record(cf_blacklist_action, {action :: kz_term:api_binary()
+                             ,caller_name :: kz_term:api_binary()
+                             ,blacklist_strategy :: kz_term:api_binary()
+                             }).
+-type cf_blacklist_action() :: #cf_blacklist_action{}.
+-type blacklist_action() :: {'ok', cf_blacklist_action()} | {'error', any()}.
+
 -type cf_exe_response() :: {'stop'} |
                            {'continue'} |
                            {'continue', integer()} |
--- a/applications/callflow/src/cf_blacklist.erl
+++ b/applications/callflow/src/cf_blacklist.erl
@@ -1,0 +1,179 @@
+%%%-----------------------------------------------------------------------------
+%%% @copyright (C) 2011-2019, 2600Hz
+%%% @doc
+%%% @author Sergey Safarov sponsored by Audian
+%%% @end
+%%% This Source Code Form is subject to the terms of the Mozilla Public
+%%% License, v. 2.0. If a copy of the MPL was not distributed with this
+%%% file, You can obtain one at https://mozilla.org/MPL/2.0/.
+%%%
+%%%-----------------------------------------------------------------------------
+-module(cf_blacklist).
+
+-export([lookup/4]).
+
+-include("callflow.hrl").
+
+-record(pattern, {action :: kz_term:ne_binary()
+                 ,caller_name :: kz_term:api_ne_binary()
+                 ,has_groups = 'false' :: boolean()
+                 ,names = [] :: kz_term:ne_binaries()
+                 ,regex :: re:mp() | 'undefined'
+                 }).
+
+-type pattern() :: #pattern{}.
+-type patterns() :: [pattern()].
+-type patterns_view_item() :: {kz_term:ne_binary(), kz_json:object()}.
+-type patterns_view_items() :: [patterns_view_item()].
+
+-type lookup_ret_number() :: {'ok', kz_term:ne_binary(), kz_term:api_ne_binary(), boolean()} | {'error', any()}.
+-type lookup_ret_patterns() :: {'ok', patterns()} | {'error', any()}.
+
+-spec lookup(kz_term:ne_binary(), kz_term:ne_binary(), kz_term:api_ne_binary(), kz_term:ne_binary()) -> lookup_ret_number().
+lookup(Number, AccountId, OwnerId, Strategy) when not is_binary(Number) ->
+    lookup(kz_term:to_binary(Number), AccountId, OwnerId, Strategy);
+lookup(<<>>, _, _, _) ->
+    {'error', 'invalid_number'};
+lookup(Number, AccountId, OwnerId, Strategy) ->
+    case kz_cache:fetch_local(?CACHE_NAME, ?BLACKLIST_NUMBER_CACHE_KEY(AccountId, OwnerId, Number)) of
+        {'ok', 'undefined'} -> {'error', 'not_found'};
+        {'ok', {Action, CallerName}} -> {'ok', Action, CallerName, 'false'};
+        {'error', 'not_found'} -> do_lookup(Number, AccountId, OwnerId, Strategy)
+    end.
+
+-spec do_lookup(kz_term:ne_binary(), kz_term:ne_binary(), kz_term:api_ne_binary(), kz_term:ne_binary()) -> lookup_ret_number().
+do_lookup(Number, AccountId, OwnerId, Strategy) ->
+    case kzd_blacklists:fetch_number(AccountId, OwnerId, Number, #{<<"enabled">> => 'true', <<"brief">> => 'true', <<"strategy">> => Strategy}) of
+        {'error', 'not_found'} when Number =/= ?NO_MATCH_BL ->
+            lookup_patterns(Number, AccountId, OwnerId, Strategy);
+        {'error', 'not_found'} = Responce -> Responce;
+        {'ok', [JObj] } ->
+            Action = kz_json:get_ne_binary_value(<<"action">>, JObj, <<"block">>),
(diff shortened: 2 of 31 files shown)


sergey-safarov:
Warning: github list commit in wrong order.
If you want look in correct order, then please use this [link](https://github.com/sergey-safarov/kazoo/commits/user_blacklist3)

sergey-safarov:
I see `kazoo_endpoints` added into FreeeSwitch mod_kazoo module.
I checked `group_confirm_key` feature works properly when call originated like
```
[group_confirm_key='1',group_confirm_cancel_timeout='true',group_confirm_read_timeout='7000',group_confirm_file='ivr/ivr-welcome_to.wav',ignore_early_media='true']kz/sip:user@domain 
```

applications/callflow/src/callflow.hrl
```diff
@@ -6,6 +6,13 @@
 -include_lib("kazoo_call/include/kapps_call_command_types.hrl").
 -include_lib("kazoo_documents/include/kazoo_documents.hrl").
 
+-record(cf_bl_action, {action :: kz_term:api_binary()
```
jamesaimonetti: I don't think `bl` is sufficient; I would prefer to see `blacklist` to immediately remind the reader what is being written about here.

applications/callflow/src/cf_blacklist.erl
```diff
@@ -0,0 +1,175 @@
...
+
+-export([lookup/4]).
+
+-include("callflow.hrl").
+
+-record(pattern, {action :: kz_term:ne_binary()
+                 ,caller_name :: kz_term:api_ne_binary()
+                 ,has_groups :: boolean()
+                 ,names = [] :: kz_term:ne_binaries()
+                 ,regex :: re:mp()
```
jamesaimonetti: `re:mp() | 'undefined'` I think; 'undefined' must be explicitly added now

applications/callflow/src/cf_blacklist.erl
```diff
@@ -0,0 +1,175 @@
...
+%%%-----------------------------------------------------------------------------
+-module(cf_blacklist).
+
+-export([lookup/4]).
+
+-include("callflow.hrl").
+
+-record(pattern, {action :: kz_term:ne_binary()
+                 ,caller_name :: kz_term:api_ne_binary()
+                 ,has_groups :: boolean()
```
jamesaimonetti: Should either default this to true or false or update type to `boolean() | 'undefined'`.

applications/callflow/src/cf_exe.erl
```diff
@@ -805,7 +805,52 @@ update_actions(Action, Call) ->
...
+            cf_module_task(CFModule, Data, UpdatedCall, AMQPWorker);
+        #cf_bl_action{caller_name = CallerName, action = ActionValue} ->
+            UpdatedCall1 = kapps_call:set_user_bl_action(ActionValue, Call),
+            UpdatedCall2 = kapps_call:set_caller_id_name(CallerName, UpdatedCall1),
+            cf_module_task(CFModule, Data, UpdatedCall2, AMQPWorker)
+    end.
+
+-spec cf_try_user_blacklists(atom(), kz_json:object(), kz_term:api_ne_binary(), kapps_call:call()) -> cf_bl_action().
+cf_try_user_blacklists(CFModule, Data, AccountBlAction, Call) ->
+    try CFModule:bl_action(Data, Call) of
```
jamesaimonetti: If this is going to be applied to all callflow action modules, it should be an optional behaviour callback of `gen_cf_action` and the code here should check if the function is exported (using `kz_module:is_exported/3`) before calling it. The only define the function in the callflow actions that actually support it.

applications/callflow/src/cf_blacklist.erl
```diff
@@ -0,0 +1,175 @@
...
+        _ -> {'ok', Action, CallerName, Number =:= ?NO_MATCH_BL}
+    end.
+
+-spec maybe_use_nomatch(kz_term:ne_binary(), kz_term:ne_binary(), kz_term:api_ne_binary(), kz_term:ne_binary()) -> lookup_ret_number().
+maybe_use_nomatch(Number, AccountId, OwnerId, Strategy) ->
+    case knm_converters:is_reconcilable(Number, AccountId) of
+        'true' -> try_nomatch(Number, AccountId, OwnerId, Strategy);
+        'false' ->
+            lager:info("can't use no_match: number not all digits: ~s", [Number]),
+            cache_number(AccountId, OwnerId, Number, 'undefined', 'undefined')
```
jamesaimonetti: This breaks the contract with `cache_number/5`. Looks like `Action` is expected to be `ne_binary()` in the spec.

applications/callflow/src/cf_blacklist.erl
```diff
@@ -0,0 +1,175 @@
...
+    end.
+
+-spec try_nomatch(kz_term:ne_binary(), kz_term:ne_binary(), kz_term:api_ne_binary(), kz_term:ne_binary()) -> lookup_ret_number().
+try_nomatch(Number, AccountId, OwnerId, Strategy) ->
+    case lookup(?NO_MATCH_BL, AccountId, OwnerId, Strategy) of
+        {'error', _} -> cache_number(AccountId, OwnerId, Number, 'undefined', 'undefined');
```
jamesaimonetti: Similar contract violation here.

core/kazoo_documents/src/kzd_blacklists.erl
```diff
@@ -5,23 +5,271 @@
...
+      ,{'key', ['null', Type, Number]}
+      ]);
+get_view_options(?LIST_BY_OWNER, OwnerId, <<"number">> = Type, Number) ->
+    props:filter_undefined(
+      [{'group', 'true'}
+      ,{'group_level', 3}
+      ,{'key', [OwnerId, Type, Number]}
+      ]).
+
+-spec format_view_results(kz_json:objects(), map()) -> kz_json:object().
```
jamesaimonetti: return type is wrong, this does not return `kz_json:object()` but `{'ok', kz_json:object()} | {'error', 'not_found'}`

core/kazoo_documents/src/kzd_blacklists.erl
```diff
@@ -5,23 +5,271 @@
...
+format_view_results(JObjs, Options) ->
+    format_view_results(JObjs, kz_json:new(), Options).
+
+-spec format_view_results(kz_json:objects(), kz_json:object(), map()) -> kz_json:object().
```
jamesaimonetti: Same bad return type

core/kazoo_documents/src/kzd_blacklists.erl
```diff
@@ -46,6 +294,99 @@ numbers(Doc, Default) ->
...
+
+-spec owner_id(doc()) -> kz_term:api_ne_binary().
+owner_id(Doc) ->
+    owner_id(Doc, 'undefined').
+
+-spec owner_id(doc(), Default) -> kz_term:ne_binary() | Default.
+owner_id(Doc, Default) ->
+    kz_json:get_ne_binary_value([<<"owner_id">>], Doc, Default).
+
+-spec set_owner_id(doc(), kz_term:ne_binary()) -> doc().
```
jamesaimonetti: Contract disagrees with the implementation; looks like `api_ne_binary()` is allowed.

core/kazoo_documents/src/kzd_blacklists.erl
```diff
@@ -46,6 +294,99 @@ numbers(Doc, Default) ->
...
+set_owner_id(Doc, 'undefined') ->
+    kz_json:delete_key([<<"owner_id">>], Doc);
+set_owner_id(Doc, OwnerId) ->
+    kz_json:set_value([<<"owner_id">>], OwnerId, Doc).
+
+-spec set_owner_id(doc(), kz_term:ne_binary(), kz_term:ne_binary()) -> doc().
```
jamesaimonetti: Similarly, `api_ne_binary()` for second arg.

applications/callflow/src/cf_blacklist.erl
sergey-safarov: I renamed record name. But modules functions still have name `bl_action`.

applications/callflow/src/callflow.hrl
sergey-safarov: I renamed record name. But modules functions still have name `bl_action`.

sergey-safarov:
PR is updated.
Looks as failed tests not related to this PR.

sergey-safarov:
PR is rebased to clean failed tests.

lazedo:
@sergey-safarov can you explain the purpose of this PR,  and why it is touching all these files ?

sergey-safarov:
Current blacklist implementation allow only block call. And blacklist granularity is account wide.
PR add features:
1. allow define user wide blacklist;
2. allow define action `pass` and now blacklist maybe used as white list (block all except allowed numbers);
3. allow define action `skip_human` and `cf_exe` will call only module that not dials end user phones;
4. allow define action 'ask_human' and when channel is established to user phone humans can make decision to accept or decline call.

About amount changed files.
Feature implemented in `callflow` application. This applications contains about 70 modules. To get user defined blacklist need get `user_id` for callflow data for each module. That is means need extend modules to extract user_id from callflow data.

sergey-safarov:
Could you look first 17 core commits, one by one. Some of this related to `make apis`.
Rest commits related to crossbar, documentation and change request from 2600hz team.
[Commit is correct order](https://github.com/sergey-safarov/kazoo/commits/user_blacklist3).

applications/callflow/src/callflow.hrl
jamesaimonetti: Please update any `bl` to `blacklist`; no need for terseness here.

jamesaimonetti:
I still want the blacklist callback tagged as optional in the behaviour and only implemented where it is actually used. That will cut down on the number of changed modules.

applications/callflow/src/callflow.hrl
sergey-safarov: Renamed except `kzd_users:bl_strategy/2`. Reason already exist `kzd_users:blacklist_strategy/2` generated by `make apis` and imported from template.
Optionally i can not import autoenerated `kzd_users:blacklist_strategy/2`.

sergey-safarov:
@jamesaimonetti blacklist callback now optional.

sergey-safarov:
Thanks Luis @lazedo for review.
```
i feel the implementation should be done higher at routing level and not action level
```
Think that is not correct.
Imagine that we have callflow where exist `cf_user for user 1` -> `cf_user for user 2`.
User 1 defined blacklist with action "block".
User 2 defined blacklist with action "ask_human".

Now we must
1. evaluate blacklist action for whole callflow and get result `ask_human`.
2. at call flow we must not call user 1 and must evaluate blacklist actions during execution again.

Think introducing blacklist at call routing level add complexity, but does not add useful things.

lazedo:
@sergey-safarov that's why we need further discussion here.
kz_endpoint:merge would make it much easier

sergey-safarov:
Yes, you a right.
I made several iteration before I able to publish PR.
Sure for 2600Hz team also need time to thing about it.

sergey-safarov:
```
allow define action skip_human and cf_exe will call only modules that not dials end user phones;
can you provide a use case for this ?
```
I say it `soft blacklist`.
Caller is able work with [IVR](https://en.wikipedia.org/wiki/Interactive_voice_response) or voicemail but newer reach humans.

sergey-safarov:
As option.
We can defined `callflow` module config option `apply_blacklist_at_routing`. If option equal `true` value and account blacklist action is `block`, then CDR for call is not created.

bazemoreb:
Hey @lazedo - I'd talked with Darren and James about this whole thing, and I've attached our rough spec that we put together here.  I think it should help to kind of describe the overall process we were going for and also contains some of the back and forth with Darren from the Jira ticket I'd done.   Let me know!  
[User Level Screening.docx](https://github.com/2600hz/kazoo/files/3370660/User.Level.Screening.docx)

jamesaimonetti:
rebase and correct conflicts?

lazedo:
@sergey-safarov @bazemoreb lets try to resume this after the repo split

PR #5863, opened 2019-06-24
PISTON-837: Added transcription to metadata & updated docs

DustinBrett:
Transcriptions are being saved outside of the metadata key which is what is returned when information about a voicemail is pulled from crossbar. I had originally considered changing the way transcriptions are saved to move them into metadata but decided on the approach of adding it to the return data upon request so that this will work with pre-existing transcriptions. I’ve updated the crossbar api doc also because now this data will be returned when applicable.

### DIFF
--- a/applications/crossbar/doc/voicemail.md
+++ b/applications/crossbar/doc/voicemail.md
@@ -557,7 +557,11 @@ curl -v -X DELETE \
         "call_id": "79959ZDNmM2I5ZTliMzA0NzA4N2FjNjlmODA5OWVkZjUxZWU",
         "folder": "new",
         "length": 3140,
-        "media_id": "201605-6aadef09f6fcf5fd8bcdfca312e923ba"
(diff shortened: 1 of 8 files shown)


DustinBrett:
Build error doesn't seem related to my changes.

DustinBrett:
Also added a `transcribed` boolean for views listing messages so that you know which ones have messages.

jamesaimonetti:
Try a rebase against master to see if that addresses build errors.

Shouldn't this PR also update `mailbox_message.json` schema with a `transcription` object and attending properties?

DustinBrett:
I will look into those comments. Thanks James.

DustinBrett:
> Try a rebase against master to see if that addresses build errors.
> 
> Shouldn't this PR also update `mailbox_message.json` schema with a `transcription` object and attending properties?

Rebased & added `transcription` to schema.

PR #5865, opened 2019-06-24
HELP-44744: allow chaining children off temporal route actions

jamesaimonetti:
Previously, actions of `menu`, `enable`, `disable`, and `reset` would
terminate the call after processing. Upon discussion, there wasn't a
compelling reason to continue doing this (and restricting the ability
to chain more actions off the action).

The usecase is that when there is an "emergency" a callflow can be
dialed to disable the main temporal route rules and enable the
particular "emergency" temporal route rule.


jamesaimonetti:
See 4.3 PR #5864

PR #5866, opened 2019-06-24
[master] HELP-42849: fixed quickcall shortdial correction when no cid-number specified

swysor:


lazedo:
@swysor quickcall uses a device/owner, it should use the device/user caller-id if one is not provided. i don't feel this is the correct approach, maybe there's something missing elsewhere

swysor:
> @swysor quickcall uses a device/owner, it should use the device/user caller-id if one is not provided. i don't feel this is the correct approach, maybe there's something missing elsewhere

It will, this is just gathering the *default* values if none are specified in the query.

swysor:
E.g. get_cid_number and get_cid_name expect a default if not found in context. THe use case for this is when you make a device quick call, but specify a shortdial number, short dial correction fails because we select the <<"internal">> cid because we could not reconcile the called number.

lazedo:
the `default` values are from device/user and are (or should be) handled in callflow (if not already) not here. the problem may be in the reconcile regex or the way it is handled in cf_route_win. again, this shouldn't be handled here. the most to do here would be to use reconcile (with the account) but event that should be handled in callflow if not already

PR #5873, opened 2019-06-25
KAZOO-6108: check order of users after sort/filter

jamesaimonetti:
Mostly adding testing to ensure sort/filter is working as expected


jamesaimonetti:
See 4.3 backport in #5874

PR #5878, opened 2019-06-27
OpenAPI v3.0 JSON

fireball460:
Swagger / OpenAPI v3.0 with some corrections


icehess:
Thank you for contribution. Unfortunately this pull request is not acceptable in this form. Swagger file is huge file and it is hard to see the changes you made.

We recently start generating valid OpenAPI 3 spec file (read more about it below) and the conversion to OpenAPI 3 in this PR is not necessary.

Other than converting to version 3, you mentioned you made some correction, and it not clear what was fixed. If you can just point us to some of the errors you found we can help you better to show you how to properly make the fixes.

Also keep in mind you can't edit the spec file directly as this file is [automatically generated](https://github.com/2600hz/kazoo/blob/master/core/kazoo_ast/src/cb_api_endpoints.erl).

As you may notice we recently generate a valid OpenAPI 3.0 spec file. It still considered as work in progress and required more work.

The spec file was break into multiple files as far as possible. The main file is [`applications/crossbar/priv/oas3/openapi.yml`](https://github.com/2600hz/kazoo/blob/master/applications/crossbar/priv/oas3/openapi.yml).

All paths are in their own file based on their endpoint name. `applications/crossbar/priv/oas3/paths/`

All schemas and parameters was move to their own file:
* `applications/crossbar/priv/oas3/oas3-parameters.yml`: Theses parameters are extracted from URL paths params.
* `applications/crossbar/priv/oas3/oas3-schemas.yml`: This file contains all JSON schemas in `applications/crossbar/priv/couchdb/schemas` (except `kapi_*` files). Schemas are converted to OpenAPI JSON Schema version.

lazedo:
@fireball460 closing this based on @icehess comments.

PR #5880, opened 2019-06-28
bputtick - ledger view unique list_by_source key

bputtick:
cb_ledger uses the "list_by_source" view to iterate through the records which has the "key" defined as [{service}, {pvt_created}] - however if there are 2 or more service entries that were created at the same time (multiple legs) this key is no longer unique and when this falls on a chunked boundary, crossbar misinterprets this as EOF returning an incomplete result.

The fix for this is to emit doc._id as the 3rd field to the key so that it remains unique and will still be sorted correctly.


jamesaimonetti:
@bputtick fixed the test and merged, thanks!

PR #5881, opened 2019-06-28
Export function to handle auto compaction triggers

harenson:


applications/tasks/src/kz_tasks_trigger.erl
```diff
@@ -258,12 +258,14 @@ browse_dbs_for_triggers(Ref) ->
...
+    lager:debug("getting databases list and sorting them by disk size"),
     Sorted = kt_compactor:get_all_dbs_and_sort_by_disk(),
     TotalSorted = length(Sorted),
+    lager:debug("finished listing and sorting databases (~p found)", [TotalSorted]),
     'ok' = kt_compaction_reporter:start_tracking_job(self(), node(), CallId, Sorted),
     F = fun({Db, _Sizes}, Ctr) ->
-                lager:debug("compacting ~p out of ~p dbs (~p remaining)",
-                            [Ctr, TotalSorted, (TotalSorted - Ctr)]
+                lager:debug("compacting ~p ~p/~p (~p remaining)",
+                            [Db, Ctr, TotalSorted, (TotalSorted - Ctr)]
```
jamesaimonetti: Can we use proper variable names like `Counter` or `DBsProcessed` instead of `Ctr`?

k-anderson:
@harenson @jamesaimonetti Does this need to go to 4.3 as well?

harenson:
@k-anderson Yes, here is the back port PR https://github.com/2600hz/kazoo/pull/5900

PR #5882, opened 2019-06-28
Add sparkly-clean command that runs all the clean commands (except test)

corporealshift:
Add a make target that includes blowing away the dependencies directory. Useful when somehow you break things really badly.


jamesaimonetti:
Merged, thanks!

PR #5889, opened 2019-07-04
Fix missing account info from parent call restrictions merge. 

Spencerhutch:
My recent change resulted in account information missing on some calls! 

Bug introduced here: https://github.com/2600hz/kazoo/pull/5695

### DIFF
--- a/core/kazoo_endpoint/src/kz_endpoint.erl
+++ b/core/kazoo_endpoint/src/kz_endpoint.erl
@@ -552,7 +552,7 @@ merge_parent_call_restrictions(Account) ->
                       {'error', _} -> Acc
                   end
           end,
-    lists:foldl(Fun, kz_json:new(), [kzd_accounts:id(Account) | kzd_accounts:tree(Account)]).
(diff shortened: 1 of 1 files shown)


core/kazoo_endpoint/src/kz_endpoint.erl
```diff
@@ -552,7 +552,7 @@ merge_parent_call_restrictions(Account) ->
                       {'error', _} -> Acc
                   end
           end,
-    lists:foldl(Fun, kz_json:new(), [kzd_accounts:id(Account) | kzd_accounts:tree(Account)]).
+    lists:foldl(Fun, Account, [kzd_accounts:id(Account) | kzd_accounts:tree(Account)]).
```
lazedo: no need to merge Account with Account.
i this should be `lists:foldl(Fun, Account, kzd_accounts:tree(Account)).`

Spencerhutch: Ahh yes! This makes sense!

PR #5890, opened 2019-07-04
PISTON-847: publish doc_edited/doc_deleted when attachments change

danielfinke:
We've noticed that when attachments are added to/deleted from a document, caches on other nodes are not cleared because kzs_publish does not publish an update for the document. If another node is checking for the available attachments on a doc (using the cache) the contents will be stale and will be missing a new attachment or containing a deleted attachment. This PR has kzs_publish send doc_edited/doc_deleted messages for when the attachments would become stale.

### DIFF
--- a/core/kazoo_data/src/kzs_attachments.erl
+++ b/core/kazoo_data/src/kzs_attachments.erl
@@ -158,9 +158,8 @@ put_attachment(#{att_handler := {Handler, Params}}=Map
             end,
             handle_attachment_handler_error(AttHandlerError, Options)
     end;
-put_attachment(#{server := {App, Conn}}, DbName, DocId, AName, Contents, Options) ->
-    kzs_cache:flush_cache_doc(DbName, DocId),
-    App:put_attachment(Conn, DbName, DocId, AName, Contents, Options).
+put_attachment(Map, DbName, DocId, AName, Contents, Options) ->
+    handle_put_attachment(Map, 'undefined', DbName, DocId, AName, Contents, Options, []).
 
 attachment_from_handler(AName, AttHandler, Size, CT) ->
     Props = [{<<"content_type">>, kz_term:to_binary(CT)}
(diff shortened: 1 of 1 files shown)


jamesaimonetti:
Good catch! Merged

core/kazoo_data/src/kzs_attachments.erl
```diff
@@ -198,16 +198,30 @@ handle_put_attachment(#{att_post_handler := 'external'}=Map, Att, DbName, DocId,
 external_attachment(Map, DbName, JObj, Att, Props) ->
     Atts = kz_json:merge_jobjs(Att, kz_json:get_value(?KEY_STUB_ATTACHMENTS, JObj, kz_json:new())),
     case kzs_doc:save_doc(Map, DbName, kz_json:set_values([{?KEY_STUB_ATTACHMENTS, Atts}], JObj), []) of
-        {'ok', Doc} -> {'ok', Doc, Props};
+        {'ok', Doc} ->
+            kzs_publish:maybe_publish_doc(DbName, JObj, Doc),
```
lazedo: @jamesaimonetti isn't this creating a `double` firing issue ?

core/kazoo_data/src/kzs_attachments.erl
```diff
@@ -174,22 +173,23 @@ attachment_handler_jobj(Handler, Props) ->
...
     kz_json:set_value(kz_term:to_binary(Handler), JObj, kz_json:new()).
 
--spec handle_put_attachment(att_map(), kz_json:object(), kz_term:ne_binary(), kz_term:ne_binary(), kz_term:ne_binary(), kz_term:ne_binary()
+-spec handle_put_attachment(att_map(), kz_term:api_object(), kz_term:ne_binary(), kz_term:ne_binary(), kz_term:ne_binary(), kz_term:ne_binary()
                            , kz_term:proplist(), kz_term:proplist()) ->
                                    {'ok', kz_json:object()} |
                                    {'ok', kz_json:object(), kz_term:proplist()} |
                                    data_error().
 
-handle_put_attachment(#{att_post_handler := 'stub'
```
lazedo: @danielfinke @jamesaimonetti why was this removed ? `att_post_handler := 'stub'`

lazedo: nm

core/kazoo_data/src/kzs_attachments.erl
danielfinke: @lazedo yeah looks like it would. I'll PR a fix in a few mins

danielfinke: @lazedo https://github.com/2600hz/kazoo/pull/5898

PR #5892, opened 2019-07-05
[4.3] On init, check if modb creation should start

jamesaimonetti:
If the tasks app is started on or after the MODB creation date, the
next month's MODBs won't be created (or may be partially created if
tasks was restarted after MODB creation had begun).

Check if the tasks app is starting up between configured creation date
and the end of the month and start a MODB creation task if so.


jamesaimonetti:
Backport of #5891

PR #5897, opened 2019-07-08
KZ00-5: move call to cast

jamesaimonetti:
To avoid blocking the calling process when the gen_server is overly
busy, move the channel updates to a cast instead of a call. In
particular, we've observed the authz worker (during route req/resp for
dialplan) getting blocked for 2s while waiting for this update to
occur.

Since the authz worker and dialplan response don't depend on the
channel cache, no sense blocking for the ETS write.


jamesaimonetti:
Backported to 4.3 in #5896

lazedo:
@jamesaimonetti lets wait before merging this as the way channels are handled in the new `event_stream` may need `sync` calls

lazedo:
@jamesaimonetti looks good

PR #5899, opened 2019-07-08
Update cf_set_variable.erl

extremerotary:
Add a new variable name for the purpose of tracking which group was sent the call


mark2600:
Please correct code formatting so this passes CI then I will merge

jamesaimonetti:
Fixed and merged.

PR #5912, opened 2019-07-16
[master] help-44967: use stored faxdb when saving new fax doc

swysor:


swysor:
When a fax is processed around the border of an modb rollover, the creation of the document uses the doc id generated when the fax is received. However create_fax_doc generates a new modb name based on the account id in the call object, not reusing the database name generated and stored in the fax_storage record when a new_request is initialized. This causes failures to save since all the other database operations use the fax_storage db name.

PR #5915, opened 2019-07-18
PISTON-772: do not start agents in acdc_queue_manager

danielfinke:
There should not be any situation where the agent needs to be started in
the queue mgr. Running state of the agents should be tied to login state
and is dealt with either on-the-fly when the agent logs in/out or on
acdc boot to resume the previous login state


jamesaimonetti:
Merged

PR #5916, opened 2019-07-18
PISTON-841: performance improvements: acdc_queue_handler:presence_probe

danielfinke:
- filter out Usernames that are not 32 char hex binary (so can't be
      queue presence probes)
    - serialize get_account_by_realm lookup to allow cache to warm when
      Kamailio nodes are restarted and many probes come in due to stale
      presence cache in Kamailio
    - request queue_size from queue manager as AMQP queue size will give
      wrong result due to unacked msgs in consumers


jamesaimonetti:
Merged. The CI failure was due to the branch being behind master, so the diff of changed files was larger than necessary. Dialyzed locally, all good.

PR #5917, opened 2019-07-18
PISTON-867: remove acdc_queue_manager monitor_call

danielfinke:
No longer need to bind to call.CHANNEL_DESTROY.<call_id> for queue
manager. Removal of calls is handled by other events. These bindings
were being leaked and consuming extra AMQP memory over time


jamesaimonetti:
Merged!

PR #5918, opened 2019-07-18
PISTON-859: always update acdc_queue_listener my_q

danielfinke:
Handles channel being lost - new queue may be provided to gen_listener


jamesaimonetti:
Merged

PR #5919, opened 2019-07-18
[4.3] HELP-44667: don't send authz resp on quick hangup

jamesaimonetti:
Processing the authz request can take longer than processing the
subsequent CHANNEL_DESTROY. This causes the flushing in j5_channels to
occur before the storing of the authorization payload. Since the
channel is gone, the authorization payload isn't flushed and counts
against the account/reseller limits.

Instead, check the j5_channels ETS table for the authorization
payload on CHANNEL_DESTROY; if none, store an entry with a
destroyed:true flag. If a subsequent authz response is attempted
to be stored, this check will keep it from happening.

Introduce a test for sending CHANNEL_DESTROY followed by authz_req and
ensuring totals remain 0 and no authz_resp is generated.

### DIFF
--- a/applications/jonny5/src/j5_authz_req.erl
+++ b/applications/jonny5/src/j5_authz_req.erl
@@ -328,6 +328,16 @@ send_response(Request) ->
               | kz_api:default_headers(?APP_NAME, ?APP_VERSION)
              ]),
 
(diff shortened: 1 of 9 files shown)


jamesaimonetti:
See master in #5914

PR #5920, opened 2019-07-23
cowboy upgrade remaining & others

lazedo:


applications/crossbar/src/api_util.erl
```diff
@@ -563,8 +554,30 @@ is_valid_request_envelope(Envelope, Context) ->
...
+
+-spec req_noun_requires_envelope(cb_context:context()) -> boolean().
+req_noun_requires_envelope(Context) ->
+    req_noun_requires_envelope(Context, cb_context:req_nouns(Context)).
+
+-spec req_noun_requires_envelope(cb_context:context(), req_nouns()) -> boolean().
+req_noun_requires_envelope(_Context, []) -> 'true';
+req_noun_requires_envelope(Context, [{Module, Params} | _]) ->
+    kz_module:is_exported(Module, 'requires_envelope', length(Params)+1)
+        andalso erlang:apply(Module, 'requires_envelope', [Context | Params]).
```
jamesaimonetti: Doesn't `Module` come in as a binary? Wouldn't this require a `kz_term:to_atom/1` call?

lazedo: fixed

PR #5910, opened 2019-07-12
[4.3] Voicemail callback

sergey-safarov:
cherry picked from commit f37d9f9b7ee0c8f69f92def1d94b0f91be9790ee

PR #5922, opened 2019-07-23
callflow: Code cleanup of cf_voicemail callback cleanup

sergey-safarov:
Reference commit f37d9f9b7ee0c8f69f92def1d94b0f91be9790ee

handle dialyzer complaints from #5910


lazedo:
thanks @sergey-safarov

PR #5926, opened 2019-07-24
[4.3] Timers for kzc_recording processes to make sure they terminate

jamesaimonetti:
Start a timer to limit recording time; stop kzc_recording process if
the timer expires

Start a timer for RECORD_STOP when recv CHANNEL_DESTROY; stop
kzc_recording if the timer expires


lazedo:
@mark2600 this PR is incomplete.  the timer is not stopped on `record_start`

PR #5931, opened 2019-07-26
HELP-44382: decrease 409s when saving accounts

jamesaimonetti:
Due to changes in how account documents are saved (in the account db,
then copied to the 'accounts' db), some systems were experiencing
increased incidences of 409 API errors when modifying the account doc
via `/v2/accounts/{ACCOUNT_ID}`.

Currently the only "smoking gun" for why some systems experienced this
elevated rate seems due to the Erlang scheduler's interleaving of
processes and the way the kzs_cache works when storing a cache entry.

Essentially the problem was distilled to the interplay of 3 processes,
A, B, and C:

A -> Cache: flush doc
A -> Couch: save rev X
B -> Couch: get doc
Couch -> B: got rev X
B -> Cache: cache rev X
Couch -> A: saved rev Y
C -> Cache: get rev X
C -> Couch: save rev X
Couch -> C: 409

See core/kazoo_data/doc/help_44382.png

So the cache was being re-populated with the soon-to-be old revision
while the save to Couch was occurring.

To address this, after a successful save operation from Couch, kzs_doc
will then cache the new document. One caveat is that if an existing
doc is already cached, the revisions of the existing and new doc are
compared and the higher revision is stored in the cache.

Another issue that arose was if _rev is included in
kzd_accounts:update/3 and is old, and ensure_saved is true, a loop
will be started where the save will 409, updates re-applied to the
actual doc (including the old _rev) and a save will again 409 - rinse
and repeat.


jamesaimonetti:
See 4.3 in #5930

PR #5932, opened 2019-07-29
HELP-45578: if already unpublished dont change authority or re-unpublish

k-anderson:


jamesaimonetti:
Seems like a unit test could capture what the issue is and prove this code change fixes it ;)

PR #5934, opened 2019-07-29
Voicemail fwd confirmation

Spencerhutch:
When a user forwards a voicemail there is no interaction with the user confirming the operation. 

This can leave a user confused, we don't want confused users. 

Kazoo-sounds change: https://github.com/2600hz/kazoo-sounds/pull/8


jamesaimonetti:
Agreed on not wanting confused users! Can you backport this to 4.3 as well. Sound file will be in the next release as well.

PR #5935, opened 2019-07-29
Voicemail fwd confirmation backport

Spencerhutch:
Backporting: https://github.com/2600hz/kazoo/pull/5934

-----

When a user forwards a voicemail there is no interaction with the user confirming the operation.

This can leave a user confused, we don't want confused users.

Kazoo-sounds change: 2600hz/kazoo-sounds#8


jamesaimonetti:
Thanks!

PR #5936, opened 2019-07-30
Bh presence master

mk1s:
This adds a new blackhole module, `bh_presence`, which emits presence events via blackhole for the following bindings:

- `dialog.*.*`
- `mwi_updates.*.*`
- `update.*.*`


lazedo:
@mk1s please rebase

PR #5938, opened 2019-07-31
add desktop feature flag

mark2600:


k-anderson:
If we try to merge the flags and features array use-cases we are going to run into problems.  Imagine if you will:

1)	I enable call forwarding, the view will now ‘push’ it into the feature array.
2)	I enable a third party feature, it takes the current feature list and adds itself to it, saving the result to the document.  Now call_forwarding is hard-coded on the document.
3)	I disable call forwarding, the view will stop pushing it into the array but since its been copied into the features array on the document it appears to be enabled and nothing will ever remove it from there….

In this particular example, a 3rd party app will then cause SmartPBX to list call_forwarding as enabled for that user regardless of if it actually is at that time.

As such, I propose separating the auto-magic features array and provide a manual array that can be used by external applications to include flags in the collection listing.

core/kazoo_documents/src/kzd_accounts.erl.src
```diff
@@ -19,6 +19,7 @@
 -export([do_not_disturb/1, do_not_disturb/2, set_do_not_disturb/2]).
 -export([do_not_disturb_enabled/1, do_not_disturb_enabled/2, set_do_not_disturb_enabled/2]).
 -export([enabled/1, enabled/2, set_enabled/2]).
+-export([flags/1, flags/2, set_flags/2]).
```
jamesaimonetti: Please migrate all the new accessor functions into the `.erl`

k-anderson: done :)

PR #5939, opened 2019-07-31
add ctu and desktop features

mark2600:


core/kazoo_apps/priv/couchdb/account/users.json
```diff
@@ -40,6 +40,8 @@
                 "    for (var feature in doc.smartpbx)",
                 "      if(doc.smartpbx[feature].enabled) features.push(feature);",
                 "  if (doc.hotdesk && doc.hotdesk.enabled) features.push('hotdesk');",
+                "  if (doc.ctu && doc.ctu.enabled) features.push('ctu');",
+                "  if (doc.desktop && doc.desktop.enabled) features.push('desktop');",
```
lazedo: these are already in `doc.smartpbx`

k-anderson: @lazedo well since they don't necessary have anything directly todo with smartpbx and it was my intention to put them at the root of the document.  We are making that change now

lazedo: @k-anderson it doesn't make sense to keep updating views on each feature addition/removal just because of this. it creates unnecessary complexity in the script and the smartpbx approach ensures that `everyone` this strategy.

core/kazoo_apps/priv/couchdb/account/users.json
```diff
@@ -35,7 +35,7 @@
             "map": [
                 "function(doc) {",
                 "  if (doc.pvt_type != 'user' || doc.pvt_deleted) return;",
-                "  var features = [];",
+                "  var features = doc.features;",
```
lazedo: maybe `var features = doc.features || [];`

mark2600: just thought of that myself :)

PR #5941, opened 2019-08-01
[4.3] fix send_file CTA and make file removal optional

k-anderson:
There is already a more expansive fix on master.

### DIFF
--- a/applications/crossbar/src/api_util.erl
+++ b/applications/crossbar/src/api_util.erl
@@ -1337,10 +1337,14 @@ final_csv_resp_type(Context, 'false') ->
 %% @doc
 %% @end
 %%------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


applications/crossbar/src/api_util.erl
```diff
@@ -1185,15 +1183,36 @@ finish_request(_Req, Context) ->
...
-    maybe_cleanup_file(cb_context:fetch(Context, 'csv_acc')),
+    _ = kz_util:spawn(fun maybe_cleanup_file/1, [Context]),
     'ok'.
 
 -spec maybe_cleanup_file(kz_csv:file_return() | 'undefined') -> 'ok'.
-maybe_cleanup_file({File, _}) ->
-    _Del = file:delete(File),
-    lager:debug("deleting ~s: ~p", [File, _Del]);
-maybe_cleanup_file(_) -> 'ok'.
+maybe_cleanup_file(Context) ->
```
jamesaimonetti: Spec is now wrong

PR #5959, opened 2019-08-13
[4.3] HELP-44944: testing of the failover_when_all_unreg config flag

jamesaimonetti:
Primarily testing that normal call-forwarded devices are not filtered
when SIP devices are registered.

Also testing that failover is working when SIP devices are
unregistered.


jamesaimonetti:
See master at #5960

PR #5969, opened 2019-08-15
[4.3] PISTON-894: fix crash on usurp publisher in ecallmgr_fs_conferences_shared for non-loopbacks

danielfinke:
Resolves a crash when using the conference dial API, dialling a number that will not cause a loopback. It looks like this has been fixed with a rewrite in master, so I'm only backporting this fix to 4.3.


lazedo:
Hi @danielfinke, can you give an example of the `api` where `dialling a number that will not cause a loopback` ? thanks

danielfinke:
Hey @lazedo. The way we are using the API is via a PUT to the conference ID to perform a conference action. The requested action is "dial". For the data property, we are sending `{"endpoints": ["some device id"]}`. I mistakenly said "number" but we are using an endpoint ID.

jamesaimonetti:
Merged, thanks!

PR #5970, opened 2019-08-15
[5.0] Upgrade to MPL 2.0

jamesaimonetti:
- add MPL 2.0 to all core/app folders for breakening
- update all source files to include MPL license info


Makefile
```diff
@@ -280,6 +280,7 @@ app_applications:
 	ERL_LIBS=deps:core:applications $(ROOT)/scripts/apps_of_app.escript -a $(shell find applications -name *.app.src)
 
 code_checks:
+	@$(ROOT)/scripts/bump-copyright-year.sh $(CHANGED_ERL)
```
icehess: edocify script is bumping the copyright year too. So can we add adding license line to edocify instead?

jamesaimonetti: what's the advantage?

icehess: None.

Just as you know with this change now both edocify and bump-copyright-year.sh are trying to bump the year.

jamesaimonetti:
@icehess i think all are caught now. Had to add exception for amqp_cron but otherwise...
Also don't think the analyze portion will be fixable :p No code changes though so I think it can be ignored.

PR #5976, opened 2019-08-20
Amqp updates

lazedo:
major change to handle assignments async

### DIFF
--- a/core/kazoo_globals/src/kz_nodes.erl
+++ b/core/kazoo_globals/src/kz_nodes.erl
@@ -712,6 +712,22 @@ handle_advertise(JObj, Props) ->
         'true' -> 'ok'
     end.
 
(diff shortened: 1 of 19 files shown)


core/kazoo_globals/src/kz_nodes.erl
```diff
@@ -708,6 +708,22 @@ handle_advertise(JObj, Props) ->
...
+-spec build_advertised_node(kz_json:object(), nodes_state()) -> kz_types:kz_node() | 'ok'.
+build_advertised_node(JObj, State) ->
+    try
+        from_json(JObj, State)
+    catch
+        _E:_R ->
+            lager:warning("error building advertised node : ~p", [{_E, _R}])
+    end.
+
+-spec update_advertised_node(kz_types:kz_node(), nodes_state()) -> 'ok'.
```
jamesaimonetti: should return `pid() | 'true'` instead of 'ok'.

PR #5977, opened 2019-08-21
Add BICS knm

corporealshift:
This is still a work in progress as the BICS API does not support searching by number prefixes, but I wanted to get some eyes on it first. I pulled in the knm-voxbone branch to use some of its iso3166 utilities. I tried to emulate the approach taken in that knm as best as possible as to keep it consistent.


corporealshift:
This branch had a bunch of extra stuff in it, reopening with a cleaner branch

PR #5980, opened 2019-08-22
HELP-45446: [4.3] move when termination handlers are running to cf_exe termination

icehess:
This is fixing the issue with missed call alert or other termination handlers when they are get executed.

There is race condition between callflow action modules, if one of the action (usually voicemail) takes a bit more to process and going down than when cf_exe is receiving and processing `channel_destory` event this cause the termination handlers run before voicemail module can update the the call record in cf_exe causing the missed call alert always being sent even though the caller left a message.

For resolving this we decided to move handlers runner to when cf_exe is stopping so at that point we're sure every other callflow modules has been already terminated or being forcefully terminate before running the handlers.

On the other hand we observe that there is another race condition between when `cf_exec` is receiving `channel_destory` event and when callflow module continuing the call to next action. In this case which is happening for modules like `cf_device`, `cf_user`  and alike, the freeswitch command like `bridge` is failed to be executed because of the caller hangup, and then the cf module genuinely advancing the call to next callflow action, `cf_exe` continues to next action still not received `channel_destory`. `cf_exe` will eventually will receive the channel destory event and takes everything down. But in that micro seconds that executing the the next callflow actions can cause some problem in some scenario that missed call alert module is being used.

This issue is silence bug, went not notice because there was nothing visible so far to show this micro/mini seconds delay of receiving the event. But if we use `device -> missed_call_alert -> voicemail` callflow chain, this could be potentially send missed call email if the caller hangup the call during ringing (device callflow is going down and advancing to next callflow which is missed call alert).

In this scenario after caller hangup, `cf_device` is going down then `cf_missed_call_alert` module and then `cf_voicemail` is executing, and *then* `cf_exe` receives the `channel_destory`, so `cf_voicemail` will be shutdown and `cf_exe` is executing the termination handlers which in now setup by missed_call_alert and since no message was left it will sent an missed call email notification.

For resolving this issue we now publish the hanup cause in dialplan error event so `cf_exe` knows that the call has been hangup so it will not executing the next callflow.

### DIFF
--- a/applications/callflow/src/cf_exe.erl
+++ b/applications/callflow/src/cf_exe.erl
@@ -21,7 +21,7 @@
 -export([hard_stop/1]).
 -export([transfer/1]).
 -export([control_usurped/1]).
--export([channel_destroyed/1]).
+-export([channel_destroyed/1, channel_destroyed/2]).
 -export([is_channel_destroyed/1]).
 -export([stop_on_destroy/1
         ,continue_on_destroy/1
@@ -72,6 +72,7 @@
                ,flows = [] :: kz_json:objects()
                ,cf_module_pid :: kz_term:api_pid_ref()
                ,cf_module_old_pid :: kz_term:api_pid_ref()
+               ,hangup_info = [] :: kz_term:proplist()
                ,status = 'init' :: callfow_status()
                ,queue :: kz_term:api_ne_binary()
                ,self = self() :: pid()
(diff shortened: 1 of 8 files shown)


applications/callflow/src/cf_exe.erl
```diff
@@ -668,14 +668,20 @@ terminate({'shutdown', 'control_usurped'}, _) ->
     lager:info("the call has been usurped by an external process");
 terminate(_Reason, #state{call=Call
                          ,cf_module_pid='undefined'
+                         ,hangup_info=HangupInfo
```
lazedo: handle `undefined` to determine if there's a need to send hangup

icehess: This is being handled in `maybe_hangup_call/1` now.

applications/callflow/src/cf_exe.erl
```diff
@@ -902,18 +928,21 @@ relay_message(Notify, Message) ->
...
 
--spec hangup_call(kapps_call:call()) -> 'ok'.
-hangup_call(Call) ->
-    hangup_call(Call, 'undefined').
+-spec maybe_hangup_call(state()) -> 'ok'.
+maybe_hangup_call(#state{hangup_info='undefined'}) -> 'ok';
+maybe_hangup_call(#state{call=Call
+                        ,hangup_info=HangupInfo
+                        }) ->
+    hangup_call(Call, kz_json:get_ne_binary_value(<<"Hangup-Cause">>, HangupInfo)).
```
lazedo: isn't the logic inverted here ?

PR #5984, opened 2019-08-22
HELP-45446: [master] move when termination handlers are running to cf_exe termination

icehess:
**This depends on https://github.com/2600hz/kazoo/pull/6022.**

This is fixing the issue with missed call alert or other termination handlers when they are get executed.

4.3: https://github.com/2600hz/kazoo/pull/5980

There is race condition between callflow action modules, if one of the action (usually voicemail) takes a bit more to process and going down than when cf_exe is receiving and processing `channel_destory` event this cause the termination handlers run before voicemail module can update the the call record in cf_exe causing the missed call alert always being sent even though the caller left a message.

For resolving this we decided to move handlers runner to when cf_exe is stopping so at that point we're sure every other callflow modules has been already terminated or being forcefully terminate before running the handlers.

On the other hand we observe that there is another race condition between when `cf_exec` is receiving `channel_destory` event and when callflow module continuing the call to next action. In this case which is happening for modules like `cf_device`, `cf_user`  and alike, the freeswitch command like `bridge` is failed to be executed because of the caller hangup, and then the cf module genuinely advancing the call to next callflow action, `cf_exe` continues to next action still not received `channel_destory`. `cf_exe` will eventually will receive the channel destory event and takes everything down. But in that micro seconds that executing the the next callflow actions can cause some problem in some scenario that missed call alert module is being used.

This issue is silence bug, went not notice because there was nothing visible so far to show this micro/mini seconds delay of receiving the event. But if we use `device -> missed_call_alert -> voicemail` callflow chain, this could be potentially send missed call email if the caller hangup the call during ringing (device callflow is going down and advancing to next callflow which is missed call alert).

In this scenario after caller hangup, `cf_device` is going down then `cf_missed_call_alert` module and then `cf_voicemail` is executing, and *then* `cf_exe` receives the `channel_destory`, so `cf_voicemail` will be shutdown and `cf_exe` is executing the termination handlers which in now setup by missed_call_alert and since no message was left it will sent an missed call email notification.

For resolving this issue we now publish the hanup cause in dialplan error event so `cf_exe` knows that the call has been hangup so it will not executing the next callflow.

### DIFF
--- a/applications/callflow/src/cf_exe.erl
+++ b/applications/callflow/src/cf_exe.erl
@@ -940,11 +940,17 @@ handle_usurp(Self, Call, JObj) ->
 
 -spec handle_error(pid(), kz_term:ne_binary(), kz_term:pids(), kz_json:object()) -> 'ok'.
 handle_error(Self, CallId, Notify, JObj) ->
-    case kz_json:get_value([<<"Request">>, <<"Call-ID">>], JObj) of
-        CallId ->
-            lager:debug("channel execution error, stopping here"),
+    case {kz_json:get_ne_binary_value([<<"Call-ID">>], JObj, kz_json:get_ne_binary_value([<<"Request">>, <<"Call-ID">>], JObj))
+         ,kz_json:get_ne_binary_value(<<"Channel-Call-State">>, JObj)
+         }
+    of
+        {CallId, <<"DOWN">>} ->
+            lager:debug("channel execution error due to call destroyed, stopping here"),
             handle_channel_destroyed(Self, Notify, JObj);
-        'undefined' -> relay_message(Notify, JObj);
+        {'undefined', <<"DOWN">>} ->
+            lager:debug("call is destroyed, stopping here"),
+            handle_channel_destroyed(Self, Notify, JObj);
+        {'undefined', _} -> relay_message(Notify, JObj);
         _Else -> 'ok'
     end.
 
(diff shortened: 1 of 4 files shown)


applications/callflow/src/cf_exe.erl
```diff
@@ -928,12 +938,14 @@ handle_usurp(Self, Call, JObj) ->
...
 
--spec handle_error(kz_term:ne_binary(), kz_term:pids(), kz_json:object()) -> 'ok'.
-handle_error(CallId, Notify, JObj) ->
+-spec handle_error(pid(), kz_term:ne_binary(), kz_term:pids(), kz_json:object()) -> 'ok'.
+handle_error(Self, CallId, Notify, JObj) ->
     case kz_json:get_value([<<"Request">>, <<"Call-ID">>], JObj) of
-        CallId      -> relay_message(Notify, JObj);
+        CallId ->
+            lager:debug("channel execution error, stopping here"),
+            handle_channel_destroyed(Self, Notify, JObj);
```
lazedo: why do you assume that an error implies a channel destroy ?

lazedo:
@icehess with #6022 merged, i think this one will only need to move the `destroy_handlers` to `terminate`

icehess:
> @icehess with #6022 merged, i think this one will only need to move the `destroy_handlers` to `terminate`

@lazedo I still think we need to handle the error event. Now with your PR we know the channel state and we can stop instead of continuing. What do you think?

lazedo:
@icehess looks good with one exception. when there's an execution error with consequent hangup, there's no way for us to reliably retrieve the hangup-code / hangup-reason at the time of the error that you're getting in handle_channel_destroyed. if the command in error does hangup the call we'll receive the channel destroy. imo, `handle_error` fun should just be removed and remove the clause in `handle_event` and let it notify the running command of the failure.

PR #5986, opened 2019-08-23
preserve initial params - 4.3

lazedo:
if there's an error before the start consume
we can potentially have initial Params
set to empty array and therefore loose the bindings

PR #5987, opened 2019-08-23
preserve bindings (#5986)

lazedo:
if there's an error before the start consume
we can potentially have initial Params
set to empty array and therefore loose the bindings

also, since we're folding the Params into bindings
the return is the list of bindings that we preserve in state,
so we still need to add the existing binding to the acumulator.

we need to set the broker when assigning existing to consumer

we need to send 'channel_lost' if we're moving the channel
to the primary broker


sergey-safarov:
I tested this PR.
Looks as connection to rabbimq properly restored after one and all rabbimq nodes failure.

PR #5990, opened 2019-08-26
how to add new country for buy number

MrAriaNet:
Hello,
i can not add number another country for buy user !
what should I do to add a new country phone number?


lazedo:
@MrAriaNet not sure what you're trying to change here. please rebase and reopen.

PR #5699, opened 2019-04-20
Google ASR mp3 to wav conversion + extra configs

arekfryz:
Google does not support mp3, media should be converted. Additional config options let you get more accurate recognition. Works with v1 and v1p1beta1 of Cloud Speech-to-Text API

PR #5999, opened 2019-08-29
Google ASR MP3 to WAV conversion & extra configs

DustinBrett:
- Moved code used for iSpeech to be shared by both ASR providers
- Added relevant Google ASR config settings


DustinBrett:
This PR builds on the code and comments in https://github.com/2600hz/kazoo/pull/5699 and is based on what we are using in production.

jamesaimonetti:
Merged! Closes #5699 in favor of this

PR #6002, opened 2019-08-29
some updates to keep progressing

lazedo:


core/kazoo_stdlib/src/props.erl
```diff
@@ -258,6 +258,9 @@ get_keys(Props) -> [as_key(KV) || KV <- Props].
 as_key(A) when is_atom(A) -> A;
 as_key({K, _}) -> K.
 
+-spec get_values(kz_term:proplist()) -> [kz_term:proplist_value()].
+get_values(Props) -> [V || {_K, V} <- Props].
```
jamesaimonetti: `get_values([foo, {bar, 1}]) = [1]` Is this the expected output? Or should it be `[true, 1]`?

lazedo: good question, for me the answer would be `[1]`,  because `foo` exists, it doesn't have a value. `foo` could also be `false` (the `key`) in which case we would be assuming that `false`=:= `true`.
but i'm `ok` with `both` ways, i guess its a defining moment :)

PR #6005, opened 2019-08-31
Loopback info

lazedo:


applications/jonny5/src/j5_authz_req.erl
```diff
@@ -16,154 +16,25 @@
...
-maybe_determine_reseller_id(Request) ->
-    case j5_request:reseller_id(Request) of
-        'undefined' -> determine_reseller_id(Request);
-        _Else -> maybe_reseller_limited(Request)
-    end.
-
--spec determine_reseller_id(j5_request:request()) -> 'ok'.
-determine_reseller_id(Request) ->
-    AccountId = j5_request:account_id(Request),
-    ResellerId = kz_services_reseller:get_id(AccountId),
```
icehess: This is the correct way to get reseller id, `kzd_accounts:reseller_id/1` is just looking at the the account's doc.

PR #6011, opened 2019-09-03
[master] fix returning queue_name from gen_listener

mark2600:
- when the queue_name is requested before it is actually done setting
up it will return 'undefined' instead of the to-be-used name, check for
that condition and return the queue_name that is configured instead


lazedo:
@mark2600 as discussed, lets not merge this as-is. if we return the `configured` queue , the receiver may try to send a msg to that queue and it may fail if the queue wasn't created yet.

jamesaimonetti:
If this (or something along these lines) is needed, let's open a new PR.

PR #6012, opened 2019-09-03
Convert kapi_websockets module to use kapi_definition module

harenson:


harenson:
`?MODULE_REQ_ROUTING_KEY` macro is being used by `blackhole_listener.erl` module by including that header file: `-include_lib("kazoo_amqp/src/api/kapi_websockets.hrl").`. That's why I didn't want to remove that header file, I don't want to touch not *kapi_* modules at the moment.

PR #6015, opened 2019-09-05
[4.3] sort user in directory consistently

zzzming:
sort user in tts directory consistently

### DIFF
--- a/applications/callflow/src/module/cf_directory.erl
+++ b/applications/callflow/src/module/cf_directory.erl
@@ -139,8 +139,9 @@ handle(Data, Call) ->
     end.
 
 -spec directory_start(kapps_call:call(), directory(), directory_users()) -> 'ok'.
-directory_start(Call, State, CurrUsers) ->
-    directory_start(Call, State, CurrUsers, 3).
+directory_start(Call, #directory{sort_by=SortBy}=State, CurrUsers) ->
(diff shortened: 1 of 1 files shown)


applications/callflow/src/module/cf_directory.erl
```diff
@@ -138,6 +139,11 @@ handle(Data, Call) ->
...
     end.
 
+-spec directory_start_sorted_users(kapps_call:call(), directory(), directory_users()) -> 'ok'.
+directory_start_sorted_users(Call, State, CurrUsers) ->
+    Users = sort_users(CurrUsers, State#directory.sort_by),
+    directory_start(Call, State, Users).
+
 -spec directory_start(kapps_call:call(), directory(), directory_users()) -> 'ok'.
 directory_start(Call, State, CurrUsers) ->
     directory_start(Call, State, CurrUsers, 3).
```
jamesaimonetti: What do you think about putting the sort call here (on `CurrUsers`); smaller changeset mostly :)

jamesaimonetti: Also, as a general preference of mine:

I like to pattern-match records in the function header instead of using the `Var#record.field` syntax. So `directory_start_sorted_users(Call, #directory{sort_by=SortBy}=State, CurrUsers) ->` then use `SortBy` in the function clause.

PR #6019, opened 2019-09-06
move fax outbound control to ra state machine

lazedo:
* starts using amqp_dist + ra
* stops using kazoo_globals


lazedo:
this will be extended in upcoming pr

PR #6020, opened 2019-09-06
[4.3] HELP-10237: improve communication about call recording settings

jamesaimonetti:
Added tests for kz_endpoint when merging call recording settings to
clarify the behaviour.

Updated documentation to clarify precedence and the duality of account
and endpoint call recording settings.


jamesaimonetti:
Will reopen when ready

PR #6040, opened 2019-09-19
replace kz_json with kzd_ from object returned by kz_datamgr:open_[ca…

zzzming:


core/kazoo_documents/src/kzd_comments.erl
```diff
@@ -26,7 +26,7 @@ new() ->
 
 -spec comments(doc()) -> kz_term:api_objects().
 comments(Doc) ->
-    comments(Doc, 'undefined').
+    comments(Doc, []).
```
jamesaimonetti: Update the spec to return `kz_json:objects()` here.

zzzming: Fixed the spec.

PR #6041, opened 2019-09-19
[4.3] KAZOO-6122: when fetching the account id by IP, return an error if an…

k-anderson:
… object is disabled

In the case of a valid document, that is enabled, and system_config set to use sip_auth, AND the associated user or account disabled kapps_util will return {ok, []}

ie: ok with no additional parameters to add to the CCVs reg_route_req assumed (happy case) that if it was found an account id was associated.  However, in this case no account id is proved leading to a route_req replay with no account id and a loop.

From an admin perspective, setting up an account with ip auth, disabling the device/user/account associated and then calling the number results in a significant spike in CPU and/or memory.


jamesaimonetti:
@k-anderson does this need to be forward-ported to master?

PR #6045, opened 2019-09-20
HELP-10392: Honor ring group members set delay

harenson:
When an endpoint appears multiple times in a ring group, treat them as separate endpoints according to their configured **delay** and **timeout** parameters. Also, sort the resolved endpoints by **delay** before sending them to FS.


make/deps.mk
```diff
@@ -43,7 +43,7 @@ endif
 
 dep_amqp_client = git https://github.com/2600hz/erlang-amqp_client.git v3.7.14
 
-dep_amqp_dist = git https://github.com/2600hz/erlang-amqp_dist.git 8af95b7dc4680da7abe125636a2fd447ce3f7f08
+dep_amqp_dist = git https://github.com/2600hz/erlang-amqp_dist.git b24862c1c64c04f0ded0844410c44632f78d92d2
```
lazedo: this is incorrect. you are reverting to an older not intended commit

harenson: Ok, let me remove that change.

jamesaimonetti:
Please add some tests to `cf_ring_group_tests` that shows that a ring group of `[Group, User]` and `[User, Group]` resolves to the same list of endpoints.

applications/callflow/test/cf_ring_group_tests.erl
```diff
@@ -28,3 +28,79 @@ weighted_random_sort_test_() ->
...
      ]
     ].
+
+%% HELP-260010392
+help_10392_test_() ->
+    UserId = <<"help-10392-user-id">>,
+    DevId = <<"help-10392-device-id">>,
+    GroupId = <<"help-10392-group-id">>,
+
+    meck:new('kz_datamgr', ['passthrough']),
```
jamesaimonetti: You are the only one using meck afaict. I would prefer tests to use fixturedb as that is where testing needing database access has gone in the project. There are a number of tests showing examples (like [webhooks](https://github.com/2600hz/kazoo/blob/master/applications/webhooks/test/webhooks_disabler_tests.erl)) of how make this happen.

applications/callflow/test/cf_ring_group_tests.erl
```diff
@@ -28,3 +28,82 @@ weighted_random_sort_test_() ->
...
+
+    Call = kapps_call:set_account_id(AccountId, kapps_call:new()),
+
+    Resp0 = cf_ring_group:resolve_endpoint_ids(help_10392_data([User, Group]), Call),
+    Resp1 = cf_ring_group:resolve_endpoint_ids(help_10392_data([Group, User]), Call),
+    Resp2 = cf_ring_group:resolve_endpoint_ids(help_10392_data([User, User]), Call),
+    Resp3 = cf_ring_group:resolve_endpoint_ids(help_10392_data([Group, Group]), Call),
+
+    Expected0 = [EndpointDelay0, EndpointDelay10],
+    Expected1 = [EndpointDelay10, EndpointDelay0],
```
jamesaimonetti: I was under the impression that we were going to sort the ring group based on the delay. So this expectation is backwards. My understanding was `resolve_endpoint_ids([User, Group], Call)` and `resolve_endpoint_ids([Group, User], Call)` would result in the same listing: `[EndpointDelay0, EndpointDelay10]`.

harenson: The sorting by delay happens at a higher level https://github.com/2600hz/kazoo/blob/a75f05de3be1a69485e42f442284e1d1df61ff65/applications/callflow/src/module/cf_ring_group.erl#L134

The functionality being tested here is just in charge of resolving the endpoints for each Ring Group member.

If you want I can add another test for testing the ordering by delay.

PR #6049, opened 2019-09-24
PISTON-914: escape CAV comma/singlequote in originate strings

danielfinke:
- also encode JSON CAVs and give json_cav_ prefix

During originate, bridge, and conference dial, CAVs with reserved characters (`,` & `'`) need to be escaped and JSON CAVs need to have the `json_cav_` prefix so that they are properly decoded in channel event handling. Commas (`,`) seem to need to be escaped even if they are within a single-quoted string otherwise FS parses the comma as a var delimiter.

I'd like to request a review of the regex I used here. It was confusing me that I needed to double-escape the whole thing like `\\\\\\g1` instead of `\\\g1`.


danielfinke:
CircleCI test failures appear unrelated to the PR

jamesaimonetti:
@danielfinke please rebase, I addressed the doc issue!

danielfinke:
@jamesaimonetti done

jamesaimonetti:
Merged

PR #6056, opened 2019-09-26
[4.3] HELP-10868: ignore channel bridge and log at info when a limit is exceeded

jamesaimonetti:


icehess:
@jamesaimonetti I think it need a rebase for the new circleci stuff.

PR #6059, opened 2019-09-27
[master] check prerequisites for make

zzzming:
Introduce pre-check for make dependency like git version. This is motivated by a recent change in mk and erlang.mk upversion that require to parse git output. The older version of git (lower than major version 2) does not match with the output parsing.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -959,6 +959,10 @@ Part of the great rename, converts Whistle-related names to Kazoo-specific names
 
 Generate API clients in multiple languages from the Swagger file.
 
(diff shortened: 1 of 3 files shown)


scripts/README.md
```diff
@@ -959,6 +959,10 @@ Part of the great rename, converts Whistle-related names to Kazoo-specific names
 
 Generate API clients in multiple languages from the Swagger file.
 
+## `make-prerequisite.sh`
+
+Check make dependency and prerequisitei such as git version.
```
jamesaimonetti: typo on prerequisite

PR #6067, opened 2019-10-02
[master] remove used_by under port_in number when its callflow is deleted

zzzming:
1. Remove used_by json object under a port_in number when the associated callflow is deleted.

### DIFF
--- a/core/kazoo_number_manager/src/knm_port_request.erl
+++ b/core/kazoo_number_manager/src/knm_port_request.erl
@@ -448,20 +448,21 @@ assign_to_app(Number, NewApp, JObj) ->
     case kz_json:get_value([?NUMBERS_KEY, Number, ?USED_BY_KEY], JObj) of
         NewApp -> {'ok', JObj};
         _OldApp ->
-            lager:debug("assigning number ~s in port request ~s to ~s"
-                       ,[Number, kz_doc:id(JObj), NewApp]
+            NumberObj = kz_json:get_value([?NUMBERS_KEY, Number], JObj),
+            lager:debug("assigning number ~s in port request ~s from ~p to ~s"
+                       ,[Number, kz_doc:id(JObj), NumberObj, NewApp]
                        ),
-            NumberJObj = kz_json:from_list(
-                           props:filter_empty(
-                             [{?USED_BY_KEY, NewApp}
-                              | kz_json:to_proplist(
-                                  kz_json:get_value([?NUMBERS_KEY, Number], JObj)
-                                 )
-                             ])
-                          ),
-            save_doc(kz_json:set_value([?NUMBERS_KEY, Number], NumberJObj, JObj))
+            NewNumberJObj = build_number_used_by(NewApp, NumberObj),
+            save_doc(kz_json:set_value([?NUMBERS_KEY, Number], NewNumberJObj, JObj))
     end.
 
+-spec build_number_used_by(kz_term:ne_binary(), kz_json:object()) -> kz_json:object().
+build_number_used_by('undefined', NumberObj) ->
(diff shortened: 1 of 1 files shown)


core/kazoo_number_manager/src/knm_port_request.erl
```diff
@@ -448,20 +448,21 @@ assign_to_app(Number, NewApp, JObj) ->
...
-                                  kz_json:get_value([?NUMBERS_KEY, Number], JObj)
-                                 )
-                             ])
-                          ),
-            save_doc(kz_json:set_value([?NUMBERS_KEY, Number], NumberJObj, JObj))
+            NewNumberJObj = build_number_used_by(NewApp, NumberObj),
+            save_doc(kz_json:set_value([?NUMBERS_KEY, Number], NewNumberJObj, JObj))
     end.
 
+-spec build_number_used_by(kz_term:ne_binary(), kz_json:object()) -> kz_json:object().
```
icehess: I was thinking something like set and remove operation like below. The code below I also protect the `assign_to_app` of adding the number accidentally if it is not there at the first place (just in case to make sure no one is messing up if they call this function manually):

```diff
diff --git a/core/kazoo_number_manager/src/knm_port_request.erl b/core/kazoo_number_manager/src/knm_port_request.erl
index d71ef6d628..982fb78911 100644
--- a/core/kazoo_number_manager/src/knm_port_request.erl
+++ b/core/kazoo_number_manager/src/knm_port_request.erl
@@ -445,21 +445,26 @@ get_user_name(AuthAccountId, UserId) ->
                            {'ok', kz_json:object()} |
                            {'error', any()}.
 assign_to_app(Number, NewApp, JObj) ->
-    case kz_json:get_value([?NUMBERS_KEY, Number, ?USED_BY_KEY], JObj) of
+    Numbers = kzd_port_requests:numbers(JObj),
+    assign_to_app(Number, NewApp, JObj, kz_json:get_json_value(Number, Numbers)).
+
+assign_to_app(_, _, _, 'undefined') ->
+    {'error', 'not_found'};
+assign_to_app(Number, NewApp, JObj, NumberJObj) ->
+    case kz_json:get_value(?USED_BY_KEY, NumberJObj) of
         NewApp -> {'ok', JObj};
         _OldApp ->
             lager:debug("assigning number ~s in port request ~s to ~s"
                        ,[Number, kz_doc:id(JObj), NewApp]
                        ),
-            NumberJObj = kz_json:from_list(
-                           props:filter_empty(
-                             [{?USED_BY_KEY, NewApp}
-                              | kz_json:to_proplist(
-                                  kz_json:get_value([?NUMBERS_KEY, Number], JObj)
-                                 )
-                             ])
-                          ),
-            save_doc(kz_json:set_value([?NUMBERS_KEY, Number], NumberJObj, JObj))
+            NumberJObj =
+                kz_json:set_value([Number, ?USED_BY_KEY]
+                                 ,NewApp
+                                 ,kz_json:delete_key([Number, ?USED_BY_KEY]
+                                                    ,kzd_port_requests:numbers(JObj)
+                                                    )
+                                 ),
+            save_doc(kzd_port_requests:set_numbers(JObj, NumberJObj))
     end.
 
 %%------------------------------------------------------------------------------
```

As you can see it removes the `used_by` key from the number (utilizing  the json key path) and then setting it to a new value. `kz_json/set_value/3` won't set the key if it is `undefined`, so both updating and deleting `used_by` key is covered.

zzzming: I'll use kzd_port_request. But the value in the number json is actually an empty object. So I would not pass undefined kz_json:set_value that will cause the entire the number removed. Instead, I just have two path to deal with undefined.

icehess:
@zzzming Please make a 4.3 PR for this.

PR #6073, opened 2019-10-04
PISTON-923: do not disable trapping exit in acdc_queue_manager

danielfinke:
This would prevent the gen_listener terminate callback from releasing
AMQP channels

This became required as of 65f6b4d3cd7812b25c6d6b1dfff4183736366108 when kz_amqp_history was removed. kz_amqp_history was responsible for capturing the DOWN message for the gen_listener and cleaning up its channels


jamesaimonetti:
Merged.

PR #6074, opened 2019-10-04
fix unbind from conference command

danielfinke:


jamesaimonetti:
Merged, obvious fix is obvious!

PR #6076, opened 2019-10-04
add account id lookup for kz_hooks_util CHANNEL_DESTROY

danielfinke:
same behaviour as CHANNEL_CREATE in case account ID must be determined by inbound destination


lazedo:
@danielfinke Hi, i think this is not suitable for `master` where all channels are identified by account, we should actually remove the part for channel create, but it doesn't hurt for now, just want to make sure you understand this and that this code and channel create too can be removed later

danielfinke:
@lazedo ok, that's good to know. This is related to changes made in Kamailio for master/v5?

PR #6082, opened 2019-10-08
PISTON-928:Added the ability to delete a doc id to kz_doc

bradfordben:
I know its rare that this is needed to be done but in the PostgreSQL driver im developing for MODB data i have a use case where i want to store the doc id in a column and the rest of the data without the doc id in another column so i need to delete the doc id. For that reason i feel this is worth adding.


jamesaimonetti:
Merged

bradfordben:
Thanks

PR #6087, opened 2019-10-10
Convert kapi_conf module to use kapi_definition module

harenson:


icehess:
@harenson Sorry, I totally forgot about this. Good catch on figuring out that `name` here is just a name for the event, and we usually used the `Event-Name` as the name.

PR #6089, opened 2019-10-10
[master] KZOO-20 and 21: allow assign rejected portin number and used_by check after completed portin

zzzming:
1. use 'phone_numbers_listing' view to get port in number assignment
2. the change applies to cb_callflow/cb_connectivity via cb_module_utils:apply_assigment_updates

3. Reconcile legacy used_by in the port-in document and trunkstore/callflow usage
4. Update portin database with the correct used_by when port is completed
5. Preserve used_by value in the new number database when a port in is completed.

### DIFF
--- a/core/kazoo_number_manager/src/knm_port_request.erl
+++ b/core/kazoo_number_manager/src/knm_port_request.erl
@@ -597,7 +597,7 @@ transition_numbers(PortReq) ->
 %% This may not be a reliable source of used_by comparing to the callflow doc.
 %% @end
 %%------------------------------------------------------------------------------
--spec app_used_by_portin(kz_term:ne_binary(), kz_json:object()) -> 'undefined' | list().
+-spec app_used_by_portin(kz_term:ne_binary(), kz_json:object()) -> kz_term:proplist().
 app_used_by_portin(Numbers, JObj) ->
     NumbersObj = kzd_port_requests:numbers(JObj),
     [{Num, kz_json:get_value([Num, ?USED_BY_KEY], NumbersObj)}
--- a/core/kazoo_number_manager/src/knm_port_request.erl
+++ b/core/kazoo_number_manager/src/knm_port_request.erl
@@ -143,14 +143,15 @@ get(DID=?NE_BINARY) ->
 %% @doc
 %% @end
 %%------------------------------------------------------------------------------
--spec get_portin_number(kz_term:ne_binary(), kz_term:ne_binary()) -> {'ok', kz_json:object()} |
+-spec get_portin_number(kz_term:ne_binary(), kz_term:ne_binary()) -> {'ok', kz_json:objects()} |
                                                                      {'error', any()}.
 get_portin_number(AccountId, DID=?NE_BINARY) ->
-    ViewOptions = [{'key', [AccountId, DID]}, 'include_docs'],
-    Result = kz_datamgr:get_single_result(?KZ_PORT_REQUESTS_DB, ?PORT_NUM_LISTING, ViewOptions),
-    case Result of
-        {'ok', Docs} ->
-            {'ok', kz_json:get_value(<<"doc">>, Docs)};
+    ViewOptions = [{'key', [AccountId, DID]}
+                  ,'include_docs'
+                  ],
+    case kz_datamgr:get_results(?KZ_PORT_REQUESTS_DB, ?PORT_NUM_LISTING, ViewOptions) of
+        {'ok', JObjs} ->
+            {'ok', [kz_json:get_value(<<"doc">>, JObj) || JObj <- JObjs]};
         {'error', _E}=Error ->
             lager:debug("failed to find portin number '~s': ~p", [DID, _E]),
             Error
--- a/core/kazoo_number_manager/src/knm_port_request.erl
+++ b/core/kazoo_number_manager/src/knm_port_request.erl
@@ -533,9 +533,9 @@ migrate() ->
 completed_port(PortReq) ->
     Numbers = kz_json:get_keys(kzd_port_requests:numbers(PortReq)),
     case transition_numbers(PortReq) of
-        {ok, Save} ->
+        {'ok', Save} ->
             reconcile_app_used_by(Numbers, PortReq),
-            {ok, Save};
+            {'ok', Save};
         Error -> Error
     end.
 
@@ -604,7 +604,7 @@ app_used_by_portin(Numbers, JObj) ->
         kz_term:is_not_empty(kz_json:get_value([Num, ?USED_BY_KEY], NumbersObj))
     ].
 
--spec reconcile_app_used_by(list(), kz_json:object()) -> 'ok'.
+-spec reconcile_app_used_by(kz_term:ne_binaries(), kz_json:object()) -> 'ok'.
 reconcile_app_used_by(Numbers, JObj) ->
     AccountId = kz_doc:account_id(JObj),
     AccountDb = kz_util:format_account_db(AccountId),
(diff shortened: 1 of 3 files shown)


core/kazoo_number_manager/src/knm_port_request.erl
```diff
@@ -260,9 +284,11 @@ states_to_scheduled(_AllowFromSubmitted='true') ->
 
 -spec transition_to_complete(kz_json:object(), transition_metadata()) -> transition_response().
 transition_to_complete(JObj, Metadata) ->
+    lager:info("JObj ~p~nMetadata ~p", [JObj, Metadata]),
     case transition(JObj, Metadata, [?PORT_PENDING, ?PORT_SCHEDULED, ?PORT_REJECTED], ?PORT_COMPLETED) of
         {'error', _}=E -> E;
-        {'ok', Transitioned} -> completed_port(Transitioned)
+        {'ok', Transitioned} ->
+            completed_port(Transitioned, JObj)
```
icehess: You don't need JObj in `completed_port`, just use the `Transitioned`

core/kazoo_number_manager/src/knm_port_request.erl
```diff
@@ -563,6 +594,68 @@ transition_numbers(PortReq) ->
...
+            {error, not_found};
+        {'ok', Doc} ->
+            {'ok', Doc};
+        E ->
+            lager:debug("Error find number ~p document view ~p error ~p", [Num, View, E]),
+            {error, E}
+    end.
+
+is_used_by_app(AccountDb, Num, App) ->
+    View = maps:get(App, ?APP_VIEW_MAP, undefined),
```
icehess: you don;t actually need this. Just pattern match in function header for callflow and trunk store

core/kazoo_number_manager/src/knm_port_request.erl
```diff
@@ -506,10 +532,15 @@ migrate() ->
...
--spec completed_port(kz_json:object()) -> transition_response().
-completed_port(PortReq) ->
-    lager:debug("transitioning numbers to active"),
-    transition_numbers(PortReq).
+-spec completed_port(kz_json:object(), kz_json:object()) -> transition_response().
+completed_port(PortReq, JObj) ->
+    AccountId = kz_doc:account_id(PortReq),
+    Numbers = kz_json:get_keys(kzd_port_requests:numbers(PortReq)),
+    lager:debug("transitioning numbers ~p to active", [Numbers]),
+    {'ok', App} = maybe_reconcile_app_used_by(lists:last(Numbers), AccountId, JObj),
```
icehess: Only checking the last number?

core/kazoo_number_manager/src/knm_port_request.erl
```diff
@@ -506,10 +532,15 @@ migrate() ->
...
+    Rc = transition_numbers(PortReq),
```
icehess: Since we can set the `used_by` at the number creation, then just call `transition_numbers(PortReq)` and if it `{ok, Save}` then process and assigned_to_app them

core/kazoo_number_manager/src/knm_port_request.erl
```diff
@@ -563,6 +594,68 @@ transition_numbers(PortReq) ->
...
+update_used_by(Numbers, App) ->
+    lager:debug("update number ~p with used_by ~p", [Numbers, App]),
+    knm_numbers:assign_to_app(Numbers, App).
+
+-spec app_used_by_portin(kz_term:ne_binary(), kz_json:object()) -> 'undefined' | kz_json:object().
+app_used_by_portin(Number, JObj) ->
+    Numbers = kzd_port_requests:numbers(JObj),
+    kz_json:get_value([Number, ?USED_BY_KEY], Numbers).
+
+maybe_reconcile_app_used_by(Num, AccountId, JObj) ->
```
icehess: Instead query db for each number you can just simply use used `[{'keys', Numbers}]` as view option.

So what you have to change here is:
1. create a proplist of form `[{Number, Appname}]` from port's numbers field
1. use `keys` view to to query dbs
1. check query db against the numbers in step one above and create a list of app assignment for each number if they have any
1. call `assign_to_app`

core/kazoo_number_manager/src/knm_port_request.erl
zzzming: Fixed

core/kazoo_number_manager/src/knm_port_request.erl
```diff
@@ -563,6 +594,66 @@ transition_numbers(PortReq) ->
...
+
+maybe_reconcile_app(App, App, _Num, _JObj) ->
+    {'ok', App};
+maybe_reconcile_app(_PortInApp, App, Num, JObj) ->
+    lager:warning("port in number ~p app ~p replaced by correct app ~p", [Num, _PortInApp, App]),
+    _Ok = assign_to_app(Num, App, JObj),
+    {'ok', App}.
+
+app_used_by([]) -> 'undefined';
+app_used_by([{_Num, App}|_T]) -> App.
```
icehess: Any Number can be assigned to any app.

Group numbers by app in previous function. return it as result. After port doc is saved loop over proplist and assign numbers to their app.

core/kazoo_number_manager/src/knm_port_request.erl
```diff
@@ -563,6 +594,66 @@ transition_numbers(PortReq) ->
...
+
+%%------------------------------------------------------------------------------
+%% @doc Apps used by in the port in document.
+%% This may not be a reliable source of used_by comparing to the callflow doc.
+%% @end
+%%------------------------------------------------------------------------------
+-spec app_used_by_portin(kz_term:ne_binary(), kz_json:object()) -> 'undefined' | list().
+app_used_by_portin(Numbers, JObj) ->
+    NumbersObj = kzd_port_requests:numbers(JObj),
+    [{Num, kz_json:get_value([Num, ?USED_BY_KEY], NumbersObj)} || Num <- Numbers].
```
icehess: Also filter numbers that have empty used_by
```
[{Num, kz_json:get_value([Num, ?USED_BY_KEY], NumbersObj)}
 || Num <- Numbers,
    kz_term:is_not_empty(kz_json:get_value([Num, ?USED_BY_KEY], NumbersObj))
]
```

core/kazoo_number_manager/src/knm_port_request.erl
```diff
@@ -563,6 +594,66 @@ transition_numbers(PortReq) ->
...
+    query_app_view(AccountDb, ?TRUNKSTORE_LIST, Numbers, <<"trunkstore">>).
+
+%%------------------------------------------------------------------------------
+%% @doc Apps used by phe number based on callflow or trunkstore docs.
+%% This is considered as a reliable source about used_by.
+%% @end
+%%------------------------------------------------------------------------------
+-spec used_by_which_app(kz_term:ne_binary(), kz_term:ne_binary()) -> list().
+used_by_which_app(AccountDb, Num) ->
+    used_by_app(AccountDb, Num, 'callflow') ++ used_by_app(AccountDb, Num, 'trunkstore').
```
icehess: Rename function to something like `get_dids_for_app`

```erlang
get_dids_for_app(AccountDb, Numbers) ->
    [{<<"callflow">>, get_dids_for_app(AccountDb, Numbers, ?CALLFLOW_LIST)}
    ,{<<"trunkstore">>, get_dids_for_app(AccountDb, Numbers, ?TRUNKSTORE_LIST)}
    ].

get_dids_for_app(AccountDb, Numbers, View) ->
    ViewOptions = [{'keys', Numbers}],
    case kz_datamgr:get_result_keys(AccountDb, View, ViewOptions) of
        {'ok', DIDs} -> DIDs;
        {'error', _} -> []
    end.
```

core/kazoo_number_manager/src/knm_port_request.erl
```diff
@@ -563,6 +594,66 @@ transition_numbers(PortReq) ->
...
+    PortInUsedBy = app_used_by_portin(Numbers, JObj),
+    AppUsedBy = used_by_which_app(AccountDb, Numbers),
+    _Ok=[maybe_reconcile_app(proplists:get_value(N, PortInUsedBy), App, N, JObj) || {N, App} <- AppUsedBy],
+    app_used_by(AppUsedBy).
+
+maybe_reconcile_app(App, App, _Num, _JObj) ->
+    {'ok', App};
+maybe_reconcile_app(_PortInApp, App, Num, JObj) ->
+    lager:warning("port in number ~p app ~p replaced by correct app ~p", [Num, _PortInApp, App]),
+    _Ok = assign_to_app(Num, App, JObj),
```
icehess: Either don't do this, or if you really want to update and fix apps in port doc before save, return the final port doc here.

I prefer to not change anything in port doc.

core/kazoo_number_manager/src/knm_port_request.erl
```diff
@@ -563,6 +594,66 @@ transition_numbers(PortReq) ->
...
+-spec app_used_by_portin(kz_term:ne_binary(), kz_json:object()) -> 'undefined' | list().
+app_used_by_portin(Numbers, JObj) ->
+    NumbersObj = kzd_port_requests:numbers(JObj),
+    [{Num, kz_json:get_value([Num, ?USED_BY_KEY], NumbersObj)} || Num <- Numbers].
+
+maybe_reconcile_app_used_by(Numbers, AccountId, JObj) ->
+    AccountDb = kz_util:format_account_db(AccountId),
+    PortInUsedBy = app_used_by_portin(Numbers, JObj),
+    AppUsedBy = used_by_which_app(AccountDb, Numbers),
+    _Ok=[maybe_reconcile_app(proplists:get_value(N, PortInUsedBy), App, N, JObj) || {N, App} <- AppUsedBy],
```
icehess: Instead of this, create a function, call it `log_reconcile_app` and tail recursion there and log numbers used by app.

core/kazoo_number_manager/src/knm_port_request.erl
```diff
@@ -508,8 +531,16 @@ migrate() ->
...
-    transition_numbers(PortReq).
+    AccountId = kz_doc:account_id(PortReq),
+    Numbers = kz_json:get_keys(kzd_port_requests:numbers(PortReq)),
+    AppUsedBy = maybe_reconcile_app_used_by(Numbers, AccountId, PortReq),
+    lager:debug("transitioning numbers ~p to active used by apps ~p", [Numbers, AppUsedBy]),
+    case transition_numbers(PortReq) of
+        {ok, Save} ->
+            update_used_by(Numbers, AppUsedBy),
+            {ok, Save};
+        _E -> _E
```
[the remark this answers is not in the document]
zzzming: Only updated this instance not the entire file.

PR #6095, opened 2019-10-15
[4.3] PROD-30: some enhancements to run custom tasks + template

icehess:
This is some changes required for PROD-30, mostly related to template rendering and tasks. 

* Move trigger-able macros to kazoo task core app
* Add option to `kz_json:from_list_recursive/2` to hint that don't treat list as ASCII chars list (printable chars) so empty list won't be converted to binary instead of a genuine empty list
* Add new type `kz_time:kazoo_month()` to refer to `{Year::kz_time:year(), Month::kz_time:month}`
* Add `kz_template:render_with_option/3` to pass render time options like `translation_fun`
* Some enhancements to ledgers and service audit view to get summary quicker

Backport of https://github.com/2600hz/kazoo/pull/6070


core/kazoo_stdlib/src/kz_json.erl
```diff
@@ -264,30 +266,45 @@ from_list(PL) when is_list(PL) ->
     ?JSON_WRAPPER([{K, utf8_binary(V)} || {K, V} <- PL, V =/= 'undefined']).
 
 -spec from_list_recursive(json_proplist()) -> object().
-from_list_recursive([]) -> new();
-from_list_recursive(L)
+from_list_recursive(L) ->
+    from_list_recursive(L, #{}).
+
+-type from_list_options() :: #{no_ascii_list => boolean()
```
zzzming: I wish the name does not have to be no double negative like `ascii_list_enforced` something like that. Can it be `ascii_list_enforced` => true as default on line 280? Or a comment or update md what no_ascii_list means.

core/kazoo_stdlib/include/kz_types.hrl
```diff
@@ -275,5 +275,24 @@
...
+%% try
+%%     ...
+%% catch
+%%     error:badarg ->
+%%         whatever;
+%%     ?STACKTRACE(E, R, Stack)
+%%         {error, {E,R, Stack}}
+%% end
+%% kz.mk defines the macro if OTP version is >= 21
+-ifdef(OTP_RELEASE).
```
jamesaimonetti: Why would this be needed? 4.3 is targeted to OTP 19.3, no?

icehess: It is not necessary needed in kazoo, I needed this for my code to support both 19 and 21.

PR #6104, opened 2019-10-18
[master] KZOO-15: ensure min value of max_recording_time_limit

zzzming:
max_recording_time_limit can be reset by values from system_config/callflow by kazoo_media_maintenance:migrate(). To prevent reset again, the original callflow (or any originals) data shall be removed after the migration.
    
It's also required that the system-config default should be 10800, as seconds in 3 hours.

### DIFF
--- a/core/kazoo_media/src/kz_media_util.erl
+++ b/core/kazoo_media/src/kz_media_util.erl
@@ -55,6 +55,7 @@
 
 -define(USE_ACCOUNT_OVERRIDES, kapps_config:get_is_true(?CONFIG_CAT, <<"support_account_overrides">>, 'true')).
 
+-define(DEFAULT_MAX_RECORDING_LIMIT, 3*?SECONDS_IN_HOUR).
 %%------------------------------------------------------------------------------
 %% @doc Normalize audio file to the system default or specified sample rate.
 %% Accepts media file content binary as input.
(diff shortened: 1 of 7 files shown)


core/kazoo_media/src/kz_media_util.erl
```diff
@@ -370,7 +371,8 @@ recording_url(CallId, Data) ->
 
 -spec max_recording_time_limit() -> ?SECONDS_IN_HOUR.
 max_recording_time_limit() ->
-    kapps_config:get_integer(?CONFIG_CAT, <<"max_recording_time_limit">>, ?SECONDS_IN_HOUR).
+    erlang:max(kapps_config:get_integer(?CONFIG_CAT, <<"max_recording_time_limit">>, ?DEFAULT_MAX_RECORDING_LIMIT)
+              ,?DEFAULT_MAX_RECORDING_LIMIT).
```
jamesaimonetti: Drop `).` to next line to align with the opening `(` from erlang:max/2

jamesaimonetti: I also am not sure about this erlang:max/2 usage. If I knowingly limit recordings to 1 hour in the system config, this will over-rule my config and use 3 hours, no?

zzzming: @jamesaimonetti Your concern is valid that this code will overwrite any number lower than the threshold, even if it's knowingly set.
@k-anderson can you help to weigh in what should be ceiling/floor of max_recording_time_limit. I explained in the PR description about why it's implemented such way. Thanks!

jamesaimonetti:
@zzzming the issue that when `migrate` commands are run, `kazoo_media_maintenance:migrate()` is run. This include migrating settings from the `system_config/callflow` doc. Looking at prod's callflow config, it sets `max_recording_time_limit` to `600`, and is migrated to the `media` doc on migrate.

So I would remove the `erlang:max` call.

The other piece is whether to update kazoo_media_maintenance to remove settings from the old doc when migrating them to the new doc (I lean to yes remove them, waiting on @k-anderson to confirm that as desirable).

jamesaimonetti:
Confirmed, please update kazoo_media_maintenance to remove the old setting.

Thanks!

PR #6106, opened 2019-10-18
PISTON-937: do not cache db statuses filtered by agent id

danielfinke:
Should not pass in Options as Agent-ID may be specified. That would mean the db statuses cache would only contain those of that agent and when reused for other reqs (to save the db hit) would mean that historical data is empty for all other agents.


jamesaimonetti:
merged

PR #6108, opened 2019-10-18
[4.3] KZOO-15: Ensure min value of max_recording_time_limit

zzzming:
max_recording_time_limit can be reset by values from system_config/callflow by kazoo_media_maintenance:migrate(). To prevent reset again, the original callflow (or any originals) data shall be removed after the migration.
    
It's also required that the system-config default should be 10800, as seconds in 3 hours.

### DIFF
--- a/applications/crossbar/priv/api/swagger.json
+++ b/applications/crossbar/priv/api/swagger.json
@@ -32242,7 +32242,7 @@
                     "type": "boolean"
                 },
                 "max_recording_time_limit": {
-                    "default": 3600,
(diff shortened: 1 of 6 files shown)


zzzming:
@k-anderson 
delete_original_config(ConfigJObj, Updates, DeleteAfter)

DeleteAfter is originally used to delete the entire system_config doc after migration. I used that to avoid updating the system_config doc if deletion is required. Therefore, the logic is we only save the doc when the flag is false which the doc won't be deleted. Makes sense?

PR #6112, opened 2019-10-22
[4.3] KZOO-20/21: fix app used_by for port_in number

zzzming:
There are a few bugs addressed in this PR.

Remove used_by json object under a port_in number when the associated callflow is deleted.

An earlier bug caused `used_by` not set when a port_in number is
in `rejected` state. Although this has been fix in this PR,
the ramification of earlier numbers without this fix still exists.
Therefore, this commit checks if `used_by` is not set properly when
the port_in transitions to `completed` state.

It also fixes another bug that a newly created phone_number as a result of
port_in completed will preserve `used_by` app. That has been missing.


icehess:
master pr: https://github.com/2600hz/kazoo/pull/6089

PR #6116, opened 2019-10-22
Add BICS as a Carrier

corporealshift:
Add the ability to use BICS as a carrier. Currently number search returns the available numbers that BICS returns, regardless of the search value entered. Until BICS provides support for area code based search there is no way to have the search work (They have it on their roadmap, it's just not implemented yet)

### DIFF
--- a/applications/crossbar/priv/api/swagger.json
+++ b/applications/crossbar/priv/api/swagger.json
@@ -34122,6 +34122,22 @@
             },
             "type": "object"
         },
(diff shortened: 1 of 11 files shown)


lazedo:
@corporealshift Hi, if you still interested in this, please rebase

lazedo:
@corporealshift Hi, please resolve the conflicts ( note that the path to kazoo_number_manager has changed to kazoo_numbers) and then reopen. thanks

PR #6117, opened 2019-10-23
Refactor kz_util and the kazoo core application

jamesaimonetti:
Move functions from kz_util into more appropriate modules. Also remove
the kazoo core application and move its modules to more appropriate
applications.

Include migration scripts for non-core apps to update their references
to moved functions and modules.


core/kazoo_config/src/kazoo_config.hrl
```diff
@@ -3,11 +3,11 @@
 -include_lib("kazoo_stdlib/include/kz_databases.hrl").
 -include_lib("kazoo_stdlib/include/kz_types.hrl").
 -include_lib("kazoo_stdlib/include/kz_log.hrl").
--include_lib("kazoo/include/kz_system_config.hrl").
+-include_lib("kazoo_apps/include/kz_system_config.hrl").
 
 -define(CONFIG_FILE_ENV, "KAZOO_CONFIG").
--define(CONFIG_FILE, "/etc/kazoo/config.ini").
--define(V4_CONFIG_FILE, "/etc/kazoo/core/config.ini").
+-define(CONFIG_FILE, "/etc/kazoo_apps/config.ini").
```
icehess: Are we going to change this too?

jamesaimonetti: good catch, errant change

jamesaimonetti:
Closing for now cause of circular deps in core

PR #6120, opened 2019-10-25
PISTON-942: allow overriding operator number on voicemail boxes

danielfinke:


applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -1754,6 +1756,8 @@ get_mailbox_profile(Data, Call) ->
                     ,media_extension = kzd_voicemail_box:media_extension(MailboxJObj)
                     ,forward_type = ?DEFAULT_FORWARD_TYPE
                     ,oldest_message_first = kzd_vmboxes:oldest_message_first(MailboxJObj)
+                    ,operator_number =
+                         kz_json:get_ne_value(<<"operator_number">>, MailboxJObj)
```
lazedo: use accessor

danielfinke:
@lazedo addressed your request to use the accessor instead
on our end, we've got to support both integers and strings for legacy reasons, but tests fail with the following:
```
Errors in file: ./applications/crossbar/priv/couchdb/schemas/vmboxes.json
path '.properties.operator_number.type': the value must be a single type and not an array of types.
```
is there a reason it's unsupported? It looks like you can have multiple types for OAS3: https://swagger.io/docs/specification/data-models/data-types/ "Mixed Types"

icehess:
> @lazedo addressed your request to use the accessor instead
> on our end, we've got to support both integers and strings for legacy reasons, but tests fail with the following:
> 
> ```
> Errors in file: ./applications/crossbar/priv/couchdb/schemas/vmboxes.json
> path '.properties.operator_number.type': the value must be a single type and not an array of types.
> ```
> 
> is there a reason it's unsupported? It looks like you can have multiple types for OAS3: https://swagger.io/docs/specification/data-models/data-types/ "Mixed Types"

No OAS3 is not supporting mixed types. This is stated in your link too:

> `type` takes a single value. `type` as a list is not valid in OpenAPI

Though you can use `anyOf` or `oneOf`.

danielfinke:
@icehess you know, I think I misunderstood what they were saying. I'll update this with `oneOf`

jamesaimonetti:
Merged

PR #6123, opened 2019-10-29
Kz endpoint

lazedo:


applications/ecallmgr/src/ecallmgr_fs_xml.erl
```diff
@@ -528,8 +528,19 @@ get_leg_vars([_|_]=Prop) ->
 get_leg_vars(JObj) -> get_leg_vars(kz_json:to_proplist(JObj)).
 
 -spec get_channel_vars(kz_json:object() | kz_term:proplist()) -> iolist().
-get_channel_vars([]) -> [];
-get_channel_vars([_|_]=Props) ->
+get_channel_vars(Param) ->
+    ["<"
```
jamesaimonetti: I assume if `Param` is `[]`, then returning `["<", [], ">"].` will not cause issues?

PR #6124, opened 2019-10-29
Store the sorted users when directory created

aisconn:
…ory functions

Conflicts:
	applications/callflow/src/module/cf_directory.erl


jamesaimonetti:
Merged @aisconn thanks! I added a commit on top of this one to use the `kzd_directories` accessor functions and added `search_fields` to the schema/kzd module as well.

PR #6126, opened 2019-10-30
PISTON-953: no need to flush endpoint cache after doc updates

danielfinke:
Flush should only flush the endpoint cache and callers of the fun do not need to do so as they are performing a doc update anyway. Recent changes to kazoo_caches remove this need.

This also resolves an infinite loop where a listener might act on a doc_{created,deleted,edited} event and request a flush of the endpoint cache before requesting the endpoint via kz_endpoint:get

### DIFF
--- a/applications/callflow/src/callflow_maintenance.erl
+++ b/applications/callflow/src/callflow_maintenance.erl
@@ -352,7 +352,8 @@ set_device_classifier_action(Action, Classifier, Uri) ->
     Update = [{[<<"call_restriction">>, Classifier, <<"action">>], Action}],
     UpdateOptions = [{'update', Update}],
 
-    {'ok', _} = kz_datamgr:update_doc(AccountDb, DeviceId, UpdateOptions).
+    {'ok', _} = kz_datamgr:update_doc(AccountDb, DeviceId, UpdateOptions),
+    'ok'.
 
 %%------------------------------------------------------------------------------
 %% @doc Checks if classifier defined in `system_config -> number_manager' doc.
--- a/applications/callflow/src/callflow_maintenance.erl
+++ b/applications/callflow/src/callflow_maintenance.erl
@@ -302,8 +302,6 @@ set_account_classifier_action(Action, Classifier, AccountDb) ->
     Update = [{[<<"call_restriction">>, Classifier, <<"action">>], Action}],
     {'ok', _} = kzd_accounts:update(AccountId, Update),
 
-    kz_endpoint:flush_account(AccountDb),
-
     io:format("  ...  classifier '~s' switched to action '~s'\n", [Classifier, Action]).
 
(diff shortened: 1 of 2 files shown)


lazedo:
@danielfinke this doesn't look right. the intent of publishing the flush is for other nodes to react to this flush. where do you see a problem with infinite loop ?

danielfinke:
@lazedo this is how to manifest the infinite loop: set up a gen_listener binding to doc_edited on a device doc. In the handler, call kz_endpoint:get/2. There is a problem here in that the endpoint cache might not be flushed yet because kz_cache_listener hasn't got around to it (both your doc_edited handler and kz_cache_listener are running in parallel). You can get around this by flushing the kazoo_endpoint_cache prior to kz_endpoint:get/2. But if you do this using the old version of kz_endpoint:flush/2, the publish of the doc_edited triggers your handler again. And then you're in the infinite loop.
This is fixed by removing the doc update, which is inconsistent with the behaviour of other cache flushes, e.g. kz_datamgr:flush_cache_docs, ecallmgr_maintenance:flush_authn, crossbar_maintenance:flush. They are per node

lazedo:
@danielfinke can you point out where in the source repo this happens ? as i said, the change to callflow_maintenance looks ok, but the one in kz_endpoint doesn't, one might want to flush the cache for an endpoint without a db change and still have the other nodes react to that.

danielfinke:
@lazedo acdc_agent_listener/handler is an example in your version, we also have a couple of internal apps that monitor for device doc changes for other reasons. Discovered the issue when I found that kz_endpoint:get/2 was returning stale results sometimes in the handlers.
I think the point about consistency applies to kz_endpoint:flush as well - all the other cache flushes have to be run per-node, so why any different for kz_endpoint?
I think what might be a good thing to add is a separate function with explicit support for a zone-wide or global flush. What do you think?

lazedo:
@danielfinke we really don't want the change to `kz_endpoint:flush`. `acdc` is community maintained (you're the top contributor to that :) ) and the changes should be made there. we welcome and appreciate any other spots that you or other members may find that can cause issues.

danielfinke:
@lazedo alright. I'll remove the change to kz_endpoint:flush.
I can't find any other apps in the open source repo that use kz_endpoint:get in a device doc_{created, edited, deleted} handler.
I _can_ solve this in acdc_agent_handler with a `kz_cache:erase_local('kazoo_endpoint_cache', {'kz_endpoint', Db, Id})` but it seems like bad form to have a cross-app reference like that (what if the cache name or cache key changed?). Hence my proposal to open up an exported function in kz_endpoint to do this. Would you accept a kz_endpoint:flush_local that contains just the kz_cache:erase_local?

lazedo:
@danielfinke thank you.
`Would you accept a kz_endpoint:flush_local that contains just the kz_cache:erase_local?` , `YES`

danielfinke:
@lazedo PR for local flushes is here: https://github.com/2600hz/kazoo/pull/6132

PR #6132, opened 2019-10-31
PISTON-955: add kz_endpoint:flush_account_local & flush_local

danielfinke:
New funs flush only the node's local cache


lazedo:
@danielfinke thanks.

PR #6135, opened 2019-11-01
move uri-related calls from kz_util to kz_http_util

jamesaimonetti:


applications/crossbar/src/cb_context.erl
```diff
@@ -421,12 +421,12 @@ method(#cb_context{method=M}) -> M.
 pretty_print(#cb_context{pretty_print = PrettyPrint}) -> PrettyPrint.
 
 -spec path_token(binary()) -> binary().
-path_token(Token) ->
-    kz_util:uri_decode(Token).
+path_token(<<"+", _/binary>>=Token) -> Token;
+path_token(Token) -> kz_http_util:urldecode(Token).
```
lazedo: @jamesaimonetti `kz_util:uri_decode` and `kz_http_util:urldecode` are very different. replacing one with the other may have undesirable side effects

jamesaimonetti: I don't disagree but they seem overlapping. So yeah, trying to uncover those inconsistencies :)

lazedo: you needed to change the `+` here. if the main intent is to discard kz_util, why not move `uri_???` to kz_http_util instead ? concerned with side effects of things we don't cover (yet) in tests.

jamesaimonetti: In this case, what happened was decoding `/v2/rates/number/+1234...` would parse into ` 1234...` and would fail `knm_converters:is_reconcilable/1` calls. I go back and forth cause `+` equaling a space is old-school while `%20` is the more "correct" way to encode spaces. And really I think `+` is only for query-string params or form bodies.

So maybe we change urldecode to account for that?

jamesaimonetti: @lazedo the more i read, the more `%20` in the path part is expected and right. I think the old `uri_decode/encode` functions should be used and a specifically named function to use urldecode sematics for `+` should be added when developers want to handle that legacy format.

Thoughts?

jamesaimonetti:
@lazedo i think we have the distinctions now. Use `legacy` if you really want to handle `+` as space; default to expected RFC. Let me know thoughts if you have any! :)

applications/crossbar/src/cb_context.erl
```diff
@@ -421,12 +421,11 @@ method(#cb_context{method=M}) -> M.
 pretty_print(#cb_context{pretty_print = PrettyPrint}) -> PrettyPrint.
 
 -spec path_token(binary()) -> binary().
-path_token(Token) ->
-    kz_util:uri_decode(Token).
+path_token(Token) -> kz_http_util:urldecode(Token, 'rfc').
```
lazedo: no need for `rfc` since its the default

lazedo:
@jamesaimonetti looks good, one minor in `cb_context:path_token`

PR #6139, opened 2019-11-05
Do not honor privacy settings for emergency calls

harenson:
If a user dials to an emergency number and the device, user, account, or carrier has `Caller-ID Privacy` enabled and/or the user uses `*67` feature code to perform the emergency call, disregard the privacy settings and use the configured caller-id for emergency calls.

### DIFF
--- a/applications/stepswitch/src/stepswitch_bridge.erl
+++ b/applications/stepswitch/src/stepswitch_bridge.erl
@@ -27,6 +27,12 @@
         ,code_change/3
         ]).
 
+-ifdef(TEST).
+-export[avoid_privacy_if_emergency_call/2
+       ,contains_emergency_endpoints/1
+       ].
+-endif.
+
 -include("stepswitch.hrl").
 -include_lib("kazoo_amqp/include/kapi_offnet_resource.hrl").
 -include_lib("kazoo_numbers/include/knm_phone_number.hrl").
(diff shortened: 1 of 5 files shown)


applications/stepswitch/src/stepswitch_bridge.erl
```diff
@@ -578,7 +579,7 @@ is_emergency_endpoint('true') ->
     'true';
 is_emergency_endpoint('false') -> 'false';
```
zzzming: I totally understand this is not the code that's updated in this PR. I raise this because I can see you try to improve the code quality. So why not they just do
`is_emergency_endpoint(B, _Endpoint) when is_boolean(B) -> B;`
Is it because they need to log? The other issue is that one variable used for both is_emergency and endpoint are confusing.  I hope there should be two parameters from the beginning. What do you think?

applications/stepswitch/src/stepswitch_bridge.erl
```diff
@@ -691,3 +692,16 @@ send_deny_emergency_response(OffnetReq, ControlQ) ->
 get_event_type(CallEvt) ->
     {Cat, Name} = kz_util:get_event_type(CallEvt),
     {Cat, Name, kz_call_event:call_id(CallEvt)}.
+
+
+%% If it is an emergency call (i.e 911 call) don't honor `privacy' settings.
```
jamesaimonetti: please use edoc style now. one day we'll generate these into the docs site :)

applications/stepswitch/src/stepswitch_bridge.erl
```diff
@@ -691,3 +692,16 @@ send_deny_emergency_response(OffnetReq, ControlQ) ->
...
+-spec avoid_privacy_if_emergency_call(boolean(), stepswitch_resources:endpoints()) -> stepswitch_resources:endpoints().
+avoid_privacy_if_emergency_call('true', Endpoints) ->
+    Values = [{?KEY_PRIVACY_HIDE_NAME, 'false'}
```
jamesaimonetti: Definitely a `lager:info` would be useful here to note the explicit setting of no privacy on the call.

lazedo:
@harenson why the changes in cf_route_req ( callflow ) ? stepswitch changes should be enough.

harenson:
@lazedo it's the way I found to handle the `*67` feature code scenario. Stepswitch change works for all the other scenarios, at least for the ones I was able to test locally.

PR #6144, opened 2019-11-08
change to use kz_time for UTC conversion

mark2600:
- this will prevent crashes or undefined behavior during "magic hour"
the time during a DST switch

### DIFF
--- a/applications/callflow/src/module/cf_temporal_route.erl
+++ b/applications/callflow/src/module/cf_temporal_route.erl
@@ -165,7 +165,7 @@ get_temporal_rules(#temporal{local_sec=LSec
 
 -spec get_temporal_rules(kz_json:path(), non_neg_integer(), kz_term:ne_binary(), kz_term:ne_binary(), rules()) -> rules().
 get_temporal_rules(Routes, LSec, AccountDb, TZ, Rules) when is_binary(TZ) ->
-    Now = localtime:utc_to_local(calendar:universal_time(), kz_term:to_list(TZ)),
+    Now = kz_time:adjust_utc_timestamp(calendar:universal_time(), kz_term:to_list(TZ)),
     get_temporal_rules(Routes, LSec, AccountDb, TZ, Now, Rules).
 
 -spec get_temporal_rules(routes(), non_neg_integer(), kz_term:ne_binary(), kz_term:ne_binary(), kz_time:datetime(), rules()) -> rules().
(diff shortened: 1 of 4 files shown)


applications/callflow/src/module/cf_temporal_route.erl
```diff
@@ -435,7 +435,7 @@ enable_temporal_rules(Temporal, [RuleId|T]=Rules, Call) ->
 %%------------------------------------------------------------------------------
 -spec load_current_time(temporal()) -> temporal().
 load_current_time(#temporal{timezone=Timezone}=Temporal)->
-    {LocalDate, LocalTime} = localtime:utc_to_local(calendar:universal_time()
+    {LocalDate, LocalTime} = kz_time:adjust_utc_timestamp(calendar:universal_time()
```
jamesaimonetti: formatting looks off

jamesaimonetti:
`make app_applications` will update the .app.src files

icehess:
Second arguments to `kz_time:adjust_utc_timestamp/2` is an integer or `kz_term:api_ne_binary()`, not `list()`.

mark2600:
Don't merge this, work still required

mark2600:
> Second arguments to `kz_time:adjust_utc_timestamp/2` is an integer or `kz_term:api_ne_binary()`, not `list()`.

That is a bad spec then, the last clause can accept a list without issue, in fact the spec for localtime:utc_to_local is for a string() and not a binary.

icehess:
> > Second arguments to `kz_time:adjust_utc_timestamp/2` is an integer or `kz_term:api_ne_binary()`, not `list()`.
> 
> That is a bad spec then, the last clause can accept a list without issue, in fact the spec for localtime:utc_to_local is for a string() and not a binary.

Well `kz_time:adjust_utc_timestamp/2` is not exactly like `localtime:utc_to_local/2`.

`adjust_utc_timestamp` is for converting UTC to local timestamp **if** you know the time difference in form of `+HH:MM`/`-HH:MM` or in seconds (integer) or you have the timezone (binary string), and it returns `kz_time:gregorian_seconds()` not `kz_time:datetime()`.

mark2600:
> Second arguments to `kz_time:adjust_utc_timestamp/2` is an integer or `kz_term:api_ne_binary()`, not `list()`.

That is a bad spec then, the last clause can accept a list without issue, in fact the spec for localtime:utc_to_local is for a string() and not a binary.

> > > Second arguments to `kz_time:adjust_utc_timestamp/2` is an integer or `kz_term:api_ne_binary()`, not `list()`.
> > 
> > 
> > That is a bad spec then, the last clause can accept a list without issue, in fact the spec for localtime:utc_to_local is for a string() and not a binary.
> 
> Well `kz_time:adjust_utc_timestamp/2` is not exactly like `localtime:utc_to_local/2`.
> 
> `adjust_utc_timestamp` is for converting UTC to local timestamp **if** you know the time difference in form of `+HH:MM`/`-HH:MM` or in seconds (integer) or you have the timezone (binary string), and it returns `kz_time:gregorian_seconds()` not `kz_time:datetime()`.

Yes, so I need to update this to use adjust_utc_datetime (to keep the return the same), but the parameters passed in should be left the same as they are passed to localtime:utc_to_local is what I was saying.

lazedo:
@mark2600 can we close this ?

PR #6147, opened 2019-11-08
Conference related updates

jamesaimonetti:
note when collection times out

bring the parent's call-id along

when a federated listener is spun up, it defaulted to all 0s for the
log ID, making it hard to correlate the listener to the parent
gen_listener.

Add the zone tag to the call-id of the parent to help associate the
listener but also differentiate it from other federated listeners.


icehess:
Needs rebase I think for `kz_log` changes :)

PR #6151, opened 2019-11-12
Notifications for Voicemail Delete Actions

noahmehl:
As a user, I may want to receive both AMQP and email notifications when a user deletes a voicemail.

AMQP messages (to the notifications exchange) are sent if voicemails are deleted via DTMF, delete after notify, and for crossbar actions.

A teletype email template: notification.voicemail_deleted is included and can be activated by:

```
# sup kapps_maintenance migrate
# sup teletype_voicemail_deleted init
# sup teletype_maintenance restore_system_templates
```

The teletype template can be made durable by adding it to system_config/notify.default.autoload_modules.

Right now, the voicemail delete template is essentially a copy of the voicemail new template with the Reason added.

This is a PR to 4.3 because we haven't been able to develop against master for a bit.  We'll submit a PR to master once we have all of the approvals for 4.3.

### DIFF
--- a/applications/callflow/src/module/cf_voicemail.erl
+++ b/applications/callflow/src/module/cf_voicemail.erl
@@ -866,7 +866,7 @@ play_messages(Messages, Count, Box, Call) ->
 
 -spec play_messages(kz_json:objects(), kz_json:objects(), non_neg_integer(), mailbox(), kapps_call:call()) ->
           'ok' | 'complete'.
-play_messages([H|T]=Messages, PrevMessages, Count, #mailbox{seek_duration=SeekDuration}=Box, Call) ->
+play_messages([H|T]=Messages, PrevMessages, Count, #mailbox{seek_duration=SeekDuration, mailbox_id=BoxId}=Box, Call) ->
     AccountId = kapps_call:account_id(Call),
     Message = kvm_message:media_url(AccountId, H),
     lager:info("playing mailbox message ~p (~s)", [Count, Message]),
@@ -890,6 +890,9 @@ play_messages([H|T]=Messages, PrevMessages, Count, #mailbox{seek_duration=SeekDu
             lager:info("caller chose to delete the message"),
             _ = kapps_call_command:flush(Call),
             _ = kapps_call_command:b_prompt(<<"vm-deleted">>, Call),
+            MessageId = kz_json:get_ne_binary_value(<<"media_id">>, H),
+            JObj = hd(kz_json:get_list_value(<<"succeeded">>, kvm_messages:fetch(AccountId, [MessageId], BoxId))),
+            kvm_util:publish_voicemail_deleted(BoxId, JObj, 'dtmf'),
             _ = kvm_message:set_folder({?VM_FOLDER_DELETED, 'false'}, H, AccountId),
             play_messages(T, PrevMessages, Count, Box, Call);
         {'ok', 'return'} ->
--- a/applications/teletype/priv/templates/voicemail_deleted.text
+++ b/applications/teletype/priv/templates/voicemail_deleted.text
@@ -1,0 +1,38 @@
+                Voicemail Deleted
+
+
+Hi{% if user.first_name %} {{user.first_name}}{% endif %},
+
+A voicemail from {{caller_id.name_number}} for
+your voicemail box at {{voicemail.vmbox_name}} (number: {{voicemail.vmbox_number}}) was deleted.
+Please find the message audio file in the attachment.
+{% if voicemail.transcription %}
+
+Deletion reason: {{reason}}
+
+=== Transcription ===
+
+{{voicemail.transcription}}
+{% endif %}
+
+=== Voicemail Message Details ===
+
+    Caller: {{caller_id.name_number}}
+    Callee: {{to.user}} (originally dialed number)
+    Received: {{date_called.local|date:"l, F j, Y \\a\\t H:i"}}
+    Length: {{voicemail.length}}{% if voicemail.file_name %}
+    File Name: {{voicemail.file_name}}
+    File Type: {{voicemail.file_type|upper}}
+    File Size: {{voicemail.file_size|filesizeformat}}{% endif %}
+
+
+
+Account Information
+
+    Account ID: {{account.id}}
+    Account Name: {{account.name}}
+    Account Realm: {{account.realm}}
+
+
+
+Sent from {{system.encoded_node}}
--- a/applications/teletype/src/templates/teletype_voicemail_deleted.erl
+++ b/applications/teletype/src/templates/teletype_voicemail_deleted.erl
@@ -1,0 +1,320 @@
+%%%-----------------------------------------------------------------------------
+%%% @copyright (C) 2014-2020, 2600Hz
+%%% @doc
+%%% @author James Aimonetti
+%%% @end
+%%%-----------------------------------------------------------------------------
+-module(teletype_voicemail_deleted).
+-behaviour(teletype_gen_email_template).
+
+-export([id/0
+        ,init/0
+        ,macros/0, macros/1
+        ,subject/0
+        ,category/0
+        ,friendly_name/0
+        ,to/0, from/0, cc/0, bcc/0, reply_to/0
+        ]).
+-export([handle_req/1]).
+
+-include("teletype.hrl").
+
+-spec id() -> kz_term:ne_binary().
+id() ->
+    <<"voicemail_deleted">>.
+
+-spec macros() -> kz_json:object().
+macros() ->
+    kz_json:from_list(
+      [?MACRO_VALUE(<<"voicemail.vmbox_id">>, <<"voicemail_vmbox_id">>, <<"Voicemail Box Id">>, <<"Which voicemail box was the message left in">>)
+      ,?MACRO_VALUE(<<"voicemail.msg_id">>, <<"voicemail_msg_id">>, <<"Voicemail Message ID">>, <<"Message Id of the voicemail">>)
+      ,?MACRO_VALUE(<<"voicemail.transcription">>, <<"voicemail_transcription">>, <<"Voicemail Message Transcription">>, <<"Voicemail Message Transcription">>)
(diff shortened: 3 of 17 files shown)


noahmehl:
@jamesaimonetti We're hitting this failure in the `build_centos7` test:

```
cp: cannot stat 'CHANGELOG': No such file or directory
```

We're not sure what's happening here.  Can you let us know if we're supposed to add something for a CHANGELOG?

core/kazoo_voicemail/src/kvm_message.erl
```diff
@@ -239,7 +239,9 @@ set_folder(Folder, Message, AccountId) ->
     end.
 
 -spec maybe_set_folder(kz_term:ne_binary(), kz_term:ne_binary(), kz_term:ne_binary(), kz_term:ne_binary(), kz_json:object()) -> db_ret().
```
icehess: Can you please fix the spec here, the second argument should be `vm_folder()`.

core/kazoo_voicemail/src/kvm_message.erl
```diff
@@ -884,6 +886,7 @@ maybe_update_meta(Length, Action, Call, MediaId, BoxId) ->
         'delete' ->
             lager:debug("attachment was sent out via notification, set folder to delete"),
             Fun = [fun(JObj) ->
+                           'ok' = kvm_util:publish_voicemail_deleted(BoxId, JObj, 'delete_after_notify'),
```
icehess: I think this is creating a bit noise, sending a delete email notification immediately after sending a new voice mail email.
If the user set the `delete_after_notify` option they already specified they don't want it in their in vmbox.

applications/callflow/src/module/cf_voicemail.erl
```diff
@@ -860,6 +860,9 @@ play_messages([H|T]=Messages, PrevMessages, Count, Box, Call) ->
         {'ok', 'delete'} ->
             lager:info("caller chose to delete the message"),
             _ = kapps_call_command:b_prompt(<<"vm-deleted">>, Call),
+            MessageId = kz_json:get_ne_binary_value(<<"media_id">>, H),
+            JObj = hd(kz_json:get_list_value(<<"succeeded">>, kvm_messages:fetch(AccountId, [MessageId], BoxId))),
+            'ok' = kvm_util:publish_voicemail_deleted(BoxId, JObj, 'dtmf'),
```
icehess: No need to pattern match `'ok'` here. Dialyzer won't complain about unused return value if it is a simple atom (like ok or boolean).

applications/teletype/priv/templates/voicemail_deleted.text
```diff
@@ -0,0 +1,38 @@
...
+
+
+Hi{% if user.first_name %} {{user.first_name}}{% endif %},
+
+A voicemail from {{caller_id.name_number}} for
+your voicemail box at {{voicemail.vmbox_name}} (number: {{voicemail.vmbox_number}}) was deleted.
+Please find the message audio file in the attachment.
+{% if voicemail.transcription %}
+
+Delete reason: {{reason}}
```
icehess: Please change the label to more readable "Deletion Reason:" string, same thing for html template file.

applications/teletype/src/templates/teletype_voicemail_deleted.erl
```diff
@@ -0,0 +1,312 @@
...
+      ,{<<"msg_id">>, kz_json:get_value(<<"voicemail_id">>, DataJObj)}
+      ,{<<"name">>, kz_json:get_value(<<"voicemail_id">>, DataJObj)} %% backward compatibility
+      ,{<<"transcription">>, get_transcription(DataJObj)}
+      ,{<<"length">>, pretty_print_length(DataJObj)}
+      ]).
+
+-spec render_vm_delete_reason(kz_json:object()) -> kz_term:api_ne_binary().
+render_vm_delete_reason(DataJObj) ->
+    case kz_json:get_atom_value(<<"reason">>, DataJObj) of
+        'dtmf' -> <<"user pressed DTMF key">>;
```
icehess: I think the reason strings here can be  more user friendly.  How about some thing like this:

* `dtmf`: Delete by calling voicemail box
* `crossbar_action`: Delete from UI

icehess:
That error in `build_centos7` should be fixed by now!

core/kazoo_voicemail/src/kvm_message.erl
noahmehl: @icehess this is happening because of the use case from @extremerotary.  They want to be able to bind to the amqp messages for deleted email, even if it happens in an automated way.

applications/teletype/src/templates/teletype_voicemail_deleted.erl
noahmehl: @icehess I'm ok with the `dtmf` string as "Deleted by calling voicemail box", however I would suggest for `crossbar_action` to be "Deleted via API" (as it could happen via a system <--> system action, and not just from MonsterUI for instance).

core/kazoo_voicemail/src/kvm_message.erl
noahmehl: Yes, we will update for this.

applications/callflow/src/module/cf_voicemail.erl
noahmehl: We will update for this.

applications/teletype/priv/templates/voicemail_deleted.text
noahmehl: We will update for this.

extremerotary:
@noahmehl - I'm fine if you want to remove the email portion

core/kazoo_voicemail/src/kvm_message.erl
noahmehl: @icehess is this acceptable?

applications/teletype/src/templates/teletype_voicemail_deleted.erl
noahmehl: @icehess would that work?

noahmehl:
As discussed with @jamesaimonetti 

We should just add a feature flag to enable/disable (disabled by default) to the system_config document that kvm (kazoo voicemail core) uses.

@icehess this should resolve the concern about amqp chattiness.  We will leave the email template disabled by default as well.

applications/teletype/src/templates/teletype_voicemail_deleted.erl
icehess: Yes, "Deleted via API" is better.

icehess:
> As discussed with @jamesaimonetti
> 
> We should just add a feature flag to enable/disable (disabled by default) to the system_config document that kvm (kazoo voicemail core) uses.
> 
> @icehess this should resolve the concern about amqp chattiness. We will leave the email template disabled by default as well.

I prefer to change `teletype_templates:init/1` and the way it initialize the template. So if the template is using  `teletype_templates:init/1` then in `teletype_templates:params/1` it calls the `Module:enabled/0` to get the default value, and then when it is updating the document in system_config if the `enabled` field is not set, it insert the default value.

In addition as you suggested, add a system_config setting to disabled the voicemail_deleted AMQP event altogether.

icehess:
That change to `teletype_templates:init/1` was a suggestion, another option is to use `teletype_util:template_system_value/3` in template module to read a setting if email should send or not with default `false`.

I think this is better and doesn't need too much code change.

@noahmehl What do you think?

noahmehl:
> That change to `teletype_templates:init/1` was a suggestion, another option is to use `teletype_util:template_system_value/3` in template module to read a setting if email should send or not with default `false`.
> 
> I think this is better and doesn't need too much code change.
> 
> @noahmehl What do you think?

We will update for this :)

icehess:
> @icehess @jamesaimonetti we've put all of the amqp messaging behind a feature flag (disabled by default). However, we're having trouble understanding this:
> 
> ```
> I prefer to change teletype_templates:init/1 and the way it initialize the template. So if the template is using teletype_templates:init/1 then in teletype_templates:params/1 it calls the Module:enabled/0 to get the default value, and then when it is updating the document in system_config if the enabled field is not set, it insert the default value.
> ```
> 
> @icehess Is there an existing use-case you can link to in the code?

Ignore this comment. Use the alternative approach of `teletype_util:template_system_value/3` in the module `handle_req/1` function to see if the template is enabled or not (by default it is not enabled).

applications/teletype/src/templates/teletype_voicemail_deleted.erl
```diff
@@ -0,0 +1,312 @@
...
+    lager:debug("invalid data for ~s", [id()]),
+    teletype_util:notification_failed(id(), <<"validation_failed">>);
+handle_req(JObj, 'true') ->
+    lager:debug("valid data for ~s, processing...", [id()]),
+
+    %% Gather data for template
+    DataJObj = kz_json:normalize(JObj),
+    AccountId = kz_json:get_value(<<"account_id">>, DataJObj),
+
+    case teletype_util:is_notice_enabled(AccountId, JObj, id()) of
```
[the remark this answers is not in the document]
icehess: @noahmehl Yes, check `teletype_util:is_notice_enabled/3` then if it was `true` check the system value.

You can look at the `teletype_fax_inbound_error_to_email_filtered` file for an example.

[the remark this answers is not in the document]
noahmehl: @icehess please review the update :)

jamesaimonetti:
@noahmehl sorry! I have time today to merge but this needs a rebase against latest origin/4.3 `git rebase origin/4.3` and correct conflicts then push.

jamesaimonetti:
And make sure to update origin to latest with `git fetch origin` (or whatever you named the 2600Hz remote).

noahmehl:
@jamesaimonetti I have rebased off of 2600hz/kazoo/4.3.

noahmehl:
@jamesaimonetti looks like we have to update something regarding `make code_checks`.  I will do that as soon as possible.

noahmehl:
@jamesaimonetti builds are passing now :)

jamesaimonetti:
Merged!

PR #6153, opened 2019-11-14
[4.3] Protect system when pagination is disabled

jamesaimonetti:
When fetching large datasets with pagination turned off, it is
possible to tank the system when the dataset causes memory pressure
during loading.

This change adds an optional system config parameter for determining
the maximum memory allowed for the Erlang process handling the API
request.

Additionally, even requests with `pagination=false` (or /v1) will
still be loaded on a per-page basis. Each page load will check the
memory consumption and error with an HTTP 416 response if memory usage
exceeds the configured maximum.

Until the maximum is reached, however, the pages will continue to be
fetched until exhausted as before.

FixtureDB has an added utility to make mirroring an existing CouchDB
database to the FixtureDB fixtures directory for easier conversion of
testing data.

### DIFF
--- a/applications/crossbar/src/crossbar_view.erl
+++ b/applications/crossbar/src/crossbar_view.erl
@@ -903,7 +903,7 @@ handle_query_result(#{page_size := PageSize
     case check_page_size_and_length(LoadMap, FilteredLength, FilteredJObjs, NewLastKey) of
         {'exhausted', LoadMap2} -> LoadMap2;
         {'next_db', LoadMap2} when PageSize =:= 'infinity', NewLastKey =/= 'undefined' ->
-            lager:info("updating new last key to ~p from ~p", [NewLastKey, _OldLastKey]),
+            lager:debug("updating new last key to ~p from ~p", [NewLastKey, _OldLastKey]),
             get_results(LoadMap2#{last_key => NewLastKey});
         {'next_db', LoadMap2} -> get_results(LoadMap2#{databases => RestDbs})
     end.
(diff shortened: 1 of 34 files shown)


applications/crossbar/src/crossbar_view.erl
```diff
@@ -826,27 +854,36 @@ get_results(#{databases := [Db|RestDbs]=Dbs
...
+    {NewLastKey, JObjs} = last_key(LastKey, Results, Limit, ResultsLength, PageSize),
 
     case apply_filter(Mapper, JObjs) of
         {'error', Reason} ->
             LoadMap#{context => cb_context:add_system_error('datastore_fault', kz_term:to_binary(Reason), Context)};
         FilteredJObjs when is_list(FilteredJObjs) ->
             FilteredLength = length(FilteredJObjs),
-            lager:debug("db_returned: ~b passed_filter: ~p next_start_key: ~p", [ResultsLength, FilteredLength, NewLastKey]),
-            handle_query_result(LoadMap, Dbs, FilteredJObjs, FilteredLength, NewLastKey);
-        FilteredJObj ->
```
icehess: I think `cb_multi_factor` is still returning a single `JObj` (a change from this PR https://github.com/2600hz/kazoo/pull/5904).

Also for master version of this PR can you add this case clause back, it seems it gets removed somehow.

applications/crossbar/src/crossbar_view.erl
```diff
@@ -973,15 +1044,23 @@ filter_foldl(Mapper, [JObj | JObjs], Acc) ->
...
-last_key(LastKey, JObjs, Limit, Returned) when Returned < Limit ->
+last_key(LastKey, JObjs, 'undefined', _Returned, _PageSize) ->
+    lager:info("no limit, re-using last key ~p", [LastKey]),
     {LastKey, lists:reverse(JObjs)};
-last_key(_LastKey, JObjs, Limit, Returned) when Returned == Limit ->
+last_key(_LastKey, JObjs, Limit, Limit, _PageSize) ->
+    lager:info("full page fetched, calculating new key"),
+    new_last_key(JObjs);
+last_key(_LastKey, JObjs, _Limit, _Returned, _PageSize) ->
+    lager:info("returned ~p < limit ~p ps: ~p lk: ~p", [_Returned, _Limit, _PageSize, _LastKey]),
```
icehess: s/lk/last_key/g is more descriptive.

applications/crossbar/src/crossbar_view.erl
```diff
@@ -991,12 +1070,21 @@ last_key(_LastKey, JObjs, Limit, Returned) when Returned == Limit ->
...
 %%------------------------------------------------------------------------------
 -spec format_response(load_params()) -> cb_context:context().
-format_response(#{total_queried := TotalQueried
-                 ,queried_jobjs := JObjs
-                 ,context := Context
-                 ,last_key := NextStartKey
-                 ,start_key := StartKey
-                 }) ->
+format_response(#{context := Context}=LoadMap) ->
+    case cb_context:resp_status(Context) of
```
icehess: Is there a situation it couldn't be `success` at this point?

icehess:
I like the `db_to_disk`, yay!

applications/crossbar/src/crossbar_view.erl
jamesaimonetti: Was going off all the specs and this type: https://github.com/2600hz/kazoo/blob/4.3/applications/crossbar/src/crossbar_view.erl#L83

Good to know about multi_factor, will look into it.

jamesaimonetti: I only see cb_multi_factor using `{'mapper', crossbar_view:map_value_fun()}`...am I missing a mapper that folds into a single JSON object instead?

jamesaimonetti: Looks like it calls `merge_summary` after the results are mapped/returned.

applications/crossbar/src/crossbar_view.erl
jamesaimonetti: Any error from kz_datamgr will be propagated in `Context` within the `LoadMap`.

applications/crossbar/src/crossbar_view.erl
jamesaimonetti: updated texts and downgraded to debug in places

applications/crossbar/src/crossbar_view.erl
icehess: You right, I don't why I was thinking it returning single `JObj` or why @lazedo added that case clause at the first place.

jamesaimonetti:
Fails are due to kz_couch_view + couchbeam dialyzer issues, not being fixed in this PR.

PR #6157, opened 2019-11-14
Remove ecallmgr cache object when TTS media expires

noahmehl:
When a TTS media object expires, we should remove the cached object from ecallmgr's cache.


lazedo:
@noahmehl this is not acceptable. you are calling a kazoo app from a core app.

noahmehl:
@lazedo would it be acceptable if media_mgr sent an amqp message for this?

If so, should ecallmgr handle it, since it's ecallmgr's cache object?

PR #6161, opened 2019-11-15
KZOO-11: Audit Summary Enhancements

mk1s:
# Audit Summary Enhancements

1. Change`../audit_summary` path spec to conform to convention `../audit/summary`
1. Updated API docs for new path spec.
1. Added `kazoo_yodb` modules for year only databases to store service audits
1. Updated `kz_services.erl` to log quantity changes to reseller YODB

## Audit Summary By Date Payload:
```
"data": [
    {
      "account": {
        "2019-12-05": {
          "addition": 1,
          "last_timestamp": 63742805754,
          "quantity": 2,
          "removal": 0,
          "sum_quantity": false
        },
        "2019-12-09": {
          "addition": 1,
          "last_timestamp": 63743140658,
          "quantity": 2,
          "removal": 1,
          "sum_quantity": false
        }
      },
      "mailbox": {
        "2019-12-09": {
          "addition": 3,
          "last_timestamp": 63743140695,
          "quantity": 3,
          "removal": 0,
          "sum_quantity": false
        }
      },
      "softphone": {
        "2019-12-09": {
          "addition": 1,
          "last_timestamp": 63743140580,
          "quantity": 3,
          "removal": 0,
          "sum_quantity": false
        }
      },
      "user": {
        "2019-12-09": {
          "addition": 1,
          "last_timestamp": 63743140695,
          "quantity": 2,
          "removal": 0,
          "sum_quantity": false
        }
      }
    }
  ],
```
## Audit Summary by Source Payload:
```
"data": [
    {
      "2019-12-09": {
        "addition": 1,
        "last_timestamp": 63743140658,
        "quantity": 2,
        "removal": 1,
        "sum_quantity": false
      }
    },
    {
      "2019-12-05": {
        "addition": 1,
        "last_timestamp": 63742805754,
        "quantity": 2,
        "removal": 0,
        "sum_quantity": false
      }
    }
  ],
```


icehess:
@mk1s I'm wondering why not add accounts to services as a quantifiable item? Though it may be a bit difficult to implement also considering some corner cases like when this you're creating the first account in the system (master account) and other corner cases.

Also why not keep the same path `item.changes.difference.billable` as the rest of the audit logs so you don't need to handle especial case for account?

applications/crossbar/src/modules/cb_accounts.erl
```diff
@@ -410,6 +410,7 @@ delete(Context, Account) ->
                                                      ,cb_context:auth_token(Context)
                                                      ),
             _ = cb_mobile_manager:delete_account(Context1),
+            maybe_audit_reseller(Context1),
```
icehess: How about moving these stuff to `kzdb_account`?

core/kazoo_modb/priv/couchdb/views/services.json
```diff
@@ -31,23 +31,34 @@
...
-                "        var changedBillable = item.changes.difference.billable;",
-                "        emit([day, item.item], {",
-                "          'addition': changedBillable > 0 ? changedBillable : 0,",
-                "          'last_timestamp': doc.pvt_created,",
-                "          'quantity': item.billable || 0,",
-                "          'removal': changedBillable < 0 ? Math.abs(changedBillable) : 0,",
-                "          'sum_quantity': false",
-                "        });",
+                "      if (doc.audit.changes.source === 'accounts') {",
+                "          var qty = doc.audit.changes.quantity;",
```
icehess: You can keep the same structure like `item.changes.difference.billable` above when creating the account's audit log so you don't need to handle especially this case.

applications/crossbar/doc/ref/services.md
```diff
@@ -198,21 +188,31 @@ curl -v -X GET \
 
 ## Fetch
 
-> GET /v2/accounts/{ACCOUNT_ID}/services/audit_summary/{SOURCE_SERVICE}
+> GET /v2/accounts/{ACCOUNT_ID}/services/audit/{AUDIT_ID}
```
jamesaimonetti: Does this section of fetching by audit ID need to migrate to the real doc?

k-anderson:
@mk1s @icehess Howdy :) so I like the thought but it breaks something I am trying to maintain, that is the quantifiers can all be resolved via the account db.  The more dbs required the more queries (lower efficiency) and more 'manipulation' (complexity) required to have an authoritative truth of the billable quantities.  That is to say, I am trying to maintain it such that in one query we have have the absolute truth that at any time can replace all known account quantities (self healing) without needing to process 'exceptions/extensions' to the view results.

However! If you hard coded an account quantity of one in a category or more simply just used the existence of the document (without enabled:false or deleted:true) to emit a 1 for a item such as "accounts/account" then the cascade quantities would automatically 'sum' the sub-accounts.  IE:
Here:
https://github.com/2600hz/kazoo/blob/master/core/kazoo_services/src/kazoo_services_maintenance.erl#L152
before the switch statement hard-code...
`emit(['accounts', 'account'], 1);`
and here:
https://github.com/2600hz/kazoo/blob/master/core/kazoo_services/src/kz_services_quantities.erl#L345
initialize the accumulator such:
`Updates = lists:foldl(fun(F, Updates) -> F(JObj, Updates end, [{[<<"accounts">>, <<"account">>], 1}], Routines)`
Those changes would cause the account to always emit an 'account' quantity of 1 for the category "accounts" and item "account" if its not disabled or deleted.  As such, without changes, the cascade quantity will always reflect the sum of those quantities, IE: the number of sub-accounts.  I believe that (for discussion) if that change is as simple as I am thinking we should do it regardless of wither it is ultimately used by this project because collecting that data on the EOM and BOM could be very interesting moving forward AND adds functionality to several other components which could be very useful in the future (for example, would provide the means to have a flat rate charge per-account or other account based charges which could be useful to other deployments with different billing models).

BUT BUT, audit logs are only written when billable quantities change, depending on the assigned service plan.   The issue with the current approach to 're-use' the existing audit logs is currently they are expected to be logs of billable changes that link to what changed on the invoice and any resulting ledgers (for prorated charges and activation charges) which also can link to transactions if a payment token needs to be invoked.  By using the audit logs in this way there are "two" types.  The billing audit logs that exist today (when you change something you are charged for, and we have no plans to charge "per-account") and these exceptions that track non-billing changes (changes that may or may not cost you money but are tracked for their history).  I would be very interested separating these concepts.  My first thought on how to approach that would be create a quantity change audit log.  Something that regardless of the billable update, as long as the quantities changed, gets stored.  The concern here would be the resulting documents, there would be an enormous amount...  I think the balance point might be to collect quantity audit logs in the MODb for select 'object' types.  For example, a system_config paramter that lists "category/item" that if a crossbar request results in a change to cause an audit log to get stored.  The idea being for our use case we populate that list with "devices/_all", "users/_all", "accounts/_all".

Thoughts?  Do we want to have a call to discuss this further? :)

core/kazoo_yodb/src/kazoo_yodb_maintenance.erl
```diff
@@ -0,0 +1,109 @@
...
+%%------------------------------------------------------------------------------
+%% @doc
+%% @end
+%%------------------------------------------------------------------------------
+-spec delete_yodbs(kz_term:ne_binary(), boolean() | kz_term:ne_binary()) -> 'ok' | 'no_return'.
+delete_yodbs(<<_/binary>> = Period, ShouldArchive) ->
+    Regex = <<"(2[0-9]{3})">>,
+    case re:run(Period, Regex, [{'capture', 'all', 'binary'}]) of
+        {'match', [_Full, Year]} ->
+            delete_yodbs(kz_term:to_integer(Year), kz_term:is_true(ShouldArchive));
```
jamesaimonetti, line 39: Think you need to update the spec to include `integer()` or `kz_time:year()`

core/kazoo_services/src/kz_services.erl
```diff
@@ -1184,6 +1185,92 @@ save_services_jobj(Services, ProposedJObj) ->
...
+    Diffs = kz_json:diff(CQ, PQ),
+    Keys = kz_json:get_keys(kz_json:flatten(Diffs)),
+    calculate_diffs({Keys, CQ, PQ}, []).
+
+%%------------------------------------------------------------------------------
+%% @doc generates the changes JObj for each diff key with changed and billable
+%% quantities
+%% @end
+%%------------------------------------------------------------------------------
+-spec calculate_diffs({kz_json:objects(), kz_json:object(), kz_json:object()}, kz_json:object()) -> kz_term:proplist().
```
jamesaimonetti, line 1261: I don't think `Keys` is a `kz_json:objects()` but `kz_json:keys()`

core/kazoo_yodb/doc/maintenance.md
```diff
@@ -0,0 +1,9 @@
+## SUP-able functions
+
+| Function | Arguments | Description |
+| -------- | --------- | ----------- |
+| `archive_yodbs/0` |  | |
```
jamesaimonetti, line 5: Can I get a description or three???

core/kazoo_yodb/src/kazoo_yodb.app.src
```diff
@@ -0,0 +1,13 @@
+{application, kazoo_yodb,
+ [
+  {description, "YODb - Utilities for an account's year-only database"},
+  {vsn, "4.0.0"},
+  {modules, []},
+  {registered, [ ]},
+  {applications, [ kernel
```
jamesaimonetti, line 7: You might try running `make apps_of_app` in the yodb directory to properly populate this file.

core/kazoo_services/src/kz_services.erl
```diff
@@ -1184,6 +1185,92 @@ save_services_jobj(Services, ProposedJObj) ->
...
+    end.
+
+%%------------------------------------------------------------------------------
+%% @doc extracts diff of services and returns audit changes JObj
+%% @end
+%%------------------------------------------------------------------------------
+-spec generate_diff_obj(kz_json:object(), kz_json:object()) -> kz_json:object().
+generate_diff_obj(CurrentServicesJObj, ProposedServicesJObj) ->
+    Key = [<<"quantities">>],
+    PQ = kz_json:get_value(Key, ProposedServicesJObj),
```
jamesaimonetti, line 1250: Can you use a typed getter here? `kz_json:get_json_value(Key, ProposedServicesJObj)` will help know what PQ should be later on.

core/kazoo_services/src/kz_services.erl
```diff
@@ -1184,6 +1185,92 @@ save_services_jobj(Services, ProposedJObj) ->
...
+    case kz_json:is_empty(Diff) of
+        'true' -> 'false';
+        _ -> 'true'
+    end.
+
+%%------------------------------------------------------------------------------
+%% @doc extracts diff of services and returns audit changes JObj
+%% @end
+%%------------------------------------------------------------------------------
+-spec generate_diff_obj(kz_json:object(), kz_json:object()) -> kz_json:object().
```
jamesaimonetti, line 1247: calc_diffs/2 returns a proplist so this function should probably have that as the return too.

core/kazoo_services/src/kz_services.erl
```diff
@@ -1184,6 +1185,92 @@ save_services_jobj(Services, ProposedJObj) ->
...
+%% quantities
+%% @end
+%%------------------------------------------------------------------------------
+-spec calculate_diffs({kz_json:objects(), kz_json:object(), kz_json:object()}, kz_json:object()) -> kz_term:proplist().
+calculate_diffs({[], _, _}, Acc) -> Acc;
+calculate_diffs({[Key | Keys], CQ, PQ}, Acc) ->
+    Proposed = kz_json:get_integer_value(Key, PQ, 0),
+    Current = kz_json:get_integer_value(Key, CQ, 0),
+    [_Source, Category, Item] = Key,
+    Obj = [{<<"category">>, Category}
```
jamesaimonetti, line 1267: Generally a list of proplists is...odd. Should this be `[kz_json:from_list(Obj) | Acc]` and the return be `kz_json:objects()`?

PR #6166, opened 2019-11-18
Update erlang formatter to 1.7.0 and format project

jamesaimonetti:


jamesaimonetti:
Just a handful of dialyzer complaints that aren't relevant; everything else passes

PR #6167, opened 2019-11-18
[4.3] Adjust page size based on chunked/pagination settings

jamesaimonetti:
Add tests for 4 scenarios from "big data" fetches:
1. paginated and chunked
2. paginated and unchunked
3. unpaginated and chunked
4. unpaginated and unchunked

Wonderfully, 3 does not trip the memory limit while 4 does, so
chunking is a nice benefit for unpaginated requests. Might consider
defaulting to `is_chunked=true` when `paginate=false`...


jamesaimonetti:
See master in #6168

PR #6169, opened 2019-11-19
Kz cache ttl

noahmehl:
This is a draft PR while we work out the specifics of how to accomplish this.

### DIFF
--- a/applications/ecallmgr/src/ecallmgr_util.erl
+++ b/applications/ecallmgr/src/ecallmgr_util.erl
@@ -38,7 +38,6 @@
 -export([media_path/1, media_path/2, media_path/3, media_path/4
         ,lookup_media/4
         ]).
--export([remove_media/1]).
 -export([unserialize_fs_array/1, unserialize_fs_props/1]).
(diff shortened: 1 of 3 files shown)


noahmehl:
As discussed with @jamesaimonetti, media_mgr can use: https://github.com/2600hz/kazoo/blob/master/core/kazoo_amqp/src/api/kapi_conf.erl to emit document changes, specifically document deleted when a TTS cache PID is removed.  We can use `TTS` as the database, and the MD5 hash as the document `ID`

We can extend ecallmgr to bind for events of document deleted for the `TTS` database, and the MD5 hash as the document `ID`, by extending cache properties here: https://github.com/2600hz/kazoo/blob/f5e88bc7821b4965e94877cf6659363e820cf8e2/applications/ecallmgr/src/ecallmgr_auxiliary_sup.erl#L31

noahmehl:
@jamesaimonetti it looks like we just needed to update the payload for kz_cache_listener to match the correct document.  Can you please take a look?

jamesaimonetti:
Merged, nice!

noahmehl:
@jamesaimonetti should we submit a PR for master as well?

PR #6170, opened 2019-11-19
[4.3] Do not honor privacy settings for emergency calls (#6139)

harenson:
Master https://github.com/2600hz/kazoo/pull/6139/

If a user dials to an emergency number and the device, user, account,
or carrier has Caller-ID Privacy enabled and/or the user uses *67
feature code to perform the emergency call, disregard the privacy
settings and use the configured caller-id for emergency calls.

### DIFF
--- a/applications/stepswitch/src/stepswitch_bridge.erl
+++ b/applications/stepswitch/src/stepswitch_bridge.erl
@@ -23,6 +23,12 @@
         ,code_change/3
         ]).
 
+-ifdef(TEST).
+-export[avoid_privacy_if_emergency_call/2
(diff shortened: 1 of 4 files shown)


applications/stepswitch/src/stepswitch_bridge.erl
```diff
@@ -23,6 +23,12 @@
         ,code_change/3
         ]).
 
+-ifdef(TEST).
+-export[avoid_privacy_if_emergency_call/2
```
jamesaimonetti: This doesn't look properly formed?

PR #6174, opened 2019-11-25
Remote callflow actions

lazedo:
* early fail with amqp
* returned value to allow continuation
* cf_module can call cf_remote_action explicitly or `remote_execution: true` can be defined as a peer or `module` in `Flow`

### DIFF
--- a/applications/callflow/src/cf_remote_action.erl
+++ b/applications/callflow/src/cf_remote_action.erl
@@ -1,0 +1,94 @@
+%%%-----------------------------------------------------------------------------
+%%% @copyright (C) 2011-2019, 2600Hz
+%%% @doc Executes remote callflow actions.
+%%%
+%%% This Source Code Form is subject to the terms of the Mozilla Public
+%%% License, v. 2.0. If a copy of the MPL was not distributed with this
+%%% file, You can obtain one at https://mozilla.org/MPL/2.0/.
+%%%
+%%% @end
+%%%-----------------------------------------------------------------------------
+-module(cf_remote_action).
+
+-export([handle/2
+        ,handle/3
+        ]).
+
+-include("callflow.hrl").
+
+-spec handle(kz_json:object(), kapps_call:call()) -> 'ok'.
+handle(Data, Call) ->
+    Action = kapps_call:kvs_fetch('remote-action', Call),
+    handle(Action, Data, Call).
+
+-spec handle(atom() | kz_term:ne_binary(), kz_json:object(), kapps_call:call()) -> 'ok'.
+handle(Action, Data, Call) ->
+    Q = kapps_call:controller_queue(Call),
+    ActionId = kz_binary:rand_hex(16),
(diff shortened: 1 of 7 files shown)


applications/callflow/src/cf_remote_action.erl
```diff
@@ -0,0 +1,81 @@
...
+
+    Routines = [{fun kapps_call:kvs_erase/2, ['cf_flow']}
+               ,{fun kapps_call:kvs_store/3, 'action-exit-keys', Keys}
+               ,fun kapps_call:clear_helpers/1
+               ],
+    kapps_call:to_json(kapps_call:exec(Routines, Call)).
+
+-spec wait_for_msg(atom() | kz_term:ne_binary(), kapps_call:call()) -> 'ok'.
+wait_for_msg(ActionId, Call) ->
+    receive
```
jamesaimonetti: It seems like a "reasonable" timeout should be included here. Might also use `kapps_call_command:receive_event/1` instead of naked receive?

core/kazoo_amqp/src/api/kapi.erl
```diff
@@ -57,3 +58,17 @@ encode_pid(Queue) ->
...
     list_to_binary(["pid://", kz_term:to_binary(Pid), "/", Queue]).
+
+-spec decode_pid(kz_term:ne_binary()) -> kz_term:api_pid().
+decode_pid(Queue) ->
+    case re:run(Queue, <<"pid://(.*)/(.*)">>, [{'capture', [1,2], 'binary'}]) of
+        {'match', [Pid, _RK]} -> kz_term:to_pid(Pid);
+        _Else -> 'undefined'
+    end.
+
+decode_pid2(<<"pid://", Pid/binary>>) ->
```
jamesaimonetti: @lazedo alternative implementation possibility

jamesaimonetti: see kz_api_tests perf test

applications/callflow/src/cf_remote_action.erl
lazedo: `kapps_call_command:receive_event/1`  doesn't handle `{'amqp_msg', {'amqp_return', _JObj, _Returned}}`

core/kazoo_amqp/src/api/kapi.erl
lazedo: i like that. only change would be to make sure it would handle `<<"pid://<0.1.0>/2bdace1ec24371/1cc5236e01d5849d18/2bdace1ec243711cc5...">>`

lazedo: it does work as intended. we can use this instead. thanks

applications/callflow/src/cf_remote_action.erl
lazedo: about the `timeout`, if there's no handler for the `action` and because of `mandatory`, we will receive the `amqp_return` message. if an application handles it, it needs to send back a message.

PR #6175, opened 2019-11-25
Revert changes on cf_route_req for e911 issue

harenson:
This module was changed to handle `*67` feature code scenario when a user calls an emergency number and uses the privacy feature code.


applications/callflow/src/cf_route_req.erl
```diff
@@ -253,15 +248,13 @@ pre_park_action(Call) ->
           kapps_call:call().
 update_call(Flow, NoMatch, ControllerQ, Call) ->
     FlowId = kz_doc:id(Flow),
-    CaptureGroup = kz_json:get_ne_value(<<"capture_group">>, Flow),
-    NewFlow = avoid_privacy_if_emergency_call(CaptureGroup, Flow),
     Props = [{'cf_flow_id', FlowId}
```
lazedo: can we fully revert it so it to stay as it was before ?
* remove ` FlowId = kz_doc:id(Flow),`
* change `Props = [{'cf_flow_id', FlowId}` to `Props = [{'cf_flow_id', kz_doc:id(Flow)}`

PR #6181, opened 2019-12-05
PISTON-974: set cid name/number on kapps_call when performing conf dial action

danielfinke:
create_call creates a kapps_call used for generating the endpoints for the dial request. Right now, with the authorizing_id set on the kapps_call, it is not possible to override the Caller ID used for ringing the endpoints. kz_attributes will choose the endpoint CID (either the name of the user that owns the conference or the conference name). By explicitly setting the caller ID to the values supplied in the HTTP request body and not setting the authorizing ID, the caller ID can be set as desired.

Looking for critique here in case there is something that I may be breaking, especially by removing the authorizing ID. I think that this proposed change results in not using the owner user's name at all. If that is fine then great, but maybe there is a better way to fix this.

### DIFF
--- a/applications/crossbar/src/modules/cb_conferences.erl
+++ b/applications/crossbar/src/modules/cb_conferences.erl
@@ -482,7 +482,7 @@ dial(Context, ConferenceId, Data) ->
 build_valid_endpoints(Context, ConferenceId, Data) ->
     case kz_json_schema:validate(<<"conferences.dial">>, Data) of
         {'ok', ValidData} ->
-            build_endpoints_to_dial(Context, ConferenceId, kz_json:get_list_value(<<"endpoints">>, ValidData));
+            build_endpoints_to_dial(Context, ConferenceId, ValidData);
         {'error', Errors} ->
             lager:info("dial data failed to validate"),
             {cb_context:failed(Context, Errors), []}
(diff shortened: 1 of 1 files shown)


jamesaimonetti:
I believe the authorizing ID is needed because we loopback into the system to check call restrictions and whatnot from the perspective of the device being called. I wonder what the effect of setting authorizing-id to the account ID instead of the device? Would you be able to set caller ID on the request then?

lazedo:
@danielfinke did you tried using `Retain-CID` ?
```
-spec build_number_endpoint(kz_term:ne_binary(), build_acc()) -> build_acc().
build_number_endpoint(Number, ?BUILD_ACC(Endpoints, Call, Context, Element)) ->
    AccountRealm = kapps_call:account_realm(Call),
    Endpoint = [{<<"Invite-Format">>, <<"loopback">>}
               ,{<<"Route">>,  Number}
               ,{<<"To-DID">>, Number}
               ,{<<"To-Realm">>, AccountRealm}
               ,{<<"Simplify-Loopback">>, 'true'}
               ,{<<"Loopback-Bowout">>, 'true'}
               ,{<<"Custom-Channel-Vars">>
                ,kz_json:from_list([{<<"Account-ID">>, kapps_call:account_id(Call)}
                                   ,{<<"Authorizing-Type">>, <<"conference">>}
                                   ,{<<"Authorizing-ID">>, kapps_call:authorizing_id(Call)}
                                   ,{<<"Loopback-Request-URI">>, <<Number/binary, "@", AccountRealm/binary>>}
                                   ,{<<"Request-URI">>, <<Number/binary, "@", AccountRealm/binary>>}
                                   ,{<<"Require-Ignore-Early-Media">>, true}
                                   **,{<<"Retain-CID">>, true}**
                                   ])

                }
               ],
```

lazedo:
@danielfinke try using `Retain-CID`

PR #6189, opened 2019-12-09
PISTON-979: fix migrating templates from system into accounts

danielfinke:
Notification template fields such as to, from, cc, bcc were being lost as only the `kz_notification:base_properties` were being written to the account-local notification doc during migration. This was being missed on the API req that performed the migration as it would succeed because the missing properties had been inherited from the parent/system during the validate step. But then subsequent requests would be missing the properties because the notification template exists in the account and inheritance is therefore not performed.


icehess:
@danielfinke Thanks. Please also backport this to 4.3.

danielfinke:
@icehess backported in #6200

PR #6190, opened 2019-12-10
PISTON-976: acdc status stats query/storage performance optimization

danielfinke:
Use ETS ordered_set table type, tuple key, and reverse select to optimize queries against acdc_stats_status ETS table. Reverse order makes "Limit n" queries very fast as they are just the first n results per agent from the result set. Also using maps instead of JObjs to make updates fast.


jamesaimonetti:
Merged!

PR #6193, opened 2019-12-11
PISTON-990: allow gen_listener implementers to hibernate

danielfinke:
gen_listener implementers' handle_call, handle_cast, and handle_info cannot hibernate the gen_listener right now.
This has caused, as an example, kz_amqp_worker to consume a lot of binary ref memory over time without an opportunity for garbage collection to release its binary refs.
This can be seen by looking at the length of the binary list for a pid via
```
{'binary', Binaries} = process_info(Pid, 'binary'),
length(Binaries).
```
and comparing the results of the same commands after manually trigger a GC on the pid via `erlang:garbage_collect(Pid)`.
This change allows kz_amqp_worker as well as other gen_listener implementers to actually get hibernation on demand.


jamesaimonetti:
Merged!

PR #6194, opened 2019-12-11
PISTON-992: remove etag from cb_queues resps that contain agent roster

danielfinke:
If API clients are using the req/resp cache headers, etag should not be set on the resp for dynamic responses as the resp in the cache may be stale.


jamesaimonetti:
Merged!

PR #6201, opened 2019-12-13
PISTON-993: Callee Field in Voicemail Notification email now matches …

CodyAntcliffe:
…the destination vmbox rather than the the caller #


jamesaimonetti:
@CodyAntcliffe thanks, merged!

PR #6204, opened 2019-12-13
Allow webhooks to be sent using PUT

jamesaimonetti:
- listen for and remove account hooks when account is deleted
- allow 'all' modifier for hook types and event types
- test put/post for webhooks


jamesaimonetti:
See 4.3 in #6202

PR #6230, opened 2020-01-02
Stagger deleting a file based on filesize

jamesaimonetti:
When using sendfile in Crossbar, there can exist a race condition
where the filename has been passed to Cowboy for sending, Cowboy
handles streaming the file asynchronously, and calls Crossbar's
finish_request callback.

Because the streaming and finish_request callback are happening in
parallel, it is possible for the file to be deleted from the local
disk before streaming has completed.

pqc_cb_ledgers will occassionally provoke this race condition.

Since deleting the file occurs in a spawned process itself, use the
rounded natural log of the file's size * 10 to calculate a small sleep
timeout before deleting the file.

### DIFF
--- a/applications/crossbar/src/api_util.erl
+++ b/applications/crossbar/src/api_util.erl
@@ -1248,23 +1248,35 @@ execute_request_results(Req, Context, _RespStatus) ->
 %% @end
 %%------------------------------------------------------------------------------
 -spec finish_request(cowboy_req:req(), cb_context:context()) -> 'ok'.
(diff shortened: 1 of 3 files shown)


lazedo:
@jamesaimonetti can you find a way to reproduce the issue so we can take a look at how cowboy handles this ?

jamesaimonetti:
@lazedo since it is hit or miss, I've instead tied the delete to a monitor of the connection PID. Presumably the send is finished when the connection PID dies :)

PR #6237, opened 2020-01-07
added reverse proxy apache instructions

yumminova:
Not clean, but i could not get it to work with letsencrypt certs


jamesaimonetti:
@yumminova cleaned up the text a bit but otherwise, thanks!

yumminova:
Ur welcome, I will update once I know how to use let's encryption. Should be simple (and not a question to you))

Verzonden met ProtonMail Mobile

-------- Oorspronkelijk bericht --------
Aan 7 jan. 2020 22:36, James Aimonetti wrote:

PR #6239, opened 2020-01-09
[4.3] Kazoo 6127: Added configuration parameters to control cluster alerts

k-anderson:
Create the following parameters on the crossbar.alerts document in system config, each is a boolean value that when false suppresses that alert category or specific type:

- check_financials
  - check_low_balance
  - check_payment_token
- check_port_requests
    - check_port_action_required
    - check_port_suspended
- check_system_alerts

For example, setting "check_financials" to false will disable both low balance and payment token alerts.  To disable low balance alerts only, leave "check_financials" and set "check_low_balance" to false.

### DIFF
--- a/applications/crossbar/priv/api/swagger.json
+++ b/applications/crossbar/priv/api/swagger.json
@@ -29922,6 +29922,47 @@
             },
             "type": "object"
         },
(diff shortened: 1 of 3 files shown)


k-anderson:
Master PR #6240

PR #6242, opened 2020-01-09
Remove kz_json:merge clauses that overwrite instead of merge nested JObjs

danielfinke:
This is a draft to discuss an oddity with `kz_json:merge_left` and `kz_json:merge_right`.
In the current version, there is inconsistent behaviour when merging nested JObjs.
Say you have A:
```
{
    "a": {
        "a1": "test",
        "a2": "test"
    }
}
{[{<<"a">>,
   {[{<<"a1">>,<<"test">>},{<<"a2">>,<<"test">>}]}}]}
```
and you merge B:
```
{
    "a": {}
}
{[{<<"a">>,{[]}}]}
```
the result is
```
{[{<<"a">>,{[]}}]}
```

But if you merge C:
```
{
    "a": {
        "a1": "overridden"
    }
}
{[{<<"a">>,{[{<<"a1">>,<<"overridden">>}]}}]}
```
the result is:
```
{[{<<"a">>,
   {[{<<"a2">>,<<"test">>},{<<"a1">>,<<"overridden">>}]}}]}
```

So in one case, the case where the node to be merged (source) is empty, _it erases all the keys in the target_. But when the node to be merged is not empty, _keys in the target are not erased, but just replaced if they happen to be in the source_.

It seems to me that the appropriate behaviour would be to always merge nested JObjs, but I could also be swayed in the direction that the source should always overwrite the target (this would make it no longer a recursive merge though). The former (always merge) would produce a result for `kz_json:merge(A, B)` like:
```
{[{<<"a">>,
   {[{<<"a2">>,<<"test">>},{<<"a1">>,<<"test">>}]}}]}
```

There's obviously a backwards-compatibility concern here. Maybe it would be good to pick a direction to go, deprecate the current implementation, and create a new, consistent implementation for future use?

### DIFF
--- a/core/kazoo_stdlib/src/kz_json.erl
+++ b/core/kazoo_stdlib/src/kz_json.erl
@@ -448,8 +448,6 @@ merge_left(_K, {'right', ?JSON_WRAPPER(_)=Right}, Options) ->
     {'ok', merge(fun merge_left/3, Right, new(), Options)};
 merge_left(_K, {'right', V}, _Options) -> {'ok', V};
 
-merge_left(_K, {'both', ?EMPTY_JSON_OBJECT=Left, ?JSON_WRAPPER(_)=_Right}, _Options) ->
-    {'ok', Left};
 merge_left(_K, {'both', ?JSON_WRAPPER(_)=Left, ?JSON_WRAPPER(_)=Right}, Options) ->
     {'ok', merge(fun merge_left/3, Left, Right, Options)};
 merge_left(_K, {'both', Left, _Right}, _Options) -> {'ok', Left}.
@@ -474,8 +472,6 @@ merge_right(_K, {'right', ?JSON_WRAPPER(_)=Right}, Options) ->
     {'ok', merge(fun merge_right/3, new(), Right, Options)};
 merge_right(_K, {'right', V}, _Options) -> {'ok', V};
 
-merge_right(_K, {'both', ?JSON_WRAPPER(_)=_Left, ?EMPTY_JSON_OBJECT=Right}, _Options) ->
-    {'ok', Right};
 merge_right(_K, {'both', ?JSON_WRAPPER(_)=Left, ?JSON_WRAPPER(_)=Right}, Options) ->
     {'ok', merge(fun merge_right/3, Left, Right, Options)};
 merge_right(_K, {'both', _Left, Right}, _Options) -> {'ok', Right}.


jamesaimonetti:
I think the debate becomes what to do here:

`kz_json:merge_right(<<"a">>,{both,{[{<<"a1">>,<<"test">>},{<<"a2">>,<<"test">>}]},{[]}})`

Right now the code explicitly chooses B from `{both, A, B}`:

```erlang
merge_right(_K, {'both', ?JSON_WRAPPER(_)=_Left, ?EMPTY_JSON_OBJECT=Right}, _Options) ->                                            
    {'ok', Right};
```

The semantics being, prefer the right value unless both are non-empty objects themselves, in which case merge Left and Right.

But I'm confused as to why you aren't satisfied with `kz_json:merge_recursive(A, B)` since that provides you with the recursion into the object at key `"a"`?

Even the [kz_json doc](https://docs.2600hz.com/dev/core/kazoo_stdlib/doc/kz_json/#merge3-vs-merge_recursive2) calls out the behaviour. So I don't think the behaviour is "inconsistent" - its working as intended. If you want to treat `{[]}` recursively use `merge_recursive`; otherwise `{[]}` is used as the merge value.

danielfinke:
@jamesaimonetti so yeah, `kz_json:merge_recursive(A, B)` does this the way I would expect. I'm just unclear on why there is the desirability for empty objects to overwrite but non-empty objects to merge. Here's a hypothetical scenario that seems weird with the merge behaviour, when using cb_configs/kapps_config_util:
1. system_config defines "category" with
```
{
    "permissions": {
        "admin": "1",
        "manager": "2",
        "sysadmin": "3"
    }
}
```
2. account defines "configs_category" as:
```
{
    "permissions": {
        "account-level custom role": "4"
    }
}
```
This works great, as they get the union of the permissions
3. Later, they don't need that custom role anymore so they remove that prop:
```
{
    "permissions": {}
}
```
But this leads to all the permissions being removed. It seems unintuitive that the entire permissions object needs to be removed instead of just the prop that it no longer needed.

A direction this could go is just that kapps_config_util may be more suited to the `merge_recursive` behaviour?

jamesaimonetti:
Yeah, I would harmonize kapps_account_config to use merge_recursive in that case, leave the implementations alone.

danielfinke:
@jamesaimonetti Sounds good.

As a separate/later PR, since it seems this particular recursion behaviour is the only important distinction between the two implementations (the other being key order, which doesn't matter), would you be open to removing merge_recursive in favour of a new merge option in `Options` of merge/merge_left/merge_right that enables the recursion?
There might be more to it (such as dealing with the predicates in some `merge_recursive` clauses), but before I dig into it I want to make sure there's openness on your end to it

jamesaimonetti:
Yeah open to making it a config option to merge! We can mark merge_recursive as deprecated for a version or two then remove it entirely once the option to merge is in place.

lazedo:
@danielfinke seems this was superseded by #6254 and its a draft, closing

PR #6245, opened 2020-01-10
Update cdrs.md

extremerotary:
Add some additional information on query string parameters and their meanings

### DIFF
--- a/applications/crossbar/doc/cdrs.md
+++ b/applications/crossbar/doc/cdrs.md
@@ -73,6 +73,20 @@ Key | Description | Type | Default | Required | Support Level
 
 ## Fetch
 
+### Prerequisites 
+When using the GET API to retrieve CDR records, the response will be paginated if there are more than 50 records. The 50 records-per-page is the default, so if you've changed your system configuration, then you may have more, but be advised, 2600hz does not recommend more than 50 records per page. 
+### Available query string parameters
+These parameters can be applied to the end of any GET CDRs API call. To add query string parameters, simply put a question mark (?) at the end of the URL followed by the parameter. Multiple parameters are strung together using the ampersand (&).
+
+
+Parameter | Description | Type | Default | Required |
+--------- | ----------- | ---- | ------- | -------- |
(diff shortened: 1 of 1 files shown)


applications/crossbar/doc/cdrs.md
```diff
@@ -73,6 +73,20 @@ Key | Description | Type | Default | Required | Support Level
 
 ## Fetch
 
+### Prerequisites 
+When using the GET API to retrieve CDR records, the response will be paginated if there are more than 50 records. The 50 records-per-page is the default, so if you've changed your system configuration, then you may have more, but be advised, 2600hz does not recommend more than 50 records per page. 
+### Available query string parameters
+These parameters can be applied to the end of any GET CDRs API call. To add query string parameters, simply put a question mark (?) at the end of the URL followed by the parameter. Multiple parameters are strung together using the ampersand (&).
```
jamesaimonetti: This all seems a bit redundant.

Querystring parameters are covered on the [Basics](https://docs.2600hz.com/dev/applications/crossbar/doc/filters/) page, [pagination sections](https://docs.2600hz.com/dev/applications/crossbar/doc/basics/#pagination) and [timestamp section](https://docs.2600hz.com/dev/applications/crossbar/doc/basics/#timestamps).

Thoughts?

extremerotary: Hey @jamesaimonetti 
I didn't notice those pages because they are in a URL that's hidden / not able to be navigated to from docs.2600hz.com . I noticed your URL is /dev instead of /supported.
I agree they are a little redundant (assuming the /dev will eventually be public / published), but i also was trying to explain in laymen's terms what these are and how they are used. I think the bit about the default pagination is important so a user knows when to expect pagination. 
Totally up to your discretion if you want that extra stuff in there, i was just trying to make the one doc a little more verbose :-)

jamesaimonetti:
@extremerotary made a couple changes, added links to the basics/filter pages. Thanks for the updates! Also the /dev is "public" in that its there for anyone to use. /supported is the "default" redirect so that customers of 2600Hz are encouraged to use what is supported by 2600Hz support.

PR #6247, opened 2020-01-15
knm refactor - part3

icehess:
* Removing `number` from `knm_state` and `knm_options` to distinguish them from `knm_numbers` better
* Moving all public face functions from both `knm_number` and `knm_numbers` to a single module`knm_numbers`.
    * This modules always returns `{'ok', Collection::knm_pipe:collection()}` or `{'dry_run', Quotes::kz_json:object()}`.
    * The caller should determine if the operation was successful or not using knm_pipe utilities. (though it needs more work, any feedback is welcomed)
    * `knm_numbers` function won't return `dry_run`, but the ok tuple with collection, during dry run on any number failure, so it's up to caller to check what is failed or what is not. ~~(probably needs better implementation like returning `{ok, Success, Failed}` to reduce dependency on on calling `knm_pipe`)~~ **(See updates 1/5/2020 about knm_numbers return)**.
* Move all collection and pipe operation functions to `knm_ops` (from `knm_numbers`).
    * `knm_ops` functions are returning `knm_pipe:collection()` instead of `ok`/`dry_run` tuples.
    * This intended to used internally by knm modules only, or if the caller app/modules want to deal with knm_pipe `do/atempt/pipe` directly.
    * For other knm usage and functionality use only the public face module `knm_numbers`.
* Remove old and unused (hopefully) `knm_numbers:reconcile/1,2`
* Experimenting with streamlining logic starting with number creation
    * Returning `ok/error` tuple instead of pipe usage as far as possible and use `kz_either` instead of knm_pipe.
    * Instead of throwing, return error tuple.
    * Fail fast if the options passed to create is not correct instead of failing after doing expensive db ops.  
* some other code cleanups

**Updates (1/15/2020)**

* `knm_numbers` now returns `{'ok', JObjs}` / `{'ok', JObjs, FailedProplist}` / `{'dry_run', Quote}`.
    * So no need to call knm_pipe at all
    * It is easier to distinguish between successful operation and failed one.
    * Returning successes and failed as list also make it easier to figure out if the operation was on a single number or multiple (this is mainly for cb_phone_numbers and kt_numbers).
    * This need future work to possibly do not depend on if the operation is for single number or not.
    * note that successful numbers are JObjs, generated by `knm_phone_number:to_public_json/1`.
    * also note that Failed is a proplist
* move failed and failed reasons types to knm_errors to abstract the type of Failed.

**Updates (2/10/2020)**

* Update teletype for TEST compile to not save smtp log, which was causing fixturedb to create the current MODB and was failing because of it was trying to refresh (this refresh thing needs alot of work, this is why I'm ignoring saving smtp log here)
* bulk read from port  requests database when when a number is not in any number (as part of rewriting `knm_ops:create/2`)
* **Kazoo FixtureDB** now would throw `{error, fixturedb_view_is_missing}` when you're trying to get a view result and you did not provide any fixtures for it in the db directory. An instruction of how to proceed to fix this problem is now shown.
* **Kazoo FixtureDB** now when getting `all_docs` will search `docs` directory of the database to fetch requested numbers. It would throw `{error, fixturedb_is_missing}` when database's directory is not exists on the file system.
* As part of the the new fixture db, all number docs from number dbs are copied to their assigned account database.
* Fix KNM tests and provide all missing document or view result which was causing the tests to fail.


lazedo:
@icehess rebase & conflict solving needed here

PR #6249, opened 2020-01-15
PISTON-1002: use collect_digits for DTMF during vm greeting/instructions

danielfinke:
When `play_greeting_intro`, `play_greeting`, or `play_instructions` play media, a noop is enqueued. `kapps_call_command:wait_for_application_or_dtmf` may continue on the noop from one of these play commands, rather than as expected due to `kapps_call_command:noop`. This means, for example, the user is unable to press a key to connect to the operator during instructions, if the greeting has already been played.

I introduced the 8-arity `collect_digits` for this purpose, too, as it allows a long initial timeout (`kapps_call_command:default_application_timeout`) but continues immediately after the expecting noop (the second argument set to 0).


lazedo:
@danielfinke rebase here

PR #6251, opened 2020-01-15
PISTON-1003: cf_util:get_operator_callflow now uses cf_flow:lookup

danielfinke:
This makes it possible for a vmbox's `operator_number` to match callflow pattern regexes


lazedo:
@danielfinke sorry for the long wait. just rebase so we can merge

lazedo:
@danielfinke thanks Daniel.

PR #6253, opened 2020-01-16
allow sup to access inner keys for kapps_config

lazedo:
* in sup commands for kapps_config, the key can be specified like A.B.C.D
  to access just D

* adds is_sup_call to process dictionary

### DIFF
--- a/core/kazoo_apps/src/kapps_config.erl
+++ b/core/kazoo_apps/src/kapps_config.erl
@@ -444,7 +444,7 @@ get_node_value(Category, Key, Default) ->
 
 -spec get_node_value(config_category(), config_key(), Default, config_node()) -> kz_json:json_term() | Default.
 get_node_value(Category, Key, Default, Node) when not is_list(Key) ->
-    get_node_value(Category, [kz_term:to_binary(Key)], Default, Node);
+    get_node_value(Category, key(Key), Default, Node);
 get_node_value(Category, Keys, Default, Node) when not is_binary(Category) ->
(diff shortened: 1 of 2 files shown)


jamesaimonetti:
@lazedo it seems like the more consistent approach would be to update `kazoo_apps_maintenance` with `config_get(Category, Key)` or something, do the split then call `kapps_config:get(Category, SplitKeys,...)`. 

Introducing SUP-specific code into kapps_config seems like a code smell.

Thoughts?

lazedo:
@jamesaimonetti something like `sup kapps_config set_default pusher User-Agents.MyApp.regex ^MyApp` is what i'm looking for, i don't see a reason for doubling the logic in kazoo_apps_maintenance.

jamesaimonetti:
But as-is, the kapps_config code is now muddled with SUP-specific bits.

Instead, I'm proposing:

```erlang
-module(kazoo_apps_maintenance).

config_set_default(Category, Key, Value) ->
    kapps_config:set_default(Category, key(Key), Value).
```

Repeat for which instantiations of kapps_config you want to support via SUP.

lazedo:
@jamesaimonetti the problem are the variations of kapps_config (set_binary, set_json, set..., get...). `sup` is already running in the vm and is part of the `core`,  the `rpc` is done to `sup` itself which then  calls `apply`, besides that, `kapps_config` could be moved out of kazoo_apps (was there because of circular references). i understand your concerns but i don't see a problem here.

PR #6254, opened 2020-01-16
PISTON-1005: make kz_json:merge_recursive an option of kz_json:merge

danielfinke:
Deprecate `kz_json:merge_recursive` in favor of `kz_json:merge/2` with the `Options` map containing 'recursive' => 'true'.

Previously discussed here: https://github.com/2600hz/kazoo/pull/6242#issuecomment-572760221

### DIFF
--- a/core/kazoo_stdlib/doc/kz_json.md
+++ b/core/kazoo_stdlib/doc/kz_json.md
@@ -74,33 +74,13 @@ You can merge two (or more) objects together and specify which objects' value wi
 -   `merge([JObj1, JObj2,...])`: Defaults to using the `merge_right` strategy where the object on the "right" will have its value used when both objects have a value at a give key path.
 -   `merge(Strategy, JObjs)`: Choose a strategy, either `fun kz_json:merge_right/2` or `fun kz_json:merge_left/2` (or define your own), and apply it to the list of objects.
 
+Both `merge_left/2` and `merge_right/2` will keep the left or right empty objects (respectively) when they exist at a given key path if the other value is also an object (empty or not). To override this behavior and always recursively merge the values when both are objects, specify `'recursive' => 'true'` in the `Options` map passed in `merge(JObjs, Options)`, `merge(JObj1, JObj2, Options)`, `merge(Strategy, JObjs, Options)`, or `merge(JObj1, JObj2, Options)`, as arg 4 to `merge/4`, or as arg 5 to `merge/5`.
+
 ### Handling `null`
 
 Sometimes it is beneficial to include the `null` atom as a value (for downstream processing perhaps) versus triggering a `delete_key` equivalent.
 
 `set_value` and `merge` can take an optional map `#{'keep_null' => 'true'}` to ensure `null` is kept in the resulting data structure.
-
-### `merge/3` vs `merge_recursive/2`
-
-There is a second, older implementation of merging objects called `merge_recursive/2`. The main difference is that it operates differently when merging a path in the second object that has an empty object as the value:
-
-```erlang
-> J1 = {[{<<"foo">>,false}]},
-> J2 = {[{<<"foo">>,false}, {<<"bar">>,{[{<<"foo">>,{[]}}]}}]}.
-> kz_json:merge(fun kz_json:merge_right/2, J1, J2).
-{[{<<"foo">>,false},{<<"bar">>,{[{<<"foo">>,{[]}}]}}]}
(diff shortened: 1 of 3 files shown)


jamesaimonetti:
@danielfinke lgtm, i'm adding more xref checking and updating some usages from merge_recursive to merge with options (see https://github.com/2600hz/kazoo/commits/voxter-pr-PISTON-1005-make-kz-json-merge-recursive-an-option-of-kz-json-merge) for CI goodness. I'll open a PR with that branch in a bit to run make-busy then merge our commits separately. Stay tuned!

jamesaimonetti:
@danielfinke so one big issue is with sub-object key ordering when using `merge/2` with `#{recursive => true}`. Because of the `lists:sort/1` calls in `merge/4`, the ordering of keys will potentially differ in the resulting merged object.

```erlang
    JObj = {[{<<"c">>,{[{<<"c.a">>,1.1},{<<"c.b">>,2.2}]}}]},
    M = kz_json:merge(JObj, JObj).

    %% M = {[{<<"c">>,{[{<<"c.b">>,2.2},{<<"c.a">>,1.1}]}}]}
```

This can be seen in practice by changing `kapps_config_doc` to use `kz_json:merge/2` and running the number manager tests - the converter regexes are in a different order so the US E164 is not checked first.

Not sure yet on a path forward there...thoughts?

danielfinke:
@jamesaimonetti since, according to the JSON spec, objects key are unordered, I don't think that we should be depending on a defined key order for anything. In cases such as the converter regexes, maybe we should migrate the schema for those to one where arrays are used (which _are_ ordered under JSON spec). So for example, the converter regexes could go from:
```
{
  "e164_converters": {
    "^\\+?1?([2-9][0-9]{2}[2-9][0-9]{6})$": {
      "prefix": "+1"
    },
    "^011(\\d*)$|^00(\\d*)$": {
      "prefix": "+"
    },
    "^[2-9]\\d{7,}$": {
      "prefix": "+"
    }
  }
}
```
to
```
{
  "e164_converters": [
    {
      "prefix": "+1",
      "regex": "^\\+?1?([2-9][0-9]{2}[2-9][0-9]{6})$"
    },
    {
      "prefix": "+",
      "regex": "^011(\\d*)$|^00(\\d*)$"
    },
    {
      "prefix": "+",
      "regex": "^[2-9]\\d{7,}$"
    }
  ]
}
```
and a manual append of the arrays from the account, reseller, and system could be performed.

In isolation, this doesn't seem too bad to me. But I'm also not familiar with how many more things across the codebase depend on the behaviour of an object merge with key order preservation. It seems like most configuration merges don't care about that off the top of my head.

jamesaimonetti:
Yeah, that's the bugaboo, where else is it implicitly depending on order? Classifiers come to mind; would definitely need to check there.

k-anderson:
@danielfinke You have discovered a dirty little secret from the start of the project ;)  At that time I was unaware that JSON object keys were unordered and as such in some places like classifiers I erroneously relied on it.  Another example is the order of time-of-day routes which is by default the order the objects appear in the children object, but for that we created a work-around where you can explicitly define an array of ids in the data object to control the order regardless of the children object ordering.  Point is I have LONG standing desire to sort that out as well as convert some things that are object of objects to array of objects (for example, all crossbar error replies should ideally be an array of error objects).  Problem is two fold: 1) Its high effort and low company value so its hard to justify the cost require right now 2) its dangerous from a migration / backward compatibility standpoint.  For example, finding all the locations in the code base, creating migrations when possible, announcements when not, and fully testing that the full set of changes has no impact on compatibility is many person hours to correct something that nobody has noticed for many years (so far) or has active tickets/complaints open.  Also, image moving the converter regexs to an array of objects.  If admins don't know to run a migration or it fails for some reason thats an outage of the entire cluster!  When you get on the phone with the client they are not going to appreciate it was because 10 years ago we relied on JSON object key to be ordered....

Good news tho! I do intended to correct this now baked in oversight via a few angles.  I plan on making a new API version endpoint (and removing v1 as supported) where we can do some non-backward compatible payload changes.  With the kzd_* modules expanding they are almost to a point that we would have a reliable place to handle a branch based on a given data structure (IE: kzd_converters could handle either an object of objects or array of objects) to facilitate a minimal impact move toward proper handling of this.  Also, major version bumps are a good time to make changes like that... *cough*

All that said, might be faster / easier to submit a change to the JSON spec.... ;)

PR #6255, opened 2020-01-17
Make auto compaction job execution louder

harenson:
- Add *YYYYMM-* prefix to the auto compaction job call id so it can be retrieved later from MoDBs.
- Make it louder by adding more log lines.

### DIFF
--- a/applications/tasks/src/modules/kt_compactor.erl
+++ b/applications/tasks/src/modules/kt_compactor.erl
@@ -316,8 +316,10 @@ do_compact_node(Node, Heuristic, APIConn, AdminConn) ->
 do_compact_node(Node, Heuristic, APIConn, AdminConn, Databases) ->
     CallId = kz_log:get_callid(),
     lists:foldl(fun(Database, Acc) ->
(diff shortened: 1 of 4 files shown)


applications/tasks/src/modules/kt_compactor.erl
```diff
@@ -474,12 +483,14 @@ get_dbs_sizes(Dbs) ->
...
                                ,{[db_and_sizes()], non_neg_integer()}
-                               , pos_integer()
+                               ,pos_integer()
                                ) -> {[db_and_sizes()], pos_integer()}.
 get_db_disk_and_data_fold(Conn, UnencDb, {_, Counter} = State, ChunkSize)
   when Counter rem ChunkSize =:= 0 ->
-    %% Every `ChunkSize' handled requests, sleep 100ms (give the db a rest).
-    timer:sleep(100),
+    %% Every `ChunkSize' handled requests, sleep `Sleep'ms (give the db a rest).
+    Sleep = 200,
```
jamesaimonetti: maybe a system_config to control this?

PR #6258, opened 2020-01-17
Ensure trunkstore uses account doc's realm

jamesaimonetti:
Connectivity docs should not include an auth_realm as it leads users
to confusion thinking they can set an realm separate from the
account's - this is not the case!

Ensure connectivity docs have the account's realm (just in case) and
ensure Trunkstore sets the account's realm when processing calls.


jamesaimonetti:
See 4.3 in #6257

PR #6259, opened 2020-01-22
PISTON-1000: refactor kapps_config_util to use kapps_account_config

danielfinke:
- also use kapps_config_cache more - no restriction on merging as the
  results are the same

Relates to discussion here: https://github.com/2600hz/kazoo/pull/6242

Depends on https://github.com/2600hz/kazoo/pull/6254 for the `'recursive'` option in `kz_json:merge`.

Wanted to refactor some similar code between `kapps_config_util` and `kapps_account_config` so that `kapps_config_util` wraps called to the other module and tacks on a few extra things only where necessary. This meant that the introduction of `get_category_with_strategy` was needed, which unlike `get_with_strategy`, gets the entire category's JSON object.
This required reworking `walk_the_walk` a bit and I observed the opportunity to be more aggressive about caching of the results. It became clear that it was OK to _always_ merge and cache the results, provided the appropriate cache origins were specified.
Another observation was that docs managed by `cb_configs` were not being considered `"pvt_type": "account_config"` so I've added `set_pvt_type` to that mod. This is important to ensure the appropriate cache clearing happens with the more aggressive caching.

Looking for feedback!

### DIFF
--- a/applications/crossbar/src/crossbar.hrl
+++ b/applications/crossbar/src/crossbar.hrl
@@ -210,6 +210,7 @@
 -define(TYPE_CHECK_OPTION_ANY, ?TYPE_CHECK_OPTION(<<"any">>)).
 
 -define(SPECIAL_EXPECTED_TYPE, [{<<"allotments">>, <<"limits">>}
+                               ,{<<"configs">>, <<"account_config">>}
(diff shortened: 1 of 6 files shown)


lazedo:
@danielfinke not against the approach but needs careful testing. lets resume after the break. sorry about that.

PR #6265, opened 2020-01-23
Beautify and multiline js codes in views

icehess:
Adding (and ran) a script to make JavaScript codes in CouchDB view files beautify and multi-line so make it more readable to developers. Also makes keep tracking changes easy during PR reviews.

`kz_datamgr:maybe_adapt_multiline/1` is updated to reverse multi-lined views back to simple string for all view's functions (such as `lists`, `shows` and of course `views`).


jamesaimonetti:
A couple small things:

1. Add to makefile as a target to run on $(CHANGED_JSON) or $(find all json)
2. CI step (or add to `make apis` as well)

PR #6266, opened 2020-01-23
PISTON-868:Storage plans will now take affect instantly after a crossbar create action

bradfordben:
Server side intervention was required after creating a storage plan through crossbar. This PR removes the need for server side intervention and will mean that once a storage plan is created through crossbar it will take affect immediately.

Updates and deletes to storage plans are left as before and will also take affect immediately.

### DIFF
--- a/applications/crossbar/src/modules/cb_storage.erl
+++ b/applications/crossbar/src/modules/cb_storage.erl
@@ -209,11 +209,15 @@ validate_storage_plan(Context, PlanId, ?HTTP_DELETE) ->
 %%--------------------------------------------------------------------
 -spec put(cb_context:context()) -> cb_context:context().
 put(Context) ->
-    crossbar_doc:save(Context).
+    Return = crossbar_doc:save(Context),
+    kzs_plan:reload(cb_context:account_id(Context)),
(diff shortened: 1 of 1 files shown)


lazedo:
@bradfordben not a fan of this. the plan will be refreshed via cache. can you provide logs ?

bradfordben:
@lazedo The plan will be refreshed by caches only after the plan has been cached. When the plan is first created there is no cash callbacks registered so the cache module is not watching for the creation of the doc.

lazedo:
@bradfordben check here https://github.com/2600hz/kazoo/blob/master/core/kazoo_data/src/kzs_plan.erl#L569

bradfordben:
Ah interesting, Ok il have a look to see why this is not working. I missed that. Thanks

bradfordben:
@lazedo I found the issue. The doc_created event is only being generated the very first time the storage doc is created, then any future deletes and recreation will not generate the doc_created but instead generate a doc_edited event. I have submitted a PR to fix this https://github.com/2600hz/kazoo/pull/6278   Please take a look and see if there is any edge case i might have missed.

PR #6267, opened 2020-01-23
Make bash script more portable

noahmehl:
There's a `declare -A` in the script, which requires bash version > 4.  The default bash version shipped with OS X is lower than that :(


jamesaimonetti:
Merged

PR #6269, opened 2020-01-24
HELP-12897: properly paginate directory's users

jamesaimonetti:
Prior, the view only indexed by the directory IDs on a user's
doc. When `paginate=false` was included on a fetch of the directory,
the directory/users_listing view did not interact nicely with the
internal pagination protections added.

The view arguments used `[{key, DirectoryId}, {limit, 50}]`; the first page would
be returned and since the next `startkey` would be `DirectoryId`
again, crossbar_doc gets stuck in a loop as there was no way to
"increment" the startkey.

Instead, the view now emits `[DirectoryId, UserId]` to allow
differentiation between rows, and view options are now passed in as
`[{startkey, [DirectoryId]}]`. This plays much nicer with pagination
regardless of whether pagination has been disabled or not.


jamesaimonetti:
See 4.3 in #6268

PR #6273, opened 2020-01-26
blackhole: Fixed crush when not defined auth_token

sergey-safarov:


core/kazoo_auth/src/kz_auth_jwt.erl
```diff
@@ -294,7 +294,9 @@ parse(JWTToken) when is_binary(JWTToken) ->
 token(JWTToken) ->
     token(JWTToken, []).
 
--spec token(kz_term:ne_binary() | map(), kz_term:proplist()) -> map().
+-spec token(kz_term:api_ne_binary() | map(), kz_term:proplist()) -> map().
+token('undefined', _Options) ->
+    {'error', 'invalid_jwt'};
```
lazedo: drop this change

sergey-safarov: This change reverted.

PR #6276, opened 2020-01-28
Add fixes and improvements to compaction jobs and to tasks app

harenson:
- Move auto compaction code from kz_tasks_trigger module to kt_compactor module.
- Make kz_tasks_trigger gen_server work as a local gen_server (not using kz_globals).
- Fix compaction job's (csv tasks/auto compaction) call_id by prefixing it with `YYYYMM-` so it can be retrieved in later months.
- Remove kz_globals usage from tasks app.


applications/crossbar/priv/couchdb/schemas/system_config.tasks.json
```diff
@@ -21,6 +21,16 @@
             "minimum": 1,
             "type": "integer"
         },
+        "compaction_list_dbs_chunk_size": {
+            "default": 20,
+            "description": "tasks compaction_list_dbs_chunk_size",
```
jamesaimonetti: Can we please start adding real descriptions, not use the placeholder auto-generated ones?

applications/tasks/src/kz_notify_resend.erl
```diff
@@ -128,9 +121,15 @@ stop() ->
 
 -spec running() -> kz_json:objects().
 running() ->
-    case kz_globals:where_is(?NAME) of
-        'undefined' -> [];
-        _ -> gen_server:call(?SERVER, 'running')
+    running(whereis(?MODULE)).
+
+-spec running(pid() | 'undefined') -> kz_json:objects().
```
jamesaimonetti: `kz_term:api_pid()`

applications/tasks/src/kz_tasks_trigger.erl
```diff
@@ -148,7 +141,7 @@ handle_info({'timeout', Ref, _Msg}, #state{day_ref = Ref}=State) ->
     {'noreply', State#state{day_ref = day_timer()}};
 
 handle_info({'timeout', Ref, _Msg}, #state{browse_dbs_ref = Ref}=State) ->
-    _Pid = kz_process:spawn(fun browse_dbs_for_triggers/1, [Ref]),
+    _Pid = kz_process:spawn(fun kt_compactor:browse_dbs_for_triggers/1, [Ref]),
```
jamesaimonetti: kt_compactor is a task executor, not a "core" task module and should not be called from within here, no?

PR #6278, opened 2020-01-29
PISTON-1091:Fixed kzs_publish bug where if a document with the same id is recreated kzs_publish should send a doc_created event and not doc_edited event

bradfordben:
There is a bug in kzs_publish that assumes when a doc is created it will all ways have a rev of 1. 

This is not the case if a doc is created and then deleted and then recreated with the same id. The rev will be set to 3-xxx and in this case and publish a doc_edited event as apposed to a doc_created event that should be published. This happens in the case of kazoo_storage, The doc is all ways created with the id set to the account id.

I have changed the logic to now match:
If the doc is marked for deletion then its a delete event
If the doc before being saved has no rev set then its a created event
If the doc before being saved has a rev set then its a edited event

I think this logic will catch every case unless there is any you any think of? 

I have removed the call to kzs_publish:publish_fields/1 in kzs_doc for 2 reasons, It seems uanessacry and i need to keep the rev (if its set) of the original doc before the db operation to calculate the event type in kzs_publish.

There is also some changes to variable names and added some comments to help others better understand in the future

### DIFF
--- a/core/kazoo_data/src/kzs_doc.erl
+++ b/core/kazoo_data/src/kzs_doc.erl
@@ -239,10 +239,13 @@ prepare_doc_for_del(Server, DbName, Doc) ->
       ]).
 
 %%------------------------------------------------------------------------------
-%% @doc Prepare doc for save and generate a doc smaller doc for a publish event
-%% Returns {PreparedDoc, PublishDoc} where
-%% PreparedDoc is a JObj containing all the necessary key values / correct format for the save or delete operation
-%% PublishDoc is the JObj supplied with a possible changed id value
+%% @doc Prepare / convert a doc for save or delete.
+%% If the doc has the key `id' set,it will be deleted and a new `id' will be
+%% generated.
+%% @returns {PreparedDoc, PublishDoc} where:
+%% PreparedDoc is a JObj containing all the necessary key values and in the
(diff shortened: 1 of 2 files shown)


lazedo:
@bradfordben the concern i have is the duplication of the document in memory because of `I have removed the call to kzs_publish:publish_fields/1 in kzs_doc`, multiply a `512Kb` document by `2` by `number of requests`, you're server memory will get more pressure. i understand the problem but i think we need a better solution.

bradfordben:
@lazedo I had though of this but didnt feel it was a big enough concern, but an alternative would be just call `kzs_publish:publish_field/1` as was done before and append the `rev` value to the result. That way the resulting memory impact is only the addition of one key value per request. Would you prefer that?  I was just trying to save one list fold but maybe memory is more important to consider here.

lazedo:
@bradfordben sorry for late reply. yes, that would be better (key vs doc)

bradfordben:
No problem, i have made that change 👍

bradfordben:
@lazedo Can you re-review this when you get a chance.

lazedo:
@bradfordben the code looks ok but `ci` is failing. maybe a rebase is needed

PR #6289, opened 2020-02-04
[4.3] KCRO-7: new filter to find missing keys OR keys with empty values

k-anderson:
If you want to find a document by missing a value for a parameter the only way currently is to use the filter for a missing key. However, this filter does not emit the document if the key is present but the value is empty/null. There is a has_value filter but no inverse of that. This adds a new query filter missing_value that will return a document if the key is missing OR the value is empty.


k-anderson:
Master PR #6288

PR #6295, opened 2020-02-07
pass relay setting as binary

icehess:
`gen_smtp_client` has been changed to accept more `relay` as a list. But it has a bug to use list comprehensive to convert a list to string, which also cause a genuine string to be converted to string char by char.

This is trying to pass the `relay` option as a binary so it would be properly converted to string by `gen_smtp_client`.

commit that changed the the `deps.mk`: https://github.com/2600hz/kazoo/commit/4aab964987de4bd1e3233aba94c2e119cc5eefeb
commit that changed the gen_smtp: https://github.com/2600hz/erlang-gen_smtp/commit/033f2d133e733304d591fef4b05ea58e8c220e92


lazedo:
@icehess no need to change here. if there's something wrong with gen_smtp, we should fix it there.

lazedo:
@icehess update this pr to reference 3d294c57153787027bea4011eeba268fee5d7165 from gen_smt

PR #6297, opened 2020-02-07
[4.3] Listener federators and memory improvements

jamesaimonetti:
Prior, when a gen_listener needed to federate its queue and bindings
to other zones, it would start_link the listener_federator (and trap
exits). However, should a listener_federator process actually exit,
the parent gen_listener would receive the EXIT message but failed to
have a handle_info clause to restart the process.

Introduce a sofo supervisor for starting listener_federator processes
and restart them if they die unexpectedly (non-normal exits).

Secondly, introduce memory consumption checks and GC gen_listeners
when they exceed 100K in total_heap_size. In addition, copy the binary
payloads as they arrive off the AMQP channel has shown to decelerate
the memory growth as the original binary can be garbage collected
sooner and not count against the gen_listener's "old" heap.

All told, these manual GC runs appear to be minimal and only affect
long-running and busy gen_listeners; even fewer runs needed to release
memory after the binary:copy/1 introduction.

### DIFF
--- a/core/kazoo_amqp/src/gen_listener.erl
+++ b/core/kazoo_amqp/src/gen_listener.erl
@@ -853,6 +853,11 @@ code_change(_OldVersion, State, _Extra) ->
 %%%=============================================================================
 %%% Internal functions
 %%%=============================================================================
+
+%%------------------------------------------------------------------------------
+%% @doc
+%% @end
+%%------------------------------------------------------------------------------
 handle_down({Pid, Ref}, Reason, #state{federators=Fs}=State) ->
     case lists:keytake({Pid, Ref}, 2, Fs) of
         'false' -> handle_callback_info({'DOWN', Ref, 'process', Pid, Reason}, State);
--- a/core/kazoo_amqp/src/gen_listener.erl
+++ b/core/kazoo_amqp/src/gen_listener.erl
@@ -104,7 +104,7 @@
         ]).
 
 -export([federated_event/4
-        ,the_federator_lives/2
+        ,notify_of_federator_listener/2
         ]).
 -export([delayed_cast/3]).
 -export([distribute_event/3]).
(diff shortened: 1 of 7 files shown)


core/kazoo_services/src/kz_services.erl
```diff
@@ -947,18 +947,15 @@ handle_fetch_options(Services, ['skip_cache'|Options]) ->
 %% @doc
 %% @end
 %%------------------------------------------------------------------------------
--spec commit_updates(kz_term:ne_binary()
-                    ,kz_services_quantities:billables() | kz_services_quantities:billable()
-                    ,kz_services_quantities:billables() | kz_services_quantities:billable()
-                    ) -> services().
+-type billable_updates() :: 'undefined' | kz_services_quantities:billables() | kz_services_quantities:billable().
+
+-spec commit_updates(kz_term:ne_binary(), billable_updates(), billable_updates()) ->
```
lazedo: i think the problem is in kz_services_quantities:billable() type spec
its defined as kz_json:api_object() and it should be kz_term:api_object()

jamesaimonetti: Hmm, possible, I just brought the master version of the specs along to silence dialzyer.

core/kazoo_amqp/src/gen_listener.erl
```diff
@@ -390,6 +395,10 @@ rm_binding(Srv, Binding, Props) ->
 federated_event(Srv, JObj, BasicDeliver, BasicData) ->
     gen_server:cast(Srv, {'federated_event', JObj, BasicDeliver, BasicData}).
 
+-spec the_federator_lives(pid(), {kz_term:ne_binary(), pid()}) -> 'ok'.
+the_federator_lives(Srv, {_Broker, _Pid}=Child) ->
```
mark2600: I would go with something more boring and to the point such as "federator_started" or "notify_federator_started"

jamesaimonetti: Where's your sense of whimsy???

PR #6306, opened 2020-02-11
[4.3] HELP-11566: clarify precedence of call forward vs failover

jamesaimonetti:
The `call_forward` object handles configuration for both call
forwarding and for failover. In its current implementation, only one
of those features can be "active".

Prior to this change, if `call_forward.failover` was `true` it would
be included on the endpoint object and ecallmgr would treat the
`call_forward` settings as a failover endpoint, regardless of the
`call_forward.enabled` flag's setting.

This change clarifies that if `call_forward.enabled` is `true`, the
settings should only be considered for call forwarding. If the setting
is `false` *and* `call_forward.failover` is `true` then consider the
settings as a failover endpoint.


jamesaimonetti:
Master at #6307

pklinus:
https://2600hz-commercial.atlassian.net/browse/MSPB-64

PR #6309, opened 2020-02-12
PISTON-1028: use startkey and endkey for get_directory_listing

danielfinke:
Fix bug introduced in a3795332061f03499498605abb629029c8564fcc where "no users have been assigned to directory" would always be hit


jamesaimonetti:
Merged, thanks!

PR #6310, opened 2020-02-12
[4.3] PISTON-1028: use startkey and endkey for get_directory_listing

danielfinke:
Fix bug introduced in 3a8ff80 where "no users have been assigned to directory" would always be hit


jamesaimonetti:
Thanks!

PR #6315, opened 2020-02-13
[4.3] KCRO-10: expose the message counts per folder when listing vmboxes

k-anderson:
When listing the voicemail boxes this adds a new object to each called "folders" if the vmbox has messages.  The object contains two properties that represent the quantity of messages in the 'new' and 'saved' folders.


k-anderson:
Master PR #6314

PR #6317, opened 2020-02-14
kzs_doc: Implemented 'doc_type' option check

sergey-safarov:
now call like 'kz_datamgr:open_cache_doc(AccountDb, {<<"user">>, ConfigId})' works properly

### DIFF
--- a/core/kazoo_data/src/kzs_doc.erl
+++ b/core/kazoo_data/src/kzs_doc.erl
@@ -62,11 +62,11 @@ maybe_soft_deleted(Doc, 'true', Options) ->
           {'error', 'not_found'}.
 maybe_type_mismatch(Doc, 'undefined') ->
     {'ok', Doc};
-maybe_type_mismatch(Doc, ReqestedType) ->
+maybe_type_mismatch(Doc, RequestedType) ->
     case kz_doc:type(Doc) of
-        ReqestedType -> {'ok', Doc};
+        RequestedType -> {'ok', Doc};
         DocType ->
-            lager:info("requested doc_type '~s', denying access to doc with type: '~s'", [ReqestedType, DocType]),
+            lager:info("requested doc_type '~s', denying access to doc with type: '~s'", [RequestedType, DocType]),
             {'error', 'not_found'}
     end.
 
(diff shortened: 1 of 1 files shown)


core/kazoo_data/src/kzs_doc.erl
```diff
@@ -42,16 +42,33 @@ open_doc(#{server := {App, Conn}}, DbName, DocId, Options) ->
...
+            lager:info("denying access to soft deleted doc: ~s", [kz_doc:id(Doc)]),
+            {'error', 'not_found'}
+    end.
+
+-spec maybe_type_mismatch(kz_json:object(), kz_term:api_ne_binary()) ->
+          {'ok', kz_json:object()} |
+          {'error', 'not_found'}.
+maybe_type_mismatch(Doc, 'undefined') ->
+    {'ok', Doc};
+maybe_type_mismatch(Doc, ReqestedType) ->
```
jamesaimonetti: s/Reqested/Requested

sergey-safarov: fixed

lazedo:
@sergey-safarov trying to understand this pr. you say this pr fixes `now call like 'kz_datamgr:open_cache_doc(AccountDb, {<<"user">>, ConfigId})' works properly` but i don't see a relation here. also, this check is already in crossbar. please provide more context on this, maybe with an example that's not working as you expected to.

sergey-safarov:
I want be sure about document type fetched from database.
For this purpose exist function [`kz_datamgr:open_cache_doc/2`](https://github.com/2600hz/kazoo/blob/master/core/kazoo_data/src/kz_datamgr.erl#L533-L534).

Functions defines `doc_type` options but implementation for this check not exist.
PR add  `doc_type` options check then document fetched from database.

lazedo:
@sergey-safarov implementation is not intended here. the `doc_type` is to help resolving dataplan access. `crossbar` does what you to put here.

sergey-safarov:
I not want create own doc_type with own `kzd` module.
Inside this module i want be sure doc_type 'opened using `kz_datamgr:open_cache_doc(AccountDb, {<<"my_doc_type">>, ConfigId})`
So crosbar here not applicable.
> doc_type is to help resolving dataplan access

Could give me reference for such usage.

lazedo:
> I not want create own doc_type with own `kzd` module.

not saying that you do. what i'm saying is that this `PR` is not acceptable, you're introducing unwanted `atm` changes in the core data access library.

if you're not using `crossbar` then you can do that in your library `foo:open_cache_doc/3` that will call `kz_datamgr:open_cache_doc /3` or other, and can do any pre-process or post-process

lazedo:
@sergey-safarov sorry for closing this but we need to move forward and merge/close pending prs as we approach 5.x

PR #6318, opened 2020-02-14
PISTON-1032: Allow SUP to handle any argument type as long as the argument is surrounded by double quotes.

bradfordben:
Allow a user to pass any data type to a function call through the sup command

EG.  
sup kz_time to_gregorian_seconds "{{2020, 01, 01}, {01, 20, 30}}" "<<\"Etc/UTC\">>"

### DIFF
--- a/core/sup/src/sup.erl
+++ b/core/sup/src/sup.erl
@@ -70,7 +70,7 @@ main(CommandLineArgs, Loops) ->
                                            halt(1)
                                    end
                            end,
-            ParsedArgs = lists:map(ParseFunArgs ,Args),
+            ParsedArgs = lists:map(ParseFunArgs, Args),
             Timeout = case props:get_value('timeout', Options) of 0 -> 'infinity'; T -> T * 1000 end,
             IsVerbose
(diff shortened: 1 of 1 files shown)


lazedo:
@bradfordben before pushing changes, make sure this works locally. `sup callmgr ecallmgr_maintenance allow_sbc kamailio-1.xxxxx 10.10.10.150 true  `

bradfordben:
@lazedo Thanks for highlighting this. I had missed the test case of `.` 

I have had to add a -e option to instruct sup to treat the args supplied as erlang terms as there is no way to handle this without changing the current way sup is used. So all existing commands will not change and if the user specifies -e then they must supply the args properly escaped and formatted.

I could not find a doc for sup. Is there one that im missing to document this? 

Erlang term arguments must be correctly formatted / escaped as below:
  * Binary: "<<\\\"Etc/UTC\\\">>"
  * Tuples: "{{2020, 01, 01}, {01, 20, 30}}"
  * Lists: "[1,2,3,4]"
  * Strings: "this is a string"
  * Atom: atom

sup kz_time to_gregorian_seconds -e "{{2020, 01, 01}, {01, 20, 30}}" "<<\\\"Etc/UTC\\\">>"

PR #6324, opened 2020-02-18
Harden blackhole against high connection rates

jamesaimonetti:
Prior, blackhole had no effective tracking of connection count
per-IP or globally. Errant clients were terminating their connection
but not closing the TCP socket, reconnecting, and causing accidental
DoS of the blackhole server.

This PR creates a websocket client in the `kazoo_proper` app to test
pings and "high" connection attempts. The PR introduces a
`system_config` flag to allow system administrators the ability to
limit concurrent connections per IP address to help limit impact of an
IP creating large amounts of connections quickly.

Additionally, update blackhole to return an error when no auth token
is present.

The kz_cache is updated to allow higher-level apps to associate a
cache key with a PID, flushing the entry on the PID's exit. A helpful
feature is to also allow some exposure to match specs for the purposes
of counting cache entries. Instead of having to fetch a list and
filter in the higher-level code (incurring copying / memory usage),
the `count` API allows the filtering criteria to be specified using
match spec grammar and done in ETS via `select_count`.


k-anderson:
@jamesaimonetti why not use token buckets here?

applications/blackhole/src/blackhole_socket_callback.erl
```diff
@@ -14,16 +14,37 @@
...
-    Context = bh_context:set_source(bh_context:new(Pid, Id), IPBin),
+    Context = bh_context:set_source(bh_context:new(Pid, SessionId), IPBin),
+
+    kz_cache:store_local(?CACHE_NAME
+                        ,?SESSION_KEY(SessionId)
+                        ,#cache_info{'start_time_s' = kz_time:now_s()
+                                    ,'socket_pid' = self()
+                                    ,'ip' = IPBin
+                                    }
+                        ,[{'expires', ?SECONDS_IN_HOUR}]
```
k-anderson: @jamesaimonetti if you expire these entries based on time, then the cache will not reflect the actual counts...

jamesaimonetti:
@k-anderson reworked connection tracking to use blackhole_tracking including PID monitoring. No expirations to sync.

lazedo:
@jamesaimonetti merging this but we need to handle proxied websockets

PR #6326, opened 2020-02-20
Syntax Corrections

JonTheWong:
Corrections for markdown, changed string for MD5, changed requirements
Added SHA example

### DIFF
--- a/applications/crossbar/doc/user_authentication.md
+++ b/applications/crossbar/doc/user_authentication.md
@@ -12,9 +12,9 @@ Provides an auth-token via user credentials
 
 Key | Description | Type | Default | Required | Support Level
 --- | ----------- | ---- | ------- | -------- | -------------
-`account_name` | The account name of the user | `string(1..128)` |   | `true` |  
+`account_name` | The account name of the user | `string(1..128)` |   | `false` |  
 `account_realm` | The account realm of the user | `string(4..253)` |   | `false` |  
-`credentials` | A hash of the user credentials | `string(1..64)` |   | `true` |  
+`credentials` | A hash of the uses credentials | `string(1..64)` |   | `true` |  
 `method` | The hash method | `string('md5' | 'sha')` | `md5` | `false` |  
 `phone_number` | A phone number assigned to the users account | `string(1..64)` |   | `false` |  
 
@@ -123,7 +123,7 @@ Send a reset-your-password email via user credentials
 
 Key | Description | Type | Default | Required | Support Level
 --- | ----------- | ---- | ------- | -------- | -------------
-`account_name` | The account name of the user | `string(ABC...Z)` |   | `false` |
+`account_name` | The account name of the user | `string(1..64)` |   | `false` |
 `account_realm` | The account realm of the user | `string(1..64)` |   | `false` |
 `phone_number` | A phone number assigned to the user's account | `string(1..64)` |   | `false` |
 `ui_url` | The UI's root URL | `string(14..2000)` |   | `true` |
--- a/applications/crossbar/doc/user_authentication.md
+++ b/applications/crossbar/doc/user_authentication.md
@@ -12,10 +12,10 @@ Provides an auth-token via user credentials
 
 Key | Description | Type | Default | Required | Support Level
 --- | ----------- | ---- | ------- | -------- | -------------
-`account_name` | The account name of the user | `string(ABC..Z)` |   | `true` |  
+`account_name` | The account name of the user | `string(1..128)` |   | `true` |  
 `account_realm` | The account realm of the user | `string(4..253)` |   | `false` |  
 `credentials` | A hash of the user credentials | `string(1..64)` |   | `true` |  
-`method` | The hash method | `string(md5..sha)` | `md5` | `true` |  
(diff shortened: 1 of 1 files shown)


applications/crossbar/doc/user_authentication.md
```diff
@@ -12,10 +12,10 @@ Provides an auth-token via user credentials
 
 Key | Description | Type | Default | Required | Support Level
 --- | ----------- | ---- | ------- | -------- | -------------
-`account_name` | The account name of the user | `string(1..128)` |   | `false` |  
+`account_name` | The account name of the user | `string(ABC..Z)` |   | `true` |
```
jamesaimonetti: `string(1..128)` just notes that the account name can be up to 128 characters. It can contain alphanumeric values too so `A-Z` wouldn't be accurate anyway.

Additionally the tables are auto-generated from the JSON schemas so editing them here in markdown is incorrect - find the corresponding JSON schema and edit the "description".

applications/crossbar/doc/user_authentication.md
```diff
@@ -12,10 +12,10 @@ Provides an auth-token via user credentials
 
 Key | Description | Type | Default | Required | Support Level
 --- | ----------- | ---- | ------- | -------- | -------------
-`account_name` | The account name of the user | `string(1..128)` |   | `false` |  
+`account_name` | The account name of the user | `string(ABC..Z)` |   | `true` |  
 `account_realm` | The account realm of the user | `string(4..253)` |   | `false` |  
-`credentials` | A hash of the uses credentials | `string(1..64)` |   | `true` |  
-`method` | The hash method | `string('md5' | 'sha')` | `md5` | `false` |  
+`credentials` | A hash of the user credentials | `string(1..64)` |   | `true` |  
+`method` | The hash method | `string(md5..sha)` | `md5` | `true` |
```
jamesaimonetti: `string(md5 | sha)` is how we denote that the value of `method` can be either `md5` OR `sha`, `..` indicating a range of values, `|` indicating an `OR`.

JonTheWong: I added the .. to keep the style the same, only reason i removed the | operator is that it doesn't show properly when previewing. In this case the OR should have been used.

applications/crossbar/doc/user_authentication.md
JonTheWong: I understand, my reasoning was that it could be alphanumeric and it wasn't clear to us. I only read this part of the doc and should have done more research before submitting.

applications/crossbar/doc/user_authentication.md
jamesaimonetti: Yeah Github won't render everything exactly right. The docs site (see https://docs.2600hz.com/dev/applications/crossbar/doc/user_authentication/ for instance) uses mkdocs to render the markdown into HTML so that's the rendering we target to make work.

jamesaimonetti:
@JonTheWong any timeline on updating the PR?

applications/crossbar/doc/user_authentication.md
JonTheWong: I understand.

applications/crossbar/doc/user_authentication.md
```diff
@@ -12,9 +12,9 @@ Provides an auth-token via user credentials
 
 Key | Description | Type | Default | Required | Support Level
 --- | ----------- | ---- | ------- | -------- | -------------
-`account_name` | The account name of the user | `string(1..128)` |   | `false` |  
+`account_name` | The account name of the user | `string(1..128)` |   | `true` |
```
jamesaimonetti: You are trying to change `support level` which is only a field for 2600Hz (and is populated based on the schema. Revert this change.
Also, `account_name` is not strictly required; `account_realm` or a `phone_number` assigned to the account serve the same purposes.

applications/crossbar/doc/user_authentication.md
```diff
@@ -116,7 +123,7 @@ Send a reset-your-password email via user credentials
 
 Key | Description | Type | Default | Required | Support Level
 --- | ----------- | ---- | ------- | -------- | -------------
-`account_name` | The account name of the user | `string(1..64)` |   | `false` |
+`account_name` | The account name of the user | `string(ABC...Z)` |   | `false` |
```
jamesaimonetti: Revert this

applications/crossbar/doc/user_authentication.md
```diff
@@ -12,9 +12,9 @@ Provides an auth-token via user credentials
 
 Key | Description | Type | Default | Required | Support Level
 --- | ----------- | ---- | ------- | -------- | -------------
-`account_name` | The account name of the user | `string(1..128)` |   | `false` |  
+`account_name` | The account name of the user | `string(1..128)` |   | `true` |  
 `account_realm` | The account realm of the user | `string(4..253)` |   | `false` |  
-`credentials` | A hash of the uses credentials | `string(1..64)` |   | `true` |  
+`credentials` | A hash of the user credentials | `string(1..64)` |   | `true` |
```
jamesaimonetti: This should be fixed in crossbar/priv/couchdb/schemas/user_auth.json, then run `make apis` from the root which should auto-update this table.

PR #6330, opened 2020-02-24
[4.3] HELP-13433: noop email conversions to avoid errors

jamesaimonetti:
On some email attachments (seemingly PDFs) the `iconv` program used to
convert between character encodings (via gen_smtp/mimemail) will fail,
causing the email to be thrown away as invalid. These emails and their
attachments are, in fact, valid and should be processed.

The gen_smtp commit hash is updated to change mimemail to consider
matching `From`/`To` encodings as a no-op and avoid the call to
`iconv` entirely. This allows the PDF to be processed out of the email
for use by fax_smtp.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,3 +86,4 @@ core/kazoo_proper/priv/mp3.mp3
 
 # asdf-vm "local" versions handler
 .tool-versions
+/applications/fax/test/pdf.pdf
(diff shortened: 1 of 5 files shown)


lazedo:
@jamesaimonetti can you add the file that failed for both ?

PR #6342, opened 2020-02-27
[4.3] KCRO-15: expose transcription capabilities in the crossbar authentication response

k-anderson:
When generating a new JWT for Crossbar the response now includes the cluster id as well as a new object 'capabilities'.  The capabilities object conveys services and settings of the cluster so the UI can modify the user experience to match.  At this time the only parameters that are exposed are voicemail.transcription.supported if a transcription service is configured with an API key and voicemail.transcription.default to inform the UI transcriptions are created by default.


k-anderson:
Master PR #6343

PR #6344, opened 2020-02-27
Minor Type-O

JonTheWong:
Minor correction.

### DIFF
--- a/applications/crossbar/priv/couchdb/schemas/user_auth.json
+++ b/applications/crossbar/priv/couchdb/schemas/user_auth.json
@@ -17,7 +17,7 @@
             "type": "string"
         },
         "credentials": {
-            "description": "A hash of the users credentials",
+            "description": "A hash of the user credentials",
             "maxLength": 64,
             "minLength": 1,
             "type": "string"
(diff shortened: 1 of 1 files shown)


applications/crossbar/priv/couchdb/schemas/user_auth.json
```diff
@@ -17,7 +17,7 @@
             "type": "string"
         },
         "credentials": {
-            "description": "A hash of the uses credentials",
+            "description": "A hash of the users credentials",
```
lazedo:
```suggestion
            "description": "A hash of the user credentials",
```

JonTheWong: done

JonTheWong:
Should i update line 35: A phone number assigned to the user account ?

lazedo:
@JonTheWong since this pr was to address this specific typo, there's no need for that. however, you're changing a json schema which in turns requires the docs and the apis to be updated. you should run `make apis` so that you get required changes locally and then you can add them to the pull request.

jamesaimonetti:
I updated the PR with `make apis` corrections and merged. Please run similar (all checks in the build_branch CI task are runnable locally) next time! Thanks for the fix!

PR #6346, opened 2020-02-28
KQUB-56: update quantifiers to support pro and basic

mark2600:
this will require a "sup qubicle_maintenance update_offering" after upgrade to correctly set the offering levels on existing qubicle entities


core/kazoo_services/src/kazoo_services_maintenance.erl
```diff
@@ -149,7 +149,11 @@ js_quantify_map() ->
        "        case 'user':"
        "            emit(['users', doc.priv_level], 1);"
        "            if (doc.qubicle && doc.qubicle.enabled) {"
-       "                emit(['qubicle', 'recipients'], 1);"
+       "                if (doc.qubicle.pvt_prod_level == 'pro') {"
+       "                    emit(['qubicle', 'pro_recipients'], 1);"
```
k-anderson: @mark2600 while this is fine, I would prefer if we user qubicle_offering for both emitted paths.  Going with that, we could simplify this to simply:

emit(['qubicle', doc.pvt_qubicle_offering || 'basic' + '_recipients'], 1);

core/kazoo_services/src/kazoo_services_maintenance.erl
```diff
@@ -149,7 +149,11 @@ js_quantify_map() ->
        "        case 'user':"
        "            emit(['users', doc.priv_level], 1);"
        "            if (doc.qubicle && doc.qubicle.enabled) {"
-       "                emit(['qubicle', 'recipients'], 1);"
+       "                if (doc.qubicle.pvt_prod_level == 'pro') {"
```
k-anderson: @mark2600 we would not want to embed a pvt_ parameter in an object.  These need to be at the root level for the the functions like kz_doc:public_fields to work.  I recommend doc.pvt_qubicle_offering.

core/kazoo_services/src/kazoo_services_maintenance.erl
```diff
@@ -207,7 +211,11 @@ js_quantify_map() ->
       ,kazoo_number_manager_maintenance:generate_js_classifiers(FunMatchBlock),
        "            break;"
        "        case 'qubicle_queue':"
-       "            emit(['qubicle', 'queues'], 1);"
+       "            if (doc.pvt_prod_level == 'pro') {"
```
k-anderson: @mark2600 same with above, this could be simplified to emit(['qubicle', doc.pvt_qubicle_offering || 'basic' + '_queue'], 1);

core/kazoo_services/src/kz_services_quantities.erl
```diff
@@ -532,8 +532,9 @@ calculate_qubicle_queue_updates(JObj, Updates) ->
     case kz_doc:type(JObj) =:= <<"qubicle_queue">> of
         'false' -> Updates;
         'true' ->
-            Key = [<<"qubicle">>, <<"queues">>],
-            [{Key, 1} | Updates]
+            BKey = [<<"qubicle">>, <<"queues">>],
+            PKey = [<<"qubicle">>, <<"pro_queues">>],
+            [{BKey, 1}, {PKey, 1} | Updates]
```
k-anderson: @mark2600 umm? for any queue encountered this adds 1 to both the queues and pro_queues count.  Need a condition in there to choose one or the other based on the offering.  However, with the above suggestions you can simply this as well to...

Key = [<<"qubicle">>, <<(kz_json:get_ne_value(<<"pvt_qubicle_offering">>, JObj, <<"basic">>))/binary, "_queues">>],
[{Key, 1}, {PKey, 1} | Updates]

core/kazoo_services/src/kz_services_quantities.erl
```diff
@@ -543,8 +544,9 @@ calculate_qubicle_recipient_updates(JObj, Updates) ->
     of
         'false' -> Updates;
         'true' ->
-            Key = [<<"qubicle">>, <<"recipients">>],
-            [{Key, 1} | Updates]
+            BKey = [<<"qubicle">>, <<"recipients">>],
+            PKey = [<<"qubicle">>, <<"pro_recipients">>],
+            [{BKey, 1}, {PKey, 1} | Updates]
```
k-anderson: @mark2600 same as above, could be simplified to:

Key = [<<"qubicle">>, <<(kz_json:get_ne_value(<<"pvt_qubicle_offering">>, JObj, <<"basic">>))/binary, "_recipients">>],
[{Key, 1}, | Updates]

core/kazoo_services/src/kazoo_services_maintenance.erl
k-anderson: @mark2600 actually, I just realized that this is emitted when processing the user document so you wouldn't know the pvt_qubicle_offering on the queue document.  However, I assume that you must be storing the parameters differently on the user document... Something like:

` 
  "qubicle": {
    "presence_id": "karl-support",
    "default_device": "75cc3551a8e3b0702b2131040c430628",
    "recipient": {
      "basic": {
        "enabled": true
      },
      "pro": {
        "enabled": true
      }
    }
  }
`

That would also make it easy to iterate the keys of the 'recipient' object and emit any enabled...

if (doc.qubicle && doc.qubicle.recipient) for (offering in doc.qubicle.recipient) if (doc.qubicle.recipient[offering].enabled) emit(['qubicle', offering + '_recipient'], 1);;

core/kazoo_services/src/kz_services_quantities.erl
k-anderson: @mark2600 this is going to have the same problem I identified on the view, the JObj is the user document so it wouldn't have pvt_qubicle_offering.  BUT if you are inline with the qubicle payload on the user document I mentioned above then this just becomes the same as the view iteration over those keys.

core/kazoo_services/src/kz_services_quantities.erl
```diff
@@ -532,7 +532,8 @@ calculate_qubicle_queue_updates(JObj, Updates) ->
     case kz_doc:type(JObj) =:= <<"qubicle_queue">> of
         'false' -> Updates;
         'true' ->
-            Key = [<<"qubicle">>, <<"queues">>],
+            Offering = kz_json:get_ne_value(<<"pvt_offering">>, JObj, <<"basic">>),
+            Key = <<Offering/binary, "_queue">>,
```
k-anderson: @mark2600 the key should be:

Key = [<<"qubicle">>, <<Offering/binary, "_queue">>],

core/kazoo_services/src/kz_services_quantities.erl
```diff
@@ -543,7 +544,8 @@ calculate_qubicle_recipient_updates(JObj, Updates) ->
     of
         'false' -> Updates;
         'true' ->
-            Key = [<<"qubicle">>, <<"recipients">>],
+            Offering = kz_json:get_ne_value([<<"qubicle">>, <<"recipient">>, <<"offering">>], JObj, <<"basic">>),
+            Key = <<Offering/binary, "_recipient">>,
```
k-anderson: @mark2600 the key should be:

Key = [<<"qubicle">>, <<Offering/binary, "_recipient">>],

core/kazoo_services/src/kazoo_services_maintenance.erl
```diff
@@ -149,7 +149,11 @@ js_quantify_map() ->
        "        case 'user':"
        "            emit(['users', doc.priv_level], 1);"
        "            if (doc.qubicle && doc.qubicle.enabled) {"
-       "                emit(['qubicle', 'recipients'], 1);"
+       "                if (doc.qubicle.recipient && doc.qubicle.recipient.offering == 'pro') {"
+       "                    emit(['qubicle', 'pro_recipient'], 1);"
+       "                } else {"
```
k-anderson: @mark2600 while there is only two offerings currently, why not just make this generic like the Erlang version (they should be functional equivalents)?

if (doc.qubicle.recipient) {
    emit(['qubicle', doc.qubicle.recipient.offering || 'basic' + '_recipient'], 1);"
}

core/kazoo_services/src/kazoo_services_maintenance.erl
```diff
@@ -149,7 +149,11 @@ js_quantify_map() ->
        "        case 'user':"
        "            emit(['users', doc.priv_level], 1);"
        "            if (doc.qubicle && doc.qubicle.enabled) {"
```
k-anderson: @mark2600 just checking, did this move? doc.qubicle.enabled to doc.qubicle.recipient.enabled?  If so double check the Erlang version please :)

k-anderson:
With the latest changes this looks good to me but mark has asked to not merge it yet (@mark2600 maybe make it a draft PR for now?)

PR #6357, opened 2020-03-04
interaction refactoring

lazedo:
dependency https://github.com/signalwire/freeswitch/pull/470

- interaction based on a single document (no more `reduce` and `re-reduce`) 
- no need to fetch keys then docs then filter then format output
- `try` to use `mango` queries with simple indexes to fetch data only the fields required for output
- allows storage & management of queries (columns, filters, etc)
- fetch by [account | user | device] + timestamp as primary keys
- possibly backport to 4.3 (only storage?) to ensure smooth transition

early example of  [cdr](https://gist.github.com/lazedo/d27a797e613ea059aaf06f50c3d2f4bb) document 

`early example of view by ownerid`
```
function(doc) {if (doc.pvt_deleted || doc.pvt_type != 'cdr' || doc.pvt_vsn != 3 || !doc.interaction_is_root) return; var ids = new Array(); var onlyUnique = function(value, index, self) {return self.indexOf(value) === index;}; for(var cf in doc.extended_data.channel_callflow) {var p = doc.extended_data.channel_callflow[cf].caller_profile; if(p.directory.endpoint_owner_id && ids.indexOf(p.directory.endpoint_owner_id) === -1) ids.push(p.directory.endpoint_owner_id); if(p.hasOwnProperty("originator") && p.originator.directory.endpoint_owner_id && ids.indexOf(p.originator.directory.endpoint_owner_id) === -1) ids.push(p.originator.directory.endpoint_owner_id);if(p.hasOwnProperty("originatee") && p.originatee.directory.endpoint_owner_id && ids.indexOf(p.originatee.directory.endpoint_owner_id) === -1) ids.push(p.originatee.directory.endpoint_owner_id);};  for(i in ids) {emit([ids[i], doc.timestamp], {id: doc._id});} 
```

example of `mango` query
```
{
   "selector": {
      "extended_data.channel_callflow": {
         "$elemMatch": {
            "caller_profile.originator.directory.endpoint_owner_id": "a4b6678410311cfb7d514ad0171a4adb"
         }
      }
   }
}
```


lazedo:
marking `ready` so it can be merged and continue with post breakening.

`TODO`

- `cdr_report` should split the payload in parts and create whatever `docs` it needs

PR #6367, opened 2020-03-05
[4.3] KCRO-14: emit the billing_seconds for the reduce

k-anderson:
The reduce view for user interactions uses the billing seconds but it is not emitted from the map function.


k-anderson:
Master PR #6366

PR #6372, opened 2020-03-09
KZOO-52: use a default empty list for messages

jamesaimonetti:
When a voicemail box is empty and deleted via API, an error would be
thrown during deletion when deciding to publish deleted
messages. kvm_messages:fetch/4 strips empty values from the results,
`succeeded` would be `[]` and thus stripped, so `get_list_value/2`
returned `undefined`. This caused the call to `lists:foreach/2` to
crash.

Instead, fetch `succeeded` with `[]` as a default and only process if
the result is a non-empty list.


jamesaimonetti:
See 4.3 version in #6371

PR #6386, opened 2020-03-18
[4.3] KZOO-56: handle conflict when moving faxes to MODB

jamesaimonetti:
The faxes DB fills up when the `move_doc` fails with a conflict when
inserting the doc into the MODB. Handle the error by checking that the
document exists in the MODB and forcing deletion of the doc in the
faxes database.


jamesaimonetti:
Testing coming, hol' up on merging

lazedo:
@jamesaimonetti i think this doesn't address the case. if the `move` fails in the attachments part, we're deleting it from `FromDB`. the simpler fix would be to hard delete the doc in `ToDB` and let it retry. but the real problem imo, is with the way `ensure_saved` works now.

PR #6387, opened 2020-03-18
KZOO-56: handle conflict when moving faxes to MODB

jamesaimonetti:
The faxes DB fills up when the `move_doc` fails with a conflict when
inserting the doc into the MODB. Handle the error by checking that the
document exists in the MODB and forcing deletion of the doc in the
faxes database.


lazedo:
@jamesaimonetti see the comments in 4.3 pr

PR #6393, opened 2020-03-19
[master] fixed circle failing when opening a PR

swysor:


.circleci/config.yml
```diff
@@ -229,3 +237,85 @@ jobs:
             else
               printf ":: no errors are found in release\n\n"
             fi
+
+  build_centos7:
+    docker:
+      - image: offical2600hz/kazoo-packager-19:1.0-centos-7
```
lazedo: should be 21 ?

swysor: oh good point, copy error, also removing the job from the workflow. Will see if that fixes, going to close and reopen this pr to test.

PR #6402, opened 2020-03-23
backport im services - 4.3

lazedo:
* compatibility with master


applications/doodle/src/doodle_listener.erl
```diff
@@ -286,66 +289,55 @@ account_from_number(#{phone_number := KNumber, request := JObj} = Map) ->
...
+    IM = kapps_im:exec(Funs, kapps_im:from_payload(SmsReq)),
     Map#{fetch_id => kapps_im:message_id(IM)
         ,message_id => kapps_im:message_id(IM)
         ,im => IM
         };
 create_im(Map) -> Map.
 
+-spec maybe_update_sender(kapps_im:im()) -> kapps_im:im().
+maybe_update_sender(Im) ->
+    case kapps_im:endpoint(Im) of
```
jamesaimonetti: Dialyzer doesn't think `undefined` will be reached, since the Im record defaults to empty JSON object. Would you prefer to test for empty JObj and return Im unmodified or just bind EP and update the From?

PR #6406, opened 2020-03-24
KZOO-57: calculate presence id for route requests

jamesaimonetti:
When a device registers, the calculated presence_id is cached and
passed along on subsequent INVITEs. If a user hotdesks into the
device, the cached presence_id (likely the SIP username of the device)
will continue to be used on call events related to that device.

This causes issues with applications (like Qubicle) that track a
user's calls based on the user's presence_id being present on dialog
update events (confirmed and terminated).

Instead, provide a uniform way to calculate a device's presence_id and
use it for setting CCVs on the route response in callflow and in
registrar for authn responses.

### DIFF
--- a/core/kazoo_directory/src/kz_directory_endpoint.erl
+++ b/core/kazoo_directory/src/kz_directory_endpoint.erl
@@ -99,7 +99,7 @@ presence_id(Endpoint, CCVs) ->
 
 -spec hold_music(kz_json:object(), kz_json:object()) -> kz_json:object().
 hold_music(Endpoint, CCVs) ->
-    case kz_json:get_value([<<"music_on_hold">>, <<"media_id">>], Endpoint) of
+    case kz_json:get_ne_binary_value([<<"music_on_hold">>, <<"media_id">>], Endpoint) of
         'undefined' -> CCVs;
         MediaId ->
             MOH = kz_media_util:media_path(MediaId, kz_doc:account_id(Endpoint)),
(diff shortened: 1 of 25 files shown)


jamesaimonetti:
@lazedo is the directory stuff done on each call? I worry that it won't necessarily reflect the most recent values if the device is hotdesked into.

lazedo:
@jamesaimonetti yes

jamesaimonetti:
OK removed and updated some missing fields from the callflows schema.

core/kazoo_directory/src/kz_directory_endpoint.erl
```diff
@@ -86,36 +86,20 @@ endpoint_id(Endpoint, CCVs) ->
...
 owner_id(Endpoint, CCVs) ->
-    case kz_json:get_ne_binary_value(<<"owner_id">>, Endpoint) of
+    case kzd_devices:owner_id(Endpoint) of
         'undefined' -> CCVs;
         OwnerId -> kz_json:set_value(<<"Owner-ID">>, OwnerId, CCVs)
     end.
 
+-spec presence_id(kz_json:object(), kz_json:object()) -> kz_json:object().
+presence_id(Endpoint, CCVs) ->
+    kz_json:set_value(<<"Presence-ID">>, kzd_devices:calculate_presence_id(Endpoint), CCVs).
```
lazedo: will revisit this and probably remove as the endpoint will already have merged the hotdesked user

sergey-safarov:
Hi @jamesaimonetti , @lazedo 

After commit dbc98c0aea7b229421d29dd35365bfe31258b160 I not able send call to FreeSwitch port  11000 from user device.
So, now if user not registered on Kamailio, say send call to different kamailio then registered, this call will fail.

lazedo:
Hi @sergey-safarov ,
I not able send call to FreeSwitch port 11000 from user device` , all calls go thru kamailio.
if user is not registered, it will be challenged at kamailio

sergey-safarov:
Thanks Luis @lazedo 
I will update kamailio

PR #6407, opened 2020-03-25
[4.3] fallback to en-US if i18n label is missing

icehess:
Fallback to en-US label of UI app when user, account and system default language is not en-US.

### DIFF
--- a/applications/crossbar/src/crossbar_util.erl
+++ b/applications/crossbar/src/crossbar_util.erl
@@ -716,14 +716,18 @@ load_apps(AccountId, UserId, Language) ->
 %%------------------------------------------------------------------------------
 -spec format_app(kz_term:ne_binary(), kz_json:object()) -> kz_json:object().
 format_app(Lang, AppJObj) ->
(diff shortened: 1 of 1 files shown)


applications/crossbar/src/crossbar_util.erl
```diff
@@ -717,13 +717,14 @@ load_apps(AccountId, UserId, Language) ->
 -spec format_app(kz_term:ne_binary(), kz_json:object()) -> kz_json:object().
 format_app(Lang, AppJObj) ->
     I18N = kzd_app:i18n(AppJObj),
-    DefaultLabel = kz_json:get_value([?DEFAULT_LANGUAGE, <<"label">>], I18N),
+    EnLabel = kz_json:get_ne_binary_value([<<"en-US">>, <<"label">>], I18N),
```
jamesaimonetti: wouldn't `kz_json:get_first_defined([[Lang, label], [default, label], [en-us, label]], I18N)` suffice?

PR #6416, opened 2020-03-27
resolve dialyzer issue with getopt:parse in sup.erl

danielfinke:
fixes

`sup.erl:204: The pattern {'ok', {_, _}} can never match the type {'error',{atom(),_}}`

`Options` will always be a list according to the `getopt` code

### DIFF
--- a/core/sup/src/sup.erl
+++ b/core/sup/src/sup.erl
@@ -227,7 +227,7 @@ print_invalid_cli_args() ->
 -spec parse_args(string()) -> {'ok', kz_term:proplist(), list()}.
 parse_args(CommandLineArgs) ->
     case getopt:parse(option_spec_list(), CommandLineArgs) of
-        {'ok', {Options, Args}} when is_list(Options) ->
+        {'ok', {Options, Args}} ->
             {'ok', Options, Args};
         {'error', {_, _}} ->
             print_help()


danielfinke:
Test failure appears unrelated to the PR.

core/sup/src/sup.erl
```diff
@@ -227,7 +227,7 @@ print_invalid_cli_args() ->
 -spec parse_args(string()) -> {'ok', kz_term:proplist(), list()}.
 parse_args(CommandLineArgs) ->
     case getopt:parse(option_spec_list(), CommandLineArgs) of
-        {'ok', {Options, Args}} when is_list(Options) ->
+        {'ok', {Options, Args}} ->
```
jamesaimonetti: getopt returns `{ok, {[getopt:option()], [string()]}}`. Reverting your change does not illicit a dialyzer complaint. What issue were you seeing?

jamesaimonetti:
Haven't recreated the dialyzer warning, the is_list line has been there since 2016, don't see a reason to keep this PR open. If you have compelling arguments for the change, please re-open!

danielfinke:
@jamesaimonetti redid this test on my own and it's no longer an issue with or without. So not sure what happened on our end here.

PR #6418, opened 2020-03-27
PISTON-1048: fix infinite loop when searching by queue ID key in queues/agents_listing

danielfinke:
Resolves a similar issue to https://github.com/2600hz/kazoo/pull/6309

### DIFF
--- a/applications/acdc/priv/couchdb/views/queues.json
+++ b/applications/acdc/priv/couchdb/views/queues.json
@@ -14,7 +14,7 @@
                 "function(doc) {",
                 "  if (doc.pvt_type !== 'user' || doc.pvt_deleted || typeof doc.queues !== 'object') return;",
                 "  for (i in doc.queues) {",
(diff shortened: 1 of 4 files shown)


applications/acdc/priv/couchdb/views/queues.json
```diff
@@ -14,7 +14,7 @@
                 "function(doc) {",
                 "  if (doc.pvt_type !== 'user' || doc.pvt_deleted || typeof doc.queues !== 'object') return;",
                 "  for (i in doc.queues) {",
-                "    emit(doc.queues[i], doc._id);",
+                "    emit([doc.queues[i], doc._id], doc._id);",
```
jamesaimonetti: One note: if you emit `null` instead of `doc._id` you can still access the id with `kz_doc:id(ViewResult)`. Would cut down on the view index's size. Update code using `get_value(<<"value">>, ViewResult)` accordingly.

jamesaimonetti:
Merged

PR #6419, opened 2020-03-27
PISTON-1049: j5_channels: handle empty string presence IDs (don't crash)

danielfinke:
Turns out, users may erroneously set their presence ID to a single space (or similar) and this leads to a blank username in the To field. j5_channels was not handling this and was crashing whenever those users placed calls.

### DIFF
--- a/applications/jonny5/src/j5_channels.erl
+++ b/applications/jonny5/src/j5_channels.erl
@@ -846,19 +846,21 @@ billing_jobj(BillingSeconds, Channel) ->
 
 -spec to_did_lookup(kz_json:object()) -> kz_term:ne_binary() | 'undefined'.
 to_did_lookup(JObj) ->
-    case kz_json:get_first_defined(
-           [<<"To">>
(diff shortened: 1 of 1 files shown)


danielfinke:
Tests appear unrelated to the content of this PR.

lazedo:
what does ` presence ID` fit in this change ?

danielfinke:
@lazedo meaning the `presence_id` field on a user
It affects the To field on the outbound leg if that user is called. In our case it was in a ring group.

lazedo:
did reproduced this in master ? would like to see that because its not supposed to

jamesaimonetti:
@danielfinke any chance you reproduced this on master?

danielfinke:
@jamesaimonetti I haven't tried yet, sorry. Been super busy. I can try to prioritize it for sometime this week and get back to you.

lazedo:
@danielfinke no worries, since this is a simple file pr, just reopen after the breakening

PR #6420, opened 2020-03-27
PISTON-1050: j5_channels: suppress logging expected ignored kz_hook events

danielfinke:
Suppress some additional events in j5_channels that may be present if blackhole registrations are in place for those events (because both j5_channels and blackhole use kz_hooks)


jamesaimonetti:
Merged

PR #6421, opened 2020-03-27
[4.3] KZOO-60: permit conference feature code

jamesaimonetti:
Rather than needing to know the extension to dial for a conference,
allow a feature code to capture the conference ID from the dialed
number. Useful for clients that are provisioned dynamically (like
WebRTC devices) that have full control over the dialed number (as
opposed to typical desk phones which only dial digits).


jamesaimonetti:
master at #6425

PR #6423, opened 2020-03-27
Added devices module to the tasks application to allow bulk creation of devices through the api

bradfordben:


bradfordben:
Pulling this PR for the moment, will submit after the "Great Breakening" this weekend

PR #6426, opened 2020-03-27
Added vmbox module to the tasks app to allow bulk creation of vmboxes

bradfordben:


bradfordben:
Pulling this PR for the moment, will submit after the "Great Breakening" this weekend

PR #6429, opened 2020-03-31
Add to some of the examples (icon, logo, etc.)

ruhnet:
Some endpoints were not fully described, and examples did not work because of missing headers or other information. I corrected a couple of these examples.

### DIFF
--- a/applications/crossbar/doc/ref/whitelabel.md
+++ b/applications/crossbar/doc/ref/whitelabel.md
@@ -49,6 +49,7 @@ curl -v -X GET \
 ```shell
 curl -v -X PUT \
     -H "X-Auth-Token: {AUTH_TOKEN}" \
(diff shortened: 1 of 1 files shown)


jamesaimonetti:
@ruhnet thanks so much!

One note (and I'll try to figure out a better way to indicate this): the docs under the "ref" folder are the auto-generated versions KAZOO builds during `make apis` (for instance). The actual docs that correspond to the docs site are a directory lower (crossbar/doc for instance).

Can you move your changes to `crossbar/doc/whitelabeling.md` instead?

ruhnet:
Ah, ok yes I wondered about the 2 sets of similar docs in some cases. So that makes sense. I'll move the changes to the proper place. 
I don't know Erlang (yet?) but I hope to be able to contribute to the documentation at least a little. 
(BTW I'm "lvlinux" on Freenode.)

PR #6438, opened 2020-04-01
[Master] - KZOO-46: Add Emergency Notification for E911 Compliance

mk1s:
This PR adds new functionality to help facilitate compliance with Kari's Law and notification for when an emergency call is placed.

An email contact list can now be set on a phone numbers emergency metadata.

Notification can now be achieved one of several ways:

* A new amqp event, `emergency_bridge` has been added to `kapi_notifications` to trigger a teletype notification email when an emergency call is placed.
* A new dedicated emergency bridge webhook can be configured

### KZOO-48: Create new teletype template for Emergency Call Notifcation
* Adds new text and html teletype templates
* Adds new teletype module to render template
* Template will distinguish between Emergency and test calls

### KZOO-49: Create AMQP Message For Emergency Notification
* Creates a new amqp definition ing `kapi_notifications` for usage when an emergency call is attempted.
* Publish callback as been added to `cb_notifications`
* Set emergency location data as well as user and device metadata

### KZOO-50: Trigger Notification When Emergency Call is Attempted
* Publishes `emergency_bridge` message when stepswitch determines an emergency call is being attempted
* Adds a new test rules object a resource to define test patterns for a resource to help differentiate between emergency and test calls.

 ### KZOO-51: Create a dedicated webhook module for emergency calls
* Adds `webhooks_emergency_bridge` which is explicitly dedicated to emergency call notification.

### DIFF
--- a/applications/crossbar/doc/phone_numbers.md
+++ b/applications/crossbar/doc/phone_numbers.md
@@ -39,6 +39,8 @@ Key | Description | Type | Default | Required | Support Level
 `e911.locality` | The locality (city) where the number is in service | `string()` |   | `true` |  
 `e911.location_id` | The e911 provisioning system internal id for this service address | `string()` |   | `false` |  
 `e911.longitude` | The e911 provisioning system calculated service address longitude | `string()` |   | `false` |  
+`e911.notification_contact_emails.[]` |   | `string()` |   | `false` |  
+`e911.notification_contact_emails` | A list of email addresses to receive notification when this number places an emergency call | `array(string())` | `[]` | `false` |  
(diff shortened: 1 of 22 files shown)


icehess:
There is no need for a dedicated webhook module, all `kapi_notifications` events are exposed as webhooks.

mk1s:
> There is no need for a dedicated webhook module, all `kapi_notifications` events are exposed as webhooks.

@icehess Thank you for having a look!  

While the hook is exposed by the existing notifications hook, one piece of the specification was:

>  We should add a new webhook that triggers from the AMQP message for a new emergency call

mk1s:
After GH outage no longer tracking recent commits, creating new one.

PR #6439, opened 2020-04-02
[4.3] Replace (and remove) ?LOG_DEV with lager:debug

harenson:
Currently on production we are not storing on disk (log files) lines that use lager's `dev` sink which makes these debugging lines useless on production servers and they are necessary when you are debugging issues.

This is why I replaced those `?LOG_DEV` macro calls with `lager:debug` which is the less important logging level on lager application.


lazedo:
@harenson this was intended to reduce log size as most of them serve as developer info. please, do not remove the `?LOG_DEV` macro.  which of those log lines do you think are useful and why ?

PR #6440, opened 2020-04-02
Replace (and remove) ?LOG_DEV macro with lager:debug

harenson:


lazedo:
@harenson comment in 4.3 pr

PR #6446, opened 2020-04-03
Master pr kz csv changes

bradfordben:
Standardized (with the lack of standards, https://tools.ietf.org/html/rfc4180) the creating and reading of CSV files. Properly escaping charters.

We needed to be able to do this as the current csv code could not handle escaped `,` correctly (Would replace them with `;`)

This was needed for a kazoo tasks callflow module (separate pr to come)to allow the user to create a callflow with a dynamic `flow` you needed to be able to take `flow` as an escaped json string.

Example json data in a csv row would be
`not json,"{""data"": { ""id"": ""2ab9ffb3bad31b8bb1d38eaf840f8dcc"", ""timeout"": 20, ""can_call_self"": false, ""delay"": 0, ""strategy"": ""simultaneous"" }, ""module"": ""user"", ""children"": {}}"`


bradfordben:
@jamesaimonetti Think the build issue here is unrelated issue ?

core/kazoo_csv/test/kz_csv_tests.erl
```diff
@@ -213,26 +218,26 @@ mapped_row_to_iolist_test_() ->
 
 mapped_row_data() ->
     [{5, <<",,,,\n">>, #{}}
-    ,{5, <<"\"a\",\"b\",,,\n">>, #{<<"1">> => <<"a">>, <<"2">> => <<"b">>}}
-    ,{2, <<"\"a\",\"b\"\n">>, #{<<"1">> => <<"a">>, <<"2">> => <<"b">>}}
-    ,{3, <<"\"a\",,\"b\"\n">>, #{<<"1">> => <<"a">>, <<"2">> => ?ZILCH, <<"3">> => <<"b">>}}
-    ,{2, <<"\"a\",\n">>, #{<<"1">> => <<"a">>, <<"2">> => ?ZILCH, <<"3">> => <<"b">>}}
-    ,{3, <<",,\"b\"\n">>, #{<<"1">> => ?ZILCH, <<"2">> => ?ZILCH, <<"3">> => <<"b">>}}
+    ,{5, <<"a,b,,,\n">>, #{<<"1">> => <<"a">>, <<"2">> => <<"b">>}}
```
jamesaimonetti, line 221: Why removing the quotes? I believe they are there because some programs export fields wrapped in single or double quotes. Why not add unquoted versions to make sure we're getting the same results regardless of the quoted-ness of the cell?

bradfordben: Ok, sorry i missed the reason for failure. Il look in to that. 

In relation to wrapping or not wrapping all cells in double quotes, i dont mind. Both are acceptable ways of doing it. I originally had 2 code paths for simple unwrapped cells vs more complex wrapped cells but then both ended up having the same order N complexity so decided to just drop the simple cell case but kept the not needing to wrap the cell. 

On 4.3 its older implementation dose not wrap every cell and I noticed that you added the wrapping every cell to your master but thought you just added it to help escape some charters / edge case. If you would prefer i can add it back in? The joys of there being no official spec of CSVs 🙄

jamesaimonetti: Yeah every time I touch CSV parsing I run away vowing never to touch it again LOL.

If the tests include it then we need to keep it because it was likely added in response to a support ticket.

bradfordben: Yeah haha! Well hopefully after this its the last for awhile!! 

Ok yeah i can re-add 👍

bradfordben:
@jamesaimonetti I have change things to always surround cells in double quotes and fixed the json to/from csv error so it should be good for a second look when you have a chance.

bradfordben:
Im not sure why make release failed, It looks like it hit a system error when trying to load a view from couch during account validation. Cant see why the CSV changes would have caused this.

bradfordben:
rebasing on master

jamesaimonetti:
Merged, thanks!

bradfordben:
@jamesaimonetti Thanks for reviewing. Would you like me to backport this to 4.3 or not?

PR #6452, opened 2020-04-06
[4.3] Handle not system_config ACLs when deleting them via SUP

harenson:


lazedo:
@harenson this looks more like a `doc` than a `code` thing, no ?

harenson:
@lazedo This change gives the user a hint about why the ACL cannot be removed. Instead of them having to search documentation based on a `ACL not found` message that says nothing about why it was not deleted.

lazedo:
@harenson sup commands are run by `admins` who should read the docs. i don't agree with this at all.
`ecallmgr` knows `nothing` about `crossbar` apis or `monster-ui` apps.

jamesaimonetti:
@harenson i agree, i would just echo out the authz type and id and suggest using an appropriate API. Something like "This entry is not managed by ecallmgr. It is of type {AUTHZ_TYPE}; please use the APIs to manage it"

PR #6465, opened 2020-04-11
Make pusher compatible with iOS 13 or higher

mehrdadep:


lazedo:
@mehrdadep thank you for your pr. the settings you are trying to force can already be set in the ExtraHeaders

mehrdadep:
@lazedo thanks, but how can we add extra headers to the pusher app considering this: `sup pusher_maintenance add_apple_app AppId Certfile Host`?

lazedo:
not with that command. use kapps_config

PR #6475, opened 2020-04-13
If the control queue isn't ready, wait for it

jamesaimonetti:


lazedo:
@jamesaimonetti i think this doesn't apply to `master` where the queue is always available because of the way init works in call control

jamesaimonetti:
So noted :)

PR #6477, opened 2020-04-14
HELP-13212: Improve check to include attachment referral information

harenson:
It seems `voicemail.file_name` is also set when `transcribe` is enabled so I added a new parameter to the template data object to take an accurate decision on whether or not to include the attachment's referral information.

[4.3 PR](https://github.com/2600hz/kazoo/pull/6476)


applications/teletype/src/templates/teletype_voicemail_to_email.erl
```diff
@@ -279,6 +279,9 @@ build_voicemail_data(DataJObj) ->
       ,{<<"box">>, kz_json:get_value(<<"voicemail_box">>, DataJObj)} %% backward compatibility
       ,{<<"vmbox_name">>, kz_json:get_value([<<"vmbox_doc">>, <<"name">>], DataJObj)}
       ,{<<"vmbox_number">>, kz_json:get_value([<<"vmbox_doc">>, <<"mailbox">>], DataJObj)}
+      ,{<<"vmbox_include_message_on_notify">>
+       ,kz_json:get_value([<<"vmbox_doc">>, <<"include_message_on_notify">>], DataJObj, 'true')
```
jamesaimonetti, line 283: use typed getters when possible (`kz_json:is_true([...], DataJObj, 'true')` for instance).

PR #6490, opened 2020-04-20
[4.3] HELP-14316: test storage plans in cluster setting

jamesaimonetti:
Report that a saving of storage plans on an API server did not cause
call-handling servers to reflect the new storage plan. Unable to
replicate but added additional helpers to make sure plans are flushed
on storage doc changes


lazedo:
@jamesaimonetti not a fun of this change. 
1) exposing `handle_new` and calling it from crossbar
2) change to `cache_callback` in `kzs_plan`

basically, we're saying that `caches` don't work and that we should call `invalidate` or `reload` routines directly.

i think we should first try to reproduce the issue if there's one, and then take action.

there's a use case not handled in 4.3 where a `delete` is issued to an account storage plan followed by a PUT/POST and the storage is not updated. it was fixed in master here https://github.com/2600hz/kazoo/pull/6278 maybe this was the case ?

jamesaimonetti:
@lazedo i'll see if backporting that PR addresses the issue; however the use case here was an edit of the doc, not a delete. I added the handling of account deletion to cb_storage because we're getting orphaned storage docs that aren't cleaned up after an account is deleted.

As for `handle_new`, in the pqc_cb_storage testing, asking the node for the storage plan would fail because the doc change hadn't propagated through to the kzs_plan cache before the pqc_cb_storage code checked for the modb dataplan. Adding a call to handle_new in cb_storage alleviated this race. 

The changes to cache_callback just unify the actions - the code didn't do different things if the cache entry was 'erase'd vs 'flush'd. I don't think those changes impact anything other than simplification.

jamesaimonetti:
Also note that I can't recreate the client's issue in our environments so this PR is mostly trying to make things more robust and tested. Its entirely possible the issue is environmental

lazedo:
` I added the handling of account deletion to cb_storage because we're getting orphaned storage docs that aren't cleaned up after an account is deleted.` , this is great.

the `handle kz_att_error returns` is also great

`The changes to cache_callback just unify the actions` , i agree

i still have reservations exposing `handle_new` and using it from crossbar just because of `pqc_cb_storage`

jamesaimonetti:
@lazedo yeah it was either expose handle_new or add timer:sleep to pqc_cb_storage, which felt weird to have to do to give the system time to reflect the expected settings.

applications/crossbar/src/modules/cb_storage.erl
```diff
@@ -596,3 +628,13 @@ decode_json(RespBody) ->
             lager:debug("error body not JSON"),
             RespBody
     end.
+
+maybe_update_plans(Context) ->
+    maybe_update_plans(Context, cb_context:resp_status(Context)).
+
+maybe_update_plans(Context, 'success') ->
+    SavedDoc = cb_context:doc(Context),
+    _ = kzs_plan:handle_new(SavedDoc),
```
lazedo, line 637: @jamesaimonetti can we find an alternative for this ? maybe a kz_caches model for pqc ?

core/kazoo_data/src/kzs_plan.erl
```diff
@@ -13,7 +13,11 @@
         ,disallow_validation_overrides/0
         ]).
 
--export([init/0, reload/0, reload/1, reload/2, flush/0]).
+-export([init/0
+        ,reload/0, reload/1, reload/2
+        ,flush/0
+        ,handle_new/1, handle_old/1
```
lazedo, line 19: is `handle_old` also related to pqc ?

jamesaimonetti:
@lazedo allowing immediate purge/reload of the storage plan when modified via API seems reasonable to me. I'm curious why its a sticking point on this PR? The client making the API call can reasonably assume that, on return of the API call, the settings are in place to use, no? Exporting the functions allows, at least for the API server, to be ready to go for the next API call from the client that would need the proper settings reflected.

I trust the caches to be flushed but also don't see the harm in getting the state of KAZOO consistent with the pqc_cb_storage expectation that settings are reflected after the API call returns.

I won't fight it if you really want the functions unexported but do want to know why exporting them is a concern? :)

jamesaimonetti:
Cleaning up logging and such atm too

lazedo:
@jamesaimonetti once exported you open the door for more commits like the one in crossbar and i think we really shouldn't be doing this, dataplan control should be black boxed inside datamgr and the same should valid for other datamgr actions like updating a document.this change would open the door to other changes like direct flushing when a document is updated in crossbar. i think we can just put a `sleep` in pqc like we did for channels until we change pqc to wait for the caches when we need to check post action values like the one we're doing to check if the storage settings were actually flushed.

i really don't want to block this as the other work is great.

jamesaimonetti:
@lazedo leaving them exported since storing `fun()`s in ETS leads to issues if kzs_plan is reloaded (the funs are invalid but nothing will flush them from ETS since the binding isn't unbound). But removed direct calls and introduced a 50ms delay on pqc_cb_storage

lazedo:
@jamesaimonetti thanks. about the kzs_plan code reload and the exporting, can you try something like
```
diff --git a/core/kazoo_data/src/kzs_plan.erl b/core/kazoo_data/src/kzs_plan.erl
index 78f9ce2..63a9512 100644
--- a/core/kazoo_data/src/kzs_plan.erl
+++ b/core/kazoo_data/src/kzs_plan.erl
@@ -17,6 +17,8 @@

 -export([reset_system_dataplan/0]).

+-on_load(initx/0).
+
 -include("kz_data.hrl").

 -define(IS_JSON_GUARD(Obj), is_tuple(Obj)
@@ -487,11 +489,16 @@ start_connection(Tag, Params) ->
         'ok' -> {Tag, kz_dataconnections:get_server(Tag)}
     end.

--spec init() -> 'ok'.
-init() ->
+initx() ->
+    lager:debug("ONLOAD"),
+    kazoo_bindings:flush_mod(?MODULE),
     reload(),
     bind().

+-spec init() -> 'ok'.
+init() ->
+    lager:debug("ONINIT").
+
 -spec bind() -> 'ok'.
 -ifdef(TEST).
 bind() -> 'ok'.
@@ -505,7 +512,7 @@ bind() ->
                         ,<<"*">>
                         ], <<".">>),
     lager:debug("binding for new storage: ~s", [RK]),
-    kazoo_bindings:bind(RK, fun handle_new/1).
+    kazoo_bindings:bind(RK, ?MODULE, fun handle_new/1).
```

the `on_load` is called in the event of a code reload.

PR #6492, opened 2020-04-21
Sup UUID

noahmehl:
This is an extension of the SUP command to create the initial master account.  We wanted to be able to add optional parameters for the uuid's of the account and user.  This is so we can set the UUIDs ahead of time for functional testing.


lazedo:
@noahmehl please create a new one (maybe add a bogus commit or even rebase) as circleci can't resolve the reference to your branch

PR #6493, opened 2020-04-21
Sup UUID 43

noahmehl:
Corresponding back port PR for: https://github.com/2600hz/kazoo/pull/6495

This is an extension of the SUP command to create the initial master account. We wanted to be able to add optional parameters for the uuid's of the account and user. This is so we can set the UUIDs ahead of time for functional testing.


noahmehl:
@lazedo @jamesaimonetti, it looks like the build checks failed because of a 500 error from Github.  Is there anyways to restart the check?

PR #6513, opened 2020-04-30
Fixed bug in tasks whereby the results csv was not being generated correctly due to changes to the csv generator to add a new line charter to the end of lines

bradfordben:
As the csv module is now adding \n to the end of lines there is no need to add one in the tasks code


bradfordben:
Unrelated build issue from what i can tell.

bradfordben:
Thank you

PR #6524, opened 2020-05-04
remove application directory completely

icehess:


jamesaimonetti:
Rolled this into #6523

PR #6533, opened 2020-05-09
Fixed detergent build

sergey-safarov:
repo https://github.com/pap/detergent no more exist


sergey-safarov:
Code fresh build error related to automatic test in my account and to `4.3` branch in my repo with extra commits.
Really kazoo build properly with this commits.

lazedo:
@sergey-safarov any reason in particular for this repo ? why not use the default which goes back to `https://github.com/devinus/detergent` ? so maybe the change is just to remove the line ?

jamesaimonetti:
I'd rather just remove it and the DTH app - long dead. See #6535

jamesaimonetti:
DTH is moved to kazoo-community and detergent is removed from deps.mk

sergey-safarov:
@lazedo without [this change](https://github.com/devinus/detergent/pull/18/commits/399e363f260751b38b978aa882b9681f395bd496) `make build-release` will fail.

PR #6537, opened 2020-05-14
Call Recording Features

brian-chamberlain:
There is 1 fix:

Add a new action 'start_on_b_leg' to 'record_call' callflow

This fixes an issue where call recording files for both a-leg and b-leg were being saved
If the agent wanted to stop recording (to get the customers credit card details for example)
and then start recording again afterwards. A total of 3 call recording files were created:

1) recording of complete call on a-leg side started by by 'record_call' callflow 'start' action
2) recording between *7 start_record and *9 stop_record
3) recording after agent *7 start_record to restart recording and end of the call.

Obviously this is a big problem, the reason to stop recording was to NOT record the CC details but the details are still recorded in (1)

A new 'record_call' action 'start_on_b_leg' now starts the call recording on the b-leg rather than the a-leg so in the same scenario only 2 call recording files are
created (2) and (3) in the list above. The CC details are no longer recorded.

And a new feature:

Our customers want to delete call recordings in the call recording app so we added a soft delete and ability to patch using recordings api to add soft_delete and soft_restore.

soft_delete and soft_restore feature:

      * PATCH a call_recording document
      * Update schema to add soft_delete and soft_restore objects
      * support for 'collections' to PATCH multiple documents
      * enable chunking by default
      * by default hide docs that have soft_delete object present
      * include soft_delete docs if 'include=soft_delete' is in the query string

### DIFF
--- a/applications/callflow/src/module/cf_record_call.erl
+++ b/applications/callflow/src/module/cf_record_call.erl
@@ -3,6 +3,11 @@
 %%% @doc Handles starting/stopping a call recording.
 %%% @author James Aimonetti
 %%% @author Sponsored by Velvetech LLC, Implemented by SIPLABS LLC
+%%%
+%%% This Source Code Form is subject to the terms of the Mozilla Public
+%%% License, v. 2.0. If a copy of the MPL was not distributed with this
+%%% file, You can obtain one at https://mozilla.org/MPL/2.0/.
+%%%
 %%% @end
 %%%-----------------------------------------------------------------------------
 -module(cf_record_call).
--- a/applications/callflow/src/module/cf_record_call.erl
+++ b/applications/callflow/src/module/cf_record_call.erl
@@ -3,11 +3,6 @@
 %%% @doc Handles starting/stopping a call recording.
 %%% @author James Aimonetti
 %%% @author Sponsored by Velvetech LLC, Implemented by SIPLABS LLC
-%%%
-%%% This Source Code Form is subject to the terms of the Mozilla Public
-%%% License, v. 2.0. If a copy of the MPL was not distributed with this
-%%% file, You can obtain one at https://mozilla.org/MPL/2.0/.
-%%%
 %%% @end
 %%%-----------------------------------------------------------------------------
 -module(cf_record_call).
@@ -38,13 +33,19 @@ handle(Data, Call, <<"start">>) ->
     Call1 = kapps_call:kvs_store('recording_follow_transfer', ShouldFollowTransfer, Call),
     lager:info("starting call recording via action (follow transfer: ~s)", [ShouldFollowTransfer]),
     cf_exe:update_call(kapps_call:start_recording(Data, Call1));
-
 handle(_Data, Call, <<"stop">>) ->
-    cf_exe:update_call(kapps_call:stop_recording(Call)).
+    cf_exe:update_call(kapps_call:stop_recording(Call));
+handle(Data, Call, <<"start_on_b_leg">>) ->
+    ShouldFollowTransfer = kz_json:is_true(<<"should_follow_transfer">>, Data, 'true'),
+    Call1 = kapps_call:kvs_store('recording_follow_transfer', ShouldFollowTransfer, Call),
+    Call2 = kapps_call:kvs_store('record_on_b_leg', 'true', Call1),
+    lager:info("starting call recording on b-leg via action (follow transfer: ~s)", [ShouldFollowTransfer]),
+    cf_exe:update_call(Call2).
 
 -spec get_action(kz_json:object()) -> kz_term:ne_binary().
 get_action(Data) ->
     case kz_json:get_ne_binary_value(<<"action">>, Data) of
         <<"stop">> -> <<"stop">>;
-        _ -> <<"start">>
+        <<"start_on_b_leg">> -> <<"start_on_b_leg">>;
+        _  -> <<"start">>
     end.
--- a/applications/crossbar/priv/couchdb/schemas/call_recordings.json
+++ b/applications/crossbar/priv/couchdb/schemas/call_recordings.json
@@ -71,6 +71,11 @@
             "name": "from",
             "type": "string"
         },
+        "id": {
+            "description": "id",
+            "name": "id",
+            "type": "string"
+        },
         "interaction_id": {
             "description": "interaction_id",
             "name": "interaction_id",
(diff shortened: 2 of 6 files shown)


applications/callflow/src/module/cf_record_call.erl
```diff
@@ -3,11 +3,6 @@
 %%% @doc Handles starting/stopping a call recording.
 %%% @author James Aimonetti
 %%% @author Sponsored by Velvetech LLC, Implemented by SIPLABS LLC
-%%%
-%%% This Source Code Form is subject to the terms of the Mozilla Public
```
jamesaimonetti, line 7: Please keep this

applications/crossbar/priv/couchdb/schemas/call_recordings.json
```diff
@@ -71,6 +71,11 @@
             "name": "from",
             "type": "string"
         },
+        "id": {
```
jamesaimonetti, line 74: please don't add `id` as a schema property, these are tracked by `_id` from Couch and leaked onto the document in the response payload. It would be weird to set "id" in your request and see the real id in the response.

applications/crossbar/priv/couchdb/schemas/call_recordings.json
```diff
@@ -121,11 +191,15 @@
             "name": "to",
             "type": "string"
         },
+        "ui_metadata": {
```
jamesaimonetti, line 194: these fields are not part of the schema. any field added by the client that isn't in the schema is stored as-is. MonsterUI adds these fields for its use but have nothing to do with the backend. please remove

core/kazoo_endpoint/src/kz_endpoint_recording.erl
```diff
@@ -29,7 +29,8 @@ maybe_record_inbound(FromNetwork, Endpoint, Call) ->
           {'true', {kz_json:path(), kz_json:object()}} | 'false'.
 maybe_record_inbound(_FromNetwork, _Endpoint, _Call, 'undefined') -> 'false';
 maybe_record_inbound(FromNetwork, Endpoint, Call, Data) ->
-    case kz_json:is_true(<<"enabled">>, Data) of
+    case kz_json:is_true(<<"enabled">>, Data) orelse
```
jamesaimonetti, line 32: KAZOO code style drops `orelse` and `andalso` to new lines (which CI will alert you to). Instead this would be:

```
case kz_json:is_true(<<"enabled">>, Data)
        orelse kapps_call:kvs_fetch('record_on_b_leg', 'false', Call)
of
```

With appropriate spacing of course.

applications/crossbar/src/modules/cb_recordings.erl
```diff
@@ -5,11 +5,6 @@
 %%% @author OnNet (Kirill Sysoev [github.com/onnet])
 %%% @author Dinkor (Andrew Korniliv [github.com/dinkor])
 %%% @author Lazedo (Luis Azedo [github.com/2600hz])
-%%%
```
jamesaimonetti, line 8: Don't remove

applications/crossbar/src/modules/cb_recordings.erl
```diff
@@ -19,7 +14,9 @@
         ,resource_exists/0, resource_exists/1
         ,content_types_provided/2
         ,validate/1, validate/2
-        ,delete/2
+        ,delete/2, patch/2
+        ,to_json/1
+        ,to_csv/1
```
jamesaimonetti, line 19: Does the API not respond with CSV when you use `Accept: text/csv`? I don't think this is needed unless there's something off with the generic CSV generator.

applications/crossbar/src/modules/cb_recordings.erl
```diff
@@ -282,3 +560,70 @@ acceptable_content_types(Context) ->
...
+                  ).
+
+
+%%------------------------------------------------------------------------------
+%% @doc
+%% @end
+%%------------------------------------------------------------------------------
+-spec set_response(result(), cb_context:context()) -> cb_context:context().
+set_response(Result, Context) ->
+    set_response(Result, Context, fun() -> Context end).
```
jamesaimonetti, line 612: What purpose does this fun provide? It is unused below?

applications/crossbar/priv/couchdb/schemas/call_recordings.json
```diff
@@ -106,6 +111,71 @@
             "name": "request",
             "type": "string"
         },
+        "soft_delete": {
```
jamesaimonetti, line 114: KAZOO already has a backend concept of soft-deletes. When a DELETE is processed the document is soft-deleted by adding `pvt_deleted` (and all the views account for this when returning the listing). If a document is deleted accidentally, it can be un-deleted by an admin by removing this key from the doc (provided the doc still exists). There is a periodic cleanup of soft-deleted documents.

As the backend doesn't use these fields, I don't think adding them to the schema makes sense. Your client can add them and KAZOO will happily store the fields for you. You can then issue the GET with the appropriate query string filter to ignore docs with `soft_delete` defined... again this is all client-side work and not something the backend needs to be responsible for.

Happy to hear your reasoning and discuss this further, of course!

alanrevans: Problem is that the schema has "additionalProperties": false so Kazoo will not happilly store the fields unless we change it to true and we assumed it was set to false for a good reason

applications/crossbar/priv/couchdb/schemas/call_recordings.json
alanrevans: The schema currently is "additionalProperties": false so we can't add arbitarry fields as you suggest. I can change it to  "additionalProperties": true if you prefer

jamesaimonetti: @alanrevans it was set because these docs are considered read-only (hence why only GET/DELETE are allowed methods). Still not entirely sure on the use case being handled by these changed?

applications/crossbar/priv/couchdb/schemas/call_recordings.json
jamesaimonetti: The soft-deleted document will contain the pvt_request_id and the account/user info of who last touched the doc (in the DELETE case, who issued the DELETE). So this seems to duplicate that functionality anyway.

brian-chamberlain: Hi James

The idea behind this was that in the call recording app we have overrides that allow users  / admins to mark documents as deleted. When they do that we patch the recording/recordings collection and mark them as soft_deleted. The reason we didn't use the existing delete is that we didn't see a way to list the deleted documents via the api to allow an admin to restore the metadata without having to go into couch to restore. If you would prefer we approach this in a different way then please let me know.

Thanks
Brian

jamesaimonetti:
@brian-chamberlain @alanrevans I would say, let's split the PR into two. One for the call recording callflow action (which doesn't seem too controversial) and then we can discuss a different way to accomplish the hiding/revealing of the recordings. That work?

brian-chamberlain:
@jamesaimonetti sounds like a plan.

PR #5066, opened 2018-08-20
Added mask/unmask call recording feature

sergey-safarov:

PR #6540, opened 2020-05-20
Add a new action 'start_on_b_leg' to 'record_call' callflow

brian-chamberlain:
This fixes an issue where call recording files for both a-leg and b-leg were being saved
If the agent wanted to stop recording (to get the customers credit card details for example)
and then start recording again afterwards. A total of 3 call recording files were created:

1) recording of complete call on a-leg side started by by 'record_call' callflow 'start' action
2) recording between *7 start_record and *9 stop_record
3) recording after agent *7 start_record to restart recording and end of the call.

Obviously this is a big problem, the reason to stop recording was to NOT record the CC details but the details are still recorded in (1)

A new 'record_call' action 'start_on_b_leg' now starts the call recording on the b-leg rather than the a-leg so in the same scenario only 2 call recording files are
created (2) and (3) in the list above. The CC details are no longer recorded.

### DIFF
--- a/core/kazoo_endpoint/src/kz_endpoint_recording.erl
+++ b/core/kazoo_endpoint/src/kz_endpoint_recording.erl
@@ -29,7 +29,8 @@ maybe_record_inbound(FromNetwork, Endpoint, Call) ->
           {'true', {kz_json:path(), kz_json:object()}} | 'false'.
 maybe_record_inbound(_FromNetwork, _Endpoint, _Call, 'undefined') -> 'false';
 maybe_record_inbound(FromNetwork, Endpoint, Call, Data) ->
-    case kz_json:is_true(<<"enabled">>, Data) of
+    case kz_json:is_true(<<"enabled">>, Data)
+         orelse kapps_call:kvs_fetch('record_on_b_leg', 'false', Call) of
         'false' -> 'false';
         'true' ->
             Values = [{<<"origin">>, <<"inbound from ", FromNetwork/binary, " to endpoint">>}
--- a/applications/callflow/doc/record_call.md
+++ b/applications/callflow/doc/record_call.md
@@ -12,7 +12,7 @@ Validator for the Record Call callflow action
 
 Key | Description | Type | Default | Required | Support Level
 --- | ----------- | ---- | ------- | -------- | -------------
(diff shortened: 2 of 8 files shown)


core/kazoo_endpoint/src/kz_endpoint_recording.erl
```diff
@@ -29,7 +29,8 @@ maybe_record_inbound(FromNetwork, Endpoint, Call) ->
           {'true', {kz_json:path(), kz_json:object()}} | 'false'.
 maybe_record_inbound(_FromNetwork, _Endpoint, _Call, 'undefined') -> 'false';
 maybe_record_inbound(FromNetwork, Endpoint, Call, Data) ->
-    case kz_json:is_true(<<"enabled">>, Data) of
+    case kz_json:is_true(<<"enabled">>, Data)
+        orelse kapps_call:kvs_fetch('record_on_b_leg', 'false', Call) of
```
jamesaimonetti, line 33: @kageds this seems backwards to me?

If `enabled` is `true` and `record_on_b_leg` is `true`, shouldn't the function return `false` to *not* start the recording on the inbound(caller) leg?

jamesaimonetti:
As I think more on it, I'm not sure I completely understand why this change is necessary.

If you don't want the inbound leg recorded, don't have the `record_call` action in the callflow. Configure the agent user (or devices) to record its calls.

jamesaimonetti:
@brian-chamberlain thoughts?

sergey-safarov:
@brian-chamberlain please look PR #5066.
When you use `mute/unmute` feature then you one recoding file for the whole call.

brian-chamberlain:
Hi @jamesaimonetti sorry been bedlam here. 

The issue here is that if you're using something like ringgroups - call queues ( either ACDC or qubicle ) and you use the start call recording call flow element in conjunction with account based recording with a storage plan it renders start / stop via konami a little impotent.

Because the call recording starts on the a leg and when you stop the recording and start the recording then it creates new call recording session for the new stared recording but it also leaves an untouched full recording of the entire call.

When you use the flag as suggested here to start on b leg this fixes this behaviour. 

Its quite possible that theres a better way to fix this by looking at the behaviour of this and fixing it so that start / stop doesn't exhibit this behaviour but we used this approach as it was easiest and wasn't causing a possible breaking change.

Brian

PR #6541, opened 2020-05-20
soft_delete and soft_restore feature

brian-chamberlain:
* PATCH a call_recording document
      * Update schema to add soft_delete and soft_restore objects
      * support for 'collections' to PATCH multiple documents
      * enable chunking by default
      * by default hide docs that have soft_delete object present
      * include soft_delete docs if 'include=soft_delete' is in the query string


brian-chamberlain:
Hi @jamesaimonetti this sounds good to me. I

Thanks

PR #6544, opened 2020-05-21
PISTON-1071: keep many call KVs up to date through cf branches

danielfinke:
- main benefit: capture groups in callflows that are branched to will actually work, such as when using cf_callflow or dialing the operator in cf_voicemail
- side benefit: tracking KVs stay up to date to understand path thru callflows
- props: cf_flow_id, cf_flow_name, cf_flow, cf_capture_group, cf_capture_groups, cf_no_match

Our current use case example for this is an operator_number on a voicemail box that points to a callflow with a capture group pattern (e.g. `\*9999(\+?[0-9]*)$`). The operator for each VM box is an offnet number specific to the voicemail box. So that number is appended to the prefix *9999 and then a generic callflow can handle the operator for that box.

Another use case could be allowing the user to pick one of a set of caller IDs, using cf_callflow to direct the caller to a callflow with cf_dynamic_cid followed by, for example, a ring group, queue, or destination number. The change also leaves room for more interesting features such as a callflow module that accepts some digits and directs the caller to a capture group pattern callflow, doing something with the entered digits.

I've also included keeping cf_flow* props up to date, as it seems kind of useful for the state to contain which callflow is currently operating for a particular call. Looking for feedback here.

Also if there might be any side-effects of the change, let me know.

### DIFF
--- a/applications/callflow/src/cf_exe.erl
+++ b/applications/callflow/src/cf_exe.erl
@@ -141,12 +141,12 @@ continue_with_flow(Flow, Call) ->
     Srv = cf_exe_pid(Call),
     continue_with_flow(Flow, Srv).
 
--spec branch(kz_json:object(), kapps_call:call() | pid()) -> 'ok'.
-branch(Flow, Srv) when is_pid(Srv) ->
-    gen_server:cast(Srv, {'branch', Flow});
-branch(Flow, Call) ->
+-spec branch(kzd_callflows:doc(), kapps_call:call() | pid()) -> 'ok'.
+branch(FlowDoc, Srv) when is_pid(Srv) ->
+    gen_listener:cast(Srv, {'branch', FlowDoc});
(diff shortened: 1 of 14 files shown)


lazedo:
@danielfinke Hi, can you please `undo` the `unrelated` variable renaming like in [here](https://github.com/2600hz/kazoo/pull/6544/files#diff-5ef370aef1f3cb35b78d5660599de006L145-R149) ? thanks

danielfinke:
@lazedo I can if you want - I changed it to be clearer that the input was now a `kzd_callflows:doc()` instead of just the flow object in the doc. (I'm going to update the spec actually) Do you still want me to change it back?

danielfinke:
spec is now changed

jamesaimonetti:
@danielfinke 5.0 commits are not being accepted to master at this time.

PR #6547, opened 2020-05-26
[4.3] Moved request validation logic out of crossbar users modules and into kazoo documents users so validation logic can be used by other parts of the code base.

bradfordben:
Backport of previous commit to master.


bradfordben:
Cant get `pqc_cb_vmboxes:seq()` to fail locally, Can you rerun the tests. i dont have access to.

bradfordben:
@jamesaimonetti When you can, can you please review this, its the same as master commit you approved a few weeks back just with some minor adjustments to suit 4.3 function calls etc

It also includes 1 last commit to add an extra return type to the system_error return. This just allows you to pass the message back as part of the error. Its not needed in this case but will be needed for callflows. I have added it as a separate PR's to kazoo5 as-well.

jamesaimonetti:
Merged

bradfordben:
@jamesaimonetti Thank you :)

PR #6548, opened 2020-05-26
ACDc Updates

noahmehl:
This is a forward port of the following branch:

https://github.com/voxter/kazoo/commit/db5e7c83f5b26f35428721f4dae08f05067c8625

We are still actively testing this PR, but want to discuss any changes outside of the ACDc application.


applications/acdc/doc/architecture.md
```diff
@@ -4,7 +4,7 @@ Agents represent the endpoints that a call in a queue will try to connect with.
 
 #### Agents
 
-Agents are comprised of two processes, a `gen_listener` (named process) and a `gen_statem`. A supervisor ensures that the `gen_listener` and `gen_statem` are kept alive together. A supervisor above that manages the list of agents.
+Agents are comprised of two processes, a `gen_listener` (named process) and a `gen_fsm`. A supervisor ensures that the `gen_listener` and `gen_fsm` are kept alive together. A supervisor above that manages the list of agents.
```
jamesaimonetti, line 7: gen_fsm is deprecated, FYI. It will be removed in future OTP versions. I would suggest reverting any gen_fsm-related changes

applications/acdc/src/acdc_agent_handler.erl
```diff
@@ -4,6 +4,8 @@
 %%% @author James Aimonetti
 %%% @author Daniel Finke
 %%%
+%%% @author James Aimonetti
```
jamesaimonetti, line 7: Why the duplication?

applications/ecallmgr/priv/mod_kazoo/events/PLAYBACK_START.xml
```diff
@@ -0,0 +1,5 @@
+<event name="PLAYBACK_START">
+        <fields verbose="true">
```
lazedo, line 2: please don't use this ` <fields verbose="true">` , it will get all headers

applications/ecallmgr/src/ecallmgr.hrl
```diff
@@ -450,6 +450,7 @@
                    ,{'bridge', ['CHANNEL_BRIDGE', 'CHANNEL_UNBRIDGE']}
                    ,{'media', ['DETECTED_TONE', 'DTMF','CHANNEL_PROGRESS','CHANNEL_PROGRESS_MEDIA']}
                    ,{'record', ['RECORD_START', 'RECORD_STOP']}
+                   ,{'playback', ['PLAYBACK_STOP']}
```
lazedo, line 453: you added playback start/stop events and only stop here ?

kageds: We only actually use PLAYBACK_STOP. It is used when an agent receives a callback call. When an agent answers a callback call, an announcement is first played and then on PLAYBACK_STOP event the member is called back. 
We could remove PLAYBACK_START but thought it makes sense to treat as a pair.

applications/acdc/doc/architecture.md
kageds: This is understood and we plan to address this in a 2nd phase of work.

applications/ecallmgr/src/ecallmgr.hrl
noahmehl: @lazedo Does it make sense to include PLAYBACK_START as well in this context?  We are happy to remove it.  Please let us know your preference.

applications/ecallmgr/priv/mod_kazoo/events/PLAYBACK_START.xml
noahmehl: Waiting on resolution of PLAYBACK_START discussion: https://github.com/2600hz/kazoo/pull/6548#pullrequestreview-418419998

noahmehl:
> Generally speaking prompts and other media go in the kazoo-sounds repo

@jamesaimonetti where do the rest of the prompts come from?  I will happily work with whomever to add consistent prompts and we can get them added to the correct place.

applications/ecallmgr/src/ecallmgr.hrl
lazedo: this doesn't make sense to me and its going to increase the event stream without the need for it, you already have channel_execute_complete fired when a command completes. not willing to accept this change as-is because it will impact everybody with the increased streaming.

noahmehl: @lazedo no problem.  We will look into `channel_execute_complete` as the way to accomplish this.

jamesaimonetti:
@noahmehl so in discussions about what to do with community 5.x PRs, we don't really have the time at the moment to review/merge them as it distracts from getting the 5.0 and app exchange released for all of the breakening to take effect.

It would be preferred if you sat on the code until the official release of 5.0 and then port the voxter commit forward. Until a 5.0 cluster can be stood up, we can't divert effort to community PRs.

Once 5.0 is out the door, breakening will be back and you will have the community acdc repo to make the PR to and 2600Hz won't need to be involved.

Hope that makes sense.

PR #6552, opened 2020-05-29
[4.3] HELP-14771: Add stepswitch.min_shortdial_destination parameter

harenson:
And also refactor short-dial correction code in order to be able to add
unit tests for that functionality per @jamesaimonetti request.

Master PRs:
- https://github.com/2600hz/kazoo-stepswitch/pull/4
- https://github.com/2600hz/kazoo-crossbar/pull/21

### DIFF
--- a/applications/crossbar/priv/api/swagger.json
+++ b/applications/crossbar/priv/api/swagger.json
@@ -35251,7 +35251,7 @@
                     "type": "boolean"
                 },
                 "fixed_length_shortdial_correction": {
-                    "description": "stepswitch fixed_length_shortdial_correction",
+                    "description": "add {FIXED} digits from caller ID to destination",
                     "minimum": 1,
                     "type": "integer"
                 },
@@ -35262,13 +35262,18 @@
                 },
                 "max_shortdial_correction": {
                     "default": 5,
-                    "description": "stepswitch maximum shortdial correction",
+                    "description": "add up to {MAX} digits from caller ID to destination number",
                     "minimum": 1,
                     "type": "integer"
                 },
                 "min_shortdial_correction": {
                     "default": 2,
-                    "description": "stepswitch minimum shortdial correction",
+                    "description": "add at least {MIN} digits from caller ID to destination number",
+                    "minimum": 1,
+                    "type": "integer"
+                },
+                "min_shortdial_destination": {
+                    "description": "do short-dial correction if dialed number's length is equal to this value",
                     "minimum": 1,
                     "type": "integer"
                 },
--- a/applications/stepswitch/src/stepswitch_util.erl
+++ b/applications/stepswitch/src/stepswitch_util.erl
@@ -16,6 +16,16 @@
 -export([resources_to_endpoints/3]).
 -export([json_to_template_props/1]).
 
+-ifdef(TEST).
+-export([maybe_constrain_shortdial_correction/3
+        ,shortdial_correction_length/3
+        ,maybe_deny_reclassified_number/3
+        ,do_correct_shortdial/3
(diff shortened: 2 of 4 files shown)


applications/stepswitch/src/stepswitch_util.erl
```diff
@@ -91,41 +101,58 @@ correct_shortdial(<<"+", Number/binary>>, CIDNum, DeniedCallRestrictions) ->
...
+               ,[Number, byte_size(Number), Length]
+               ),
+    'undefined'.
+
+-spec should_correct_shortdial(kz_term:ne_binary()) -> {boolean(), pos_integer()}.
+should_correct_shortdial(Number) ->
+    Length = kapps_config:get_integer(?SS_CONFIG_CAT, <<"shortdial_correction_if_length_is">>, 0),
+    %% The length is also included on the response just to avoid having to read this configuration
+    %% parameter again when printing the debug line. So it is only useful for the `{false, N}' branch.
+    {0 =:= Length
```
jamesaimonetti, line 131: I would prefer the logic portion (comparing Length and Number) in its own function:

```erlang
should_correct_shortdial(Number) ->
    Length = kapps_config:(...),
    should_correct_shortdial(Number, Length).

should_correct_shortdial(_Number, 0) -> {'true', 0};
should_correct_shortdial(Number, Length) ->
    {byte_size(Number) =:= Length, Length}.
```

This is also easily testable without using meck or other means.

applications/crossbar/priv/api/swagger.json
```diff
@@ -35262,16 +35262,21 @@
...
                 "min_shortdial_correction": {
                     "default": 2,
-                    "description": "stepswitch minimum shortdial correction",
+                    "description": "add at least {MIN} digits from caller ID to destination number",
                     "minimum": 1,
                     "type": "integer"
                 },
+                "min_shortdial_destination": {
+                    "default": 0,
+                    "description": "do short-dial correction if dialed number's length is equal to this value ONLY when it is greater than 0",
```
jamesaimonetti, line 35277: I still think this should not default to 0. Instead, I think you could put a `minumum:1` constraint. So if `min_shortdial_destination` is unset/null, it is ignored. If set, the value must be greater than 0 already so use it.

harenson: Yeah, I forgot to remove that, the code is not relying on `0` as the default value anymore, that's a leftover from the previous changes.

applications/stepswitch/src/stepswitch_util.erl
```diff
@@ -91,61 +101,100 @@ correct_shortdial(<<"+", Number/binary>>, CIDNum, DeniedCallRestrictions) ->
...
-            'undefined';
-        'false' ->
-            CorrectedNumber
-    end.
+    lager:debug("corrected shortdial ~s via CID ~s to ~s", [Number, CIDNum, CorrectedNumber]),
+    CorrectedNumber.
 
--spec should_deny_reclassified_number(kz_term:ne_binary(), kz_json:object()) -> boolean().
-should_deny_reclassified_number(CorrectedNumber, DeniedCallRestrictions) ->
+-spec maybe_deny_reclassified_number(kz_term:ne_binary(), kz_term:ne_binary(), kz_json:objects()) -> kz_term:api_binary().
```
jamesaimonetti, line 149: Spec is wrong, `DeniedCallRestrictions` is a `kz_json:object()` no? At least, previous calls have that var tagged as such.

jamesaimonetti: Return should be `kz_term:api_ne_binary()` as well.

PR #6554, opened 2020-05-31
[4.3] acdc - serialize all agent spawns thru acdc_queues_sup

danielfinke:
- also add dedicated force-restart API for agents


jamesaimonetti:
Merged

PR #6558, opened 2020-05-31
[4.3] acdc - PISTON-55: acdc distributed member_connect_win

danielfinke:
- better synchronizes states of acdc_agent_fsm per agent

plan to make `member_connect_win` more DRY later


jamesaimonetti:
Merged

PR #6559, opened 2020-06-01
[4.3] acdc - PISTON-371: handle ringing cancellations better when originate_uuid hasn't come back yet

danielfinke:


jamesaimonetti:
Merged

PR #6572, opened 2020-06-03
[4.3] PISTON-1081: expose `created` and `status` on fax API doc GETs

danielfinke:
- matches the additional info provided when listing all faxes - doesn't make sense to get less back from the individual endpoint
- put them in _read_only so they don't get added to the fax doc if using the put actions "resubmit" and "forward"


jamesaimonetti:
Merged

PR #6581, opened 2020-06-08
[4.3] PISTON-1084: fail pivot req immediately on a req error

danielfinke:
Without publishing a pivot failed event, errors on pivot HTTP reqs such as DNS resolution failures will hang around until the 10s timeout in cf_pivot, which is pointless


jamesaimonetti:
Merged

PR #6588, opened 2020-06-17
[4.3] acdc ignore/cancel/exit overhaul

danielfinke:
- PISTON-179: re-implement exit key support
- PISTON-191: clear queue fsm when ignoring members
- PISTON-588: allow external member_call_cancels


jamesaimonetti:
Merged

PR #6593, opened 2020-06-23
[4.3] Dialyzer issues

bradfordben:


applications/crossbar/src/modules/provisioner_v5.erl
```diff
@@ -56,16 +56,11 @@ maybe_add_device_defaults(JObj) ->
...
-      kz_json:from_list(
-        [case KV of
-             {[<<"advanced">>, <<"expire">>], 'undefined'} -> {Path, 360};
-             {[<<"advanced">>, <<"srtp">>], 'undefined'} -> {Path, 'false'};
-             {[<<"basic">>, <<"enabled">>], 'undefined'} -> {Path, 'true'};
-             _ -> KV
-         end
-         || {Path,_}=KV <- kz_json:to_proplist(kz_json:flatten(LineJObj))
-        ])).
+    Defaults = [{[<<"advanced">>, <<"expire">>], 360}
```
jamesaimonetti, line 59: Is this being changed in response to something or just a backport? I don't think this is fixing a bug is it?

core/kazoo_number_manager/src/knm_numbers.erl
```diff
@@ -487,10 +486,7 @@ new(Options, ToDos, KOs) -> new(Options, ToDos, KOs, 'not_reconcilable').
 new(Options, ToDos, KOs, Reason) ->
     #{'todo' => ToDos
      ,'ok' => []
-     ,'ko' => case is_function(Reason, 1) of %%FIXME: find something better than Reason/1.
-                  'false' -> maps:from_list([{KO, Reason} || KO <- KOs]);
-                  'true' -> maps:from_list([{KO, Reason(KO)} || KO <- KOs])
-              end
+     ,'ko' => maps:from_list([{KO, Reason} || KO <- KOs])
```
jamesaimonetti, line 489: Same here, are you fixing a bug?

bradfordben:
@jamesaimonetti No they are all just changes that popped up due to dialyzer. No bug fixes here. We have changed our build ci to run dialyze-hard and not dialyze-changed as it was not catching some simple errors. 

But now its catching errors from other areas of the code base when we make any changes. So this is just dialyze errors that popped up. 

Would you consider changing your dialyze checks to run dialyze-hard ? or is there a reason why you have not chosen to do this? We chose to run hard as without it, it missed a simple type bug and caused an issue in production for us.

lazedo:
@bradfordben but the changes are breaking things, no ? specially the knm_numbers.

bradfordben:
@lazedo No the changes should not be breaking. knm_numbers change is a case that dialyzer was complaining to say that as the new/4 function is only called with an atom as Reason so the function check is not needed.

core/kazoo_number_manager/src/knm_numbers.erl
harenson: I faced this same issue when I started using _dialyze-hard_ instead of _dialyze-changed_ as it was not catching most of the warning, even the most evident ones.

As part of fixing those warning, I searched around _core/kazoo_number_manager_ application and I could not find any _Reason_ that would cause that case statement to go through the "true" branch.

I am sorry for resurrecting this very old comment, I just saw it and wanted to add my findings 😄

PR #6597, opened 2020-06-29
[4.3] PISTON-1088: more DRY Makefile dialyze targets

danielfinke:
- also make the output of check-dialyzer.escript consistently "nothing"
  whether using --bulk, --hard, or no flags at all

### DIFF
--- a/scripts/check-dialyzer.escript
+++ b/scripts/check-dialyzer.escript
@@ -10,7 +10,6 @@
 
 main([]) ->
     print_help();
-main([_KazooPLT]) -> 'ok';
 main([KazooPLT | CommandLineArgs]) ->
     {'ok', Options, Args} = parse_args(CommandLineArgs),
     handle(KazooPLT, Options, Args).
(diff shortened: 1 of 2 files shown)


scripts/check-dialyzer.escript
```diff
@@ -10,7 +10,6 @@
 
 main([]) ->
     print_help();
-main([_KazooPLT]) -> 'ok';
```
danielfinke, line 13: This clause was previously used to handle when no files were passed

scripts/check-dialyzer.escript
```diff
@@ -38,8 +37,7 @@ print_help() ->
     getopt:usage(option_spec_list(), "ERL_LIBS=deps/:core/:applications/ " ++ Script ++ " .kazoo.plt [args] [file.beam | path/ebin/ ...]"),
     halt(1).
 
-handle(_KazooPLT, _Options, []) ->
-    print_help();
+handle(_KazooPLT, _Options, []) -> 'ok';
```
danielfinke, line 40: But this clause is needed to handle when no files were passed AND there was a flag such as `--bulk` or `--hard` passed. And it will handle the L13 clause as well.

jamesaimonetti:
Merged!

PR #6598, opened 2020-06-30
[4.3] Moved callflow validation logic out of crossbar and into kazoo core so it can be used by other modules

bradfordben:
Moved validation logic so it can be used by future kazoo tasks callflow module


jamesaimonetti:
Merged

bradfordben:
Thanks @jamesaimonetti

PR #6601, opened 2020-07-02
[4.3] PISTON-1002: use collect_digits for DTMF during vm greeting/instructions

danielfinke:
Backport of https://github.com/2600hz/kazoo/pull/6249. Going to make a change in master as well as the 8-arity version of kapps_call_command:collect_digits wasn't actually needed.


jamesaimonetti:
Merged

PR #6602, opened 2020-07-06
fix missing variable_Call-Control-Queue for kazoo originated calls

alanrevans:
@lazedo @noahmehl
This is a draft PR to solicit feedback.

There appears to be a problem when Kazoo originates a call to an Endpoint using ecallmgr_originate. The channel variable 'Call-Control-Queue' is not defined in dialstring sent to FS which in turn causes CHANNEL_EXECUTE_COMPLETE events to be blocked due to the filter in CHANNEL_EXECUTE_COMPLETE.xml [https://github.com/2600hz/kazoo/blob/master/applications/ecallmgr/priv/mod_kazoo/events/CHANNEL_EXECUTE_COMPLETE.xml#L4]

With this quick fix, the variable is defined and the events start working.

### DIFF
--- a/applications/ecallmgr/src/ecallmgr_fs_channel.erl
+++ b/applications/ecallmgr/src/ecallmgr_fs_channel.erl
@@ -136,7 +136,8 @@ is_bridged(UUID) ->
                 ],
     case ets:select(?CHANNELS_TBL, MatchSpec) of
         ['undefined'] -> lager:debug("channel is not bridged"), 'false';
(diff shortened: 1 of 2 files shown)


noahmehl:
@lazedo we see this being set on inbound calls, in a similar fashion: https://github.com/2600hz/kazoo/blob/d8dec0dafbea07cc6f05d0f73d255db967412227/applications/ecallmgr/src/call_control/ecallmgr_call_control.erl#L652

jamesaimonetti:
See https://github.com/2600hz/kazoo/pull/6649 comment

PR #6609, opened 2020-07-15
Prevent Agent in Queue from Logging-out for ever

ykh:
When queue agent doesn't answers to call until the queue goes timeout, ACDC will kicked him out from queue forever.

It seems there is an issue in `restart_agent`.

Scenario:
- Make a call to queue.
- The Agent doesn't answers to call in the current call-queue while call timeout.
- Agent will be logged-out from queue for ever and doesn't back to queue!

Fix:
Use restart_child method instead of delete_child.

Questions:
- Is it a right change in code to fix?
- Is this really an issue or it's normal behavior of ACDC?

### DIFF
--- a/applications/acdc/src/acdc_agents_sup.erl
+++ b/applications/acdc/src/acdc_agents_sup.erl
@@ -82,7 +82,7 @@ stop_agent(AcctId, AgentId) ->
     case supervisor:terminate_child(?SERVER, Id) of
         'ok' ->
             lager:info("stopping agent ~s(~s)", [AgentId, AcctId]),
-            supervisor:delete_child(?SERVER, Id);
+            supervisor:restart_child(?SERVER, Id);
         E ->
             lager:info("no supervisor for agent ~s(~s) to stop", [AgentId, AcctId]),
             E


jamesaimonetti:
@danielfinke any feedback here?

ykh:
@danielfinke not yet.

danielfinke:
@ykh this isn't a suitable fix for the problem, though I haven't reproduced it myself yet. We need to retain the ability for agent processes to be stopped, such as for logout.

Also, I think this change would have been in the wrong spot anyway - it's in `stop_agent` not `restart_agent`

ykh:
@danielfinke 
You right, it's not the right fix.

But, why agent will be logged-out from queue for ever if it doesn't answer calls for three times? (the number is configurable, I know that).

Thanks

danielfinke:
@ykh the reason this function is present is to avoid situations where a queue call continues to ring an agent whose device(s) is/are unregistered. This may not be very impactful when using the `round_robin` queue strategy, but will manifest as a call stuck on the agent (and a rapidly increasing missed call count) when his devices are unregistered, the queue is using the `most_idle` queue strategy, and he is determined to be the most idle.
It will also help if you are in an environment where agents clock in/out by logging in/out of the call center. If an agent forgets to log out at the end of the day, they will eventually be automatically logged out after missing x calls.

ykh:
@danielfinke thank you for the explanation. it makes sense :) 

In that case the agent forgot to logout from call-center and he logged-out from queue after x missed calls, how can he logged back to the queue?

Does he can do this by DTMF or some way like this? (I couldn't found a way)

danielfinke:
@ykh you have a couple of options.

1. Add an `agent[action=login]` callflow module, set a number for the callflow, and have the agent call it to log in.

2. You can use the Set Agent Status API (`/agents/AID/status`) described here: https://docs.2600hz.com/supported/applications/crossbar/doc/agents/. `AID` is the agent's user ID. Your payload would look something like
```
{
  "data":{
    "status":"login",
    "presence_id":<optional>
  }
}
```
`"presence_id"` does not have to be present, but you can set a custom presence ID if desired.

The latter option is ideal if you can add interactive buttons to your phone hardware to trigger the API or if you are using a softphone that you wrote (and can therefore add the functionality to the softphone to call the API).

ykh:
thank you @danielfinke, it will helps me!

danielfinke:
@ykh no problem. Is it good with you if this PR is closed?

PR #6610, opened 2020-07-20
kzd_accounts: fixed failed notification …

sergey-safarov:
… when account have configured notification preferences


jamesaimonetti:
@sergey-safarov I'll discuss with the team but "notify" is the legacy setting for existing accounts. You have to explicitly enable "teletype" on an account to get it. For 5.x it may be appropriate to flip this but I'm not sure how many folks on 4.x have accounts configured to still use the "notify" templates.

jamesaimonetti:
https://github.com/2600hz/kazoo/pull/6649

PR #6615, opened 2020-07-30
[4.3] Upstream 4.3 telnyx carrier api issues

bradfordben:
Fixes to the Telnyx carrier module

The best effort configuration setting did not work, My guess is that this was written and then telnyx changed the api response and the response to no longer included the flag “any_best_effort” so the old code would never filter the best effort out. This improvement will now pass the best effort flag to telnyx and remove the need to filter the response.

Requests for TF numbers was failing as the request built and passed to Telnyx was incorrect. It passed the TF NAP as `prefix` in the body and it should be `nap` as there api doc states. I have also added the option to filer TF numbers by NXX.

Also found that the crossbar quantity option was always overwritten to 500 so every time you requested 15 numbers through CB it requested 500 from telnyx resulting in longer query times. The fix was to only set the default values if they were not set.

Kazoo5 PR https://github.com/2600hz/kazoo-core/pull/48

### DIFF
--- a/core/kazoo_number_manager/src/carriers/knm_telnyx.erl
+++ b/core/kazoo_number_manager/src/carriers/knm_telnyx.erl
@@ -65,23 +65,31 @@ check_numbers(_Numbers) -> {'error', 'not_implemented'}.
 %%------------------------------------------------------------------------------
 -spec find_numbers(kz_term:ne_binary(), pos_integer(), knm_search:options()) ->
           {'ok', knm_number:knm_numbers()}.
-find_numbers(<<"+1", Prefix:3/binary, _/binary>>, Quantity, Options)
-  when ?IS_US_TOLLFREE(Prefix) ->
-    Results = numbers('tollfree', Quantity, Prefix, 'undefined'),
+find_numbers(<<"+1", NPA:3/binary, _/binary>>=Num, Quantity, Options)
+  when ?IS_US_TOLLFREE(NPA) ->
+    Results = numbers('tollfree', Quantity, NPA, get_nxx(Num)),
     {'ok', numbers(Results, Options)};
 
 find_numbers(<<"+1", NPA:3/binary, _/binary>>=Num, Quantity, Options) ->
-    NXX = case byte_size(Num) >= 2+3+3 of
-              'true' -> binary:part(Num, 2+3, 3);
-              'false' -> 'undefined'
-          end,
-    Results = numbers('npa', Quantity, NPA, NXX),
+    Results = numbers('npa', Quantity, NPA, get_nxx(Num)),
     {'ok', numbers(Results, Options)};
 
 find_numbers(<<"+",_/binary>>=_InternationalNum, Quantity, Options) ->
     Country = knm_search:country(Options),
     Results = numbers('region', Quantity, Country, 'undefined'),
     {'ok', international_numbers(Results, Options)}.
+
+%%------------------------------------------------------------------------------
+%% @doc For a given Number, parse out the NXX value.
+%% Return `undefined' if the NXX value can not be parsed.
+%% @end
+%%------------------------------------------------------------------------------
+-spec get_nxx(kz_term:ne_binary()) -> kz_term:ne_binary() | 'undefined'.
+get_nxx(Num) ->
+    case byte_size(Num) >= 2+3+3 of
+        'true' -> binary:part(Num, 2+3, 3);
+        'false' -> 'undefined'
+    end.
 
 %%------------------------------------------------------------------------------
 %% @doc Acquire a given number from the carrier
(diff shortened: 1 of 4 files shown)


core/kazoo_number_manager/src/carriers/knm_telnyx.erl
```diff
@@ -65,24 +68,32 @@ check_numbers(_Numbers) -> {'error', 'not_implemented'}.
...
     Country = knm_search:country(Options),
     Results = numbers('region', Quantity, Country, 'undefined'),
     {'ok', international_numbers(Results, Options)}.
 
+%%------------------------------------------------------------------------------
+%% @doc For a given Number, parse out the NXX value.
+%% Return `undefined' if the NXX value can not be parsed.
+%% @end
+%%------------------------------------------------------------------------------
+-spec get_nxx(kz_term:ne_binary()) -> kz_term:ne_binary() | 'undefined'.
```
jamesaimonetti, line 90: api_ne_binary

core/kazoo_number_manager/src/carriers/knm_telnyx.erl
```diff
@@ -65,24 +68,32 @@ check_numbers(_Numbers) -> {'error', 'not_implemented'}.
...
+get_nxx(Num) ->
+    case byte_size(Num) >= 2+3+3 of
```
jamesaimonetti, line 92: Pulled out the ole TI-82 and computed `2+3+3 = 8`! Is there a readability thing necessitating keeping the addition? :)

Also I wonder if you can just do:

```erlang
get_nxx(<<_PlusOneAreaCode:5/binary, NXX:3/binary, _/binary>>) -> NXX;
get_nxx(_) -> 'undefined'.
```
This seems a bit more readable than guessing why `2+3+3` is important.

core/kazoo_number_manager/src/knm_telnyx_util.erl
```diff
@@ -66,13 +62,10 @@ req(Method, Path) ->
...
 req('post', ["number_searches"], JObj) ->
-    case kz_json:get_value([<<"search_descriptor">>, <<"prefix">>], JObj) of
-        <<"800">> -> rep_fixture("telnyx_tollfree_search.json");
-        'undefined' ->
-            case kz_json:get_value([<<"search_descriptor">>, <<"country_iso">>], JObj) of
-                <<"GB">> -> rep_fixture("telnyx_international_search.json");
-                'undefined' -> rep_fixture("telnyx_npa_search.json")
-            end
+    case kz_json:get_integer_value([<<"search_type">>], JObj) of
+        1 -> rep_fixture("telnyx_npa_search.json");
```
jamesaimonetti, line 66: Generally dislike `magic` numbers like these. Can you `-define(NICE_NAME, 1)` and use `?NICE_NAME` instead? I have no idea what 1, 2, and 3 mean in this context...

core/kazoo_number_manager/src/knm_telnyx_util.erl
```diff
@@ -142,9 +135,7 @@ http_options() ->
 rep({'ok', 200=Code, _Headers, <<"{",_/binary>>=Response}) ->
     ?DEBUG_APPEND("Response:~n~p~n~p~n~s~n", [Code, _Headers, Response]),
 
-    Routines = [fun(JObj) -> maybe_remove_best_effort(?SHOULD_KEEP_BEST_EFFORT, JObj) end
-               ,fun(JObj) -> maybe_filter_rates(?SHOULD_FILTER_RATES, JObj) end
-               ],
+    Routines = [fun(JObj) -> maybe_filter_rates(?SHOULD_FILTER_RATES, JObj) end],
```
jamesaimonetti, line 138: I like using named functions when possible. Could we make this be:

```erlang
Routines = [fun maybe_filter_rates/1],
...

maybe_filter_rates(JObj) -> maybe_filter_rates(JObj, ?SHOULD_FILTER_RATES).
```

Correcting arg order in `maybe_filter_rates/2`.

Thoughts?

jamesaimonetti (changes requested):

core/kazoo_number_manager/src/knm_telnyx_util.erl
bradfordben: Yeah think that is a nicer way of doing it 👍

core/kazoo_number_manager/src/knm_telnyx_util.erl
bradfordben: Yeah definitely a cleaner way of doing it. Will make the change

jamesaimonetti:
Merged

bradfordben:
Thanks

PR #6616, opened 2020-08-04
[4.3] Upstream 4.3 knm converters normalize crashes

bradfordben:
Crash fixes related to knm_converters:normalize/1

Kazoo5 PRs:
https://github.com/2600hz/kazoo-jonny5/pull/4
https://github.com/2600hz/kazoo-core/pull/54

### DIFF
--- a/core/kazoo_events/src/kz_hooks_util.erl
+++ b/core/kazoo_events/src/kz_hooks_util.erl
@@ -251,7 +251,10 @@ lookup_account_id(JObj) ->
     case kz_call_event:account_id(JObj) of
         'undefined' ->
             Number = get_inbound_destination(JObj),
-            case kz_cache:peek_local(?HOOKS_CACHE_NAME, cache_key_number(Number)) of
+            case not kz_term:is_empty(Number)
+                andalso kz_cache:peek_local(?HOOKS_CACHE_NAME, cache_key_number(Number))
+            of
+                'false' -> {'error', 'not_found'};
                 {'ok', _AccountId}=Ok -> Ok;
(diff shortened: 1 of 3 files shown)


core/kazoo_events/src/kz_hooks_util.erl
```diff
@@ -251,7 +251,10 @@ lookup_account_id(JObj) ->
     case kz_call_event:account_id(JObj) of
         'undefined' ->
             Number = get_inbound_destination(JObj),
-            case kz_cache:peek_local(?HOOKS_CACHE_NAME, cache_key_number(Number)) of
+            case not kz_term:is_empty(Number)
```
jamesaimonetti, line 254: We're trying not to allow the right-hand side of `andalso` and `orelse` to be non-boolean returns (even though technically allowed).

I'd write this using a function call instead:

```erlang
        Number = get_inbound_destination(JObj),
        lookup_account_id_by_number(Number, kz_term:is_empty(Number));
...

lookup_account_id_by_number(_Number, 'true') -> {'error', 'not_found'};
lookup_account_id_by_number(Number, 'false') ->
    case kz_cache:peek_local(...)
```

jamesaimonetti (changes requested):

core/kazoo_events/src/kz_hooks_util.erl
bradfordben: 👍  No problem, it will also remove the nested case so thats also a plus. Il make the change now.

jamesaimonetti:
Merged

PR #6617, opened 2020-08-07
[4.3] acdc - PISTON-1112: fix validation of agent pause timeout/time-limit

danielfinke:
- schema validates system_config.acdc default_agent_pause_timeout for non-negative integers or the "infinity" special string
- refactored out ?DEFAULT_AGENT_PAUSE_TIMEOUT into shared header
- fixed acdc_maintenance expecting integer, despite possibility of <<"infinity">> value
- moved Time-Limit validation in kapi_acdc_agent into PAUSE_* macros rather than AGENT_*


jamesaimonetti:
merged

PR #6619, opened 2020-08-07
PISTON-1104: Exposes voicemail forward-type option per mailbox.

rpn-voxter:
* Migrates the 'vm_message_forward_type' option from the 'system_config.callflows` schema to the 'system_config.voicemail' schema.
* Adds the 'vm_message_forward_type' option to the 'vmboxes' schema.
* Revises the implementation to look for the 'vm_message_forward_type' option per-mailbox and inherit from account configuration.


jamesaimonetti:
@rpn-voxter closing as master dev is elsewhere

PR #6620, opened 2020-08-13
Piston 1117: Add .vscode directory to .gitignore.

rpn-voxter:
Visual Studio Code uses the .vscode directory for workspace settings that should not be under source control.


jamesaimonetti:
Merged

PR #6621, opened 2020-08-19
[4.3]KZOO-70: use pvt_offering or the literal string and append _reci…

k-anderson:
…pient/_queue to the result


k-anderson:
master pr #6622

PR #6624, opened 2020-08-25
SUPP-14: fix losing origination-call-id when faing account to account

lazedo:
When sending fax, fax_worker sets `Origination-Call-ID` but will be discarded by `kapi_call` on the loopback. This will take care of fax events to honor this field.
Also it puts job_id in fax_worker handle_event as call_id so we can grab logs by job_id for all stages in fax_workers.

Increasing timeouts for fax_worker stages to give it a more relax time if something taking more times to finish like sending the page.


lazedo:
this needs more work, please don't merge

PR #6631, opened 2020-09-17
[4.3] kz json schema filter add additional properties support JObjs with the schema's additionalProperties field.

bradfordben:
kz_json_schema:filter/2 was not respecting the schemas “additionalProperties” field when filtering JObjs. As a result if a schema defined an object with “additionalProperties” set to an object then kz_json_schema:filter/2 would all ways filter out the objects.

Eg, if you defined a property of a schema as below and called filter/2 it would all ways be filtered out.

  ``` 
     "new_property": {
        "additionalProperties": {
            "properties": {
                "additional_property_1": {
                    "description": "Example additional property",
                    "type": "string"
                }
            },
            "additionalProperties": false,
            "type": "object"
        },
        "description": "Example new property with additional properties",
        "type": "object"
    }
```

With the new implementation, for it filter out additional properties from a JObj that are not defined in the schemas properties, you must define "additionalProperties": false. This is because by default "additionalProperties" is assumed to be true allowing additional properties to be defined without breaking the schema.

Kazoo5 PRs:
https://github.com/2600hz/kazoo-core/pull/77
https://github.com/2600hz/kazoo-crossbar/pull/47

### DIFF
--- a/applications/crossbar/priv/api/swagger.json
+++ b/applications/crossbar/priv/api/swagger.json
@@ -74,6 +74,7 @@
                         "default": {
                             "description": "conferences controls default",
                             "items": {
+                                "additionalProperties": false,
                                 "properties": {
                                     "action": {
                                         "description": "what action to assign to the star-code",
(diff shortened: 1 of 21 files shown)


lazedo (changes requested):
just blocking as we're not going to merge this. you need to find another solution in the core pr

jamesaimonetti:
4.3 is closed meow

bradfordben:
👍

PR #6634, opened 2020-10-01
[4.3] PISTON-1134: enforce non-empty CF number when CF is enabled

danielfinke:
Do not allow users to enable call forward if they have entered an empty forwarding number

### DIFF
--- a/applications/callflow/src/module/cf_call_forward.erl
+++ b/applications/callflow/src/module/cf_call_forward.erl
@@ -22,7 +22,7 @@
 -define(MOD_CONFIG_CAT, <<(?CF_CONFIG_CAT)/binary, ".call_forward">>).
 -define(KEY_LENGTH, 1).
 
--define(MIN_CALLFWD_NUMBER_LENGTH, kapps_config:get_integer(?MOD_CONFIG_CAT, <<"min_callfwd_number_length">>, 3)).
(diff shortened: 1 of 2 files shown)


applications/callflow/src/module/cf_call_forward.erl
```diff
@@ -22,7 +22,7 @@
 -define(MOD_CONFIG_CAT, <<(?CF_CONFIG_CAT)/binary, ".call_forward">>).
 -define(KEY_LENGTH, 1).
 
--define(MIN_CALLFWD_NUMBER_LENGTH, kapps_config:get_integer(?MOD_CONFIG_CAT, <<"min_callfwd_number_length">>, 3)).
+-define(MIN_CALLFWD_NUMBER_LENGTH, max(kapps_config:get_integer(?MOD_CONFIG_CAT, <<"min_callfwd_number_length">>, 3), 1)).
```
jamesaimonetti, line 25: Why floor the config to 1? While I appreciate the guard rails, configs generally are foot guns if used improperly :)

danielfinke: @jamesaimonetti I was matching the minimum length of 1 that is being added to `kzd_users`. Let me know if you'd prefer I omit this part.

jamesaimonetti: no its fine, was just curious.

harenson:
:+1:

jamesaimonetti:
Merged

PR #6635, opened 2020-10-01
[4.3] fix `make validate-js`

danielfinke:
`make validate-js` was always failing with
```
Usage: ./scripts/validate-js.sh file.json+
```


jamesaimonetti:
Merged

PR #6636, opened 2020-10-01
[4.3] PISTON-1135: compare existing account doc to saved account doc, not req data

danielfinke:
Prior to this change, the check of whether the account’s "pvt_enabled" property changed was always succeeding. This caused the registrations for the account to be flushed on every account save.


jamesaimonetti:
Merged

PR #6637, opened 2020-10-02
[4.3] Moved callflow validation logic (#6598)

jamesaimonetti:
Out of crossbar and into kazoo core so it can be used by other
modules.


jamesaimonetti:
@bradfordben just running for CI after rebasing here. If all passes, will merge

PR #6638, opened 2020-10-08
PISTON-1137: Fixes inaccurate stack trace.

rpn-voxter:
If an exception is caught within the `kvm_message:prepend_and_notify/5` function, the correct stack trace was potentially overwritten by an exception occuring within the `kvm_message:remove_malform_vm/2` function. The stack trace is now captured before calling that function.


jamesaimonetti:
Merged

PR #6642, opened 2020-10-15
[4.3] PISTON-1141: VM setup/configurability fixes

danielfinke:
- vmboxes that are not set up should still offer setup when in announcement-only mode
- in announcement-only mode, if configurability is disabled, go straight to mailbox
  - if configurability is enabled, go to config menu. The expectation is the primary action they will want is to disable announcement-only mode. They can get to mailbox with '0'


jamesaimonetti:
Merged

PR #6643, opened 2020-10-16
[4.3] PISTON-1142: continue config menu after deleting VM temp greeting

danielfinke:
- Resolves crash after deleting temporary greeting


harenson:
Hi @danielfinke may you please provide the steps to reproduce the crash? Thanks.

danielfinke:
Hey @harenson, the crash will be possible, currently, if the config menu is entered due to `"announcement_only": true` on the voicemail box. `check_mailbox_menu/2` skips the main menu and goes right to the config menu, as it should. The crash will occur in the `<<"check">>` clause of `handle/2` when the temporary greeting is deleted. Rather than add an extra clause to handle a mailbox return at the top level `handle/2` call, which would have to go thru `check_mailbox/2` etc. all over again, I just changed the post-delete action to go back to the config menu, like other config options.

PR #6644, opened 2020-10-16
[4.3] PISTON-1144: add missing entries for kapi_notifications for voicemail_deleted

danielfinke:
- Causes crashes without these if Preview mode is used for the template via cb_notifications


jamesaimonetti:
Merged

PR #6648, opened 2020-10-18
PISTON-1146: handle delayed job start replies from fax_worker

danielfinke:
- if job start/error reply comes in after the `kz_amqp_worker:call/3`
  timeout, need to handle removing jobs from serialized list in fax_jobs
  to avoid looping of job start attempts. Job will never be released
  from fax_jobs
  - job start/error reply becomes increasingly likely to be delayed as
    number of zones increases - kz_globals takes longer to register the
    globals


jamesaimonetti:
Merged, thanks! Good find...

PR #6649, opened 2020-10-19
Fixed crushes when used AWS ALB instead of haproxy

sergey-safarov:
When CouchDB server not available AWS load balancer returns 502 error code.
Added handles for this error code.


jamesaimonetti:
@sergey-safarov not to discourage you but the master branch here is not actively worked on. At the moment, while the infrastructure components are put in place for app exchange and whatnot, 5.x is being done in private repos. These repos will be opened up once the app exchange is in place for folks to sell and buy the apps they want to run in their KAZOO clusters.

I've updated the announcements and README to point this out. Please keep these PRs ready for the launch of 5.0!

PR #6650, opened 2020-10-19
kz_network_utils:to_cidr added IPv6 support

sergey-safarov:
PR allow auto-discover Kamailio IPv6 interfaces and properly populate ecallmgr ACL object.


jamesaimonetti:
See https://github.com/2600hz/kazoo/pull/6649 comment

PR #6651, opened 2020-10-19
crossbar: added VP9 codec into schemas

sergey-safarov:


sergey-safarov:
I do not understand how added allowed codec value breaks the JSON unit test.
Please help me understand.

jamesaimonetti:
See https://github.com/2600hz/kazoo/pull/6649

PR #6656, opened 2020-10-30
PISTON-1143: Reconciles the voicemail schema documentation.

tetrachromium:
At some point, the `voicemail` document was renamed as `vmboxes`, but the related documentation files were not  renamed to match. As a result, a hand-edited `voicemail.md` document diverged considerably from the auto-generated `ref/vmboxes.md` document. The contents of the `voicemail.md` have been reordered to reduce the degree of difference from the auto-generated file, and some minor corrections were made to the text.

### DIFF
--- a/applications/crossbar/doc/voicemail.md
+++ b/applications/crossbar/doc/voicemail.md
@@ -68,7 +68,7 @@ Key | Description | Type | Default | Required | Support Level
 
 
 
-## Fetch
+## Fetch all voicemail boxes for an account
 
 > GET /v2/accounts/{ACCOUNT_ID}/vmboxes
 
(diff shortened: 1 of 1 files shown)


rpn-voxter:
Looking at the differences within the `voicemail.md` file is likely to just be confusing. To understand the reasons for the changes and check for correctness, the revised `voicemail.md` file should be compared with the `ref/vmboxes.md` file. I suspect that the amount of difference between the two files has crept in due to the fact that they no longer have the same name. It was a while before I realized the two files were related. If you are willing to rename the file and redirect from the old URL, let me know and I will do that as a separate commit.

PR #6662, opened 2020-11-05
Add temporal_route end-date

supapo:


supapo:
What is the error? 
I run
git clone https://github.com/supapo/kazoo.git -b 4.3 kz
cd kz
./scripts/check-unstaged.bash

But there is no output

jamesaimonetti:
@supapo you can run the `circle-codechecks` and `circle-docs` targets on your branch. They will most likely generate the missing code from your PR

PR #6669, opened 2020-11-23
[4.3] KZOO-76: added support for IBM speech to text

k-anderson:


core/kazoo_speech/src/asr/kazoo_asr_ibm.erl
```diff
@@ -0,0 +1,154 @@
+%%%-----------------------------------------------------------------------------
+%%% @copyright (C) 2018-2020, 2600Hz
+%%% @doc
+%%% @author Sergey Safarov <s.safarov@gmail.com>, Sponsored by Audian
```
lazedo: i know copy+paste is a bless but can we change this ?

PR #6672, opened 2020-12-08
PISTON-1163: Refactors cf_acdc_agent status update logic.

rpn-voxter:
Voxter has introduced an additional `end_wrapup` action that requires `cf_acdc_agent` changes. The existing logic seems somewhat harder to follow than necessary; in inconsistent in various ways; and seems to still include obsolete code. These changes only refactor the status update logic to give the code a more consistent structure and simplify future changes. Other than making log messages more consistent, these changes should have no functionally observable effect.


jamesaimonetti:
Merged

PR #6676, opened 2020-12-16
[4.3] SUPP-33: add maintenance view to all dbs, clean soft deleted from all dbs

icehess:
We should remove soft deleted docs from all configured dbs. To this we also need
to add maintenance view to all dbs.

Also `maintenance/soft_deletes` is update to have id and rev so deleting won't
required lookup_rev one by one.


harenson (approved):
:+1:

PR #6688, opened 2021-01-14
[4.3] PISTON-1177: new approach to determine initial availability for agents of queue

danielfinke:
- when acdc_queue_manager boots, do not immediately assume all member agents are available
- when manager proc is ready, pub msg so that agents send an availability update
- any subsequent agent starts/availability changes will already be picked up by existing bindings


danielfinke:
Issues are only copyright-related

jamesaimonetti:
merged

PR #6690, opened 2021-01-18
[4.3] add disposition value to cdr/interaction endpoint

supapo:


supapo:
It failed because of the transformation to 2021 in the code check.
The modifications are only two additional rows

jamesaimonetti:
merged, sorry for the delay!

PR #6691, opened 2021-01-20
PISTON-1186: Fixes 'make dialyze_changed' failure when no Erlang file changes

rpn-voxter:
The `check-dialyzer.escript` script exits with a non-zero status when no Erlang files are specified on the command line. To prevent this from causing the make `dialyze-changed` target (or targets that depend on it) to fail when no Erlang files have been changed, the makefile now filters the list of changed files.


jamesaimonetti:
merged

PR #6694, opened 2021-02-01
Pulled in improvements to validate-js that will fix the failing test due to incorrectly parsing of the word 'function' in JSON schema files

bradfordben:
@jamesaimonetti as we discussed.


jamesaimonetti:
Merged

PR #6695, opened 2021-02-03
[4.3] KZOO-154: make sure e911 email format is email

holy-batman:
when invalid email is entered into email section it crashes 911 notifications.


harenson (approved):
:+1:

PR #6701, opened 2021-02-11
PISTON-1191 Add Queue name and options to tasks_listener

bradfordben:
After the removal of kz_globals from the tasks app the tasks app was creating N number of copies of an object where N was the number of tasks applications in the cluster.

Kazoo 5.x PR https://github.com/2600hz/kazoo-tasks/pull/13


bradfordben:
Thanks @k-anderson

PR #6702, opened 2021-02-20
HELP-19729: handle rapid carrier retry that leaves cf_exe processes

k-anderson:
There is potential for a callflow executor leak when sending a default error response.  In the field we found inbound calls that were forwarded to a number which canceled the request almost immediately.  When this happens callflows gets a message back from stepswitch that the bridge was unsuccessful then looks up the children in the callflow to see if any handle that specific hangup cause.  In this case (and most configurations) there are no special branches so it performs a default action which is to send back a SIP error message.  However, while that process is happening FreeSWITCH has already forwarded the cancel from the B leg back to the A leg and out to the carrier.  The particular carrier in this incident then immediately retries the same call in another zone.  When that call gets created in the other zone the ecallmgr call control usurps any other call control processes for that call id, terminating the ecallmgr call control in the first zone.  It is only then that the first callflow sends the command to publish the SIP error but with no ecallmgr control process in that zone anymore nothing is there to process it.  Callflows then waits indefinitely to hear back if the SIP error message was processed.

As a general avoidance of similar situations the all infinite wait times in call command have also been reduced to one day by default, however if the cluster is servicing things like turret phones they can set it back to infinity.

### DIFF
--- a/applications/callflow/src/cf_util.erl
+++ b/applications/callflow/src/cf_util.erl
@@ -431,7 +431,7 @@ send_default_response(Cause, Call) ->
                 {'error', 'no_response'} ->
                     lager:debug("failed to send default response for ~s", [Cause]);
                 {'ok', NoopId} ->
-                    _ = kapps_call_command:wait_for_noop(Call, NoopId),
+                    _ = kapps_call_command:wait_for_noop(Call, NoopId, 2 * ?MILLISECONDS_IN_SECOND),
                     lager:debug("sent default response for ~s (~s)", [Cause, NoopId])
             end
     end.
--- a/applications/crossbar/priv/api/swagger.json
+++ b/applications/crossbar/priv/api/swagger.json
@@ -29791,6 +29791,11 @@
                     "description": "call_command interdigit timeout",
                     "type": "integer"
                 },
+                "max_wait_ms": {
+                    "default": 86400000,
+                    "description": "call_command max_wait_ms",
+                    "type": "integer"
+                },
                 "message_timeout": {
                     "default": 5000,
                     "description": "call_command message timeout",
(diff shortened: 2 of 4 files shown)


applications/callflow/src/cf_util.erl
```diff
@@ -431,7 +431,7 @@ send_default_response(Cause, Call) ->
                 {'error', 'no_response'} ->
                     lager:debug("failed to send default response for ~s", [Cause]);
                 {'ok', NoopId} ->
-                    _ = kapps_call_command:wait_for_noop(Call, NoopId),
+                    _ = kapps_call_command:wait_for_noop(Call, NoopId, 2000),
```
jamesaimonetti, line 434: Use `2 * ?MILLISECONDS_IN_SECOND` to avoid magic numbers

applications/crossbar/priv/api/swagger.json
```diff
@@ -29791,6 +29791,11 @@
                     "description": "call_command interdigit timeout",
                     "type": "integer"
                 },
+                "max_wait": {
```
jamesaimonetti, line 29794: Rename to `max_wait_ms` to indicate units

core/kazoo_call/src/kapps_call_command.erl
```diff
@@ -237,6 +239,8 @@
 
 -type store_fun() :: kz_term:ne_binary() | fun(() -> kz_term:ne_binary()).
 
+-define(MAX_WAIT, kapps_config:get(?CONFIG_CAT, <<"max_wait">>, ?MILLISECONDS_IN_DAY)).
```
jamesaimonetti, line 242: Rename here, ignore prev comment lol

jamesaimonetti: Also, probably best to use `get_integer`

applications/callflow/src/cf_util.erl
k-anderson: good call, thanks :)

PR #6706, opened 2021-03-09
PISTON-1184: Correctly escape and wrap values when passing to FS to prevent special characters from causing issues

bradfordben:


bradfordben:
Master PR https://github.com/2600hz/kazoo-ecallmgr/pull/82

jamesaimonetti:
Merged

PR #6722, opened 2021-06-21
PISTON-1289: create_caller_id_name_if_undefined

bradfordben:
Added new attribute to the users object 'create_caller_id_name_if_undefined' to allow the creation of the users
caller id name to be optional when no caller id is defined.


lazedo:
@bradfordben Hi, can't you do this on the client side before submitting ?

bradfordben:
@lazedo Im not sure what you mean? Maybe im missing something?  

So for an example where a kazoo device calls some endpoint B,
Kazoo Device --- A leg ---> Kazoo/FS etc ---B leg---> Some endpoint B

Without the change if the kazoo device sets the caller id on the A leg it wont be used on the B leg as kazoo generates the default username if no caller id is set on the endpoint and uses that caller id name on the B leg.

PR #6728, opened 2021-07-26
PISTON-1267 add a handler for action 'move' to channel's crossbar API

alexey-khabulyak:
Add ability to move a call(call-flipping) via crossbar API
kazoo5 PR: https://github.com/2600hz/kazoo-crossbar/pull/203


jamesaimonetti:
Merged, thanks!

PR #6737, opened 2021-09-11
Fix to remove concatenated media names

kdamas93:


applications/cdr/src/cdr_util.erl
```diff
@@ -54,3 +59,16 @@ do_save_cdr(AccountMODb, Doc) ->
 -spec register_views() -> 'ok'.
 register_views() ->
     kz_datamgr:register_views_from_folder('cdr').
+
+-spec check_media_names(list()|kz_term:api_binary()) -> list().
+check_media_names(MediaNames) when is_list(MediaNames) ->
```
jamesaimonetti, line 64: You can do this all in the function head:

```erlang
check_media_name([HeadMN | TalMN]) ->
```
Implies the `is_list` check, and binds the head and tail of the list, all in one go!

applications/cdr/src/cdr_util.erl
```diff
@@ -54,3 +59,16 @@ do_save_cdr(AccountMODb, Doc) ->
...
+    [HeadMN|TailMN] = MediaNames,
+    re:split(HeadMN, ",") ++ check_last_media_name(TailMN);
+check_media_names(MediaNames) ->
+    re:split(MediaNames, ",").
+
+-spec check_last_media_name(list())->list().
+check_last_media_name([FirstMediaName | []])->
```
jamesaimonetti, line 71: `[FirstMediaName | []]` is the same as `[FirstMediaName]` - the `| []` is implied.

applications/cdr/src/cdr_util.erl
```diff
@@ -54,3 +59,16 @@ do_save_cdr(AccountMODb, Doc) ->
...
+    re:split(FirstMediaName, ",");
+check_last_media_name([FirstMediaName | LasttMediaName])->
```
jamesaimonetti, line 73: `[FirstMediaName | LastMediaName]` means that `LastMediaName` will be a list of binaries. Then you `re:split` on it below which is likely to crash.

Where are the tests showing this functionality working as expected?

applications/cdr/src/cdr_util.erl
```diff
@@ -54,3 +59,16 @@ do_save_cdr(AccountMODb, Doc) ->
 -spec register_views() -> 'ok'.
 register_views() ->
     kz_datamgr:register_views_from_folder('cdr').
+
+-spec check_media_names(list()|kz_term:api_binary()) -> list().
+check_media_names(MediaNames) when is_list(MediaNames) ->
+    [HeadMN|TailMN] = MediaNames,
+    re:split(HeadMN, ",") ++ check_last_media_name(TailMN);
+check_media_names(MediaNames) ->
```
jamesaimonetti, line 67: Won't this crash is `MediaNames` is `'undefined'` ?

jamesaimonetti (changes requested):
A first pass. Need to see tests showing this functionality working as expected for various formats of media names (non-existent, one binary, a list of one binary, and a list of two or more binaries, at a minimum.

PR #6739, opened 2021-09-21
Fixed chunked encoding of potential json terms.

lordnull:
This was discovered when the `page_size` property for cdr's was sometimes
a string, and some times a number.

When finishing a chunked return using json, the trailer properties were
encoded by first turning the property values into binaries, and then
wrapping those binaries in quotes. This meant the erlang term `43` always
become `"43"`.

By checking if the erlang term is already a json term, and only turning to
a binary if it is not, then we can avoid encoding things like json objects,
`null`, booleans, and numbers into strings, and things like tuples, pids
(and more) are still handled as before.

The spec doesn't fully reflect the above, however, as we don't yet have
anything other than strings and numbers.


lordnull:
As per a comment on https://github.com/2600hz/kazoo-crossbar/pull/216 , this is unlikely to be merged as it is not considered a bug. Thus, I'll close it to avoid excess clutter.

PR #6745, opened 2021-12-07
acdc fix pickup and exit by dtmf

vlitvinov:
1) Add the possibility to make pickup (call replace) to agent's calls.
2) Fix an issue with an exit from the queue by pressing DTMF key.


applications/acdc/src/acdc_agent_listener.erl
```diff
@@ -1032,6 +1033,7 @@ maybe_connect_to_agent(MyQ, EPs, Call, Timeout, AgentId, _CdrUrl) ->
                                   ,{<<"Retain-CID">>, <<"true">>}
                                   ,{<<"Agent-ID">>, AgentId}
                                   ,{<<"Member-Call-ID">>, MCallId}
+                                  ,{<<"originate_signal_bond">>, MCallId}
```
jamesaimonetti, line 1036: I can't speak to the ACDc-specific portions, but this name should match the format of other keys: `Originate-Signal-Bond` in this case.

PR #6748, opened 2022-01-19
Add restrict_presence_event_publisher ecallmgr system_config option

timeodonovan:
When enabled it prevents all but the ecallmgr node handling a call from sending presence updates. This stops duplicate notifies being sent to devices which can result in them returning 500 errors, removing their subscriptions and subsequently breaking BLFs.

A caveat to this is that the initial state change will not be forwarded to subscribers when a device is dialling, as no ecallmgr node will be defined at this stage. This is usually for a very short period as the following state changes will be at the point where the call is being handled by an ecallmgr node.


jamesaimonetti:
@timeodonovan thanks, taking a look. Will address CI issues and merge to 4.3 when passing here.

jamesaimonetti:
Merged, thanks!

PR #6754, opened 2022-07-27
Update how to created MD5 hash

nkarunarathne:
Since it is a little bit confusing the wording of "usernames are always in lower case", did change to a more descriptive version.


nkarunarathne:
changes did on another pull request

PR #6762, opened 2024-05-15
rxfax called before answer fix

hmmhesays:
There is a possibility of rxfax getting called before answer it it arrives in the queue ahead of answer.


jamesaimonetti:
merged, thanks!

